PR #103, opened 2018-08-02
feat: select and autucomplete component, rfc

blackswanny:
#### Major: Select and  Search\Autocomplete component


.prettierignore
```diff
@@ -1,3 +1,4 @@
 node_modules
 dist
 coverage
+chore: rollup build (#2)
```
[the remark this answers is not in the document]
blackswanny: that came from `master` during merge conflict. Idk if I should leave it.

rfcs/select.md
```diff
@@ -0,0 +1,132 @@
...
+* `onMouseEnter: func`:
+  handler for events on trigger element
+* `onMouseLeave: func`:
+  handler for events on trigger element
+* `onFocus: func`:
+  handler for events on trigger element
+* `onBlur: func`:
+  handler for events on trigger element
+
+### `Label` API
```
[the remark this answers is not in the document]
blackswanny: cleaned up

rfcs/select.md
```diff
@@ -0,0 +1,132 @@
...
+} from './index';
+import {withStyle} from 'styletron-react';
+
+const CustomOption = withStyle(Option, {
+  textColor: 'red',
+});
+
+export default () => {
+  return (
+    <div>
```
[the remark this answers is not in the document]
blackswanny: ok, but no difference

src/components/select/constants.js
```diff
@@ -0,0 +1,26 @@
...
+export const ICON = {
+  CLEAR_ALL: 'CLEAR_ALL',
+  CLEAR_TAG: 'CLEAR_TAG',
+  SELECT: 'SELECT',
+  SELECTED: 'SELECTED',
+  LOOP: 'LOOP',
+};
+
+export const STATE_TYPE = {
+  select: 'SELECT',
```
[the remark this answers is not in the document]
blackswanny: I just following the pattern we use in this type of constants

.prettierignore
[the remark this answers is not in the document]
gergelyke: it does not seem like a valid blob pattern, we should delete it imho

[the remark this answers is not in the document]
blackswanny: ok

rfcs/select-component.md
```diff
@@ -0,0 +1,191 @@
...
+* `StyledSelectInput`
+* `StyledOption`
+* `StyledDropDown`
+* `StyledLabel`
+* `StyledHint`
+
+### `Select` and `StatefulSelect` API
+
+* `type: TYPE.SEARCH | TYPE.SELECT`:
+  type of component to be in select or search mode (with lookup input)
```
schnerd: What's the downside of always allowing search?

Even for normal HTML `<select>` elements, if you open it and start typing an option prefix, the browser starts selecting items that match. I think if our Select doesn't support that, it's inferior to the native select – so might as well just make every select searchable?

rfcs/select-component.md
```diff
@@ -0,0 +1,191 @@
...
+* `selectedOptions: Array<Object>`:
+  Current selected options. Every option object has `id: string` and `label: React$Node`. Label is defaulted to display for selected option, otherwise see `getSelectedOptionLabel` method
```
schnerd: I think a better type for this might be `string | Array<string>`
- For a single select, forcing people to use an array seems like an annoyance
- The selectedOptions will usually be passed in from the state of the parent form, and that state usually only includes keys, not labels. People will need to map the selected keys back to the corresponding option object, which is extra work for them – we should automatically do that internally. `string | Array<string>` also seems closer to what people expect from a `value` prop on a input.

rfcs/select-component.md
```diff
@@ -0,0 +1,191 @@
...
+* `placeholder: ?string`:
+  Placeholder text if nothing is selected. Default is `Choose one...`
+* `rows: ?number`:
+  Represents maximum visible length of options, all other will be scrolled. If not defined, all options will be visible.
+* `options: Array<Object>`:
+  All Options in dropdown. Should be provided for Select and Search mode equally. Every option object has `id: string` and `label: React$Node`. Label is defaulted to display for option in dropdown, otherwise see `getOptionLabel` method. Optional `disabled: boolean` for option to be disabled from selection.
+* `label: ?React$Node`:
+  Component or String representing label for Select component. Default is `''`
+* `caption: ?React$Node`:
+  Component or String representing hint message for Select component. Default is `''`
```
schnerd: I think Nadiia's work in #125 will handle this outside of components. Can we remove these?

rfcs/select-component.md
```diff
@@ -0,0 +1,191 @@
...
+* `textValue: ?string`:
+  Initial text value for Search to lookup for text.
```
schnerd: I think `textValue` should be internal component state of `Select`. The vast majority the time I use form controls, I'm using the stateless version, because I need to pass in the current value based on redux or parent component state. We shouldn't force users to create a `textValue` state property in their parent component, the `Select` component should just handle this automatically.

rfcs/select-component.md
```diff
@@ -0,0 +1,191 @@
...
+* `error: ?string`:
+  Error message and error state. Default is `''`
```
schnerd: Error message should ideally be displayed by FormControl in #125. I think this should just be a boolean that sets input border color to red (or whatever the error style is here)

rfcs/select-component.md
```diff
@@ -0,0 +1,191 @@
...
+* `multiple: ?boolean`:
+  Sets if multiple choices are allowed in Select component. Default is `false`
+* `autoFocus: boolean`:
+  make the control focused (active). Default is `false`
+* `disabled: boolean`:
+  Disable control from being changed
+* `required: boolean`:
+  Mark control as required
+* `overrides: {}`
+  * `DropDown: ?React$Node` component to use for dropdown list
```
schnerd: Should we call this `Menu` for consistency with Anh's component?

rfcs/select-component.md
```diff
@@ -0,0 +1,191 @@
...
+  * `Option: ?React$Node` component to use for options in dropdown list
+  * `Root: ?React$Node` component to use for most top of the select component
+  * `Input: ?React$Node` component for Input showing current selected value(s). See `Input` Control of this framework for reference to override it's functionality.
```
schnerd: I think these should all be React.ComponentType, not React.Node

rfcs/select-component.md
```diff
@@ -0,0 +1,191 @@
...
+  * `SearchIcon: ?React$Node` component for all icons appearing in Select component. It's provided `$type: ICON.LOOP | ICON.CLEAR_TAG | ICON.CLEAR_ALL | ICON.SELECTED` to setup corresponding icon of Select component
```
schnerd: I kind of think each of these icons should be individually override-able. If I just want to override the CLEAR_TAG, I don't want to have to wire up a bunch of defaults for the other three icons.

rfcs/select-component.md
blackswanny: Currently we have no such UI design for Search + single, it's always Search+multiple. Search+single will require to change logic of our Input. Otherwise no issues, so anyone can switch on Search mode by allowing this type

rfcs/select-component.md
```diff
@@ -0,0 +1,191 @@
...
+  * `Tag: ?React$Node` component for selected options Tags shown in Input for multiple mode selection
+* `onSelect: func(e: SyntheticInputEvent, params: Object)`:
+  handler for events on trigger element when option is selected. `params` has `id` and `label` of selected option and `selectedOptions` array of all of selected.
+* `onUnSelect: func(e: SyntheticInputEvent, params: Object)`:
+  handler for events on trigger element when option is unselected (for multiple choices). `params` has `id` and `label` of unSelected option and `selectedOptions` array of all of selected.
```
schnerd: I think we should just merge these two into a single `onChange(selected: Object | Array<Object>)`. People almost never need the event, so it's pointless to expose it (we can hide it as a second argument if people ask for it). 

And passing people the selected object or array of obejcts (for multiple) makes it easier to store this back in form state.

This also prevents me from needing to add both `onSelect` and `onUnSelect` handlers for multi-select.

react-select uses a single onChange with this signature as well: https://react-select.com/props

rfcs/select-component.md
```diff
@@ -0,0 +1,191 @@
...
+                    style={{
+                      borderRadius: '50%',
+                      height: '75px',
+                    }}
+                    src="3.jpg"
+                  />Third
+                </span>
+              ),
+            },
+          ],
```
schnerd: Putting react components in state seems like an anti-pattern.
http://web.archive.org/web/20150419023006/http://facebook.github.io/react/docs/interactivity-and-dynamic-uis.html

 I think instead of allowing people to pass label as `React.Node`, they should just specify a component override for `Item` or we should expose a renderProp.

rfcs/select-component.md
schnerd: I think this is search + single, no?

image

Can we just remove this prop and allow search for all types?

src/components/select/constants.js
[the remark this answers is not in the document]
schnerd: I think we decided to lowercase for all of these (see input/constants.js, popover/constants.js, and Jun 11th meeting notes).

rfcs/select-component.md
blackswanny: this one is not yet single, it can be both. I saw only single pick for Select mode.
We can allow search for all, but as per current idea they are separate with Select
@nguyensomniac correct me if I am wrong

rfcs/select-component.md
blackswanny: Native `Select` has imperfect interface. It's controversial. For single option it contain it in `value`, but for multiple options it is in `selectedOptions` and in totally different format. So I wouldn't reference to native Select as good practice. So I unified it to be always Array, so people no need to guess, what is the format of output. 
Also selectedOptions cannot come for Select from outside (parent) as they are a value of it's internal state (like for any input) and lots of routines based on them, if they always come from outside we cannot be sure they still support the same format, and more markup will be involved in re-rendering.
Mapping back is a concern, but if only `label` doesn't contain what is needed, and usually it does contain. You can look at `onSelect\onUnSelect` functions, which send back both `id` and `label`, so mapping is not needed unless we have data in other arrays or fields. `label` itself is a `React$Node`, which maybe anything valid: string, component, tag, etc

rfcs/select-component.md
blackswanny: only if it's not a part of Input yet. But so far it is a part of it, which makes it easier for me to use

rfcs/select-component.md
blackswanny: in this case it would be impossible to pre-setup input value to some initial state, right?

rfcs/select-component.md
blackswanny: error is just passed to Input as is, cause only Input subcomponent is used for error state. I can rename it in Select outer interface to be boolean, but there has to be message here to comply with Input

src/components/select/select.js
```diff
@@ -0,0 +1,304 @@
...
+
+  componentDidMount() {
+    const document = require('global/document');
+    document.addEventListener('click', this.handleClickEvent, {
+      capture: true,
+    });
+  }
+
+  componentWillUnmount() {
+    const document = require('global/document');
```
schnerd: Can we move this require to the top of the file?

rfcs/select-component.md
blackswanny: it is in fact derived from Ahn's component, but more recognized terminology in scope of controls is Dropdown. So Menu seems to be more confusing in this case

rfcs/select-component.md
blackswanny: fixed

src/components/select/select.js
```diff
@@ -0,0 +1,304 @@
...
+      });
+      this.props.onUnSelect(e, {id, label, selectedOptions});
+    }
+  };
+
+  onKeyUp = (e: SyntheticInputEvent<HTMLInputElement>) => {
+    const newTextValue = e.target.value;
+    this.setState({textValue: newTextValue});
+    this.props.onKeyUp(e, {textValue: newTextValue});
+  };
```
schnerd: I think the Select element should do the option filtering, similar to react-select. IMO that's one of the biggest benefits of a reusable autocomplete component – customers ideally shouldn't have to implement that filtering logic on their own.

rfcs/select-component.md
blackswanny: there is only one prop to override for icon ($type)
Otherwise we would have to add 5 more different components just to support all of them.
With `override` approach it would be harder to change code everywhere and not mistaken if we wish to maintain or add changes here

rfcs/select-component.md
blackswanny: I have event as first arg, because it's a standard we use across our code so far, but agree it's not needed here.
We do pass all current `selectedOptions` back with it, all in `params` object, but we may still need to know, which one was selected\unselected, that's why we provide it as well.
I can remove UnSelect and Clear handlers, but have to add some extra param of action, like it happens in `react-select` as well

src/components/select/select.js
```diff
@@ -0,0 +1,304 @@
...
+    onMouseUp: () => {},
+    error: '',
+    autoFocus: false,
+  };
+
+  state = {
+    textValue: '',
+    selectedOptions: [],
+    isDropDownOpen: this.props.type === TYPE.SEARCH,
+    ...this.props,
```
schnerd: Why put all props in state?

rfcs/select-component.md
blackswanny: `label` is whatever renderable is inside of Option. We do not even care, what's there giving total flexibility. So there is nothing to override, otherwise we shall have default layout for it, but we don't have it by design. The same approach is used in Menu, actually it is made for Menu as we re-use it.
Yes, state shouldn't contain elements, but from perspective of Select it is a data to be shown.
Being able to support a dozen different component in labels (especially for Search results) seems to be more better, than do this work in one `override` class function

src/components/select/select.js
blackswanny: yes, seems like it's friendly to server-side rendering

src/components/select/select.js
blackswanny: But all options are provided from outside as it's totally custom thing, there might be ajax call or other logic to recalculate all options. So at the moment of `keyUp` we just don't have actual and correct options to filter. Moreover, whatever is in text value in most commercial implementation (for instance lookup for product) is not exact text to look for, it's just a key and exact algorithm to find and filter results is known only to special search engine logic

rfcs/select-component.md
schnerd: Yup. And I think that's perfectly fine – I've never seen a product where that's a requirement.

If necessary we could have an "initialTextValue" prop, but I don't think it's needed at this point.

rfcs/select-component.md
schnerd: Nadiia will be removing those from input I think, in favor of FormControl. FormControl just wraps existing components, so you don't have to add anything here to support it.

rfcs/select-component.md
schnerd: Okay, I think @nadiia may change this soon when FormControl lands (correct me if I'm wrong)

rfcs/select-component.md
schnerd: Does anyone have a use case for supporting a dozen different labels inside a Select? I'd argue they don't.

Additionally, if we want to support filtering the options based on keywords the user entered, this label needs to be a string.

src/components/select/select.js
schnerd: The vast majority of the time I use select, I have a list of many options, and I just want basic lowercase string matching capability (maybe tokenized by word). This is exactly with react-select offers out of the box, we shouldn't burden customers by making them implement this logic themselves every time they use a select component.

Also, I posted this in a separate thread, but I think for async options we chould accept `options` with the following signature: `(query: string) => Promise<Option[]>`. In this case it would skip our naive string filtering and call the users `options` callback with the current query. We show a loading indicator until the promise resolves.

blackswanny: you try to put too many concerns in one place. From one side you want a simple filter, but as I mentioned we have a mixed select\search control, and if it works for select mode, it won't work for search mode. If customers want primitive lowercase search, it is a one liner function, I don't think we need to implement this and support conditionally. On the other hand supporting promise signature also seems to much for me, as remember, it is just a control, ot shouldn't comprehend all logic delivering data to it.

src/components/select/select.js
blackswanny: removed

rfcs/select-component.md
blackswanny: There are examples in internet. Not dozens, but a few. I can do as you wish. I can treat label as a `string` by default and if it is object containing data than customer should provide override to treat this data correspondingly.
image

rfcs/select-component.md
blackswanny: fixed

rfcs/select-component.md
blackswanny: removed caption

src/components/select/select.js
schnerd: > you try to put too many concerns in one place.

This is a fair point. I don't mean to advocate for shoving all this logic into `<Select>`, I mostly just want a solution that works out of the box. That could mean that in the future we split some of this behavior into a separate wrapper component like `<AsyncSelect>`, [which is what react-select does](https://react-select.com/async).

> If customers want primitive lowercase search, it is a one liner function

This kind of oversimplifies the issue a bit in my opinion. Here's the minimum amount of code I would need to write to support this on my own:

```js
class UserSelect extends React.Component {
  state = {
    filteredOptions: this.props.options,
  };
  onKeyUp(params) {
    this.setState({
      filteredOptions: this.props.options.filter(o => (
        o.label.toLowerCase().indexOf(params.textValue.toLowerCase()) >= 0;
      ));
    });
  }
  clearFilter() {
    this.setState({
      filteredOptions: this.props.options
    });
  }
  render() {
    return <Select 
      options={this.filteredOptions}
      onKeyUp={this.onKeyUp} 
      onClear={this.clearFilter}
      onSelect={this.clearFilter} />;
  }
}
```

While this isn't a gigantic amount of code, it does require at least one extra piece of state, I have to subscribe to all three onKeyUp/onClear/onSelect, and in practice should would probably need to be debounced or have better keyword matching. It would get old re-implementing this every time I use a select in my app. I believe this is why react-select offers this behavior out of the box.

If the concern is giving too many concerns to one component, then maybe we keep Select dumb, but also offer an AutocompleteSelect export that has this behavior built in.

It might be interesting to get feedback from some other potential customers too. I think I'm the only one that's reviewed this RFC/PR so far.

rfcs/select-component.md
schnerd: Yeah the Amazon example is an interesting case. I'd probably still lean towards handling this via an override component or `renderOption` prop, but I agree it's not cut and dry.

Maybe we can get feedback from some more folks.

src/components/select/select.js
blackswanny: ok, here are some solutions for your code
1) To avoid extra state we can still use `options`, but they will be not so efficient as options are wiped every time we clean up. To address that we can have some inner cache strategy. However, `filtered` is not needed here
2) To not subscribe to so many we may do on `onChange` handler at least for `clearAll`\`onSelect`\`onUnselect` and provide extra action telling what happened instead like select, unselect, clear (btw, we don't have `clearfilter` right now for this control as Lily wants this X button to clean selected options only)

rfcs/select-component.md
blackswanny: I changed towards label to be pure string or object of data, in second case customer is obliged to provide custom code to return label layout based on label data

benshope:
I have a hard time evaluating this as a single RFC because stateful autocompletes are so complex.  I would love to see us break this apart into several RFCs and land the stateless parts first:

1. Create stateless style components for the inputs, the list items, and the popover (and create stories showing them used individually and as a stateless autocomplete)
2. Create a virtual scroll component (and create stories showing it's usage)
3. Create string filter functions (perhaps with memoization) (and create stories showing their usage)
4. Create stateful stories showing all these ^ components composed together as an autocomplete
5. Create a stateful autocomplete component informed by what we learned in step 4 writing stateful autocomplete stories out of these individual components

What we gain by breaking this apart is a more human development process with steady progress, clear goals, and great visibility into each stage.

benshope:
Although if we do want to go for it all in one go, we could completely avoid duplicating native props with an API like this:  ```
<StatefulSelect value="merp" onChange={this.onSelectChange} filterFunction={(inputValue, options => options.filter(option => option.contains(inputValue)))} renderSelect={(inputValue, onInputChange) => (<Select>
      {['hello', 'we', 'are', 'some', 'pills'].map(pillText => (
        <Pill text={pillText} onRemove={action(`pill/onRemove ${pillText}`)} />
      ))}
      <ExpandingInput
        text="yeee"
        input={<input onChange={onInputChange} value={inputValue} />}
      />
      <Suggestions>
        {['these', 'are', 'some', 'suggestions'].map((suggestion, i) => (
          <Suggestion
            className={i === 1 ? 'selected' : ''}
            onClick={action('autocomplete/onClick')}
            key={suggestion}
          >
            {suggestion}
          </Suggestions>
        ))}
      </Suggestions>
    </Select>)}>
```

File: src/components/select/select.js [hunk 18]
```diff
@@ -0,0 +1,321 @@
...
+import type {
+  DefaultPropsT,
+  LabelT,
+  OptionT,
+  PropsT,
+  StatelessStateT,
+  ChangeActionT,
+} from './types';
+import {getOverride} from '../../helpers/overrides';
+import document from 'global/document';
```
rtsao: This package is entirely bloat assuming `document` is never used outside of browser-only React lifeycle methods, which seems to be the case.

blackswanny: what's the solution? We have lint complaining about inline requires and no approach to have bundle based on browser\server code

src/components/select/select.js
schnerd: Seems like these still haven't been moved?

blackswanny: I took it back to inline as @rtsao said I should avoid import for those, which work only on client-side

schnerd: Unless you guys had a separate conversation, it sounds like @rtsao said to just use eslint comments:

```
componentWillUnmount() {
  /* eslint-disable-next-line noundef  */
  document.addEventListener('click', this.handleClickEvent, {
      capture: true,
    });
}
```

Personally I'd be fine with just adding a `/* global document */` to the top of the file, and relying on tests to catch cases where we're incorrectly using document server-side.

package.json
```diff
@@ -100,6 +100,7 @@
     "styletron-standard": "^1.0.3"
   },
   "dependencies": {
+    "global": "^4.3.2",
```
rtsao: Was this intentional?

blackswanny: yes. it was used as inline dependency

src/components/select/select.js
rtsao: Yeah, I think including this as an actual dependency is pure bloat

src/components/select/constants.js
```diff
@@ -0,0 +1,26 @@
...
+
+export const ICON = {
+  CLEAR_ALL: 'CLEAR_ALL',
+  CLEAR_TAG: 'CLEAR_TAG',
+  SELECT: 'SELECT',
+  SELECTED: 'SELECTED',
+  LOOP: 'LOOP',
+};
+
+export const STATE_TYPE = {
```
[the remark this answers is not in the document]
schnerd: Would be good to stay consistent. I prefer `STATE_CHANGE_TYPE` since technically they refer to types of changes, not to types of state.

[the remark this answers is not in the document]
blackswanny: state change type is long , also "change" is a specific event , whereas this one states for many different events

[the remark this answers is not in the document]
schnerd: I believe Anh, myself, and Nadiia have all been using STATE_CHANGE_TYPE in our components. Can we just go with that for the sake of consistency for now? If you want to propose changing to STATE_TYPE in the future, maybe we can open that a separate issue/PR.

gergelyke:
@blackswanny do you think we can break this PR down to smaller PRs? I believe if we would do that we could move faster, as we could review things in a shorter amount of time.

blackswanny:
@gergelyke i cannot split to smaller PRs right now as it has already been reviewed and there are just a few requests left for existing code in this PR. I should have split to smaller PRs before all of that.

src/components/select/constants.js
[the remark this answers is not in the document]
blackswanny: ok

package.json
blackswanny: removed

src/components/select/constants.js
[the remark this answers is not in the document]
blackswanny: renamed

src/components/select/dropdown.js
```diff
@@ -0,0 +1,113 @@
...
+            $type={type}
+            $isOpen={isDropDownOpen}
+          />
+        ),
+        ListItem: DropDownItem,
+      }}
+      items={options}
+    />
+  ) : (
+    <div />
```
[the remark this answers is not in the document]
blackswanny: this is recognized as constructor function, and eslint complains that `null` value is lost if returned from function with `new`

src/components/select/select.js
```diff
@@ -0,0 +1,322 @@
...
+    const document = require('global/document');
+    document.removeEventListener('click', this.handleClickEvent, {
+      capture: true,
+    });
+  }
+
+  handleClickEvent = (event: MouseEvent) => {
+    // eslint-disable-next-line react/no-find-dom-node
+    const el = findDOMNode(this);
+    // $FlowFixMe
```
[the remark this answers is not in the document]
blackswanny: no, in here we have different types of React return and expected code here. For some reason React type doesn't look like DOM element with method `contains`

src/components/select/stateful-select-container.js
```diff
@@ -0,0 +1,134 @@
...
+    onMouseEnter: () => {},
+    onMouseLeave: () => {},
+    onFocus: () => {},
+    onBlur: () => {},
+  };
+
+  constructor(props: StatefulContainerPropsT) {
+    super(props);
+    const {initialState} = this.props;
+    this.state = {
```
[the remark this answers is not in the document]
blackswanny: ok

src/components/select/stateful-select-container.js
```diff
@@ -0,0 +1,134 @@
...
+    const newState = stateReducer(type, nextState, this.state, e, params);
+    this.setState(newState);
+  };
+
+  render() {
+    const {
+      children = (childProps: {}) => null, // eslint-disable-line no-unused-vars
+      initialState, // eslint-disable-line no-unused-vars
+      stateReducer, // eslint-disable-line no-unused-vars
+      ...rest
```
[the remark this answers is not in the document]
blackswanny: we used to have a different pattern, just followed it

src/components/select/select.js
```diff
@@ -0,0 +1,340 @@
...
+      }
+      case STATE_CHANGE_TYPE.keyUp: {
+        // $FlowFixMe
+        const newTextValue = e.target.value;
+        this.setState({textValue: newTextValue});
+        if (this.props.simpleFilter) {
+          let filteredOptions = this.props.options.filter(
+            option =>
+              typeof option.label === 'string' &&
+              option.label.indexOf(newTextValue) >= 0,
```
schnerd: Can we do a case-insensitive match here? A lot of Selects have capitalized option names, whereas most users just enter a lowercase search query.

```
option.label.toLowerCase().indexOf(newTextValue.toLowerCase())
```

blackswanny: sure

src/components/select/select.js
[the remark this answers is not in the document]
blackswanny: fixed

src/components/select/styled-components.js
```diff
@@ -0,0 +1,162 @@
...
+      }
+    : {
+        paddingTop: scale300,
+        paddingBottom: scale300,
+        paddingRight: '18px',
+        paddingLeft: '18px',
+      };
+  return {
+    ':hover': {
+      cursor: disabled ? 'not-allowed' : 'text',
```
nguyensomniac: Cursor should be 'pointer' for selectable options

nguyensomniac:
To clarify—tags in the Storybook are placeholders and will be updated when we have a proper tag to use, yes?

ardok:
@blackswanny 
Do you have an example of how a dev would use this component to build a hierarchy select dropdown with search box?
Just a rough idea would be enough.

blackswanny:
@ardok You mean nested levels options? No, currently it's not supported, as `click` automatically selects option. But if option is replaced with the same nested dropdown (it is easy as its a separate subcomponent) and `click` is redefined to show dropdown, than it would be easy. If you think it's needed, we can make such change here

nguyensomniac:
Some more design feedback

frame 16

src/components/select/styled-components.js
blackswanny: ok

src/components/select/styled-components.js
```diff
@@ -0,0 +1,162 @@
...
+});
+
+export const Input = styled('input', props => {
+  return {
+    ...getInputStyles({...props, $size: SIZE.default}),
+    cursor: 'pointer',
+    width: 'auto',
+    flexGrow: '1',
+    ':hover': {
+      cursor: 'pointer',
```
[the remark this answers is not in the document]
blackswanny: ok

blackswanny:
@nguyensomniac fixed all 1-5 items , except some in 3). The styles are mostly after original Input. It has line-height 24 for main input exact control (line height is applicable only for inline elements). However, there is some padding for parent container, so total size is 62px, not 40px. So if I make 40px of most outer container, it will be totally different and small
image

blackswanny:
As Tags work already started here, with @nguyensomniac we decided to extract Tags from this component to separate one and reuse it here

nguyensomniac:
40px in the outermost container is correct; these need to match input styles whenever possible. May help to look with 24px tall tags—i.e. text at 12px font-size, 20px line-height, 2px each for vertical padding

ardok:
@blackswanny nah, just for future consideration because hierarchy dropdown is pretty big for my team. Especially for location and departments.

We have this currently:
aug-29-2018 17-00-26

blackswanny:
@ardok interesting case, I think we can just change Dropdown itself, no need to change Select logic. Do you want me to try PoC for this right now?

blackswanny:
@nguyensomniac fixed

PR #119, opened 2018-08-11
[RFC] Modal Component

schnerd:
RFC for modal component (Click "Files changed" above). Attempts to take all the best API design bits from other modals on the market (material-ui, antd, palantir, etc), as well as accessibility best practices.

Two other lingering topics that I'd appreciate your feedback on:

### 1. Opinionated Header/Footer

People can render whatever arbitrary content they want, but we also want to make it easy to create a beautiful standard modal. After some thought, I think one of the two following APIs would be best:

#### Option A: Top-level title/actions props
```js
<Modal
  isOpen={this.state.open}
  title="Hello world!" // string|React.node
  actions={[{
    // Actions are mapped to `<Button>` instances, essentially same props
    type: 'secondary'
    label: 'Close'
    onClick: this.handleClose
  }, {
    type: 'primary'
    label: 'Close'
    onClick: this.handleSubmit
    isLoading: this.state.isSubmitting
  }]}
>
  Lorem ipsum dolor sit amet
</Modal>
```

#### Option B: Use exported ModalHeader, ModalBody, ModalFooter components.
```js
<Modal isOpen={this.state.open}>
  <ModalHeader>Hello world!</ModalHeader>
  <ModalBody>Lorem ipsum dolor sit amet</ModalBody>
  {/* ModalFooter can take same actions array as Option A */}
  <ModalFooter actions={[{
    type: 'secondary'
    label: 'Close'
    onClick: this.handleClose
  }, /*  more buttons */ ]} />
  {/* Or you can just pass children if you only want spacing/border styling */}
  <ModalFooter>
    My custom footer content
  </ModalFooter>
</Modal>
```

Whichever option we choose, I think the `actions` array is nice because it prevents you from needing to do all the margin between buttons, and right-alignment yourself.

### 2. `Modal.method()` Imperative API

[Antd has a pretty innovative alternate modal API](https://ant.design/components/modal/#Modal.method()) for quickly rendering modals in a callback handler. Here's what it looks like:

```js
handleSubmit() {
  this._modal = Modal.confirm({
    // Mostly same props as <Modal/>
    title: 'Are you sure delete this task?',
    content: 'Some descriptions',
    okType: 'danger',
    onOk() { console.log('OK'); },
    onCancel() { console.log('Cancel') },
  });
  // Sometime later...
 this._modal.destroy();
}
```

Internally, it just attaches a new react root to the page with a modal inside of it.

In my opinion, this imperative API actually better matches the mental model of modals compared to the declarative approach of always having a `<Modal/>` in your render() method and controlling `isOpen`. 

I think we should offer this utility API **in addition to** the default `<Modal/>` API, but curious to see if anyone has strong opinions against that.

### DIFF
--- a/rfcs/modal-component.md
+++ b/rfcs/modal-component.md
@@ -2,69 +2,69 @@
 
 ### Exports
 
-* `Modal`
-* `ModalHeader`
-* `ModalBody`
-* `ModalFooter`
-* `StyledRoot`
-* `StyledBackdrop`
-* `StyledDialog`
-* `StyledDialogInner`
-* `StyledHeader`
-* `StyledBody`
-* `StyledFooter`
-* `StyledFooterAction`
-* `StyledClose`
-* `SIZE`
-* `ROLE`
+- `Modal`
+- `ModalHeader`
+- `ModalBody`
+- `ModalFooter`
+- `StyledRoot`
+- `StyledBackdrop`
+- `StyledDialog`
+- `StyledDialogInner`
+- `StyledHeader`
+- `StyledBody`
+- `StyledFooter`
+- `StyledFooterAction`
+- `StyledClose`
+- `SIZE`
+- `ROLE`
 
 ### <Modal/> API
 
-* `children: React.node | ({close: fn}) => React.node` - Required
+- `children: React.node | ({close: fn}) => React.node` - Required
   Modal content. The children-as-function API may be preferable for performance reasons (wont render until opened) and also gives users access to a `close()` callback that they can add to buttons
-* `isOpen: boolean` - Required
+- `isOpen: boolean` - Required
   Controls whether the modal is visible or not
-* `closeable: boolean` - Optional, Defaults to `true`
+- `closeable: boolean` - Optional, Defaults to `true`
   Whether the modal should be closeable by the user (either via escape, backdrop click, etc). You can set this to `false` if your modal has an action that the user _must_ take before closing.
-* `onClose: function` - Optional
-  A callback that is invoked when the modal will close
-* `size: SIZE.{small|medium|large|full|cover|auto}|string|number`, Defaults to SIZE.default
+- `onClose: (source: CloseSource) => void` - Optional
+  A callback that is invoked when the modal will close. Callback is passed a constant identifying what triggered the close.
+- `size: SIZE.{small|medium|large|full|cover|auto}|string|number`, Defaults to SIZE.default
   Controls the size of the modal (primarily width). Can be a SIZE constant or css `width` property value.
-* `aria-labelledby: string` - Optional, Default to ''
+- `aria-labelledby: string` - Optional, Default to ''
   If specified, will be passed to the dialog element
-* `aria-describedby: string` - Optional, Defaults to ''
+- `aria-describedby: string` - Optional, Defaults to ''
   If specified, will be passed to the dialog element
-* `aria-label: string` - Optional, Defaults to ''
+- `aria-label: string` - Optional, Defaults to ''
   If specified, will be passed to the dialog element (described the modal for screen readers)
-* `id: string` - Optional, Defaults to ''
+- `id: string` - Optional, Defaults to ''
   If specified, will be passed to the dialog element
-* `role: 'dialog'|'alertdialog'|string` - Optional, Defaults to 'dialog'
+- `role: 'dialog'|'alertdialog'|string` - Optional, Defaults to 'dialog'
   Which accessibility role this modal should have
-* `mountNode: HTMLElement | React.Ref` - Optional, Defaults to document.body
+- `mountNode: HTMLElement | React.Ref` - Optional, Defaults to document.body
   Where to mount the modal
-* `overrides: {Root, Backdrop, Dialog, DialogInner, Close}` - Optional
+- `overrides: {Root, Backdrop, Dialog, DialogInner, Close}` - Optional
   Overrides for presentational components. See "Presentational Components Props API" below.
-  * `[ComponentName]: ReactComponent | {props: {}, style: {}, component: ReactComponent}` - Optional
+  - `[ComponentName]: ReactComponent | {props: {}, style: {}, component: ReactComponent}` - Optional
 
 ### <ModalHeader/> API
 
 ModalHeader is just a styled component with spacing / typography
 
-* `$style: () => {} | {}` - Optional
+- `$style: () => {} | {}` - Optional
   Style overrides object or function
 
 ### <ModalBody/> API
 
 ModalBody is just a styled component with the correct spacing
 
-* `$style: () => {} | {}` - Optional
+- `$style: () => {} | {}` - Optional
   Style overrides object or function
 
 ### <ModalFooter/> API
 
 ModalBody is just a styled component with spacing and top border
 
-* `$style: () => {} | {}` - Optional
+- `$style: () => {} | {}` - Optional
   Style overrides object or function
 
 ### <ModalButton/> API
@@ -74,22 +74,26 @@ See <Button/> docs for available props and usage.
 
 ### SIZE Constant
 
-* `compact` - 400px width
-* `default` - 550px width
-* `full` - Span full viewport, minus some margin
-* `auto` - Based on content width
+- `compact` - 400px width
+- `default` - 550px width
+- `full` - Span full viewport, minus some margin
+- `auto` - Based on content width
 
-^ Open to ideas here. Too many options?
+### CLOSE_SOURCE Constant
+
+- `close_button`
+- `escape`
+- `backdrop`
 
 ### Presentational Components Props API
 
 Next properties are passed to every presentational (styled) component
 
-* `$isOpen: boolean`
-* `$size: string|number`
-* `$role: string`
-* `$closeable: boolean`
-* `$theme: theme`
+- `$isOpen: boolean`
+- `$size: string|number`
+- `$role: string`
+- `$closeable: boolean`
+- `$theme: theme`
 
 More props likely TBA as development continues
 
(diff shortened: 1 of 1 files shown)


rfcs/modal-component.md
```diff
@@ -0,0 +1,154 @@
...
+  If specified, will be passed to the dialog element
+* `aria-describedby: string` - Optional, Defaults to ''
+  If specified, will be passed to the dialog element
+* `aria-label: string` - Optional, Defaults to ''
+  If specified, will be passed to the dialog element (described the modal for screen readers)
+* `id: string` - Optional, Defaults to ''
+  If specified, will be passed to the dialog element
+* `role: 'dialog'|'alertdialog'|string` - Optional, Defaults to 'dialog'
+  Which accessibility role this modal should have
+* `mountNode: HTMLElement` - Optional, Defaults to document.body
```
blackswanny: may be more flexible if they provide DOM or React. Otherwise they will have to use ReactDOM to find dom element

blackswanny: are we going to use `Portals` to mount it somewhere else? Also if parent element is hidden (overlflow: hidden) than modal will not be visible as well.

rfcs/modal-component.md
```diff
@@ -0,0 +1,154 @@
...
+TBD
+
+### SIZE Constant
+
+* `small` - 400px width
+* `medium` - 550px width
+* `large` - 800px width
+* `full` - Span full viewport, minus some margin
+* `cover` - Span full viewport, no margin
+* `auto` - Based on content width
```
blackswanny: shall we define "scroll" of content type?

rfcs/modal-component.md
```diff
@@ -0,0 +1,154 @@
...
+
+TBD
+
+### SIZE Constant
+
+* `small` - 400px width
+* `medium` - 550px width
+* `large` - 800px width
+* `full` - Span full viewport, minus some margin
+* `cover` - Span full viewport, no margin
```
blackswanny: we may combine with `full` as it's just a matter of one style

rfcs/modal-component.md
```diff
@@ -0,0 +1,154 @@
...
+^ Open to ideas here. Too many options?
+
+### Presentational Components Props API
+
+Next properties are passed to every presentational (styled) component
+
+* `$isOpen: boolean`
+* `$size: string|number`
+* `$role: string`
+* `$closable: boolean`
```
blackswanny: what does this one mean?

rfcs/modal-component.md
```diff
@@ -0,0 +1,154 @@
...
+        <Button onClick={() => this.setState({open: true})}>Open</Button>
+        <Modal
+          isOpen={this.state.open}
+          close={() => this.setState({open: false})}
+        >
+          {({close}) => (
+            <>
+              <ModalTitle>Whatsup!</ModalTitle>
+              <ModalBody>This is a modal</ModalBody>
+              <ModalFooter>
```
blackswanny: footer and header may be parts of body. Otherwise we may need to sync their styles (margin\padding) in most cases.

rfcs/modal-component.md
```diff
@@ -0,0 +1,154 @@
...
+* `Escape` key will close the modal
+* Click on backdrop (anywhere outside dialog) will hide modal
+* Background will not be scrollable while modal is open (position: fixed)
+* Focus should ideally be transitioned back to last focused element upon closing
+
+### Possible Future Props
+
+The following props could be useful in the future but are low priority for V1
+
+* `onBackdropClick: function` - Callback handler for when backdrop is clicked specifically
```
blackswanny: imo, too many handlers. We may need only two (open\close) and provide reason\source of action in parameters of it

rfcs/modal-component.md
```diff
@@ -0,0 +1,154 @@
...
+* `backdrop: boolean` - Set to false to not show a transparent backdrop
+* `backdropCloseable: boolean` - Set to false if click on backdrop should not close modal
+* `onEscapeKey: function` - Callback handler for when escape key is hit
+* `escakeKeyCloseable: boolean` - Set to false to disable escape key closing modal
+* `onOpenComplete: function` - handler after opening (after animation, etc)
+* `onCloseComplete: function` - handler after closing (after animation, etc)
+* `autofocus: boolean` - Set to false if modal shouldn't autofocus on its content
+* `enforceFocus: boolean` - Set to false if we should force focus to stay in modal
+* `usePortal: boolean` - Set to false to not use a portal, and just mount as direct child
```
blackswanny: if mount node is provided, I think it's enough condition to use Portal or not

rfcs/modal-component.md
schnerd: Yes, by default we will use a portal attached to document.body, so `overflow: hidden` shouldn't be a problem.

Accepting a React.Ref here could be a good idea, will consider that.

rfcs/modal-component.md
schnerd: so people will just use overrides and set `margin: 0`? that could work.

rfcs/modal-component.md
schnerd: can you clarify the question? the content will scroll vertically if it needs to.

rfcs/modal-component.md
schnerd: Able to be closed. From above:

Whether the modal should be closable by the user (either via escape, backdrop click, etc). You can set this to `false` if your modal has an action that the user _must_ take before closing.

rfcs/modal-component.md
schnerd: There will be another wrapper element internally (StyledDialog or StyledDialogInner) that has the outer margin/padding to create spacing around the content. ModalTitle may have a marginBottom, and ModalFooter may have a marginTop, but otherwise these components won't have many styles.

rfcs/modal-component.md
schnerd: Good point

rfcs/modal-component.md
schnerd: If `closeable` is false, then using `onClose` to detect backdrop clicks doesn't make sense. Either way, these extra props here won't be in the initial version, just food for thought for the future.

rfcs/modal-component.md
blackswanny: yes

rfcs/modal-component.md
blackswanny: sometimes they want content-based width and \height, however if content is not known sometimes (most CMS systems) then they want it to be scrolled either.

schnerd: If someone wants more control over scroll they could probably override styles and set explicit width/height and `overflow: auto` where applicable.

rfcs/modal-component.md
blackswanny: ok. Makes sense. Usually there are 3 ways to close it: backdrop, X button and user button. Btw, user action may need to be promise in case user needs to send ajax and may not close modal based on result of call

schnerd: For the ajax case – generally they'll just listen for `onClick` on the submit button, fire their XHR, and then set `state.isModalOpen = false`, which they then pass to `isOpen={false}`.

See also idea #2 in the PR description.

If you have other ideas on how a promise-based API might look, can you write some example code?

rfcs/modal-component.md
ardok: I think this looks good.
Some designers like to have a line separator that spans from one end to the other between the actual body and footer.
I assume I'd be able to "sneak" this line separator (`div`) between `ModalBody` and `ModalFooter` ?
```
<ModayBody>...</ModalBody>
<div style={{ ...some thin horizontal line styles...}} />
<ModalFooter>...</ModalFooter>
```

rfcs/modal-component.md
```diff
@@ -0,0 +1,154 @@
...
+      </div>
+
+    );
+  }
+}
+```
+
+### Key Design Decisions
+
+* Modal content will not render server side. This is because portals don't exist server-side, and realistically your page shouldn't be loading with a modal open.
```
ardok: Wouldn't there be a case when you actually want to do it?
You get some user state from the server, if something matches some condition, open the modal on that page?
Or like the "What's New" modal?

rfcs/modal-component.md
```diff
@@ -0,0 +1,154 @@
...
+* `ModalBody`
+* `ModalFooter`
+* `StyledRoot`
+* `StyledBackdrop`
+* `StyledDialog`
+* `StyledDialogInner`
+* `StyledHeader`
+* `StyledBody`
+* `StyledFooter`
+* `StyledFooterAction`
```
ardok: I assume this is the container to put those "action" buttons on the right side?
So, these 2 buttons will be rendered side by side on the right side?
```
<StyledFooterAction>
  <button>...</button>
  <button>...</button>
</StyledFooterAction>
```

rfcs/modal-component.md
schnerd: I suppose there may be valid cases for it. The question is how do you implement this safely? Normally you want to add modals as a direct child of (or close to) document.body to avoid [z-index conflicts with other elements on the page](https://codepen.io/anon/pen/qyGZqN), but you can't use portals server side. And if you change where it's mounted on the client, you get the almighty `React attempted to reuse markup in a container but the checksum was invalid` error.

Open to ideas if you have any.

rfcs/modal-component.md
schnerd: I was thinking StyledFooter would be a flex box with `justify-content: flex-end`, and then StyledFooterAction would wrap _each_ button to create a margin between them.

rfcs/modal-component.md
ardok: Eh, not sure.
One thing come to my mind is to use `componentDidMount` ?

```
constructor(props) {
  super(props);
  this.state = {
    isMounted: false,
  };
}

componentDidMount() {
  this.setState({isMounted: true});
}

render() {
  return this.state.isMounted && <SomePortalElement />;
}
```

Not sure whether it will give you that almighty error?

abmai:
Might be a bit late to the party here, I like option `1B` - but what if we take it a step further?

```js
<Modal isOpen={this.state.open}>
  <ModalHeader>Hello world!</ModalHeader>
  <ModalBody>Lorem ipsum dolor sit amet</ModalBody>
  <ModalFooter>
    <ModalButton type="primary" ... />
    <ModalButton type="secondary" ... />
  </ModalFooter>
</Modal>
```

Since the convention here is already to use the individual pieces together, it might be slightly odd to expect `Footer` to take an `actions` array. Thoughts?

I also like the antd approach to rendering the modal, though. If we're already going to rely on `Portals` it might make more sense to go with their approach.

rfcs/modal-component.md
schnerd: >If portals disabled server-side by React for some reason, I think we should try to do it via DOM itself, although it's outside of React world a little.

It's not that "portals" are disabled. It's just that no document.body exists on the server since it's just rendering the root react component to a string. Can you clarify what you mean by "we should try to do it via DOM itself"?

blackswanny: I thought you just cant use Portal at all there. So it's either goes to stringified version on server or will be mounted later via client-side scripts. Now I am confused and try to identify problem here.
By DOM I meant something like 
```js
ReactDOM.render(modal, document.body);
```

schnerd: Yeah I guess you can't use portal there at all because it requires a dom node to mount into.

Code like the following is what causes the server/client mismatch error:
```
render() {
  if (__IS_BROWSER__) {
    return ReactDOM.createPortal(this.renderModal(), document.body);
  }
  return this.renderModal();
}
```

Ardo's example code is likely the approach we should take, however since componentDidMount isn't called on the server, the modal still will not render on the server.

I don't love the idea of rendering as a direct child on the server, and then updating to a portal on the client, it feels like there could be z-index conflicts or jank before/after the switch.

My vote is that we just support client-rendering for now, and then when a customer requests support for server-side rendering, we add extra props/styles where necessary to facilitate this.

schnerd:
@abmai That's a good point. Having the `actions` array was mainly so we could map to buttons that automatically have margin between them, but your `ModalButton` solves that I think.

With 1b the Modal import statement becomes quite a bit gnarlier:
```
import {Modal, ModalHeader, ModalBody, ModalFooter, ModalButton} from '@uber/baseui/modal'
```
but I think it's still acceptable.

Based on the feedback it sounds like 1b is the way to go, possible also supporting the antd-style API. If anyone is opposed to this please share thoughts, otherwise that's what we'll go with.

rfcs/modal-component.md
blackswanny: Ok, so the issue is that in Portals API there is no "stringify and append to <body> tag" method for serverside rendering. So Ardo's will work, but it's almost the same as to say "if __BROWSER__". Can we try to apply decorator here only for serverside? So it will silently amend logic and modal code itself will not be affected?

https://medium.com/@gigobyte/enhancing-react-components-with-decorators-441320e8606a

benshope:
Helping users make stateless components is a great thing to do because once a component becomes stateful, it is likely that users will tack on more state and not refactor into smaller components.  State (`isOpen` in this case) is the single broken window in the neighborhood that starts a graffiti and crime spree (more state and large components).  Is there anything that `1b` can do that `antd-style` cannot?   I would be inclined to expose just `antd` if that's possible.

schnerd:
> Is there anything that 1b can do that antd-style cannot? I would be inclined to expose just antd if that's possible.

Ben and I chatted in person, but to keep the discussion going here...
The main downside of the antd approach I can think of is the user likely has to remember to clean up their modal by calling `destroy()` at some point, and they probably need to do it in componentWillUnmount as well as a safety net for if the parent unmounts. Ben is going to experiment and see if there's a way to avoid that.

Some people may just prefer the declarative API as well 🤷‍♂️ , and we'd probably use it internally anyway so might as well export it. In the docs we could pick which API we encourage people to use.

rfcs/modal-component.md
ardok: @schnerd just found this:
https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/NoSsr/NoSsr.js

Similar concept, that is display it on componentDidMount

schnerd: Nice, that looks like a good approach.

Warbring3r:
Throwing in a vote from Uber Freight, we could definitely use this module. Any ETA or can we help get this landed?

I'm voting for the declarative API ("isOpen={isOpen}") because that's the standard React way and there's no need to remember calling .destroy().

schnerd:
@Warbring3r Implementation (of the declarative API at least) is about 50% done, hoping to have this reviewed, landed, published at some point next week. 

@abmai @blackswanny can I get a final review and approval or comments? I've addressed pretty much all the comments and updated the RFC.

rfcs/modal-component.md
```diff
@@ -0,0 +1,169 @@
...
+* `SIZE`
+* `ROLE`
+
+### <Modal/> API
+
+* `children: React.node | ({close: fn}) => React.node` - Required
+  Modal content. The children-as-function API may be preferable for performance reasons (wont render until opened) and also gives users access to a `close()` callback that they can add to buttons
+* `isOpen: boolean` - Required
+  Controls whether the modal is visible or not
+* `closeable: boolean` - Optional, Defaults to `true`
```
blackswanny: the modal is always closable. So little confusing. We can combine custom action and default actions (close button, X, backdrop) to allow user to choose. Action can be provided with default close handler and user can just call it, or do some other things in it.

schnerd: There may be use cases where the user *must* take an action before they can close the modal–for example, authenticating before accessing a page. By setting closeable to false, this removes the close button, and prevents backdrop from being clickable.

blackswanny: correct. I mentioned my code example, where if we click on X or backdrop nothing happens by default, but only if user does custom action (async or not) and calls default `close` method intentionally whenever user needs. That saves some extra properties and logic

schnerd: ahhh okay I missed your code example. here it is for context:

```
onCloseClick(close, source) => {
   switch (source) {
     case 'ESCAPE': close(result);
     case 'BACKDROP': //do something else or nothing
   }
}
```

I agree, instead of having onBackdropClick, onEscapePress, etc, we can just have keep the single `onClose` and pass a source attribute.

That said, we still need a way to hide the close button, right? No point in showing it if clicking it does nothing. That's what `closeable` is for.

blackswanny: ok, let's leave it for this purpose

PR #124, opened 2018-08-13
docs(readme): remove styled() from default example

schnerd:
I don't know if we want to promote using `styled` from baseui (at least in the primary setup example). I think it could create confusion about when to use `styled` from `fusion-plugin-styletron-react` vs `styletron-react` vs `baseui`. 

It seems like generally we'd want to encourage people to use the one they directly depend upon in their app. The only time it might make sense to use `styled` from baseui is if you:
1. need baseui theme values (though they could wire these up to their own styled impl)
2. are passing a new styled component through overrides and want to maintain it's ability to accept style override objects through `$style`. 

We could probably document these use cases in a more advanced guide though.

Curious to hear others thoughts about this though. @rtsao @nadiia @abmai @gergelyke

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,22 +17,13 @@ npm install @uber/baseui
 ```
 
 ```javascript
-import {LightTheme, ThemeProvider, styled} from '@uber/baseui';
+import {LightTheme, ThemeProvider} from '@uber/baseui';
 import {StatefulInput} from '@uber/baseui/input';
-
-const Centered = styled('div', {
-  display: 'flex',
-  justifyContent: 'center',
-  alignItems: 'center',
-  height: '100%',
-});
 
 export default function Hello() {
   return (
     <ThemeProvider theme={LightTheme}>
-      <Centered>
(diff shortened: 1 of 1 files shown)


nadiia:
I'd guess it's more common to reuse theme values while creating an in-app styled component, and it can be more confusing is two `styled` functions are used all over the place. I think it's ok to promote the baseui's `styled` wrapper as a way to go in order to get the benefits of the ThemeProvider. We may need to explain it a little better in the example.

nadiia:
Also, it's a good example for the question "How do I access the theme values for my custom components?"

schnerd:
You both raise good points–I guess people will maybe want to access theme values when building out other parts of their app (colors, fonts, spacing, etc), even if they aren't using components directly, in which case it probably does make sense to use baseui's `styled`.

Maybe for now we can leave this here and we'll remember to write documentation for the trade-offs of which `styled` you use. Thanks for the replies.

PR #125, opened 2018-08-14
feat: form control component that adds label and caption

nadiia:
Implements the `FormControl` component for adding label and caption to form elements.

- [x] Main component
- [x] Exported styled components
- [x] Unit tests
- [x] Examples for storybook
- [x] Design review


src/components/form-control/form-control.js
```diff
@@ -0,0 +1,97 @@
...
+      error,
+      children,
+    } = this.props;
+
+    const onlyChildProps = React.Children.only(children).props;
+    const sharedProps = mapStyletronProps(
+      onlyChildProps,
+      STYLETRON_PROP_MAPPER,
+    );
+    sharedProps.$error = this.props.error || sharedProps.$error;
```
schnerd: It seems like this solution turned out pretty elegant. 👍

src/components/form-control/form-control.js
```diff
@@ -0,0 +1,97 @@
...
+          {(caption || error) && (
+            <Caption {...sharedProps} {...getOverrideProps(CaptionOverride)}>
+              {error && typeof error !== 'boolean'
+                ? typeof error === 'function' ? error(sharedProps) : error
+                : typeof caption === 'function'
+                  ? caption(sharedProps)
+                  : caption}
+            </Caption>
+          )}
+        </StyledControlContainer>
```
schnerd: It looks like StyledControlContainer has a margin-bottom by default. I think this makes sense, but we definitely want to make it easy to remove this margin if necessary, and I think the best way to do that is expose the Container via overrides.

Also curious, why leave label outside the container, but put caption inside? Maybe it makes sense to put both inside and just make the Container the root element (instead of using fragment)

schnerd:
@nguyensomniac do we have designs for what a "required" form item looks like? (asterisk, red dot, etc).

src/components/form-control/__tests__/__snapshots__/form-control.test.js.snap
```diff
@@ -0,0 +1,143 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
abmai: Should we use snapshots for these? In my experience, over time it could get tedious to review snapshots. It looks like most of these are very similar in structure as well, which would require clear context to review (i.e. what are `correct props` and `correct error value`).

Would it be clearer to just write assertion for the specific case we're catching?

src/components/form-control/styled-components.js
```diff
@@ -0,0 +1,69 @@
...
+*/
+// @flow
+import {styled} from '../../styles';
+
+export const Label = styled('label', props => {
+  const {$disabled, $theme: {colors, sizing, typography}} = props;
+  return {
+    ...typography.font350,
+    fontWeight: 500,
+    color: $disabled ? colors.mono700 : colors.mono1000,
```
nguyensomniac: For labels, cursor should be `not-allowed` when `$disabled`

src/components/form-control/form-control.js
nadiia: Oh, it definitely has to be a part of overrides! I added it last minute and totally missed to add to the overrides. The label is left outside in order to make it easier to move label around, for example, place it on the left while a caption always aligned with it's control.

nguyensomniac:
Nailed the presentation, great job! ✨

Could you add an example for the `htmlFor` prop for labels? Wanted to make sure it works as expected.

Also, will we be able to use these for individual control labels for checkboxes and radios too?

nadiia:
@nguyensomniac I added a story example for all controls we currently have: input, textarea, checkbox and radio group. I'll add one for a custom id/htmlFor prop.

nguyensomniac:
Yup—but looks like the labels inside the individual components are unchanged. If we can use `StyledLabel` from here for those labels I'd advocate for those being the default.

Also, any way we could have the behavior that HTML `for` enables [(see example)](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label) as the default when you use these for labels/textareas/the like?

nguyensomniac:
Talking about the ones in red, btw.

frame 15

src/components/form-control/form-control.js
schnerd: I suppose the discussion in #131 is actually sort of relevant to this. 

There may be cases where someone may want to add a wrapper element between their FormControl and Input element, or if someone wants two inputs side-by-side under the same label (credit card + cvv code, etc). React context doesn't help us here, but maybe for those cases maybe we still want to support optioanlly passing `disabled`, `size`, `required`, etc, to the top-level form-control?

PR #132, opened 2018-08-17
feat(menu): add profile list to Menu component; tweak styles

abmai:
Added Profile List to Menu component. Some outstanding items to be done below, but otherwise good for review.

* [x] Initial Component Implementation
* [x] Highlight Style


src/components/menu/menu-profile.js
```diff
@@ -0,0 +1,77 @@
...
+        // $FlowFixMe
+        const {key, ...itemProps} = mapStyletronProps(requiredProps);
+        const ItemImg = getProfileItemImg(item);
+        const {title, subtitle, body} = getProfileItemLabels(item);
+        return (
+          // Need to be explicit with `key` otherwise eslint throws error?
+          <ListItemProfile key={key} {...itemProps}>
+            <ProfileImgContainer>
+              {ItemImg &&
+                (typeof ItemImg === 'string' ? (
```
abmai: How do we better detect that the user is passing in a React component vs. just an Image URL here? There are libraries out there that we could use, or we could roll our own?

src/components/menu/menu.js
```diff
@@ -24,11 +23,9 @@ export default function Menu({
   return (
     <List $ref={rootRef}>
       {items.map((item, index) => {
+        const requiredProps = getRequiredItemProps(item, index);
         // $FlowFixMe
-        const {key, ...itemProps} = mapStyletronProps(
-          getRequiredItemProps(item, index),
-          STYLETRON_PROP_MAPPER,
-        );
+        const {key, ...itemProps} = mapStyletronProps(requiredProps);
```
blackswanny: do we still want styletron to be optional in base ui? if so, we should mention it only in `styled-components` file

src/components/menu/menu-profile.js
blackswanny: React.isValidElement(component)

abmai: I think that only tests for elements (i.e. after JSX has been parsed), in this case we're expecting a component class/function.

src/components/menu/menu.js
abmai: good call! i'm not sure if this mapper is needed anymore tbh, how are we handling passing of styletron props down to styled components elsewhere - are we just hand mapping say `isHighlighted` to `$isHighlighted`?

src/components/menu/menu-profile.js
schnerd: This seems like a good enough check to me? `typeof ItemImg === 'function'` might be an okay check for component since that covers stateless functional components and classes, but I've had trouble getting flow to accept that on occasion.

src/components/menu/menu-profile.js
```diff
@@ -0,0 +1,77 @@
...
+              {title && <ProfileTitle>{title}</ProfileTitle>}
+              {subtitle && <ProfileSubtitle>{subtitle}</ProfileSubtitle>}
+              {body && <ProfileBody>{body}</ProfileBody>}
+            </ProfileLabelsContainer>
+          </ListItemProfile>
+        );
+      })}
+    </ListProfile>
+  );
+}
```
schnerd: I could be missing something, but it seems like instead of exporting a `MenuProfile` we should just be exporting a `MenuProfileOption` that renders each row, and people pass that class to Menu/Select via overrides or a render prop.

```js
// Most people will be using Menu via Select (I think?)
import {Select, SelectOptionProfile} from 'baseui/select';

function ProfileOption(props) {
  return (
    <SelectOptionProfile
      {...props} 
      imageSrc={() => props.item.imageSrc}
      label={() => props.item.label}
      sublabel={() => props.item.subtitle}
    />
  );
}

export () => (
<Menu items={users} overrides={{Option: ProfileOption}} />
);
```

This might be more in line with how an end user would think about customizing the layout of a select/menu option.

But again, I might be missing something. How were you thinking the code would look for end users to utilize MenuProfile when using Select?

src/components/menu/menu.js
schnerd: good point oleg, i'd agree styletron should ideally only be mentioned in styled.js. That said, you could still have this function and rename it to something unbranded like `mapStyleProps`.

I've just been doing the mapping manually:
```js
getSharedProps(): $Diff<SharedStylePropsArgT, {children: React.Node}> {
    const {isOpen, showArrow} = this.props;
    const {isAnimating, arrowOffset, popoverOffset, placement} = this.state;
    return {
      $showArrow: Boolean(showArrow),
      $arrowOffset: arrowOffset,
      $popoverOffset: popoverOffset,
      $placement: placement,
      $isAnimating: isAnimating,
      $isOpen: isOpen,
    };
}
```
But no reason _not_ to use a helper function if it makes sense for a given component.

src/components/menu/stateful-menu-profile.js
```diff
@@ -0,0 +1,39 @@
...
+            getProfileItemLabels={getProfileItemLabels}
+            getProfileItemImg={getProfileItemImg}
+            getProfileItemImgText={getProfileItemImgText}
+            overrides={overrides}
+          />
+        )}
+      </StatefulContainer>
+    );
+  }
+}
```
schnerd: Similar to comment above, it seems kind of redundant to have both a StatefulMenu and StatefulMenuProfile. Why not just StatefulMenu with a custom option renderer?

src/components/menu/menu-profile.js
blackswanny: look here, it's oficial
https://www.npmjs.com/package/react-is

src/components/menu/styled-components.js
```diff
@@ -11,34 +11,88 @@ type StyledListItemPropsT = {
...
     cursor: 'pointer',
     transitionProperty: 'color',
     transitionDuration: $theme.animation.timing100,
     transitionTimingFunction: $theme.animation.easeOutCurve,
     ':hover': {
       color: $theme.colors.primary400,
     },
+    ':not(:last-child)': {
+      marginBottom: $theme.sizing.scale600,
+    },
```
schnerd: Should we use padding instead of margin for these? Otherwise there will be space in-between items where I can't click to select the item, right?

src/components/menu/menu-profile.js
```diff
@@ -0,0 +1,77 @@
...
+  const ProfileSubtitle =
+    getOverride(overrides.ProfileSubtitle) || StyledProfileSubtitle;
+  const ProfileBody = getOverride(overrides.ProfileBody) || StyledProfileBody;
+
+  return (
+    <ListProfile $ref={rootRef}>
+      {items.map((item, index) => {
+        const requiredProps = getRequiredItemProps(item, index);
+        // $FlowFixMe
+        const {key, ...itemProps} = mapStyletronProps(requiredProps);
```
schnerd: I think $FlowFixMe gets stripped in our dist, so consumers might end up getting warnings for our internal flow errors (at least that's what was happening in #110). Any chance we can avoid the FlowFixMe here?

src/components/menu/menu-profile.js
schnerd: Nice, hadn't seen that

src/components/menu/styled-components.js
```diff
@@ -11,34 +11,88 @@ type StyledListItemPropsT = {
...
+}));
+
+export const ListItemProfile = styled('li', {
+  position: 'relative',
+  display: 'flex',
+  alignItems: 'center',
+  ':not(:last-child)': {
+    marginBottom: '18px',
+  },
+});
```
schnerd: Probably already on your to-do list, but it seems like we're missing hover styles and cursor: pointer for these options.

src/components/menu/menu-profile.js
abmai: Ooh, hadn't seen that either. Thanks, @blackswanny

src/components/menu/menu.js
abmai: It seems to me that these functions all basically do a mapping of `{someKey: value}` => `{$someKey: value}`. Should we make this a helper?

src/components/menu/menu-profile.js
abmai: They would import `MenuProfile` instead of `Menu`.

I had thought about the approach you mentioned, but I couldn't work out exactly how people would in-turn override the internals of `MenuProfile` without chaining multiple `overrides` props.

src/components/menu/styled-components.js
abmai: Yeah I need to sync with @nguyensomniac on that.

src/components/menu/menu-profile.js
abmai: Oh interesting, will update this code. Thanks!

src/components/menu/styled-components.js
abmai: Good call!

src/components/menu/menu-profile.js
abmai: Hm, with `react-is`, this code seems overly verbose. Any potential issue with leaving this as-is and let React errors take care of the last mile (i.e. if users don't pass in a valid React component React will just error out anyway).

src/components/menu/menu-profile.js
schnerd: >They would import MenuProfile instead of Menu.

Gotcha. If I'm using Select, do I have to import MenuProfile and then pass MenuProfile in as an override?

>I had thought about the approach you mentioned, but I couldn't work out exactly how people would in-turn override the internals of MenuProfile without chaining multiple overrides props.

Not sure if this helps, but tooltip has an example of passing along overrides to another nested component (it uses Popover internally) using the mergeOverrides helper:
https://github.com/uber-web/baseui/blob/master/src/components/tooltip/tooltip.js

Happy to jam on this together if it'd be helpful (i'm remote much of this week but could zoom).

schnerd: My hunch was that `MenuProfile` wouldn't exist, and we'd only export a `MenuProfileOption` that can be easily passed as an override to `Menu` or `Select`.

```js
<Menu overrides={{Option: MenuProfileOption}}>
```

If for some reason you need more control over mapping props or overrides, you could wrap MenuProfileOption:

```js
const CustomOptionRenderer = props => (
  <MenuProfileOption 
    {...props}
    label={props.option.user.name}
    imgSrc={props.option.user.avatar}
    overrides={{Label: {style: {color: 'red'}}}}
  />
);

<Menu overrides={{Option: CustomOptionRenderer}}>

// or with Select, which would just pass override to Menu:
<Select overrides={{Option: CustomOptionRenderer}}>
```

That's sort of the approach I was imagining, but there could definitely be some things I'm missing. Is there behavior that `MenuProfile` offers on top of this?

abmai: Yep, I was concerned about exactly whether that would be a cleaner approach vs. just exporting a different top-level component. It seemed odd to be defining a new component just to override the internals, but maybe I'm over thinking it.

I can actually code it out and see how it looks though because I agree that it would make more sense contextually.

schnerd: Gotcha, yeah I see your thinking – definitely feels a bit strange at first exporting something just for use inside of overrides. Might feel a bit more natural as a render prop like `renderOption={profileOptionRenderer}`, but alas, we've sort of gravitated towards overrides for doing these kinds of things for better or worse.

One other thought why this strikes me as the right approach – if we come up with other standard option layouts in the future like `MenuFooOption`, `MenuBarOption`, etc, it might be a little bit redundant to have to also develop and export `MenuFoo`, `StatefulMenuFoo`, `MenuBar`, `StatefulMenuBar`.

But anyways, definitely try it out – there may be things I haven't considered.

abmai:
Updated with the implementation changes discussed before

src/helpers/overrides.js
```diff
@@ -75,6 +75,21 @@ export function toObjectOverride<T>(override: OverrideT<T>): OverrideT<T> {
   return override;
 }
 
+/**
+ * Get a convenient override object that will always have {component, props}
+ */
+export function getOverrideObject<T>(
```
abmai: I added this because it seems like in most cases you'll need to pass-thru the `props` and `style` anyway. Thoughts here? @schnerd

schnerd: I suppose this does feel a bit cleaner. I'm a fan.

src/menu/stories.js
```diff
@@ -27,26 +44,108 @@ const ITEMS: Array<{label: string}> = [
...
+          },
+        },
+        Option: {
+          component: OptionProfile,
+          props: {
+            getProfileItemLabels: ({title, subtitle, body}) => ({
+              title,
+              subtitle,
+              body,
+            }),
```
schnerd: This sort of just looks like an identity function – is it just for demonstration purposes or is this actually required?

schnerd:
>Updated with the implementation changes discussed before. Seems like going all-in on overrides is not bad at all.

Woohoo! Glad to hear it worked, I think this simplifies things.

Can you update the RFC too to remove MenuProfile, etc?

Also it seems like the ProfileOption is missing cursor: pointer and hover styles. Thoughts on adding this?

src/menu/stories.js
abmai: It's required, in this case the input `item` has `title, subtitle, body` as object keys, but in other scenarios it could be

```js
getProfileItemLabels: ({name, jobTitle, selfDescription}) => ({title: name, subtitle: jobTitle, body: selfDescription})
```

src/helpers/overrides.js
abmai: Great, I'll add tests

abmai:
Still waiting on hover styles from @nguyensomniac - we might have to circle back to that so that this PR doesn't just sit here.

Maybe I can also go with a first pass + add cursor style so that we have something tho.

nguyensomniac:
Just talked about this with @schnerd, will update everything by eod

src/select/dropdown.js
```diff
@@ -60,48 +60,59 @@ const SelectDropDown = function(props: DropDownPropsT) {
...
-        List: listProps => (
-          <DropDown
-            {...listProps}
-            $rows={rows}
-            $type={type}
-            $isOpen={isDropDownOpen}
-          />
-        ),
-        ListItem: DropDownItem,
+        List: {
```
abmai: @blackswanny updated integration point for `Menu` here, please take a look

abmai:
@nguyensomniac can you help sign off on the current styles when you get a chance? It's missing the condensed style, but I'm planning to add that in a separate PR

nguyensomniac:
ooo, will take a look friday

abmai:
Oh oops, gotcha. I'll merge then and you can do a whole review once I have the condensed style too

abmai:
FYI this changes the API of Menu, so we should probably do a major version bump for the next one cc @gergelyke

PR #138, opened 2018-08-20
[docs] Update npm install command

kahwee:
Since `yarn add` saves, `npm install` should save too.

#### License

MIT

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ Add `baseui` to your project:
 yarn add baseui
 
 # using npm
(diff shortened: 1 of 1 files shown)


gergelyke:
npm automatically adds it to package.json too - https://docs.npmjs.com/cli/install

now you explicitly have to tell npm not to save with `--no-save: Prevents saving to dependencies.`

kahwee:
Got it. Thanks!

PR #149, opened 2018-08-23
feat(button): change type to kind

kahwee:
* refactor `$type` to `$kind`
* add compact `$size`
* more examples

I didn't get $kind and $size to work as expected. It seems that `styled` didn't persist the props from overridden BaseButton. Will appreciate if you can take a look.


src/components/button/button.js
```diff
@@ -14,14 +14,21 @@ export default function Button({
   startEnhancer,
   endEnhancer,
   overrides = {},
+  $size,
+  $kind,
```
schnerd: We only use the $ prefix on styled components. For the main exports you should just use normal props `size` and `kind`. When they are passed to "BaseButton", "ButtonLabel", etc, that's when we prefix with a dollar.

src/components/button/styled-components.js
```diff
@@ -1,30 +1,38 @@
...
+  ...$theme.typography[$size === BUTTON_SIZE.compact ? 'font200' : 'font300'],
   alignItems: 'center',
   justifyContent: 'space-between',
   border: 'none',
-  borderRadius: $type === BUTTON_TYPE.round ? '50%' : $theme.borders.radius200,
+  borderRadius: $kind === BUTTON_KIND.round ? '50%' : $theme.borders.radius200,
   outline: 'none',
   WebkitAppearance: 'none',
+  transitionDuration: $theme.animation.timing100,
+  transitionTimingFunction: $theme.animation.easeOutCurve,
```
schnerd: Should we be explicit about which properties are being transitioned? (background color, etc).

kahwee: Good point, I think it should only be the background color. I thought the immediate change of background color looks too harsh.

src/components/button/button.js
kahwee: Gotcha, updating to address this.

abmai:
Looks good! I think once this is merged you can just run with whatever's on `anh.button` and work on it. It's missing flow types (cuz I was speeding), and proper component overrides usage iirc.

Thanks, @kahwee!!

PR #155, opened 2018-08-28
Radio option should be disabled if the group or itself is disabled

nudge:
It is not possible to disable an option in a `RadioGroup`, this change correctly disables a radio option if the `RadioGroup` is disabled or itself is disabled.


gergelyke:
thanks for the contribution David!

@blackswanny can you take a look please?

gergelyke:
released as `1.0.2`

gergelyke:
thanks for the contrib @nudge 🎉

PR #158, opened 2018-08-28
feat: create toggle version of checkbox

blackswanny:

### DIFF
--- a/rfcs/checkbox-component.md
+++ b/rfcs/checkbox-component.md
@@ -2,72 +2,74 @@
 
 ### Exports
 
-* `StatefulCheckbox`
-* `StatefulCheckboxContainer`
-* `Checkbox`
-* `StyledRoot`
-* `StyledCheckmark`
-* `StyledLabel`
-* `StyledInput`
+- `StatefulCheckbox`
+- `StatefulCheckboxContainer`
+- `Checkbox`
+- `StyledRoot`
+- `StyledCheckmark`
+- `StyledLabel`
+- `StyledInput`
 
 ### `Checkbox` API
 
-* `checked: ?boolean`:
+- `checked: ?boolean`:
   check or uncheck the control. Default is `false`.
-* `autoFocus: boolean`:
+- `autoFocus: boolean`:
   make the control focused (active). Default is `false`
-* `children: React$Node`:
+- `checkmarkType: STYLE_TYPE.default | STYLE_TYPE.toggle`:
+  the style to be applied to a checkmark. Default is `STYLE_TYPE.default`
+- `children: React$Node`:
   Component or String value for label of checkbox. Default is empty string
-* `disabled: boolean`:
+- `disabled: boolean`:
   Disable control from being changed
-* `required: boolean`:
+- `required: boolean`:
   Mark control as required
-* `isError: boolean`:
+- `isError: boolean`:
   Sets control into isError state. Default is `false`
-* `isIndeterminate: boolean`:
+- `isIndeterminate: boolean`:
   Indeterminate control value. checked is ignored.
-* `overrides: {Root: (props: {[string]: any}) => React$Node, Checkmark: (props: {[string]: any}) => React$Node, Label: (props: {[string]: any}) => React$Node, , Input: (props: {[string]: any}) => React$Node}`
-  * `Label` to render. Optional.
-  * Custom `Checkmark` (exact control). If used, most of handlers may not work.
-  * `Root` wrapper element for the whole checkbox control to apply styles
-  * `Input` may replace original functionality of checkbox. If provided should follow the same interface and logic or may break the whole component
-* `labelPlacement: 'top' | 'right' | 'bottom' | 'left'`:
+- `overrides: {Root: (props: {[string]: any}) => React$Node, Checkmark: (props: {[string]: any}) => React$Node, Label: (props: {[string]: any}) => React$Node, , Input: (props: {[string]: any}) => React$Node}`
+  - `Label` to render. Optional.
+  - Custom `Checkmark` (exact control). If used, most of handlers may not work.
+  - `Root` wrapper element for the whole checkbox control to apply styles
+  - `Input` may replace original functionality of checkbox. If provided should follow the same interface and logic or may break the whole component
+- `labelPlacement: 'top' | 'right' | 'bottom' | 'left'`:
   How to position the label relative to the checkbox itself. Default is `right`
-* `onChange: func`:
+- `onChange: func`:
   handler for events on trigger element
-* `onMouseEnter: func`:
+- `onMouseEnter: func`:
   handler for events on trigger element
-* `onMouseLeave: func`:
+- `onMouseLeave: func`:
   handler for events on trigger element
-* `onFocus: func`:
+- `onFocus: func`:
   handler for events on trigger element
-* `onBlur: func`:
+- `onBlur: func`:
   handler for events on trigger element
 
 ### `StatefulCheckboxContainer` API
 
-* `initialState: {}`
+- `initialState: {}`
   Initial state of an uncontrolled popover component.
-  * `checked` - an initial checked state. Check or uncheck the control. Default is `false`.
-* `stateReducer: (type: text, nextState: {}, currentState: {}, e: any) => nextState`
+  - `checked` - an initial checked state. Check or uncheck the control. Default is `false`.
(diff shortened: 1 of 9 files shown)


src/components/checkbox/styled-components.js
```diff
@@ -127,7 +130,43 @@ export const Root = styled('label', props => {
...
+      top: '-8px',
+      transitionProperty: 'margin-left',
+      transitionDuration: animation.timing700,
+      transitionTimingFunction: animation.easeOutCurve,
+      marginLeft: $checked ? '16px' : '-4px',
+      content: `url('data:image/svg+xml;utf8,` + toggleSVG + `')`,
+      color: 'black',
+    },
+  };
+};
```
schnerd: It looks like the notch on the toggle handle is a little bit off-center:
image

Should be perfectly centered based on mocks:
image

schnerd: Text also looks like it's too close to the toggle element (when it's off). @nguyensomniac how many pixels of margin should we have here?

schnerd: Also, looks like shadow is still showing when control is disabled:
image

Mocks have the shadow removed here:
image

schnerd: Hover style in this diff seems to just change opacity:
image

I think its supposed to make the track and notch a darker blue color?
image

blackswanny: @nguyensomniac please, check the mocks. Seems like they are not consistent. The color of dash is not consistent with color of toggle itself. Shadow is not everywhere

src/components/checkbox/styled-components.js
```diff
@@ -163,6 +202,13 @@ export const Checkmark = styled('span', props => {
     marginLeft: $theme.sizing.scale200,
     marginRight: $theme.sizing.scale200,
   };
+};
+
+export const Checkmark = styled('span', props => {
+  const {$checkMarkType} = props;
```
schnerd: Should we be consistent with naming? The styled checkmark export is `Checkmark`, but the checkmark type is `checkMarkType`. Can we use `checkmarkType` instead?

src/components/checkbox/checkbox.js
```diff
@@ -96,10 +97,11 @@ class StatelessCheckbox extends React.Component<PropsT, StatelessStateT> {
   };
 
   render() {
+    const {checkMarkType} = this.props;
```
schnerd: Thoughts on just making this a boolean prop called `toggle`?

<Checkbox toggle/>

Seems like a slightly simpler API for end users

blackswanny: there are plenty of checkbox styles. Just because we have only two so far doesn't make it boolean

src/components/checkbox/styled-components.js
blackswanny: ok, sure

src/components/checkbox/styled-components.js
blackswanny: @schnerd fixed obvious issues with styles

rfcs/checkbox-component.md
```diff
@@ -2,72 +2,74 @@
...
 ### Exports
 
-* `StatefulCheckbox`
-* `StatefulCheckboxContainer`
-* `Checkbox`
-* `StyledRoot`
-* `StyledCheckmark`
-* `StyledLabel`
-* `StyledInput`
+- `StatefulCheckbox`
```
gergelyke: what's the reason for changing it from `*` to `-`?

src/components/checkbox/styled-components.js
schnerd: Nice, looks centered correctly now.

>Seems like they are not consistent. The color of dash is not consistent with color of toggle itself. Shadow is not everywhere

Maybe those are on purpose? In my opinion the disabled toggle looks better without shadow–pretty sure that's intentional.

Also, should things be aligned left in the flexbox?
image

After we resolve those I'm happy to stamp.

gergelyke:
should we add a storybook example too? probably folks will look for a Toggle component, and they may miss it under Checkbox

rfcs/checkbox-component.md
blackswanny: was fixed by lint

gergelyke: do we lint markdown files? it seems that all other RFCs follow the `*` syntax

src/components/checkbox/styled-components.js
blackswanny: In my experience usually label and toggle are left and right aligned in container given. I wait for @nguyensomniac to check this assumption

blackswanny:
@gergelyke there are 2 examples under checkbox group 
image

gergelyke:
Yeah, I get that - my questions is more about if we should add a new section for it? So people will see a top level Toggle element

src/components/checkbox/styled-components.js
nguyensomniac: Shadows shouldn't change. I'll investigate

nguyensomniac: Fyi, the color of the thumb becomes `primary400` when hovered/active, as in different from the color of the background

src/components/checkbox/styled-components.js
```diff
@@ -127,7 +130,43 @@ export const Root = styled('label', props => {
...
+    height: '16px',
+    marginTop: '4px',
+    marginBottom: '4px',
+    marginLeft: '4px',
+    marginRight: '4px',
+    ':after': {
+      position: 'absolute',
+      top: '-8px',
+      transitionProperty: 'margin-left',
+      transitionDuration: animation.timing700,
```
nguyensomniac: Let's make this faster. Maybe 400 here.

src/components/checkbox/styled-components.js
blackswanny: What is thumb? The small dash in center?
The small dash has no change in color (or almost no change) even if the slider color changes to gray. Is it correct?

src/components/checkbox/styled-components.js
blackswanny: ok

src/components/checkbox/styled-components.js
nguyensomniac: Yes, the dash in the center

blackswanny: fixed

gergelyke:
@nguyensomniac do we need to change anything else, or looks good as is?

src/components/checkbox/styled-components.js
```diff
@@ -127,7 +130,73 @@ export const Root = styled('label', props => {
...
+    `"/></g><defs><filter id="filter0_d" x="0" y="0" width="32" height="32" filterUnits="userSpaceOnUse" color-interpolation-filters="sRGB"><feFlood flood-opacity="0" result="BackgroundImageFix"/><feColorMatrix in="SourceAlpha" type="matrix" values="0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 127 0"/><feOffset dy="1"/><feGaussianBlur stdDeviation="2"/><feColorMatrix type="matrix" values="0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0.32 0"/><feBlend mode="normal" in2="BackgroundImageFix" result="effect1_dropShadow"/><feBlend mode="normal" in="SourceGraphic" in2="effect1_dropShadow" result="shape"/></filter></defs></svg>`;
+  return {
+    position: 'relative',
+    flex: '0 0 auto',
+    borderRadius: $theme.borders.useRoundedCorners
+      ? $theme.borders.radius200
+      : null,
+    display: 'inline-block',
+    backgroundColor: getBackgroundColor(props),
+    width: '40px',
```
nguyensomniac:
```
transitionDuration: animation.timing100,
transitionTimingFunction: animation.easeOutCurve,
```

for `backgroundColor` property

src/components/checkbox/styled-components.js
```diff
@@ -89,11 +92,11 @@ function getBackgroundColor(props) {
     }
   } else {
     if ($isActive || $isFocused) {
-      return colors.mono500;
+      return isToggle ? colors.mono800 : colors.mono500;
     } else if ($isHovered) {
       return colors.mono400;
```
nguyensomniac: `isToggle ? colors.mono700 : colors.mono400;`

nguyensomniac:
Two small tweaks—otherwise lgtm. 👍

PR #161, opened 2018-08-29
build(components): remove rollup

gergelyke:
https://github.com/uber-web/baseui/issues/104


README.md
```diff
@@ -22,7 +22,7 @@ npm install baseui
 
 ```javascript
 import {LightTheme, ThemeProvider, styled} from 'baseui';
-import {StatefulInput} from 'baseui/input';
+import {StatefulInput} from 'baseui/components/input';
```
gergelyke: note - because of this, we'll have to do a major release

package.json
```diff
@@ -52,6 +56,7 @@
     "babel-core": "^6.26.0",
     "babel-eslint": "^9.0.0",
     "babel-jest": "^23.4.0",
+    "babel-loader": "7",
```
gergelyke: will be able to update to the latest once we change to babel@7

README.md
abmai: This seems pretty long imo, is there no way we can keep `baseui/input`?

gergelyke: if we don't split themes/foundation/components into multiple packages, I don't think so

also, it is kinda consistent with what materiaul does: `import Button from '@material-ui/core/Button';`

schnerd: Haha, yeah bumping shouldn't be an issue this early in the project.

I did a quick search and surprisingly can't find any existing babel plugins or other tooling to move folders around and update relative paths accordingly. Either my google skills are lacking or something like this doesn't already exist and is sorely needed. I'd imagine we could write a plugin that does this, but I don't know if its worth the effort right now. Seems like we should just take the path of least resistance.

abmai: Maybe micro-optimizing but what if we rename the folder to `core` aswell. It's so much easier to type `core` vs. `components` lol.

Just trying to make it easier when eventually our customers' import lines become:

```js
import {Button} from 'baseui/components/button';
import {Input} from 'baseui/components/input';
import {Modal} from 'baseui/components/modal';
```

gergelyke: yeah. that's a good point, thanks Ryan! do we want to do that? @schnerd @abmai

abmai: By hoisting you mean adding the exports to `src/index.js` or physically moving the folders into `src`?

abmai: If all it takes is to move the exports, I'm all for it. Unless there are other reasons we don't want those in `src/index.js`.

rtsao: If you want:

```js
import {Input} from 'baseui/input';
import {Modal} from 'baseui/modal';
```

then moving directories should suffice. This might be nice for the case of something like:

```js
import * as button from 'baseui/button';
import * as input from 'baseui/input';
import * as modal {Modal} from 'baseui/modal';

<input.BaseButton/>
```

In this case it is convenient to import all related components but don't want to bundle the entirety of base ui.

With a single entry point, I don't think tree shaking is necessarily smart enough to prune unused properties of import star.

gergelyke: thanks! so based on this, would you be ok with essentially rolling back https://github.com/uber-web/baseui/commit/3a0310bf7fb3e314fa3ac0ae61a5a6cd833d7183? cc @abmai @schnerd 

as it seems, we added it because of rollup

abmai: Ahhh got it. @schnerd what was the original reason for moving the components under `/components`? Is it worth moving them back out to `/src` to save keystrokes on imports? What about renaming it to `/core`? Lol.

schnerd: I think we moved to `/components` for rollup. Moving things back to `/src` for now seems like the right move and won't impact customers. We can always revisit in the future.

gergelyke: updated ✌️

PR #164, opened 2018-08-30
feat(rfc): Buttons

kahwee:
Chatted with @abmai on this and I extracted the gist of a previous branch we collaborated on and added additional feedback from @schnerd (on `"shape"`).

Here's an overall look at how buttons would be used:

```javascript
import {
  Button, 
  StyledStartEnhancer, 
  StyledEndEnhancer, 
  SIZE, 
  KIND, 
  SHAPE
} from 'baseui/button';

export default () => <Button
  startEnhancer={<StyledStartEnhancer><SvgIcon /></StyledStartEnhancer>}
  endEnhancer={<StyledEndEnhancer><SvgIcon /></StyledEndEnhancer>}
  shape={SHAPE.round}
  kind={KIND.secondary}
  size={SIZE.compact}
  type="submit"
>Hello</Button>;
```

For `label`, we can use `StyledButtonLabel` as well which is essentially a div.

The DOM makeup of this is:

```
+------------------------------------------------------------------+
| BaseButton (display: flex; justifyContent: space-between)        |
| +-----------------+-----------------------+---------------------+|
| |  startEnhancer  |         label         |     endEnhancer     ||
| |  display=flex   |                       |     display=flex    ||
| |                 |                       |                     ||
| |                 |                       |                     ||
| |                 |                       |                     ||
| +-----------------+-----------------------+---------------------+|
+------------------------------------------------------------------+
```


rfcs/button-component.md
```diff
@@ -0,0 +1,81 @@
...
+### <Button/> API
+
+* `label: node | function` - Optional
+  A label in the button.
+* `startEnhancer: node | function` - Optional
+  A helper rendered at the start of the button.
+* `endEnhancer: node | function` - Optional
+  A helper rendered at the end of the button.
+* `disabled: boolean` - Optional
+  Defines if the button disabled.
```
schnerd: Should we list `kind`, `size`, `shape` as well as their defaults here?

We also usually list the `overrides` prop here (see other component RFCs.

rfcs/button-component.md
```diff
@@ -0,0 +1,81 @@
...
+* `StyledStartEnhancer`
+* `StyledEndEnhancer`
+* `KIND`
+* `SIZE`
+* `SHAPE`
+
+### <Button/> API
+
+* `label: node | function` - Optional
+  A label in the button.
```
schnerd: What's the thought behind using `label` instead of just using `children`? It seems like material, antd, and native HTML buttons all use `children`, should we just stick with that?

rfcs/button-component.md
schnerd: We should probably also make sure that people can pass through `type="submit"` to their buttons.

Actually it looks like there are a lot of other form/a11y related props people may need to pass through: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button

Maybe we should just pass through `...rest` props to the root element so users can pass whatever accessibility / aria attributes they might need?

rfcs/button-component.md
```diff
@@ -0,0 +1,81 @@
...
+
+### Usage
+
+Basic usage:
+
+```javascript
+import * as React from 'react';
+import {
+  Button, 
+  StyledStartEnhancer,
```
gergelyke: if we call it basic usage, should it be really basic, without enchancers, etc? it looks a bit scary if we reference it as basic, imho. how about adding a really basic one, and this?

rfcs/button-component.md
kahwee: Added!

kahwee: I agree to pass `...rest`. Additionally I imagine it's not uncommon for users to change the `StyledButton` to a link using `$as="a"` (Styletron's `$as`) and pass in `href` and `target`.

rfcs/button-component.md
kahwee: Good point! Thanks.

schnerd:
Looks good to me, thanks for taking feedback into account. 

One thing I think we definitely need, but we can punt on for now, is an `isLoading` prop that automatically displays a spinner in the endEnhancer. But we don't even have the spinner component yet so kind of a moot point right now.

kahwee:
@abmai, mind giving this a glance? Just in case you spot anything that didn't fit your use case.

PR #172, opened 2018-08-31
feat: create tag rfc

blackswanny:
#### Minor: add Tag RFC

### DIFF
--- a/rfcs/tag-component.md
+++ b/rfcs/tag-component.md
@@ -1,0 +1,53 @@
+# Tag Component
+
+## Exports
+
+* `Tag`
+* `StyledRoot`
+* `StyledAction`
+
+## `Tag` API
+
+* `color: STYLE.blue(theme.colors.primary470) | STYLE.orange(theme.colors.warning400) | STYLE.green(theme.colors.positive400) | STYLE.red(theme.colors.negative400)`:
+  the color theme to be applied to a Tag. Default is `STYLE.blue`.
+* `children: React$Node`:
+  Component or String value for label of tag. Default is empty string
+* `disabled: boolean`:
+  Disable control from being changed
+* `overrides: {Root: (props: {[string]: any}) => React$Node, Action: (props: {[string]: any}) => React$Node}`
+  * Custom `Action` button.
+  * `Root` wrapper element for the whole tag control to apply styles
+* `onActionClick: func(evt => void)`:
+  handler for events on Action button element (default is Cross).
+
+## Usage
+
+```js
+import {Tag, Root, Action} from './index';
+import {styled} from '../styles/index';
+
+const CustomRoot = styled(Root, {
+  textColor: 'red',
+});
+
+const CustomAction = styled(Action, {
+  backgroundColor: 'green',
+});
+
+export default () => {
+  return (
+    <div>
+      <Tag
+        onActionClick={event => {}}
+        disabled={false}
+        overrides={{
+          Root: CustomRoot,
+          Action: CustomAction,
+        }}
+      >
+        I am Tag
+      </Tag>
+    </div>
+  );
+};
+```


rfcs/tag-component.md
```diff
@@ -0,0 +1,96 @@
...
+
+- `StatefulTag`
+- `StatefulTagContainer`
+- `Tag`
+- `StyledRoot`
+- `StyledCross`
+
+### `Tag` API
+
+- `style: STYLE.blue | STYLE.orange | STYLE.green | STYLE.red`:
```
kahwee: `style` is a standard HTML attribute. Can we call it something else? 

Also, can we take the colors from the [theme](https://github.com/uber-web/baseui/blob/master/src/themes/light-theme.js)?

rfcs/tag-component.md
```diff
@@ -0,0 +1,96 @@
+# Tag Component
+
+### Exports
+
+- `StatefulTag`
```
kahwee: I'm curious what's your use case here. Why do you want to have states in a Tag.

schnerd: +1. Seems like a tag should be stateless.

rfcs/tag-component.md
```diff
@@ -0,0 +1,96 @@
...
+        disabled={false}
+        onMouseEnter={this.onTagHover}
+        onChange={this.onTagChange}
+      >
+        {childrenProps => {
+          return (
+            <Tag
+              {...childrenProps}
+              overrides={{
+                CustomRoot: CustomRoot,
```
kahwee: Do you mean `Root: CustomRoot`?

rfcs/tag-component.md
blackswanny: ok, called `color`

rfcs/tag-component.md
blackswanny: removed

rfcs/tag-component.md
blackswanny: yes

rfcs/tag-component.md
```diff
@@ -0,0 +1,61 @@
+# Tag Component
+
+### Exports
+
+- `Tag`
+- `StyledRoot`
+- `StyledCross`
```
schnerd: Thoughts on calling this `StyledAction`? I like the idea of calling it "Action" because in theory someone could override the action to be whatever they want – for example a pencil icon if they want an "Edit" modal to popup.

rfcs/tag-component.md
```diff
@@ -0,0 +1,61 @@
...
+  the color theme to be applied to a Tag. Default is `STYLE.blue`.
+- `children: React$Node`:
+  Component or String value for label of tag. Default is empty string
+- `disabled: boolean`:
+  Disable control from being changed
+- `overrides: {Root: (props: {[string]: any}) => React$Node, Cross: (props: {[string]: any}) => React$Node}`
+  - Custom `Cross` button.
+  - `Root` wrapper element for the whole tag control to apply styles
+- `onCrossClick: func(remove => {})`:
+  handler for events on Cross button element. `remove` argument method can be called any time to remove tag.
```
schnerd: How were you imagining `remove` works? I don't think the tag can remove itself, it's up to the parent component to remove the tag. 

Maybe we should just change this to `onActionClick: (evt) => void`?

rfcs/tag-component.md
```diff
@@ -0,0 +1,61 @@
...
+- `onMouseEnter: func`:
+  handler for events on trigger element
+- `onMouseLeave: func`:
+  handler for events on trigger element
+- `onFocus: func`:
+  handler for events on trigger element
+- `onBlur: func`:
+  handler for events on trigger element
```
schnerd: Is it worth making these a part of the explicit API? Technically the user can just do overrides for whatever event handler they want:

```js
<Tag overrides={{Root: {
  props: {onMouseEnter: myHandler}
}}}>
  Foo
</Tag>
```

abmai: +1 I think this isn't necessary and could just be pass-thru via props override

rfcs/tag-component.md
```diff
@@ -0,0 +1,61 @@
...
+
+### Exports
+
+- `Tag`
+- `StyledRoot`
+- `StyledCross`
+
+### `Tag` API
+
+- `color: STYLE.blue(theme.colors.primary470) | STYLE.orange(theme.colors.warning400) | STYLE.green(theme.colors.positive400) | STYLE.red(theme.colors.negative400)`:
```
abmai: Is it worth introducing semantics here rather than color names? i.e. `STYLE.info, STYLE.warning, STYLE.positive, STYLE.negative`

rfcs/tag-component.md
blackswanny: technically it's possible, but raises concerns. I will remove this
http://codetheory.in/remove-a-mounted-react-component-with-react-unmountcomponentatnode/

rfcs/tag-component.md
blackswanny: ok

rfcs/tag-component.md
blackswanny: ok

rfcs/tag-component.md
blackswanny: semantics allow user to choose predefined default styles rather than going to theme and trying to find corresponding color, which is not so clear

gergelyke:
can you update with the latest master to have markdownlinter running on this too?

rfcs/tag-component.md
```diff
@@ -0,0 +1,53 @@
...
+  Component or String value for label of tag. Default is empty string
+* `disabled: boolean`:
+  Disable control from being changed
+* `overrides: {Root: (props: {[string]: any}) => React$Node, Action: (props: {[string]: any}) => React$Node}`
+  * Custom `Action` button.
+  * `Root` wrapper element for the whole tag control to apply styles
+* `onActionClick: func(evt => void)`:
+  handler for events on Action button element (default is Cross).
+
+## Usage
```
nadiia: Looks like we put `Usage` before the exports and API for docs reasons

blackswanny: ok, will fix this

PR #173, opened 2018-08-31
docs(example): add readme docs

gergelyke:


kahwee:
You can do this instead:

```
  .addDecorator(withInfo)
  .add('Controlled and uncontrolled input', () => <ControlledInput />, {
    info: Readme,
  })
```

You're using the previous storybook addon pattern.

gergelyke:
yeah, this wont work sadly - for this we have to wait for version 4 (it is in alpha now)

gergelyke:
generally speaking, what do you think about the idea?

kahwee:
I'm assuming you mean the method where every sub-story has to load an `.md` file.

Did you do this to get the nice syntax highlighting in the `.md` files?

I am not a fan of having that many of these files to be honest. It gets unwieldy quickly as each story file (`stories.js`) can contain multiple sub stories and each of these sub stories would then have to load the relevant `.md` files.

PR #178, opened 2018-09-02
[RFC] Pagination Component

abmai:
Throwing this up for discussion, some things I really need help thinking through:

* Is it safe to assume that this component will always be dealing with numeric pages?
* Is it a good idea to encapsulate all dropdown/select logic to the `Select` component? i.e. if users want to override fault dropdown behavior (menu open), they will have to pass-thru props to `Select` via `overrides`.
* Do we export the used icons here as well? This component will need at least 3 default icons.

Screenshot of design for easier reference:

image

### DIFF
--- a/rfcs/pagination-component.md
+++ b/rfcs/pagination-component.md
@@ -21,13 +21,21 @@
   Defaults to "of"
 - `overrides: {PrevButton, NextButton, Select}` - Optional
   Presentational override props, all are `React.ComponentType<*> | {component, props, style}`
+- `onBackClick: (event) => void` - Optional
+  Callback for prev button click
+- `onNextClick: (event) => void` - Optional
+  Callback for next button click
 
 ### StatefulContainer API
 
 - `numPages: number` - Required
   Max number of pages
+- `stateReducer: (changeType, changes, currentState) => Object`
+  State reducer function
 - `initialState: {currentPage: number}` - Optional
   The initial state
+- `onPageChange: (nextPage, prevPage) => void` - Optional
+  Callback for when page changes
 
 ### StatefulPagination API
 
@@ -37,6 +45,10 @@
   Defaults to "of"
 - `overrides: {PrevButton, NextButton, Select}` - Optional
   Presentational override props, all are `React.ComponentType<*> | {component, props, style}`
+- `onBackClick: (event) => void` - Optional
+  Callback for prev button click
+- `onNextClick: (event) => void` - Optional
+  Callback for next button click
 
 ### Usage
 
--- a/rfcs/pagination-component.md
+++ b/rfcs/pagination-component.md
@@ -6,7 +6,7 @@
 - `StatefulPagination`
 - `StatefulContainer`
 - `StyledRoot`
-- `StyledBackButton`
+- `StyledPrevButton`
 - `StyledNextButton`
 - `StyledMaxLabel`
 - `STATE_CHANGE_TYPE`
@@ -45,7 +45,7 @@
   Defaults to "of"
 - `overrides: {PrevButton, NextButton, Select}` - Optional
   Presentational override props, all are `React.ComponentType<*> | {component, props, style}`
-- `onBackClick: (event) => void` - Optional
+- `onPrevClick: (event) => void` - Optional
   Callback for prev button click
 - `onNextClick: (event) => void` - Optional
   Callback for next button click
--- a/rfcs/pagination-component.md
+++ b/rfcs/pagination-component.md
@@ -56,7 +56,7 @@ Basic usage:
 
 ```javascript
 import * as React from 'react';
-import {StatefulPagination} from 'baseui/paginate';
+import {StatefulPagination as Pagination} from 'baseui/paginate';
 
 export default () => <Pagination numPages={6} />;
 ```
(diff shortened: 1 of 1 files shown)


rfcs/pagination-component.md
```diff
@@ -0,0 +1,55 @@
...
+- `prepositionLabel: string` - Optional
+  Defaults to "of"
+- `overrides: {PrevButton, NextButton, Select}` - Optional
+  Presentational override props, all are `React.ComponentType<*> | {component, props, style}`
+
+### StatefulContainer API
+
+- `numPages: number` - Required
+  Max number of pages
+- `initialState: {currentPage: number}` - Optional
```
gergelyke: do we want to add props like

* onChange (page change callback)
* pageSize (items per page)?

I was looking at https://github.com/react-component/pagination#api

abmai: I knew I missed something ;). Updated to add `onPageChange` handler as well as a couple of other ones.

I don't think we should handle complex situations like `pageSize` for now since the current design doesn't support it. My original thinking was that this component's responsibilities are:

* Show current and max number of pages
* Allow users to navigate 1 back/forward (buttons) or arbitrary page (dropdown)

How the application hooks into that behavior is up to the user. Thoughts?

rfcs/pagination-component.md
```diff
@@ -0,0 +1,67 @@
...
+`StatefulContainer` API and the following
+
+- `prepositionLabel: string` - Optional
+  Defaults to "of"
+- `overrides: {PrevButton, NextButton, Select}` - Optional
+  Presentational override props, all are `React.ComponentType<*> | {component, props, style}`
+- `onBackClick: (event) => void` - Optional
+  Callback for prev button click
+- `onNextClick: (event) => void` - Optional
+  Callback for next button click
```
schnerd: Should StatefulPagination have `onPageChange` also? (Instead of, or in addition to onBackClick/onNextClick)

abmai: Yep, it has everything that `StatefulContainer` has

rfcs/pagination-component.md
```diff
@@ -0,0 +1,70 @@
...
+- `STATE_CHANGE_TYPE`
+
+### Pagination API
+
+- `numPages: number` - Required
+  Max number of pages
+- `currentPage: number` - Required
+  The current page
+- `prepositionLabel: string` - Optional
+  Defaults to "of"
```
schnerd: One thing we haven't thought a ton about is internationalization. In theory with `overrides` you can change everything, but it seems like having props like the one you have here would be very useful for that. Maybe we should extend this to allow customizing labels for "Prev" and "Next"?

[Atlaskit has an i18n object for this](https://atlaskit.atlassian.com/packages/core/pagination). It seems like a `labels: {next, prev, preposition}` object might be more generic. Or we could stay with `prepositionLabel` and add `prevLabel`, `nextLabel`. Thoughts?

rfcs/pagination-component.md
```diff
@@ -0,0 +1,70 @@
...
+- `overrides: {PrevButton, NextButton, Select}` - Optional
+  Presentational override props, all are `React.ComponentType<*> | {component, props, style}`
+- `onBackClick: (event) => void` - Optional
+  Callback for prev button click
```
schnerd: nit: Should we use `onPrevClick` so terminology is consistent with the default button label?

schnerd:
> Is it safe to assume that this component will always be dealing with numeric pages?

I think so. We can probably revisit this in the future if a customer needs it.

> Is it a good idea to encapsulate all dropdown/select logic to the Select component? i.e. if users want to override fault dropdown behavior (menu open), they will have to pass-thru props to Select via overrides.

I think this makes sense.

> Do we export the used icons here as well? This component will need at least 3 default icons.

In my opinion, we should not export icons. I think once we have an `Icon` component, we'll just be using those internally. So if a user wants to import the icon they can import the actual Icon component instead of the one from Pagination. I don't have too strong an opinion here though.

rfcs/pagination-component.md
abmai: I was thinking that the user could just override the button itself directly (and pass in the labels that way) vs. surfacing a high level prop here. What do you think? If that's too much, then the object approach here is good.

rfcs/pagination-component.md
abmai: Oh yeah there's some inconsistency here. I'll update everything to use `prev`.

rfcs/pagination-component.md
schnerd:
```js
<Pagination overrides={{
  StyledBackButton: props => (
    <StyledBackButton {...props}>{app.strings.previous}</Button>
  ),
  StyledNextButton: props => (
    <StyledNextButton {...props}>{app.strings.next}</Button>
  )
}}>
```

That's not too bad.

I looked a bit at other designs to see if other components will run into this. It seems like datepicker is one of the only other components that has _visible_ static labels (there's probably plenty of invisible aria labels for close/next buttons, etc).

image

Overriding 7 components for the "Quick Select" and each button might get a bit tedious. The object format is a slight improvement on this.

All that said, I did a bit more research on i18n in component libraries, and I think [ant's approach](https://ant.design/docs/react/i18n) makes sense. Instead of passing in override strings at the callsite, you have a json file full of strings that are passed through via context. So maybe supporting overrides for all libraries here isn't so necessary. Feel free to just make a decision and land. I'll open up a separate issue about general i18n strategy.

abmai:
> In my opinion, we should not export icons. I think once we have an Icon component, we'll just be using those internally. So if a user wants to import the icon they can import the actual Icon component instead of the one from Pagination. I don't have too strong an opinion here though.

Cool, I also think it's easier and better to not be exporting the icons so let's go with that for now. Thanks!

PR #179, opened 2018-09-04
feat(modal): initial modal implementation

schnerd:
## Modal Component

image

This PR includes most of the basic functionality for Modal's declarative API as defined in #119

**What's not included:**
- `Modal.method()` imperative API (as defined in #119)
- Set `aria-labeledby` of Dialog element pointing to the `<ModalTitle>` child element if present.
- Best practices for accessibility when multiple modals are visible at once. Certain aria props need to be added/removed from the bottom modal. We need a sort of `ModalManager` class to facilitate this.
- Proper close icon once `Icon` component is ready. Currently using a placeholder.
- Proper button once `Button` component is ready. Currently using a placeholder.

These features can be added in the future in a separate PR.


rfcs/modal-component.md
```diff
@@ -20,29 +20,23 @@
...
+- `size: SIZE.{default|full|auto}|string|number`, Defaults to SIZE.default
   Controls the size of the modal (primarily width). Can be a SIZE constant or css `width` property value.
-- `aria-labelledby: string` - Optional, Default to ''
-  If specified, will be passed to the dialog element
-- `aria-describedby: string` - Optional, Defaults to ''
-  If specified, will be passed to the dialog element
-- `aria-label: string` - Optional, Defaults to ''
-  If specified, will be passed to the dialog element (described the modal for screen readers)
-- `id: string` - Optional, Defaults to ''
-  If specified, will be passed to the dialog element
```
schnerd: `aria-labeledby` will be handled in a separate PR. All these others a11y-related props can be passed in via overrides if desired. the modal by default has the necessary default a11y practices without exposing these props explicitly.

rfcs/modal-component.md
```diff
@@ -20,29 +20,23 @@
...
 - `role: 'dialog'|'alertdialog'|string` - Optional, Defaults to 'dialog'
   Which accessibility role this modal should have
-- `mountNode: HTMLElement | React.Ref` - Optional, Defaults to document.body
```
schnerd: Allowing a React.Ref here proved problematic. If your mount node unmounts before your modal component unmounts, react throws some scary looking errors (this was happening as I was switching stories in storybook). It seems safest to just encourage people to use an HTMLElement for now (likely one outside of their react tree). If really necessary they can still use `findDomNode` to use a react node.

e2e/tests.js
```diff
@@ -1,11 +1,34 @@
+/*
```
abmai: Do we want to just go with a truncated MIT license? cc @gergelyke @rtsao 

```
/** Copyright (c) [YEAR] Uber Technologies, Inc.
 *
 * This source code is licensed under the MIT license found in the
 * LICENSE file in the root directory of this source tree.
 */
```

gergelyke: yeah, we will - it's on my plate, will send a pr today

src/modal/e2e.js
```diff
@@ -0,0 +1,71 @@
...
+
+      await driver.sleep(100);
+
+      // Open modal
+      let openModalButton = await driver.findElement(
+        By.css('.open-modal-button'),
+      );
+      openModalButton.click();
+
+      await driver.sleep(100);
```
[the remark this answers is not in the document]
schnerd: Fair point, was essentially just waiting for the dialog to animate in, updated to use the webdriver helpers.

src/modal/modal.js
```diff
@@ -0,0 +1,336 @@
...
+      return mountNode;
+    }
+    // Flow thinks body could be null (cast through any)
+    // eslint-disable-next-line flowtype/no-weak-types
+    return ((document.body: any): HTMLBodyElement);
+  }
+
+  getChildren() {
+    const {children} = this.props;
+    return typeof children === 'function' ? children() : children;
```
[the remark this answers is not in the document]
schnerd: If a modal is hidden, it seems a tad wasteful to still pass all the children–even if its essentially just a bunch of React.createElement calls. I was thinking optionally supporting children as function was a cheap addition and allows people to lazily render children only when the modal is open. That said, I don't feel super strongly here, if you think it could lead to confusion we can remove it.

PR #186, opened 2018-09-05
fix(menu): only polyfill smoothscroll in browser env

schnerd:
Fix for an issue reported by @nudge:
```
ReferenceError: window is not defined
    at Object.polyfill (/Users/davidng/Uber/src/riderpayments/web-payments-experience-2/node_modules/smoothscroll-polyfill/dist/smoothscroll.js:8:13)
    at Object.<anonymous> (/Users/davidng/Uber/src/riderpayments/web-payments-experience-2/node_modules/baseui/menu/utils.js:18:32)
7:33 AM
```
> i think because the smoothscroll polyfill is being called regardless of browser/backend inside menu/utils.js

Wrapping this call in a `__BROWSER__` should do the trick.

### DIFF
--- a/src/menu/utils.js
+++ b/src/menu/utils.js
@@ -8,7 +8,9 @@ LICENSE file in the root directory of this source tree.
 /* eslint-disable import/prefer-default-export */
 import smoothscroll from 'smoothscroll-polyfill';
 
-smoothscroll.polyfill();
+if (__BROWSER__) {
+  smoothscroll.polyfill();
+}
 
 /**
  * Given a props object and a mapper dictionary of prop keys, we will prepend


gergelyke:
do we want to add test coverage for this?

schnerd:
>do we want to add test coverage for this?

Yes we should, though it's maybe not as straightforward as adding a new test case. It seems like we'd need to run `jest` a second time with `node` env instead of `jsdom`, and we'd only want to run basic render tests (since only renderToString is called on the server).

We've historically used [unitest](https://github.com/rtsao/unitest) for things like this. But I think that is more for tap than jest (I could be wrong). Either way we'd probably want to discuss a strategy for this separate from this PR. I can open up an issue.

PR #187, opened 2018-09-05
docs(rfcs): reorder to better fit docs page

gergelyke:


rfcs/button-component.md
```diff
@@ -1,5 +1,50 @@
...
+  <Button
+    startEnhancer={
+      <StyledStartEnhancer>
+        <SvgIcon />
+      </StyledStartEnhancer>
+    }
+    endEnhancer={
+      <StyledEndEnhancer>
+        <SvgIcon />
+      </StyledEndEnhancer>
```
schnerd: Unrelated to @gergelyke 's changes but @abmai @kahwee why do we need StyledEnhancer here? Shouldn't it automatically be using these internally already? Seems like you should only need to import `Styled*` when you want to use the `overrides` mechanism.

rfcs/button-component.md
```diff
@@ -1,5 +1,50 @@
 # Button Component
 
+### Usage
+
+#### Basic usage
```
schnerd: Nice, I like the idea of showing both basic usage and advanced usage.

rfcs/button-component.md
```diff
@@ -1,5 +1,50 @@
...
+
+```javascript
+import * as React from 'react';
+import {Button, KIND} from 'baseui/button';
+
+export default () => <Button kind={KIND.primary}>Submit</Button>;
```
schnerd: Should we still show some sort of basic onClick handler here?

```js
<Button onClick={() => console.log('clicked')} kind={KIND.primary}>
  Submit
</Button>
```

rfcs/checkbox-component.md
```diff
@@ -1,5 +1,75 @@
...
+### Usage
+
+#### Basic usage
+
+```javascript
+import * as React from 'react';
+import {Checkbox} from 'baseui/checkbox';
+
+export default () => <Checkbox />;
+```
```
schnerd: I think for form controls the basic usage should hopefully still show passing a value in and change handlers

```js
<Checkbox 
  checked={props.isEnabled}
  onChange={evt => props.setEnabled(evt.target.checked)}
>
  Enabled
</Button>
```

rfcs/input-component.md
```diff
@@ -1,133 +1,145 @@
...
+### Usage
+
+#### Basic usage
+
+```javascript
+import * as React from 'react';
+import {Input} from 'baseui/input';
+
+export default () => <Input />;
+```
```
schnerd: Same here, maybe something like:

```js
<Input 
  value={props.name}
  onChange={evt => props.setName(evt.target.value)}
/>
```

rfcs/button-component.md
abmai: Hm my guess is probably just an oversight in writing these. It should be used internally already yes. cc @kahwee

PR #189, opened 2018-09-05
docs(rfcs): reorder to better fit docs page, second batch

gergelyke:


rfcs/select-component.md
```diff
@@ -1,92 +1,43 @@
...
+        imgSrc: '1.jpg',
+      },
+    },
+    {
+      id: '2',
+      disabled: true,
+      label: {
+        text: 'Second',
+        imgSrc: '2.jpg',
+      },
```
schnerd: We should use string labels here – i'm not sure the object format works without specifying a `getOptionLabel` callback.

rfcs/select-component.md
```diff
@@ -1,92 +1,43 @@
...
+    },
+  ];
+  return
+    <Select
+      options={options}
+      placeholder="Start typing to select a destination"
+    />
```
schnerd: `selectedOptions` and `onChange`? Select is sort of useless without these.

rfcs/textarea-component.md
```diff
@@ -1,5 +1,43 @@
...
+
+### Basic usage
+
+```javascript
+import * as React from 'react';
+import {StatefulTextarea as Textarea} from 'baseui/textarea';
+
+export default () => {
+  return (
+    <Textarea name="description"/>
```
schnerd: `value` + `onChange`?

rfcs/popover-component.md
```diff
@@ -2,7 +2,31 @@
...
+
+export default () => {
+  const popoverContent = <div>Hello world</div>;
+  return (
+    <React.Fragment>
+      <Popover
+        placement="topLeft"
+        content={popoverContent}
+      >
+        <span>Hover me!</span>
```
schnerd: I think maybe my original example was bad here, but by default Popovers are triggered by click, so maybe this should be "Click me!". Maybe in advanced usage we can add `triggerType="hover"`.

PR #196, opened 2018-09-05
docs(spinner): add spinner rfc

gergelyke:


rfcs/spinner-component.md
```diff
@@ -0,0 +1,30 @@
+# Spinner
```
schnerd: for accessibility it seems like adding `role="progressbar"` should be sufficient.

rfcs/spinner-component.md
```diff
@@ -0,0 +1,30 @@
...
+    />;
+};
+```
+
+## Exports
+
+* `Spinner`
+* `VARIANT`
+
+## `Spinner` API
```
schnerd: `size` appears to be missing below though its in the usage example above.

In addition to size I think we probably want some story around customizing the color. This could probably be done via `overrides`, but it seems like overriding the color may be a common enough use case that supporting a `color: string` prop may be worth it (superfine had something similar).

```js
<Spinner size={10} color="white" />
```

nudge: Would be great to have a `text` prop also which is placed below the spinner for loading text.

In addition, a `success` variant would also be useful (see our usage below).

We have built a custom spinner component internally in BaseUI theme in pure CSS, but we don't have the bandwidth to contribute back for a couple of weeks. Do let us know if you need help or want to see our implementation. Cheers!

spinner-with-success mov

gergelyke: cool, thanks David - I'll update the PR, and ping you
also, will follow up in private re your implementation

gergelyke: should the success variant be a separate component, maybe an icon?

PR #197, opened 2018-09-06
docs(icon): icon rfc

schnerd:
image

### DIFF
--- a/rfcs/icon-component.md
+++ b/rfcs/icon-component.md
@@ -1,0 +1,69 @@
+# Icon Component
+
+This is the set of ~16 icons used internally within baseui that are safe to open source. This is not the set of uber-specific icons published internally as @uber/icons (though the API is intended to align with that to reduce confusion).
+
+![image](https://user-images.githubusercontent.com/875591/45131331-51e8c500-b141-11e8-898d-ead83d44b345.png)
+
+## Usage
+
+```javascript
+import * as React from 'react';
+import {LeftArrow} from 'baseui/icon/left-arrow';
+
+export default () => (
+  <LeftArrow size="10px" color="#ccc" />
+);
+```
+
+## Exports
+
+* `Icon` (Base Icon component used by other exports below)
+
+From left-to-right based on the image above:
+
+* `DeleteAlt`
+* `Search`
+* `Delete`
+* `Check`
+* `ArrowUp`
+* `ArrowDown`
+* `ArrowLeft`
+* `ArrowRight`
+* `TriangleRight`
+* `TriangleDown`
+* `Upload`
+* `Plus`
+* `Grab`
+* `Alert`
+* `Menu`
+* `ChevronRight`
+* `ChevronLeft`
+* `Spinner`
+* `Overflow`
+* `TriangleLeft`
+* `TriangleUp`
+* `Filter`
+* `CheckIndeterminate`
+
+## `Icon` API
+
+* `size: number|string` - Optional, Defaults to `scale500` (12px)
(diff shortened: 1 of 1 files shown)


rfcs/icon-component.md
```diff
@@ -0,0 +1,69 @@
+# Icon Component
+
+This is the set of ~16 icons used internally within baseui that are safe to open source. This is not the set of uber-specific icons published internally as @uber/icons (though the API is intended to align with that to reduce confusion).
+
+![image](https://user-images.githubusercontent.com/875591/45131331-51e8c500-b141-11e8-898d-ead83d44b345.png)
```
gergelyke: do we wan't to have this image in the docs before the usage?

also, i'd use a more descriptive alt text, maybe like `baseui-icons`?

rfcs/icon-component.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Usage
+
+```javascript
+import * as React from 'react';
+import {LeftArrow} from 'baseui/icon/left-arrow';
```
kahwee: This is almost identical to our other icons API. 👍

rfcs/icon-component.md
```diff
@@ -0,0 +1,69 @@
...
+
+## `Icon` API
+
+* `size: number|string` - Optional, Defaults to `scale500` (12px)
+  Size of element, will be passed to the svg width/height style. Can also be a baseui `theme.sizing` constant like `sizing500`.
+* `color: string` - Optional, Defaults to `currentColor`
+  Color of icon, will be used as svg fill
+* `title: string` - Optional
+  Allows you to override the SVG `<title>` label, which is used for accessibility
+* `overrides: {Svg}`
```
kahwee: What's the use-case of overriding the SVG with another SVG?

rfcs/icon-component.md
```diff
@@ -0,0 +1,69 @@
...
+* `CheckIndeterminate`
+
+## `Icon` API
+
+* `size: number|string` - Optional, Defaults to `scale500` (12px)
+  Size of element, will be passed to the svg width/height style. Can also be a baseui `theme.sizing` constant like `sizing500`.
+* `color: string` - Optional, Defaults to `currentColor`
+  Color of icon, will be used as svg fill
+* `title: string` - Optional
+  Allows you to override the SVG `<title>` label, which is used for accessibility
```
kahwee: Should we just default to `Plus` and `ArrowLeft` etc? I think developer might forget to change this.

kahwee:
Do we return an SVG or a container with the SVG inside?

rfcs/icon-component.md
schnerd: It's less to override the code, and more for overriding styles or props:

```js
<LeftArrow overrides={{
  Svg: {
    props: {viewBox: '0 0 24 20'},
    style: {position: 'absolute', top: '0px', left: '10px'}
  }
/>
```

rfcs/icon-component.md
schnerd: yeah, we'll provide a default name for each component. this prop is mainly available for override in case the name title doesn't accurately fit the specific use case for the icon.

schnerd:
>Do we return an SVG or a container with the SVG inside?

Just the SVG–trying to keep it simple. If there's a good reason to add a container then we could consider it though.

rfcs/icon-component.md
kahwee: I see! So this only overrides the top Element's props I suppose?

It's probably clearer if users create a container to house the SVG themselves. I don't think touching the viewBox is a good idea, users might want to touch `preserveAspectRatio` though. 

https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/preserveAspectRatio

PR #198, opened 2018-09-06
fix(select): correctly handle disabled state for select component

nudge:
#### Patch: Bug Fix

The `select` component shows up as disabled styling by default (ie. greyed out -- see https://baseui.netlify.com/?selectedKind=Select&selectedStory=In%20Select%20single%20mode&full=0&addons=1&stories=1&panelRight=0&addonPanel=REACT_STORYBOOK%2Freadme%2Fpanel).

This pull request correctly displays the enabled state as well as correctly handles the disabled state by adding a `disabled` prop to the top level and disabling changes if `true`.

#### Tests Added + Pass?

Yes, updated snapshot tests.

#### License

MIT


rfcs/select-component.md
```diff
@@ -150,7 +150,9 @@ export default () => {
   Lets you control what label is rendered for the specified option. The default behavior returns `option.label`.
 * `getSelectedOptionLabel?: (option: Object) => React$Node`:
   Lets you control what label is rendered for the selected option. If not specified, defaults to `getOptionLabel`.
-* `autoFocus: boolean`:
+* `disabled: boolean`:
+  Make the control disabled. Default is `false`
```
schnerd: It looks like we already list `disabled` here two lines below?

src/select/select.js
```diff
@@ -55,7 +55,7 @@ class Select extends React.Component<PropsT, StatelessStateT> {
     filteredOptions: null,
     textValue: this.props.textValue,
     selectedOptions: this.props.selectedOptions,
-    isDropDownOpen: this.props.type === TYPE.search,
+    isDropDownOpen: !this.props.disabled && this.props.type === TYPE.search,
```
schnerd: @blackswanny why does this default to `true` for search to begin with? i've never seen a select element or search bar that has the dropdown open by default. maybe I'm missing something.

blackswanny: probably in case if we have some populated. Ok, seems like it works by different condition. I will remove this

src/select/select.js
```diff
@@ -189,21 +189,24 @@ class Select extends React.Component<PropsT, StatelessStateT> {
...
       <div
         onClick={() => {
-          this.setState({isDropDownOpen: !this.state.isDropDownOpen});
+          if (!disabled) {
+            this.setState({isDropDownOpen: !this.state.isDropDownOpen});
+          }
         }}
       >
         <InputComponent
-          disabled={true}
```
blackswanny: if we don't set it to true, we will be able to input symbols in Select mode, which is not what expected by design. Only in Search mode it should be available.

src/select/select.js
```diff
@@ -256,7 +261,12 @@ class Select extends React.Component<PropsT, StatelessStateT> {
           InputContainer: InputContainer,
           After: () => (
             <SearchIcon
-              onClick={e => this.onChange(e, STATE_CHANGE_TYPE.clearAll)}
+              disabled={disabled}
```
blackswanny: this attribute will leak to DOM  tag, but it's not supported by this DOM element, that's why we name them `$disabled` if we want it to be ignored

src/select/select.js
```diff
@@ -256,7 +261,12 @@ class Select extends React.Component<PropsT, StatelessStateT> {
...
+              onClick={e => {
+                if (!disabled) {
```
blackswanny: if all Input component were disabled, we wouldn't need to add this condition everywhere as event handlers shouldn't be called

src/select/select.js
```diff
@@ -289,16 +300,19 @@ class Select extends React.Component<PropsT, StatelessStateT> {
           />
         )}
         {selectedOptions.map(option => (
-          <Tag key={option.id} $multiple={multiple}>
+          <Tag key={option.id} $multiple={multiple} disabled={disabled}>
```
blackswanny: this attribute will leak to DOM  tag, but it's not supported by this DOM element, that's why we name them `$disabled` if we want it to be ignored

src/select/select.js
```diff
@@ -289,16 +300,19 @@ class Select extends React.Component<PropsT, StatelessStateT> {
...
             {multiple && (
               <SearchIcon
+                disabled={disabled}
                 onClick={e => {
-                  this.setState({
-                    selectedOptions: this.state.selectedOptions.filter(
-                      selectedOption => selectedOption.id !== option.id,
-                    ),
-                  });
+                  if (!disabled) {
```
blackswanny: there is a better way to prevent events from happening. You capture event on top level DOM element and prevent it from going down 
image

src/select/select.js
blackswanny: @gergelyke can you set this boolean to false here, and if nothing breaks, lets leave it like this

src/select/select.js
gergelyke: what do you recommend, what we should do here?

src/select/select.js
gergelyke: did you wanted to mention @nudge ? I'd rather wait for him and not modify this PR

src/select/select.js
blackswanny: so setting it to disabled was intentional as we reuse Input component and it takes care of disabled state. So functionally it is correct. Unfortunately it causes a change in styles, which we don't have in Select by design. So I suggest to have disabled=true, but override styles of each subcomponent of Input by adding something like this:
```
overrides: {
        Input: InputOverride,
      },
    } = this.props;

InputOverride = Object.assign(InputOverride, {
    style: {
        //styles to override for disabled state of Input
     }
});

        <Input
          {...getOverrideProps(InputOverride)}
        >
```

src/select/select.js
blackswanny: sure. I think this condition is overriden or ignored in initial render, so it doesn't cause issues, however it is better to remove it
cc @nudge

src/select/select.js
gergelyke: I am not sure if I understand why do we need to set disabled to true on the initial render - can you help me what am I missing?

blackswanny: In Select mode we are not allowed to input anything, only display. So we set Input component to disabled to avoid any work with keyboard

schnerd: Can we just use a `div` instead of an Input for select mode? Seems like the input is only needed for search mode.

gergelyke:
@schnerd @blackswanny what do you think, what should we fixed in the scope of this PR, and what should we handle separately?

src/select/select.js
blackswanny: I tried. However, input has more than half of styles needed for Select. It's not just a div, but some subcomponents. It seems obvious not to DRY and reuse Input , which already supports styles and subcomponents. Otherwise we need to repeat all of it.

blackswanny:
@gergelyke  the way it fixes right now may introduce some new bugs. For instance, what if we forgot some styles to be changed in Input, which were previously changed by enforced "disabled=true" set? Also the fix contains more code, for the same feature, which is harder to maintain. For, instance, `if !disabled` is mentioned in a few places, whereas it's better to do it once for events handler for instance. So all comments are related to current fix, and I am not sure if we can split it into fix and some extra features refactoring as it's all solid.

src/select/select.js
schnerd: |   |   |
|---|---|
| screen shot 2018-09-06 at 11 58 44 am | screen shot 2018-09-06 at 11 59 00 am  |

So they look similar, but how many styles do they actually share? Maybe the padding, typography, and border radius? Repeating those styles doesn't sound that bad, especially if it removes a whole class of random bugs that crop up (like this `disabled` one) by trying too hard to be 100% DRY.

blackswanny: styles:
padding, fonts, background color, border
functional:
Before and After icons and their styles, Placeholder and selected value
So if we wish it separate then for Select mode the same functional part should be repeated and maintained outside of Input and transferred to Select itself

gergelyke:
There seems to be we have a lot of discussion on this thread - @blackswanny can you put together a list of action items we should in your opinion to fix this issue? I'd like to make sure that when tomorrow morning David gets up he has a clear set of tasks he has to do on the PR.

blackswanny:
@gergelyke @nudge ok, here is the list:
1) Return `disable=true` to Input, but instead update styles only by overriding them for subcomponents like this:
```
overrides: {
        Input: InputOverride,
      },
    } = this.props;

InputOverride = Object.assign(InputOverride, {
    style: {
        //styles to override for disabled state of Input
     }
});

        <Input
          {...getOverrideProps(InputOverride)}
        >
```

2) Avoid to check in any handler for `if (!disabled) {`, but otherwise just capture click on very parent element and stop propagating down, like example below:
```
const events = option.disabled ?
                  {
                    onClickCapture: e => e.stopPropagation(),
                  }
```

src/select/select.js
nudge: Thanks for the explanation @blackswanny!

rfcs/select-component.md
nudge: Nice catch, removed.

src/select/select.js
nudge: Removed this condition.

src/select/select.js
nadiia: I think we need to rethink the select implementation here since using a disabled input feels wrong semantically-, visually-, and accessibility-wise

src/select/select.js
nadiia: There should be a set of shared props that are passed to all subelements, I'd say the `disabled` props should be one of them, all props in the shared set should be prefixed with `$` sign

gergelyke:
@nudge @blackswanny do we need anything else to be changed on this PR, or is it ready to be merged?

blackswanny:
@gergelyke `disabled` is still `false`, so I anticipate bugs due to inconsistency of styles, when Input is not disabled

gergelyke:
if we change that to true, won't we end up with a disabled default state?

gergelyke:
@nudge can you make that last change? we can jump on a quick call too, if that helps

nudge:
Hey team, I've been out in Bangalore for the past week. Thanks for the discussion surrounding this, will make changes when I get a chance!

gergelyke:
Thanks David! 👍

src/select/select.js
```diff
@@ -289,27 +313,38 @@ class Select extends React.Component<PropsT, StatelessStateT> {
...
-        {selectedOptions.map(option => (
-          <Tag key={option.id} $multiple={multiple}>
-            {this.getSelectedOptionLabel(option)}
-            {multiple && (
-              <SearchIcon
-                onClick={e => {
+        {selectedOptions.map(option => {
+          const events = disabled
+            ? {
+                onClickCapture: e => e.stopPropagation(),
```
schnerd: What benefit does stopPropagation have for all of these cases?

It seems like there might be cases where you still want the vent to bubble up to the document. For example, if I have a popover that's visible, and it uses onClickOutside-style logic to close when the document is clicked. In theory clicking a disabled select should still cause probably still be able to trigger this behavior.

blackswanny: do we try to do two actions via one click? I am not sure that click inside of Select should trigger close of popover somewhere else. Otherwise in general practice activating two event handlers at the very same moment may cause issues with controversy of application state if both change this state in the same data

schnerd: >I am not sure that click inside of Select should trigger close of popover somewhere else. 

This seems pretty common. For instance, in gmail click on the user dropdown to open that popover and then click on search - the popover will close. Twitter/facebook have similar behavior.

In these cases clicking anywhere outside the popover is essentially simulating a "blur" event for the popover, which doesn't natively support blur. By stopping propagation here you could be preventing that from happening.

Is there any downside to just not specifying a handler here when select is disabled? (instead of explicitly capturing the event and stopping propagation)

blackswanny: Yes, if we don't specify handler, the click will sink down and activates logic of inner used component.
Unfortunately, we cannot stop propagation only downwards, and allow it bubble up on parent element. Event broadcast cycle stops it for all the elements and document is somewhere down the hole

schnerd: > The click will sink down and activates logic of inner used component.

That seems fine though? If the inner Input is `disabled` then it shouldn't have any interaction events. And if it does that means Input is broken and should be fixed.

Is there anything that fundamentally doesn't work correctly when you remove the stopPropagation? Seems like we can test this. If it works fine then maybe we should leave it out since it could break onClickOutside logic for other components.

blackswanny: so about this particular case. Currently we don't send `disabled` to `<Tag>` so it means that tag will never deactivate it's cross button click itself. So there are two options:
- current solution assuming we don't trust Tag to handle `disabled` state
- pass `disabled` to Tag and assume it knows what to do, so we don't need to capture it ahead.

schnerd: It seems like we should pass disabled to Tag. That way it doesn't show a `pointer` cursor on the tag when the select element is disabled.

src/select/select.js
```diff
@@ -115,10 +115,16 @@ class Select extends React.Component<PropsT, StatelessStateT> {
...
     label?: LabelT,
+    disabled?: boolean,
   ) => {
     const multiple = this.isMultiple();
     const selected = this.state.selectedOptions.find(tag => tag.id === id);
     let selectedOptions;
+
+    if (disabled) {
+      return;
+    }
```
nudge: Not sure what the latest decision is on the event handling during disabled state, but added this which early return to stop changes from occurring.

blackswanny:
we usually rebase, not merge

gergelyke:
should we close it in favor of https://github.com/uber-web/baseui/pull/292 @blackswanny ?

blackswanny:
> should we close it in favor of #292 @blackswanny ?

@gergelyke  I moved the code from here to that PR

gergelyke:
cool, thanks Oleg!

PR #200, opened 2018-09-06
feat(card): implement card component

MarkReeder:
#### Minor: Add Card component
#199 

#### Tests Added + Pass?

Yes

#### License

MIT

Add the card component

### DIFF
--- a/src/card/stories.js
+++ b/src/card/stories.js
@@ -61,7 +61,7 @@ storiesOf('Card', module)
     </Card>
   ))
   .add('Text only', () => (
-    <Card style={{width: cardWidth}}>
+    <Card>
       <StyledBody>{placeholderText}</StyledBody>
     </Card>
   ))


src/card/stories.js
```diff
@@ -0,0 +1,130 @@
...
+  hasThumbnail: (props: {+thumbnail?: string}) => Boolean(props.thumbnail),
+  overrides: {},
+};
+
+function Button(props) {
+  return <button {...props} />;
+}
+
+storiesOf('Card', module)
+  .addDecorator(story => (
```
gergelyke: can you add the RFC as the readme? just like here: https://github.com/uber-web/baseui/blob/master/src/select/stories.js#L19

gergelyke:
@MarkReeder can you point us to the RFC please?

MarkReeder:
@gergelyke RFC is here: https://github.com/uber-web/baseui/blob/8466aeefc9ebc741a0d0112f839727ad1f21e831/rfcs/card-component.md

src/themes/light-theme.js
```diff
@@ -59,6 +59,38 @@ const mono1000 = '#000000';
 const primaryFontFamily = '"Helvetica Neue", arial, sans-serif';
 
 export default {
+  border: {
```
schnerd: We already have a `borders` key in theme, should we consolidate? May also need to update `ThemeT` in src/styles/types.js

rfcs/card-component.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Usage
+
+Basic usage:
+
+```javascript
+import * as React from 'react';
+import {Card} from 'baseui/card';
+
+export default () => <Card title="This is a card" />;
```
schnerd: Might be good to flesh this out a little bit since we're currently using our RFCs as documentation for the time being. We've also started moving "Usage" to the top for that reason (#195).

schnerd:
The flexibility of having a prop-based _and_ composable API is pretty cool. At first I was a bit concerned that it could lead to confusion about the "right" way to use it, but I think as long as we adequately document both it should be fine.

rfcs/card-component.md
MarkReeder: Definitely- I'll flesh this out a bit more today. I mostly wanted to get the functional component up for review before doing so since the props/composable API is a bit of a departure from other things we've done. I'll add some more here and some tests today.

src/themes/light-theme.js
MarkReeder: Ah, that may have been added between my initial add and now. It looks like they can be pretty cleanly merged- the existing `borders` key is strictly concerned with border radius. I'll merge.

src/card/card.js
```diff
@@ -0,0 +1,107 @@
...
+  const Body = getOverride(BodyOverride) || StyledBody;
+  const Contents = getOverride(ContentsOverride) || StyledContents;
+  const HeaderImage = getOverride(HeaderImageOverride) || StyledHeaderImage;
+  const Root = getOverride(RootOverride) || StyledRoot;
+  const Thumbnail = getOverride(ThumbnailOverride) || StyledThumbnail;
+  const Title = getOverride(TitleOverride) || StyledTitle;
+
+  const $hasThumbnail = hasThumbnail(props);
+  return (
+    <Root {...getOverrideProps(RootOverride)} {...otherProps}>
```
nadiia: Let's add `getOverrideProps(RootOverride)` last to make sure those overrides always take precedence

src/card/stories.js
```diff
@@ -0,0 +1,135 @@
...
+};
+
+function Button(props) {
+  return <button {...props} />;
+}
+
+storiesOf('Card', module)
+  .addDecorator(withReadme(CardReadme))
+  .addDecorator(story => (
+    <ThemeProvider theme={LightTheme}>{story()}</ThemeProvider>
```
nadiia: No need to add a ThemeProvider here as it's handled in the storybook's config

src/card/stories.js
```diff
@@ -0,0 +1,135 @@
...
+
+import type {Props} from './card';
+
+const cardWidth = '328px';
+const placeholderTitle = 'Card Title Entry';
+const placeholderText =
+  'Proin ut dui sed metus pharetra hend rerit vel non mi. Nulla ornare faucibus ex, non facilisis nisl. Maecenas aliquet mauris ut tempus cursus. Etiam semper luctus sem ac blandit. ';
+
+const Card = (props: Props) => (
+  <BaseCard style={{width: cardWidth}} {...props} />
```
nadiia: How about passing style overrides explicitly through overrides for a specific presentational component?

rfcs/card-component.md
```diff
@@ -0,0 +1,56 @@
...
+  * `Root: OverrideObjectT` - Optional
+  * `Thumbnail: OverrideObjectT` - Optional
+  * `Title: OverrideObjectT` - Optional
+    Overrides for presentational components.
+* `thumbnail: string` - Optional
+  `src` of the `StyledThumbnail` component
+* `title: React.Node` - Optional
+  Contents of the `StyledTitle` component
+
+## Usage
```
gergelyke: can we move the usage to be the first section to keep consistency with other RFCs?

src/card/stories.js
```diff
@@ -0,0 +1,153 @@
...
+storiesOf('Card', module)
+  .addDecorator(withReadme(CardReadme))
+  .add('Headline & Text', () => (
+    <Card>
+      <StyledTitle>{placeholderTitle}</StyledTitle>
+      <StyledBody>{placeholderText}</StyledBody>
+    </Card>
+  ))
+  .add('Text only', () => (
+    <Card style={{width: cardWidth}}>
```
nadiia: no need for passing this style anymore since it's handled in the `Card` custom component

MarkReeder: good catch

PR #201, opened 2018-09-06
fix: styled should work properly for regular components again

abmai:
Fixes https://github.com/uber-web/baseui/issues/184


gergelyke:
released as https://github.com/uber-web/baseui/releases/tag/v1.3.1

PR #202, opened 2018-09-06
fix(input): remove duplicate IDs from default props

### LINKED ISSUE #192: [Accessibility] Duplicate IDs are present in Input and Textarea components
* [x] I have searched the [issues](https://github.com/uber-web/baseui/issues) of this repository and believe that this is not a duplicate.

## Expected Behavior

An expected behavior for a standard `Input` (with an associated `Label`) is to have a unique ID for each pairing.  

## Current Behavior

The current `Input` and `Textarea` components are providing a `defaultProp` for ID using the `getBuiId` utility. Since `defaultProps` is a static method and it isn't called on instances of the class, the ID is being duplicated across all usages.

#### Storybook a11y failures:

screen shot 2018-09-05 at 12 47 47 pm

## Context

I'm not too familiar with the layout of this repo, but I have a simple commit (https://github.com/austin94/baseui/commit/c39199a8745065d796eb8643de3333a8d071abcf) that includes some fixes for this issue. 

It includes:
* Remove ID defaultProp from all components
* Generate a unique ID for each instance, similar to the `Popover` implementation
  * Not applied to `BaseInput` and `TextArea` since there are no associated Labels that require it
* Update snapshots and add `htmlFor` assertion

If any of this seems reasonable I can open a PR if wanted. Thanks 👋!

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | master  |
| React   |    ^16.4.0     |
| browser |    Chrome     |
| node.js |    v9.11.1     |

## Steps to Reproduce

https://baseui.netlify.com/?selectedKind=Input&selectedStory=Controlled%20and%20uncontrolled%20input&full=0&addons=1&stories=1&panelRight=0&addonPanel=REACT_STORYBOOK%2Freadme%2Fpanel

austin94:
Fixes #192 

#### Patch: Bug Fix

This PR removes duplicate IDs that are present in the `BaseInput`, `Input`, and `Textarea` components. From the discussion #192, this is due to the `static` nature of `defaultProps`.

#### Tests Added + Pass?

Snapshots have been updated as necessary and an additional `htmlFor` assertion has been included.

#### License

MIT

## Additional Questions

I'm not familiar with this codebase so there are some assumptions in this PR that will probably need to change 😄These include:

* Removing the generated ID from the `BaseInput` and `Textarea` components
  * ID's are traditionally only necessary if there are associated label, hint, caption, etc. components.
  * Since that is normally determined at a higher abstraction, `Input` in this case. I've left this out.


austin94:
@schnerd from the original branch... 

> Looks great! Since Textarea uses BaseInput you may also need the generatedId logic there as well.

I might be overlooking this in the Storybook examples, but is there an example of `Textarea` with an associated `Label`? It looks like the `Textarea` doesn't include any `Label` logic.

schnerd:
>I might be overlooking this in the Storybook examples, but is there an example of Textarea with an associated Label? It looks like the Textarea doesn't include any Label logic.

I think you're right actually, we're moving labels out of the controls themselves. Ignore that comment then.

src/input/input.js
```diff
@@ -82,7 +93,7 @@ class Input extends React.Component<InputPropsT, InternalStateT> {
           <Label
             {...getOverrideProps(LabelOverride)}
             {...sharedProps}
-            htmlFor={this.props.id}
+            htmlFor={this.getInputIdAttr()}
```
nadiia: Label was extracted, so can revert this change for an easier rebase

src/input/types.js
```diff
@@ -65,7 +65,7 @@ export type BaseInputPropsT = {
   autoFocus: boolean,
   disabled: boolean,
   error: boolean | React.Node | ((props: PropsT) => React.Node),
-  id: string,
+  id: ?string,
```
nadiia: Maybe keep it not nullable?

src/input/input.js
austin94: Removed after the rebase.

src/input/types.js
austin94: This change is no longer needed. Removed.

austin94:
> Label was extracted, so can revert this change for an easier rebase

It looks like there has been a lot of movement in the components since I opened this PR. Now that label is extracted the changes are a little less dramatic. With the `Label` extraction the only real change is the defaultProp removal.

@nadiia might be worth a follow-on review now that I've rebased.

gergelyke:
@austin94 @nadiia do we need anything else before merging?

nadiia:
Going to merge this one

PR #203, opened 2018-09-06
docs(toast): toast notification rfc

nadiia:
screen shot 2018-09-06 at 11 43 50 am

### DIFF
--- a/rfcs/toast-component.md
+++ b/rfcs/toast-component.md
@@ -1,0 +1,117 @@
+# Toast Notification Component
+
+Toast notifications are used to display a brief alert, warning, or confirmation message or a banner.
+
+## Usage
+
+### Basic usage
+
+```javascript
+import * as React from 'react';
+import {Toast} from 'baseui/toast';
+
+export default () => <Toast>This is a notification message.</Toast>;
+```
+
+### Advanced usage
+
+```javascript
+import * as React from 'react';
+import {Toast} from 'baseui/toast';
+
+export default () => {
+  return (
+    <React.Fragment>
+      <Toast
+        onClose={()=>{console.log('Toast notification was dismissed.')}}
+        autoHideDuration={5000}
+        overrides={{
+          Body: props => <CustomContainer>{props.children}</CustomContainer>,
+        }}
+      >
+      This a dismissable info notification that is hidden after 5 seconds automatically.
+      </Toast>
+      <Toast
+        closeable={false}
+        kind={KIND.error}
+      >
+      {({dismiss}) => {
+        return (
+          <div>
+            This is an error notification with an action button
+            <button onClick={() => {
+              dismiss();
+              console.log('Action button clicked');
+            }}>
+              Take action
+            </button>
+          </div>
+        );
+      }}
+      </Toast>
+    </React.Fragment>
+  );
+}
+```
+
+## Exports
+
+* `Toast`
+* `StyledRoot`
+* `StyledBody`
+* `StyledCloseButton`
+* `KIND`
+* `PLACEMENT`
+
+## `Toast` API
+
+* `children: React.node | ({dismiss}) => React.node` - Required
+  Toast notification content. The children-as-function receives a `dissmiss` method that can be called to dismiss the notification and can be used as a handler for an action inside the toast content
+* `closeable: boolean` - Optional. Defaults to true
+  When set to true a close button is displayed and the notification can be dismissed by a user
+* `kind: 'info' | 'success' | 'warning' | 'error'` - Optional. Defaults to 'info'
+  Defines the type of notification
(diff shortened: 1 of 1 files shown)


rfcs/toast-component.md
```diff
@@ -0,0 +1,116 @@
...
+import {Toast} from 'baseui/toast';
+
+export default () => {
+  return (
+    <React.Fragment>
+      <Toast
+        onClose={()=>{console.log('Toast notification was dismissed.')}}
+        autoHideDuration={5000}
+        overrides={{
+          Bosy: props => <CustomContainer>{props.children}</CustomContainer>,
```
gergelyke: `Body`?

rfcs/toast-component.md
```diff
@@ -0,0 +1,116 @@
...
+* `bottomLeft` - positioned relative to the browser window
+
+## Dependencies
+
+Does this component depend on any 3rd party packages or other internal components?
+
+## Accessibility
+
+* Toast notification container element will have `role="alert"` set on it
+* When multiple alerts are displayed and positioned relative to the browser window they won't overlap but displayed in a column way
```
gergelyke: do we have a max number of Toast that can be stacked up on each other? what would happen if there would be more than the screen can fit?

rfcs/toast-component.md
```diff
@@ -0,0 +1,116 @@
...
+* `$theme: theme`
+
+## KIND Constant
+
+* `info` - Generally used for messages with an informational context
+* `success` - Generally used as a confirmation of a succesful action or operation  
+* `warning` - Generally used for messages with an warning context
+* `error` - Generally used as a notification of an error happened as a result of an action or operation
+
+## PLACEMENT Constant
```
blackswanny: `Dock`?

rfcs/toast-component.md
```diff
@@ -0,0 +1,116 @@
...
+* `kind: 'info' | 'success' | 'warning' | 'error'` - Optional. Defaults to 'info'
+  Defines the type of notification
+* `onClose: function` - Optional
+  A callback function called when a notification is dismissed
+* `overrides: {Root, Body, CloseButton}` - Optional
+  Overrides for presentational components. See "Presentational Components Props API" below.
+  * `[ComponentName]: ReactComponent | {props: {}, style: {}, component: ReactComponent}` - Optional
+* `placement: 'inline' | 'topLeft' | 'top' | 'topRight' | 'bottomRight' | 'bottom' | 'bottomLeft'` - Optional. Defaults to 'top'
+  Position of a toast notification relative to the browser window. If set to 'inline' a notitication is positioned according to a normal flow of the document
+* `autoHideDuration: number` - Optional. Defaults to 0
```
blackswanny: we can call just `duration`.

rfcs/toast-component.md
blackswanny: i think toasts should be controlled by some management tooling, which would avoid them to overlap each other, be hidden or outside of screen

rfcs/toast-component.md
schnerd: typo: Bosy -> Body

rfcs/toast-component.md
```diff
@@ -0,0 +1,116 @@
...
+* `StyledCloseButton`
+* `KIND`
+* `PLACEMENT`
+
+## `Toast` API
+
+* `children: React.node | ({dismiss}) => React.node` - Required
+  Toast notification content. The children-as-function receives a `dissmiss` method that can be called to dismiss the notification and can be used as a handler for an action inside the toast content
+* `closable: boolean` - Optional. Defaults to true
+  When set to true a close button is displayed and the notification can be dismissed by a user
```
schnerd: I use `closeable` with an `e` in modal. it seems like neither is "correct", but we should probably be consistent (I have no preference)

nadiia: I'll use your version - `closeable` - it looks closer to a correct way

rfcs/toast-component.md
nadiia: That's a good question. I looked through some examples, there are cases where everything that overflows the browser window (usually older/earlier notifications) is hidden. I think scrolling the notifications list would be an option. Will add this case to this section to consider for development.

rfcs/toast-component.md
nadiia: Not sure what you mean by `Dock`, could you clarify?

rfcs/toast-component.md
nadiia: I'd say `autoHideDuration` gives more context on its usage. But happy to consider other options or accept this one if it gets more support.

rfcs/toast-component.md
nadiia: Let's keep it in sync with the prop name

PR #204, opened 2018-09-07
docs(slider): add rfc for slider

blackswanny:
#### Minor: Add RFC for Sliders component

### DIFF
--- a/rfcs/slider-component.md
+++ b/rfcs/slider-component.md
@@ -1,0 +1,112 @@
+# Slider Component
+
+Represents control with sliding axis to choose a single value or range between min and max values.
+
+## Usage
+
+### Basic usage
+
+```javascript
+import * as React from 'react';
+import {Slider} from 'baseui/slider';
+
+export default () => <Slider range={[
+  {value: 0, label: '0, min'},
+  {value: 100, label: '100, max'}
+]} />;
+```
+
+### Advanced usage
+
+```javascript
+import * as React from 'react';
+import {StatefulSlider} from 'baseui/slider';
+
+export default () => {
+  return <StatefulSlider
+    range={[
+      {value: 0, label: '0, min'},
+      {value: 20, label: '20'},
+      {value: 40, label: '40'},
+      {value: 60, label: '60'},
+      {value: 80, label: '80'},
+      {value: 100, label: '100, max'}
+    ]}
+    initialState={{
+      value: [{value: 20, label: '20'}, {value: 40, label: '40'}]
+    }}
+    onChange={(event, value)=>{console.log(value)}}
+    overrides={{
+      Thumb: props => <div style={{
+        color: 'red',
+        borderRadius: '5px',
+      }}></div>,
+    }}
+    />
+}
+```
+
+## Exports
+
+* `Slider`
+* `StatefulSlider`
+* `StatefulSliderContainer`
+* `StyledRoot`
+* `StyledAxis`
+* `StyledTick`
+* `StyledThumb`
+
+## `Slider` API
+
+* `range: Array<Point>` - Required
+  All points present on slider axis. First and last point represent min and max value. Each `Point` can be a value itself (for primitive type) or if `Point` is `Object` it should have `value` and `label`. If more than 2 elements are present in array, they represent all ticks on axis and `step` property is ignored.
+* `value: Point | Array<Point>` - Required
+  Selected points chosen on axis. It can be a single point (one thumb) or 2 points array (range thumbs)
+* `step: number` - Optional
+  If `range` contains only min and max points (2 elements) step is to shift thumb every time user moves it left or right. If `step` in place `value` for each `Point` should be of primitive type
+* `disabled: boolean`:
+  Disable control from being changed  
+* `overrides: {Root: (props: {[string]: any}) => React$Node, Axis: (props: {[string]: any}) => React$Node, StyledTick: (props: {[string]: any}) => React$Node, Thumb: (props: {[string]: any}) => React$Node}`
+  * `Root` container element to render.
+  * `Axis` is the range line element to render.
+  * `Tick` to render a tick(label) for min, max and optional points on axis and on top of `Thumb`
+  * `Thumb` to render a thumb sliding over axis.
+* `onChange: func(event, value)` - Optional
+  handler for events on trigger element, each time thumbs change selection, which is passed in `value`
+
+## `StatefulSlider` API
+
+* `initialState: {}`
+  Initial state of an uncontrolled slider component.
+  * `value: Point | Array<Point>` - an initial state of selected points on axis. It can be a single point (one thumb) or 2 points array (range thumbs)
+* `stateReducer: (type: text, nextState: {}, currentState: {}, e: any) => nextState`
+  A state change handler.
+  * `type` - state change type
+  * `nextState` - a new state changes that will be set
+  * `currentState` - current full state of the component
+* `range: {Array<Point>}` - Required
+  All points present on slider axis. First and last point represent min and max value. Each `Point` can be a value itself (for primitive type) or if `Point` is `Object` it should have `value` and `label`. If more than 2 elements are present in array, they represent all ticks on axis and `step` property is ignored.
+* `step: number` - Optional
+  If `range` contains only min and max points (2 elements) step is to shift thumb every time user moves it left or right. If `step` in place `value` for each `Point` should be of primitive type
+* `disabled: boolean`:
+  Disable control from being changed  
+* `onChange: func(event, value)` - Optional
+  handler for events on trigger element, each time thumbs change selection, which is passed in `value`
+
+## Presentational components props API
+
+These properties are passed to every presentational (styled) component that is exported:
+
+* `$disabled: boolean`
+  Disable control from being changed
+* `$isHovered: boolean`
+  If component is hovered
+* `$isActive: boolean`
+  If component is active
+* `$thumbIndex: number`
+  For `Thumb` only. Index of thumb, which is rendered (0 for first thumb and 1 for second thumb)
+
+## Accessibility
+
+Should support keyboard hotkeys: `tab` - to switch between thumbs, `left` and `bottom` to decrease value of thumb, `right` and `up` to increase thumb value on axis
+Accessibility best practices for this component (`aria-valuenow`, `aria-valuetext`, `aria-valuemin`, `aria-valuemax`, `role=slider`)


rfcs/sliders-component.md
```diff
@@ -0,0 +1,122 @@
...
+## Exports
+
+* `Sliders`
+* `StatefulSliders`
+* `StatefulSlidersContainer`
+* `StyledRoot`
+* `StyledAxis`
+* `StyledTick`
+* `StyledThumb`
+* `CONSTANT`
```
nadiia: is this needed?

rfcs/sliders-component.md
```diff
@@ -0,0 +1,122 @@
...
+## `Sliders` API
+
+* `range: {Array<Point>}` - Required
+  All points present on slider axis. First and last point represent min and max values. Each `Point` can be a value itself (for primitive type) or if `Point` is `Object` it should have `value` and `label`. If more than 2 elements are present in array, they represent all ticks on axis and `step` property is ignored.
+* `thumbs: number` - Optional
+  Amount of thumbs representing how many values can be selected. For even quantity of thumbs selected range is considered in between them. Default is `1`
+* `selectedValues: {Array<Point>}` - Optional
+  Selected array of points from left to right chosen on axis. Default is `[]`
+* `step: number` - Optional
+  If `range` contains only min and max points (2 elements) step is to shift thumb very time user moves it left or right. If `step` in place `value` for each `Point` should be of primitive type
```
nadiia: `very` => `every`

rfcs/sliders-component.md
```diff
@@ -0,0 +1,122 @@
...
+* `disabled: boolean`:
+  Disable control from being changed  
+* `overrides: {Root: (props: {[string]: any}) => React$Node, Axis: (props: {[string]: any}) => React$Node, StyledTick: (props: {[string]: any}) => React$Node, Thumb: (props: {[string]: any}) => React$Node}`
+  * `Root` container element to render.
+  * `Axis` or the range line element to render.
```
nadiia: extra `or` here?

rfcs/sliders-component.md
```diff
@@ -0,0 +1,122 @@
+# Sliders Component
```
schnerd: Why use "Sliders" instead of "Slider"? Seems like the latter makes more sense and is the more common choice among other frameworks.

rfcs/sliders-component.md
```diff
@@ -0,0 +1,122 @@
...
+### Basic usage
+
+```javascript
+import * as React from 'react';
+import {Sliders} from 'baseui/sliders';
+
+export default () => <Sliders range={[
+  {value: 0, label: '0, min'},
+  {value: 100, label: '100, max'}
+]} />;
```
schnerd: Can we show the "selectedValues" here too? Seems like that still belongs as part of basic usage for a stateless component.

rfcs/sliders-component.md
```diff
@@ -0,0 +1,122 @@
...
+* `StyledTick`
+* `StyledThumb`
+* `CONSTANT`
+
+## `Sliders` API
+
+* `range: {Array<Point>}` - Required
+  All points present on slider axis. First and last point represent min and max values. Each `Point` can be a value itself (for primitive type) or if `Point` is `Object` it should have `value` and `label`. If more than 2 elements are present in array, they represent all ticks on axis and `step` property is ignored.
+* `thumbs: number` - Optional
+  Amount of thumbs representing how many values can be selected. For even quantity of thumbs selected range is considered in between them. Default is `1`
```
schnerd: Seems like supporting more than 2 thumbs might complicate things quite a bit. I'm trying to figure out how a slider with 4 or 5 thumbs would work. You'd either need to have a bunch of logic/styles to deal with overlapping ranges, or restricting the min of one range to the max of the other. 

What do you think about just simplifying to either 1 or 2 thumbs?

rfcs/sliders-component.md
schnerd: Also, maybe we can use the numeric Point syntax to for the basic example
```js
<Slider range={[0, 100]} />
```

rfcs/sliders-component.md
```diff
@@ -0,0 +1,122 @@
...
+* `selectedValues: {Array<Point>}` - Optional
+  Selected array of points from left to right chosen on axis. Default is `[]`
```
schnerd: Thoughts on changing this to `value`, more like the `<input type="range" />`?

Also, what do you think about supporting `Point | Array<Point>`.  `<Slider value={5}/>` feels a lot more natural for the common single thumb use case.

rfcs/checkbox-component.md
```diff
@@ -126,7 +126,7 @@ export default () => {
 ## `StatefulCheckboxContainer` API
 
 * `initialState: {}`
-  Initial state of an uncontrolled popover component.
+  Initial state of an uncontrolled checkbox component.
```
gergelyke: should we fix it a new PR? it seems to be out of context of this PR

rfcs/sliders-component.md
```diff
@@ -0,0 +1,122 @@
...
+  {value: 0, label: '0, min'},
+  {value: 100, label: '100, max'}
+]} />;
+```
+
+### Advanced usage
+
+```javascript
+import * as React from 'react';
+import {Foo} from 'baseui/foo';
```
gergelyke: did you mean `slider?`

rfcs/sliders-component.md
gergelyke: I'd go for a simplified version too as a first approach - if later needed, we can figure it out then.

rfcs/sliders-component.md
blackswanny: removed

rfcs/sliders-component.md
blackswanny: ok

rfcs/sliders-component.md
blackswanny: fixed

rfcs/sliders-component.md
blackswanny: fixed

rfcs/sliders-component.md
blackswanny: Numeric is also mentioned as `Point` type can be primitive. I can rephrase.
`selectedValues` is the result. It's optional if we don't want to provide initial value for it. We will just put thumb to min and max

rfcs/sliders-component.md
blackswanny: Yes, more than 2 seems to be very rare case, for instance cut couple samples on timeline or signal axis. However, I see the ways to do it without complicating logic to support it. As the benefit, the interface would not care about how many thumbs there, all unified. No need to check for "magic numbers" 1 or 2

rfcs/sliders-component.md
blackswanny: yes, I can do it, but user need to check if its array or just a value, if amount of thumbs is a variable
renamed to `values`

rfcs/sliders-component.md
schnerd: Maybe we can discuss this at weekly meeting on monday?

rfcs/checkbox-component.md
blackswanny: ok

rfcs/sliders-component.md
blackswanny: fixed

rfcs/sliders-component.md
blackswanny: ok

rfcs/slider-component.md
```diff
@@ -0,0 +1,121 @@
...
+* `StyledTick`
+* `StyledThumb`
+
+## `Slider` API
+
+* `range: {Array<Point>}` - Required
+  All points present on slider axis. First and last point represent min and max values. Each `Point` can be a value itself (for primitive type) or if `Point` is `Object` it should have `value` and `label`. If more than 2 elements are present in array, they represent all ticks on axis and `step` property is ignored.
+* `thumbs: number` - Optional
+  Amount of thumbs representing how many values can be selected. For even quantity of thumbs selected range is considered in between them. Default is `1`
+* `values: {Array<Point>}` - Optional
```
nadiia: How about `value` (not plural) as a name and `Point | Array<Point>` as a value here?

nadiia:
Should we consider a vertical slider option?

rfcs/slider-component.md
```diff
@@ -0,0 +1,121 @@
...
+  * `Tick` to render a tick(label) for min, max and optional points on axis and on top of `Thumb`
+  * `Thumb` to render a thumb sliding over axis.
+* `onChange: func(event, values)` - Optional
+  handler for events on trigger element, each time thumbs change selection, which is passed in `values`
+
+## `StatefulSlider` API
+
+* `initialState: {}`
+  Initial state of an uncontrolled slider component.
+  * `values: {Array<Point>}` - an initial state of selected points array from left to right chosen on axis.
```
nadiia: same comment on a `value` naming here

rfcs/slider-component.md
blackswanny: ok

rfcs/slider-component.md
blackswanny: ok

blackswanny:
> Should we consider a vertical slider option?
i didn't see design for it. I saw vertical sliders only in some scientific apps
cc @nguyensomniac

rfcs/sliders-component.md
nadiia: Let's do a more widely used case (two thumbs) for the initial implementation and then we can look into the improvements once we have a design and use case for it.

kahwee: We can do `multiple=true` to mean you want a slider with 2 knobs. Anything more than 2 knobs sounds like a really obscure use case to me. The `<select>` in [HTML](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select) also has `multiple` attribute, I'm not sure if we should call it `multiple` because of this.

`react-select` calls this `isMulti`.

schnerd: Cool, thanks for the insights everyone. Sounds like we should simplify to just 1 or 2 (both pretty common use cases).

I'm leaning towards just implicitly switching to two thumbs if `value` is an array (instead of a single value), rather than having an explicit `thumbs` prop. Curious if anyone else has thoughts here.

schnerd: Thanks for the insights all.

I like the idea of having a boolean prop like kahwee mentioned instead of `thumbs: number`. Antd uses `range: boolean`, but we're already using `range` for something else, so maybe `multiple` is good? In theory we could also implicitly detect this if `value` is an array.

blackswanny: I added explicit `thumbs` as initial value of selected range was optional, so if user doesn't provide it, we can't say, how many thumbs needed. I can do multiple or make initial value required as it should be more required than optional. otherwise will have to make knobs min amd max, which looks wierd

blackswanny: so removed `thumb` property and made `value` as required, in this case it can be 1 or 2 point value indicating what to expect from component

rfcs/slider-component.md
```diff
@@ -0,0 +1,113 @@
...
+  {value: 0, label: '0, min'},
+  {value: 100, label: '100, max'}
+]} />;
+```
+
+### Advanced usage
+
+```javascript
+import * as React from 'react';
+import {Slider} from 'baseui/slider';
```
schnerd: This should be StatefulSlider if you're using `initialState` below, right?

blackswanny: correct

rfcs/slider-component.md
```diff
@@ -0,0 +1,113 @@
...
+      {value: 20, label: '20'},
+      {value: 40, label: '40'},
+      {value: 60, label: '60'},
+      {value: 80, label: '80'},
+      {value: 100, label: '100, max'}
+    ]}
+    initialState={{
+      value: [{value: 20, label: '20'}, {value: 40, label: '40'}]
+    }}
+    step={20}
```
nadiia: Below it says that step is ignored in the case of > 2 values passed in the `range` prop. Should probably remove it here to avoid a confusion.

rfcs/slider-component.md
```diff
@@ -0,0 +1,113 @@
...
+* `StyledRoot`
+* `StyledAxis`
+* `StyledTick`
+* `StyledThumb`
+
+## `Slider` API
+
+* `range: Array<Point>` - Required
+  All points present on slider axis. First and last point represent min and max value. Each `Point` can be a value itself (for primitive type) or if `Point` is `Object` it should have `value` and `label`. If more than 2 elements are present in array, they represent all ticks on axis and `step` property is ignored.
+* `value: Point | Array<Point>` - Required
```
nadiia: Do we actually need it as a required prop or can we make it equal to the min value of the range by default?

schnerd: I think it's required because we determine whether or not to render two thumbs based on whether it's an array or not. If we want to make this optional then we should add a `multiple={true}` prop or something.

rfcs/slider-component.md
blackswanny: removed

rfcs/slider-component.md
blackswanny: David is right. To reduce amount of props we use this array to find out amount of thumbs to show. We can default number of thumbs, but cannot default values on axis for these thumbs.
About making it to min. Maybe , but usually they set it to some most appropriate , but not a min, also in case of default min we need to introduce prop to show that it is 2 thumb type of slider

benshope:
If you guys are interested in staying consistent with the native range input API that people have worked with before, here is a fun implementation of a dual-handle range slider that I put together.  It is very concise because it focuses on exposing the native input component:

```
const StatelessRangeSlider = props => (
  <RangeSliderStyles>
    {[0, 1].map(sliderIndex =>
      React.createElement(
        'input',
        Object.keys(props).reduce(
          (oneSliderProps, propName) => ({
            ...oneSliderProps,
            [propName]: props[propName][sliderIndex]
          }),
          {type: 'range'}
        )
      )
    )}
  </RangeSliderStyles>
);

class RangeSlider extends React.Component {
  state = {};

  constructor(props) {
    super(props);
    this.state.value = props.value || props.defaultValue;
  }

  render() {
    return StatelessRangeSlider({
      ...this.props,
      value: this.state.value,
      onChange: this.props.onChange.map((onChange, handleIndex) => e => {
        const targetValue = parseInt(e.target.value, 10);
        const newValue = this.state.value.map(
          (value, i) => (handleIndex === i ? targetValue : value)
        );
        const validNewValue = newValue[0] <= newValue[1];
        if (validNewValue) {
          this.setState({
            value: newValue
          });
          onChange(newValue);
        }
      })
    });
  }
}
```

And here is what using it looks like:
```
<RangeSlider
      key={[0, 1].map(i => `slider-${i}`)}
      min={[0, 0]}
      max={[50, 50]}
      value={[15, 45]}
      onChange={[action('onChangeLeft'), action('onChangeRight')]}
/>
```

gergelyke:
@schnerd @nadiia @blackswanny is there anything else we want to address in this RFC?

rfcs/slider-component.md
blackswanny: @nadiia would you still request changes to have another interface for thumbs and value?

nadiia: I'd add a separate prop for the `single | multi` thumbs setting, but it's ok for now.

rfcs/slider-component.md
```diff
@@ -0,0 +1,112 @@
...
+### Advanced usage
+
+```javascript
+import * as React from 'react';
+import {StatefulSlider} from 'baseui/slider';
+
+export default () => {
+  return <StatefulSlider
+    range={[
+      {value: 0, label: '0, min'},
```
benshope: I'm just going to comment because I don't know how things are done here - but to give a bit more concrete feedback, the idea of a `Point: {value: number, label: string}` with a `label` is a great example of something that increases new, non-native API surface while adding relatively little value because most of the time the label can be inferred from the `step` property.   Alteration to tick placement and labels can live in the overrides:
```
<RangeSlider
      key="my-range-slider"
      min={0}
      max={50}
      step={5}
      value={[15, 45]}
      onChange={action('onChangeValues')]}
      overrides={{
         ticksAt: (min, max, step) => ([0, 20, 40, 60, 80, 100]),
         tickFormat: (value) => (<div>(value === 0 ? `${value}, min` : value)</div>)
      }}
/>
```

blackswanny: `Overrides` are optional. They are mostly used for styles and not for functional amendment. In your example we need to use them required.
By native you mean original `input type="range"`? So original inputs are very weak and insufficient, that's why we have these custom components.
Also there are lot's of sliders (in Surveys, settings, etc) that do not use `step` and need to know all the range of values, for instance:
```js
<Caption>To what extend you satisfied?</Caption>
<StatefulSlider
    range={[
      {value: 1, label: 'I am not satisfied'},
      {value: 2, label: 'I am disappointed'},
      {value: 3, label: 'Neutral'},
      {value: 4, label: 'I am satisfied'},
      {value: 5, label: 'I am very satisfied'}
    ]} />
```

benshope: Yes, absolutely overrides should be optional, but not using them for functional aspects of the component seems like a dogma that could lead to limiting architectural choices.  I disagree that "original inputs are very weak and insufficient."  There was a lot of though that went into native component APIs and many have stood up surprisingly well over time.   I would write the particular example you describe out as:
```
<RangeSlider
      key="my-range-slider"
      min={1}
      max={5}
      step={1}
      defaultValue={3}
      onChange={action('onChangeValues')]}
      overrides={{
         tickFormat: (value) => ({1: 'I am not satisfied', 2: 'I am disappointed', ...})(value))
      }}
/>
```
This would preserve all native slider props and only overrides would be different.

blackswanny: >  seems like a dogma that could lead to limiting architectural choices.
It's not a dogma. We can use it for functional change. However, default behavior and usage of any component is not intended to use `overrides`, otherwise it neglects our architecture. `input type=range` is very rare to be used unlike `input type=text`, which has become very powerful. Native range still doesn't support multiple thumbs, and ticks are supported through `<datalist>` with provided `id` reference, which is not the way of abstraction modern frameworks use. Just complying with native API would be not so necessary need to befriend it with current design requirements

benshope: Okay so you're saying that a user will almost always want to format the label on a stepped range's tick values.  I'm not entirely sure that's the case and, even if it is, it might be best to just bump bump that `override` up to a new `prop` rather than creating a structure like `Point`: 
```
<RangeSlider
      key="my-range-slider"
      min={1}
      max={5}
      step={1}
      defaultValue={3}
      onChange={action('onChangeValue')]}
      tickFormat={(value) => ({1: 'I am not satisfied', 2: 'I am disappointed', ...})(value))}
/>
```
I wouldn't argue that you should use reproduce anything janky like I assume `<datalist />` is.  I'm just saying that API surface reuse is an awesome thing to maximize - both to make the component easier to use and to simplify the implementation.  I think it is likely that you can satisfy your design requirements while being just a bit more aligned with the `input type=range`.

blackswanny: @benshope  Thanks for your feedback. The team had no majority in preferences on the API naming and other stuff, so we left it as is so far for future consideration

ericsoco:
Sorry, late the party -- are there plans here to display slider values in editable text inputs? I don't see it in the RFC. E.g. Kepler:

screen shot 2018-09-18 at 3 29 46 pm

We have specific internal use cases I can elaborate on if useful.

blackswanny:
@ericsoco we don't have it in design. Technically you can still do it by overriding `Tick` subcomponent with your own input, and add handlers to update `thumb` value when input new value. Do you want this input to be placed under the thumb? Cause currently Tick offset is measured automatically. Mix and min tick are placed right and left, others somewhere in between according to space available. Otherwise `Thumb` subcomponent may contain such input itself, so it will be always under thumb.

ericsoco:
Ok, that makes sense. It's less important to our use cases to be able to see the numeric value adjacent to the thumb, and more important to:
- see the numeric value of the thumb's current location
- be able to enter a specific numeric value

So I guess we'd override the Tick component to display an input for the min and max ticks only.

PR #205, opened 2018-09-07
feat(button): initial button component implementation

abmai:
### Button Component

This is the initial implementation for `Buttons`, outlined in https://github.com/uber-web/baseui/pull/164

I already had much of this code implemented elsewhere; and this is blocking `Pagination` so trying to fast-track it through.

[Storybook Preview](https://deploy-preview-205--baseui.netlify.com/?selectedKind=Button&selectedStory=Buttons&full=0&addons=1&stories=1&panelRight=0&addonPanel=REACT_STORYBOOK%2Freadme%2Fpanel)

#### Not In Scope

* Button Groups
* Special kind for `spinner`
* A11Y Check-thru
* Design Review


rfcs/button-component.md
```diff
@@ -80,7 +80,7 @@ Rough DOM composition:
   * `EndEnhancer: ReactComponent` - Optional
 * `size: 'default' | 'compact'`
   Defines the size of a button.
-* `shape: 'round' | 'square'`
+* `shape: 'default' | 'round' | 'square'`
```
abmai: Having an additional `default` shape allows us to cater specific styling to the truly `square` shape:

image

kahwee:
There's one violation in the Accessibility panel about the button background color contrast.

https://dequeuniversity.com/rules/axe/2.6/color-contrast?application=axeAPI

Place foreground as white, background as `30, 102, 240`

src/button/examples.js
```diff
@@ -0,0 +1,175 @@
...
+};
+
+export default {
+  [examples.BUTTON]: function Story1() {
+    return (
+      <ButtonContainer>
+        <ButtonRow>
+          <Button>Primary</Button>
+        </ButtonRow>
+        <ButtonRow>
```
kahwee: I tried this:

```
          <Button $as="a" href="#" kind={KIND.secondary}>
            Secondary
          </Button>
```

There's an underline for the `a` element. Is there any guidelines on whether the underline should be there? Or should we set a `text-decoration: none`.

blackswanny: `$as` is not supported. Imho, it shouldn't, cause button has exact semantic tag <button>, which serves all needs. If we need to have `a`, we can wrap button in this `<a>` tag

src/button/styled-components.js
```diff
@@ -0,0 +1,115 @@
...
+          paddingLeft: $theme.sizing.scale600,
+          paddingRight: $theme.sizing.scale600,
+        }),
+    // Kind style override
+    ...getStyleForKind({$theme, $kind}),
+  }),
+);
+
+export const EndEnhancer = styled('div', ({$theme}) => ({
+  display: 'flex',
```
blackswanny: end element should dock to right always. Middle element can have `flex=auto` or there are other tricks
https://www.google.com/search?q=flex+last+element+right

src/button/styled-components.js
```diff
@@ -0,0 +1,115 @@
...
+// @flow
+import {styled} from '../styles';
+import {KIND, SIZE, SHAPE} from './constants';
+
+export const BaseButton = styled(
+  'button',
+  ({$theme, $size, $kind, $shape}) => ({
+    ...$theme.typography[$size === SIZE.compact ? 'font200' : 'font300'],
+    display: 'flex',
+    alignItems: 'center',
```
blackswanny: so all elements are centered, whereas first and last elements should stick to borders
I think another approach should work:
enhancers should have `flex=1`, and middle should have `flex=auto`

src/button/examples.js
abmai: `$as` works here because the `Button` itself is a styled component. We can either always apply `text-decoration: none` or leave it up to the user.

```js
<Button $as="a" style={{textDecoration: 'none'}} ... />
```

blackswanny: which framework supports `$as`? Styletron or React?
Imho, changing the very basic tag of component may cause issues as tags have their default behavior, styles and semantic meaning for browser

abmai: It's a `styletron` feature: https://github.com/styletron/styletron/tree/master/packages/styletron-react#as

Again, since this is not an API we offer and you would have to understand how `Button` is built to do this, I don't think we need to support it unless it's easy to do so.

schnerd: If we want to support buttons as links, I'd rather we avoid the whole `$as="a"` nonsense and just explicitly support an `href` prop that causes it to start using an anchor internally. It looks like [antd does this too](https://ant.design/components/button/).

schnerd: But either way, seems fine to add textDecoration: 'none' as a default style.

src/button/types.js
```diff
@@ -0,0 +1,26 @@
...
+
+export type OverridesT = {
+  BaseButton?: OverrideT<*>,
+  StartEnhancer?: OverrideT<*>,
+  EndEnhancer?: OverrideT<*>,
+};
+
+export type ButtonPropsT = {
+  startEnhancer?: () => React$Node | React$Node,
+  endEnhancer?: () => React$Node | React$Node,
```
schnerd: Why the double `React$Node`?

Also, can you educate me on what the difference between React$Node and React.Node from `import * as React`? Seems like we use both in various places.

src/button/styled-components.js
blackswanny: order - yes. But are you sure, that you will not have this case when right side will be of too much padding?
image

src/button/styled-components.js
```diff
@@ -0,0 +1,115 @@
...
+        },
+        ':active:enabled': {
+          backgroundColor: $theme.colors.mono500,
+        },
+      };
+    case KIND.minimal:
+      return {
+        color: $theme.colors.primary,
+        backgroundColor: 'transparent',
+        ':hover:enabled': {
```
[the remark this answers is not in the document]
abmai: Hover/active styles shouldn't show up for disabled, at least from what I can see in the design. This helps ensure that these only show up when the element is not in a disabled state.

src/button/styled-components.js
```diff
@@ -0,0 +1,115 @@
...
+            $theme.sizing[$size === SIZE.compact ? 'scale400' : 'scale500'],
+        }
+      : {
+          paddingTop:
+            $theme.sizing[$size === SIZE.compact ? 'scale200' : 'scale300'],
+          paddingBottom:
+            $theme.sizing[$size === SIZE.compact ? 'scale200' : 'scale300'],
+          paddingLeft: $theme.sizing.scale600,
+          paddingRight: $theme.sizing.scale600,
+        }),
```
schnerd: image

The padding feels a bit off here compared to the designs. I suppose in the designs the buttons of a minimum with, but it still feels like theres too much top/bottom padding compared to left and right. Can we have lily take a pass at these styles when she's back next week?

src/button/button.js
```diff
@@ -0,0 +1,61 @@
...
+
+export default function Button({
+  children,
+  disabled,
+  startEnhancer,
+  endEnhancer,
+  overrides = {},
+  size = SIZE.default,
+  kind = KIND.primary,
+  shape = SHAPE.square,
```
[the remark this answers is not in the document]
abmai: You mean to explicitly do this?
```
Button.defaultProps = {
  ...
};
```

src/button/button.js
```diff
@@ -0,0 +1,61 @@
...
+  const {component: EndEnhancer, props: endEnhancerProps} = getOverrideObject(
+    overrides.EndEnhancer,
+    StyledEndEnhancer,
+  );
+  return (
+    <BaseButton
+      disabled={disabled}
+      $size={size}
+      $kind={kind}
+      $shape={shape}
```
nadiia: The tree above props should probably be shared and passed to all subelements.

src/button/button.js
```diff
@@ -0,0 +1,61 @@
...
+      {...baseButtonProps}
+      {...restProps}
```
nadiia: I feel like override props (`{...baseButtonProps}`) should go last.

src/button/button.js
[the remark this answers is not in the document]
schnerd: Yes. 

It can be useful if someone wants to create their own button wrapper because they can do things like:
```
FancyButton.defaultProps = Button.defaultProps
```

Additionally, some automated documentation tools will read defaultProps when generating documentation (though maybe they can also read default param values too?)

src/button/types.js
blackswanny: afaik , React.Node comes from React files itself, whereas React$Node may come from flow-typed folder ( I don't see it, but maybe somewhere in deeper imports)
To use React.Node we need to import like this only:
`import React from 'react';`
https://flow.org/en/docs/react/types/#toc-react-node

schnerd: Makes sense. Curious if there is a "best practice" for which one to use.

src/button/styled-components.js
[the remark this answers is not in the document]
blackswanny: yes. But did you test, that just `:hover` will apply styles for `disabled` also?

src/button/examples.js
nadiia: +1 for adding `textDecoration` style by default. I think overriding a base tag is fine and considered an advanced usage so whoever does it needs to understand the consequences.

src/button/button.js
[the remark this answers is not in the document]
abmai: Sounds good, will update!

src/button/types.js
abmai: So it's either a function that returns a `React$Node` here or a `React$Node` itself. I was trying to follow the convention from `Input` here. Though maybe we should also talk about what we really expect for these `enhancers`?

src/button/styled-components.js
abmai: Yeah definitely needs a design review. I can also tighten some of the padding here first, will take a pass through.

src/button/button.js
abmai: Good point! Will update

src/button/styled-components.js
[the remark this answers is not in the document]
abmai: Yeah a regular `:hover` will work even with `disabled={true}`.

image

src/button/styled-components.js
```diff
@@ -0,0 +1,115 @@
...
+        },
+      };
+    case KIND.minimal:
+      return {
+        color: $theme.colors.primary,
+        backgroundColor: 'transparent',
+        ':hover:enabled': {
+          backgroundColor: $theme.colors.mono200,
+        },
+        ':active:enabled': {
```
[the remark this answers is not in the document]
abmai: This is the one thing that was not clear to me...the only discussion I could follow on StackOverflow seemed to lean towards *not* assuming that `:active` and `:disabled` are mutually exclusive:

https://stackoverflow.com/questions/12591966/html-disabled-button-getting-active-css-pseudo-class

The spec also sucks. Thoughts?

src/button/examples.js
abmai: Kewl. Let's do that. We can follow up with the `href` + auto `a` tag later.

src/button/types.js
blackswanny: this one is more common, it includes them both, function also
https://github.com/facebook/flow/blob/v0.79.1/lib/react.js#L152

src/button/styled-components.js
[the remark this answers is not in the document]
blackswanny: @abmai check experimentally. Try to focus by `tab` on this element while disabled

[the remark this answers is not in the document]
abmai: Yes but is that consistent everywhere? I guess the other question is, does it hurt to be explicit here and do `:active:enabled`?

src/button/types.js
abmai: You mean we should switch to accepting components for `enhancers`?

abmai:
@kahwee I'm gonna save the `a11y` checks for a follow up PR/task.

src/button/styled-components.js
```diff
@@ -0,0 +1,116 @@
...
+  'button',
+  ({$theme, $size, $kind, $shape}) => ({
+    ...$theme.typography[$size === SIZE.compact ? 'font200' : 'font300'],
+    display: 'flex',
+    alignItems: 'center',
+    justifyContent: 'space-between',
+    border: 'none',
+    borderRadius: $shape === SHAPE.round ? '50%' : $theme.borders.radius200,
+    textDecoration: 'none',
+    outline: 'none',
```
schnerd: Forgot to mention this – if we're removing outline, can we add another style for focus? When tabbing through buttons right now I can't tell which one is focused.

abmai: Thanks for bringing this up. It's most likely an a11y concern that I need to dig further into. Going to clump it with the a11y todos here: https://github.com/uber-web/baseui/issues/207

src/button/types.js
blackswanny: no, just use flow type, which is more common

PR #206, opened 2018-09-07
chore(select): remove unused label prop, fix error prop type

schnerd:
`label` prop was never actually used internally, and we have the FormControl component now.

`error` prop is never actually used as a string internally and is always casted as a boolean. accepting a string here could mislead the user into thinking we display it somewhere.


schnerd:
>will it be a major bump, or is it just a fix?

In theory if someone was passing an error string before then this new version would cause a flow error, but no functionality would be broken. Since it's just a flow change primarily and its unlikely that anyone is using this prop at the moment anyway, I'm inclined to just land and do a patch/minor bump.

PR #209, opened 2018-09-08
feat(select): update dropdown style for select mode

abmai:
### Select Component (Select Mode)

* Update dropdown style for select mode
* Remove usage of style helper functions from Menu (since `styled` can override other styled components properly now)
* Minor code refactors along the way

#### Preview

image

### DIFF
--- a/src/select/select.js
+++ b/src/select/select.js
@@ -5,6 +5,7 @@ This source code is licensed under the MIT license found in the
 LICENSE file in the root directory of this source tree.
 */
 // @flow
+/* global document */
 import * as React from 'react';
 import {findDOMNode} from 'react-dom';
 import {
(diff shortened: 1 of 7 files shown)


src/select/select.js
```diff
@@ -76,17 +77,19 @@ class Select extends React.Component<PropsT, StatelessStateT> {
   }
 
   componentDidMount() {
-    /* eslint-disable-next-line cup/no-undef  */
-    document.addEventListener('click', this.handleClickEvent, {
-      capture: true,
-    });
+    if (__BROWSER__) {
```
nadiia: Is this check needed for linting?

abmai: Following convention in other components where we only attach DOM events when on the browser. Though looking at it now, I think `componentDidMount` is only called on client-side anyway so maybe we can forgo this check here?

PR #214, opened 2018-09-10
feat(pagination): initial component implementation (part 3)

abmai:
### Pagination Component (part 3 / last part)

This is the last part of the Pagination PR series. It includes:

* Index w/ exports
* Stateful Container / Stateful Pagination
* Storybook Stories
* Modifications to the RFC (WIP pending review on [part 2](https://github.com/uber-web/baseui/pull/213))

We can also see the storybook preview only on this PR.

#### Preview


KevinGrandon:
Just letting you know that it seems the arrow is broken in Firefox. Probably something about using the SVG as a data URI. Perhaps something needs to be escaped?

image

src/pagination/__tests__/stateful-container.test.js
```diff
@@ -0,0 +1,74 @@
...
+      ...getSharedProps(),
+      initialState: {
+        currentPage: 5,
+      },
+    };
+    const component = mount(<StatefulContainer {...props} />);
+    expect(component.state('currentPage')).toBe(5);
+  });
+
+  test('onPageChange', () => {
```
blackswanny: there are different human friendly names for unit test with explanation, why it's needed.
This one doesn't seems to comply with any of those. It contains 3 `expect`, which do not belong to one logical `assert`, they check for different things
https://dzone.com/articles/7-popular-unit-test-naming

abmai: thanks for calling out my laziness :p, will update the name to be clearer here

abmai:
> Just letting you know that it seems the arrow is broken in Firefox. Probably something about using the SVG as a data URI. Perhaps something needs to be escaped?

Thanks for pointing this out! I copied the method from `Select` and looks like it's an issue for that component too. It's easy to fix for `Pagination` since I can turn it into an SVG. But maybe a more generic fix is needed for `Select`? cc @blackswanny

nadiia:
Need to set a z-index for the options list (`ul`) since the dropdown items' background with no rounded corners on hover stacks on top of rounded corners of the list.

abmai:
> Need to set a z-index for the options list (ul) since the dropdown items' background with no rounded corners on hover stacks on top of rounded corners of the list.

It's an issue inside `Menu` i believe, setting `overflow: auto` on `List` will fix it, will throw up a separate PR for that.

src/pagination/examples.js
```diff
@@ -0,0 +1,20 @@
...
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+// @flow
+import * as React from 'react';
+import StatefulPagination from './stateful-pagination';
+
+export const suite = 'Pagination Test Suite';
+export const examples = {
+  STATEFUL_PAGINATION: 'Stateful Pagination',
```
nadiia: Shall we add a stateless example here as well? And an advanced example with overrides.

src/pagination/stateful-container.js
```diff
@@ -0,0 +1,59 @@
...
+
+  // Internal set state function that will also invoke stateReducer
+  internalSetState(
+    changeType: $Keys<typeof STATE_CHANGE_TYPE>,
+    changes: StatefulContainerStateT,
+  ) {
+    const {stateReducer} = this.props;
+    if (stateReducer) {
+      this.setState(stateReducer(changeType, changes, this.state));
+    }
```
nadiia: The `.setState` should be called in any case, even if a `stateReducer` is not provided

abmai:
> Need to set a z-index for the options list (ul) since the dropdown items' background with no rounded corners on hover stacks on top of rounded corners of the list.

Just kidding, you are correct - seems like `z-index` is the only way to fix this for now. I'll just go ahead and include it here.

abmai:
updated

PR #215, opened 2018-09-10
fix(button): use inline-flex instead of flex

schnerd:
By default, native `<button>` elements are `display: inline-block`. Our new Button components are `display: flex`, presumably so that the internal enhancers can be aligned properly. However, this means that two buttons next to each other will be displayed stacked instead of inline:

image

Additionally, in the example above `text-align: right;` was supposed to be applied but the buttons are aligned left, another artifact of using `display: flex`.

This PR changes button to be `display: inline-flex` by default, which should be more aligned with the behavior that users would expect from a native button.


abmai:
Nice catch, thanks for the fix!

PR #217, opened 2018-09-10
docs(datepicker): add datepicker RFC

blackswanny:
#### Minor: add datepicker RFC


rfcs/datepicker-component.md
```diff
@@ -0,0 +1,130 @@
...
+import {DatePicker} from 'baseui/datepicker';
+
+export default () => {
+  return <DatePicker
+    initialState={{
+      value: [new Date(2018, 1, 15), new Date(2018, 8, 25)]
+    }}
+    isRange={true}
+    dateFormat={'MM/DD/YY'}
+    minDate={new Date(2000, 1, 1)}
```
gergelyke: can you add it to the DatePicker API? also, should we add defaults too?

blackswanny: thank you. That is old example, will change it soon

rfcs/datepicker-component.md
```diff
@@ -0,0 +1,130 @@
...
+    maxDate={new Date(2020, 1, 1)}
+    displayMonths={2}
```
gergelyke: do we want to document this one too?

blackswanny: ok

nadiia: Here are some additional ideas on this option naming: `rangeMode | rangeType | rangeDisplayMode | monthRange: ('single' | 'double') | (1 | 2)`, or can do something like `rangeDoubleMode` to accept a boolean value. Don't have a strong name preference here tho.

blackswanny: monthRange so far

gergelyke:
@schnerd @abmai can you take a look too?

rfcs/datepicker-component.md
```diff
@@ -0,0 +1,148 @@
...
+* `StyledMonthSelect`
+* `StyledYearSelect`
+* `StyledMonthGrid`
+* `StyledMonthGridLabel`
+* `StyledMonthGridDay`
+* `StyledQuickSelect`
+
+## `DatePicker` API
+
+* `isRange: boolean` - Optional
```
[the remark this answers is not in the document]
blackswanny: it means how many values to select, one or two. So there is no ambiguity here yet

rfcs/datepicker-component.md
```diff
@@ -0,0 +1,148 @@
...
+* `value: Date | Array<Date>` - Optional
+  Selected `Date`. If `isRange` is set, `value` is an array of 2 values.
+* `minDate: Date` - Optional
+  minimum date available for selection. Default is `Jan 01 1970`
+* `maxDate: Date` - Optional
+  maximum date available for selection. Default is current date + 3 years.
+* `disabled: boolean`:
+  Disable control from being changed
+* `monthRange: number`
+  Amount of minth grids to be shown. Accepts 1 or 2. Default is `1`
```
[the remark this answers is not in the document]
blackswanny: they have different meaning. This one is just to tell how many month to display, not select.
So we can do boolean, but I saw examples of 3 months as well. For big screens is totally fine, and from implementation perspective won't cost us anything.
@nguyensomniac what do you think?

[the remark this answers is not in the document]
nguyensomniac: We don't have an example of more than 2 months in the base UI library. Is there a use case you had in mind?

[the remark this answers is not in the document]
blackswanny: @nguyensomniac i dodn't find example among those web sites I use, but if you google, there is a demand for such datepickers, even a grid of months. It usually is needed for long range dates and more exact scheduling
https://www.google.com/search?q=3+month+datepicker&source=lnms&tbm=isch&sa=X&ved=0ahUKEwi_85LwgsXdAhXYJTQIHXpwBiMQ_AUIDigB&biw=2560&bih=1241

rfcs/datepicker-component.md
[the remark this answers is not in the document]
caseyhebebrand: Ok I  just think it is more specific nomenclature since you can also select a month range

rfcs/datepicker-component.md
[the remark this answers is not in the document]
caseyhebebrand: If we want to support multiple months in the future then we can just increase the range of numbers that `monthRange` accepts. If we only support 1 or 2 months then I am proponent for the `isMonthRange: boolean` option.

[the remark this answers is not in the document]
blackswanny: ok, I changed copy not to be confused

rfcs/datepicker-component.md
```diff
@@ -0,0 +1,148 @@
...
+  MonthGridDay: (props: {[string]: any}) => React$Node, QuickSelect: (props: {[string]: any}) => React$Node}`
+  * `Root` - parent element of whole datepicker
+  * `DateLabel` - label above date input
+  * `DateInput` -  date input showing current selected date
+  * `MonthAndYearBar` - a bar\panel showing current selected month and year
+  * `MonthSelect` - select component to pick a month
+  * `YearSelect` - select component to pick a year
+  * `MonthGrid` - a grid of days in month
+  * `MonthGridLabel` - label for day of week in grid
+  * `MonthGridDay` - day number in a grid
```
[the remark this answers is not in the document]
blackswanny: yes, it's a default behavior. So range consists of one date only and inRange should be `true` for it. Do you have access to UX design of BaseUI?
cc @nguyensomniac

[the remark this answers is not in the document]
junsuhlee: I see, is there way if we want to highlight one date without making that minDate-maxDate? We probably can have some use case to customize highlighted date cell by date, if we can have access to date on MonthGridDay, what's your thought on here?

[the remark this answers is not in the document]
blackswanny: min and max dates are not for the range mode, they only indicate how far you can swipe left and right to pick the date. `isRange` actually indicates what would be the pick up, is it one date or a range, so highlight works correspondingly. Currently we expose next thisn to `MonthGridDay`. It might be enough to calculate exact date as you may also need to know month and year. However, if you see issues with props below, we can add more

* `$inRange: boolean`
  For `MonthGridDay` only. True if day is in range between min and max selected date value
* `$value: number`
  For `MonthGridDay` only. Number of day in month

rfcs/datepicker-component.md
```diff
@@ -0,0 +1,148 @@
...
+import * as React from 'react';
+import {DatePicker} from 'baseui/datepicker';
+
+export default () => {
+  return <DatePicker
+    initialState={{
+      value: [new Date(2018, 1, 15), new Date(2018, 8, 25)]
+    }}
+    isRange={true}
+    dateFormat={'MM/DD/YY'}
```
ajbogh: How would this work with I18N? Does the string parsing include time options?

rfcs/datepicker-component.md
```diff
@@ -0,0 +1,148 @@
...
+### Advanced usage
+
+```javascript
+import * as React from 'react';
+import {DatePicker} from 'baseui/datepicker';
+
+export default () => {
+  return <DatePicker
+    initialState={{
+      value: [new Date(2018, 1, 15), new Date(2018, 8, 25)]
```
ajbogh: Why a value array? Is this min/max? What about an object with min/max keys?

rfcs/datepicker-component.md
```diff
@@ -0,0 +1,148 @@
...
+* `StyledQuickSelect`
+
+## `DatePicker` API
+
+* `isRange: boolean` - Optional
+  If set, datepicker is to select 2 dates range, so there are two date input and `value` comes as Array of two dates, min and max. Otherwise only one date is selected. Default is `false`
+* `dateFormat: string` - Optional
+  Date format that is accepted by `Date.parse` function. RFC2822 tools.ietf.org/html/rfc2822#section-3.3 standard. Default is `DD/MM/YYYY`
+* `value: Date | Array<Date>` - Optional
+  Selected `Date`. If `isRange` is set, `value` is an array of 2 values.
```
ajbogh: "is an array of 2 Date values: [min, max]"

rfcs/datepicker-component.md
```diff
@@ -0,0 +1,148 @@
...
+* `minDate: Date` - Optional
+  minimum date available for selection. Default is `Jan 01 1970`
+* `maxDate: Date` - Optional
+  maximum date available for selection. Default is current date + 3 years.
```
ajbogh: Is the default max limit due to some other system or is it imposed by this component? The question I have is why is there a max at all?

rfcs/datepicker-component.md
```diff
@@ -0,0 +1,148 @@
...
+* `disabled: boolean`:
+  Disable control from being changed
+* `monthRange: number`
+  Amount of minth grids to be shown. Accepts number of visible months(mostly 1 or 2). Default is `1`
```
ajbogh: s/minth/month/

ajbogh: Suggestion, change this sentence so it makes more sense:

"The number of calendar months to show at once. Default is `1`"

rfcs/datepicker-component.md
```diff
@@ -0,0 +1,148 @@
...
+* `overrides: {Root: (props: {[string]: any}) => React$Node, DateLabel: (props: {[string]: any}) => React$Node,
+ DateInput: (props: {[string]: any}) => React$Node, MonthAndYearBar: (props: {[string]: any}) => React$Node,
+  MonthSelect: (props: {[string]: any}) => React$Node, YearSelect: (props: {[string]: any}) => React$Nod,
+  MonthGrid: (props: {[string]: any}) => React$Node, MonthGridLabel: (props: {[string]: any}) => React$Node,
+  MonthGridDay: (props: {[string]: any}) => React$Node, QuickSelect: (props: {[string]: any}) => React$Node}`
```
ajbogh: Please fix the formatting on this, make it an object with line breaks and spacing.

rfcs/datepicker-component.md
```diff
@@ -0,0 +1,148 @@
...
+  * `Root` - parent element of whole datepicker
```
ajbogh: Shouldn't root be optional? If it's not provided then the root is the current elem, otherwise it should use the Root property.

rfcs/datepicker-component.md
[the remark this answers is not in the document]
junsuhlee: Yes, I wonder if we can support highlight a exact date without selecting? I see we can definitely highlight the date or range we've selected, but want to see if we can highlight without selecting.

On our service, we often have use case for 'recommending' a date by highlighting a date - which is not necessarily today or any calculable date but it can be a random date.

rfcs/datepicker-component.md
blackswanny: it's imposed. In some cases they don't want user to input already not valid date for some use case. So why not to limit by max date, otherwise they will have to check and validate it later.

rfcs/datepicker-component.md
blackswanny: ok

blackswanny: ok

rfcs/datepicker-component.md
blackswanny: ok

rfcs/datepicker-component.md
blackswanny: all overrides are optional

rfcs/datepicker-component.md
```diff
@@ -0,0 +1,148 @@
...
+  * `DateLabel` - label above date input
+  * `DateInput` -  date input showing current selected date
+  * `MonthAndYearBar` - a bar\panel showing current selected month and year
+  * `MonthSelect` - select component to pick a month
+  * `YearSelect` - select component to pick a year
+  * `MonthGrid` - a grid of days in month
+  * `MonthGridLabel` - label for day of week in grid
+  * `MonthGridDay` - day number in a grid
+  * `QuickSelect` - set of buttons or any other components. Default contains quick selection of current week, month, year and other common ranges
```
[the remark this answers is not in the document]
blackswanny: they are mentioned right above in `override` declaration

rfcs/datepicker-component.md
blackswanny: ok, but not min and max. Only those to be picked or set as default values

rfcs/datepicker-component.md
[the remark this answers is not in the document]
blackswanny: would you like to highlight with another color?
so those date will have $value for day of month and $inRange will be false
It still should be enough for override logic to make it happen.
@nguyensomniac  do you want to add it as default feature?

blackswanny:
@ajbogh 
> +    initialState={{
> +      value: [new Date(2018, 1, 15), new Date(2018, 8, 25)]
> Why a value array? Is this min/max? What about an object with min/max keys?
as these are not min and max , but just values selected ( or preselected) they don't have these names. Also this can have one value only, not an array if it's not range.

blackswanny:
@ajbogh 
> +    isRange={true}
> +    dateFormat={'MM/DD/YY'}
> How would this work with I18N? Does the string parsing include time options?

It doesnt include time as it's only date picker, not time picker. 
Component currently doesn't work with i18n or locale features as it is delegated to outer code. It only takes care of proper visual representation of dates

rfcs/datepicker-component.md
```diff
@@ -0,0 +1,161 @@
...
+               Root: (props: {[string]: any}) => React$Node,
+               DateLabel: (props: {[string]: any}) => React$Node,
+               DateInput: (props: {[string]: any}) => React$Node,
+               MonthAndYearBar: (props: {[string]: any}) => React$Node,
+               MonthSelect: (props: {[string]: any}) => React$Node,
+               YearSelect: (props: {[string]: any}) => React$Nod,
+               MonthGrid: (props: {[string]: any}) => React$Node,
+               MonthGridLabel: (props: {[string]: any}) => React$Node,
+               MonthGridDay: (props: {[string]: any}) => React$Node,
+               QuickSelect: (props: {[string]: any}) => React$Node
```
nadiia: The type of all the overrides above should be `{props: {}, style: {}, component: ReactComponent} | ReactComponent`

rfcs/datepicker-component.md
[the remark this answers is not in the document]
nadiia: I think it makes sense to highlight a specific day, a good example of the day highlighting would be a current day and maybe we can make it customizable. The highlighted should probably have a separate style from what is selected since `hightlighted` doesn't mean `selected`.

rfcs/datepicker-component.md
```diff
@@ -0,0 +1,161 @@
...
+  * `YearSelect` - select component to pick a year
+  * `MonthGrid` - a grid of days in month
+  * `MonthGridLabel` - label for day of week in grid
+  * `MonthGridDay` - day number in a grid
+  * `QuickSelect` - set of buttons or any other components. Default contains quick selection of current week, month, year and other common ranges
+* `onChange: func(event, value)` - Optional
+  handler for events on trigger element, current selected date `value` (single or 2 dates)
+* `onMonthSelect: func(event, month)` - Optional
+  handler for events on trigger element, current month selected in Select dropdown. From 0 to 11.
+* `onYearSelect: func(event, year)` - Optional
```
nadiia: We'd also need an onChange handler when an actual date selected

rfcs/datepicker-component.md
```diff
@@ -0,0 +1,161 @@
...
+* `StyledMonthGrid`
+* `StyledMonthGridLabel`
+* `StyledMonthGridDay`
+* `StyledQuickSelect`
+
+## `DatePicker` API
+
+* `isRange: boolean` - Optional
+  If set, datepicker is to select 2 dates range, so there are two date input and `value` comes as Array of two dates, min and max. Otherwise only one date is selected. Default is `false`
+* `dateFormat: string` - Optional
```
nadiia: How about we make it `string | function` type? So a custom format function can be passed in.

rfcs/datepicker-component.md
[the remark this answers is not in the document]
junsuhlee: Yes it would be different color other than color for 'selected'. Correct, highlighing today is one example, but we can also have different date highlighted.

rfcs/datepicker-component.md
blackswanny: should we tell that `ReactComponent` may be declared as `(props: {[string]: any}) => React$Node` ?

rfcs/datepicker-component.md
blackswanny: `onChange` is couple lines above

rfcs/datepicker-component.md
blackswanny: ok

rickyp-uber:
Looking through the RFC it does look like the [airbnb/react-dates](https://github.com/airbnb/react-dates/) component supports most (if not all, only did a quick scan) of the functional scenarios requested here.

The only question would be if we can get it styled correctly in a way you wish.

Is this a consideration for the implementation of this RFC?

blackswanny:
@rickyp-uber ok, I checked it. Just some example. For instance we need to change that `top` margin of DayPicker
image

which is a 66px inlined style

image

which is made by this line of code

https://github.com/airbnb/react-dates/blob/096453d2d9405e61d085f9b0d0a44a548e1b0e87/src/components/DateRangePicker.jsx#L419

so we need to override two vars `inputHeight` and `verticalSpacing`
1) https://github.com/airbnb/react-dates/blob/096453d2d9405e61d085f9b0d0a44a548e1b0e87/src/utils/getInputHeight.js   - we need to override a dozen of props to make sure it will work as we expect

2) https://github.com/airbnb/react-dates/blob/096453d2d9405e61d085f9b0d0a44a548e1b0e87/src/constants.js#L32 - is impossible to override, afaik

rfcs/datepicker-component.md
benshope: @blackswanny hey!  just following up on the focus group - is this RFC ready for feedback?  I'd also be happy to contribute

benshope:
@rickyp-uber it sounds like you're interested in using an uber style version of `react-dates`.  That is likely a great choice, but the best way to do it might be to create an Uber-style fork of AirBnb's react-dates in a separate repo from `baseui` because `react-dates` uses a single-field input, MomentJS, and it scrolls horizontally

A repo that has done a great job with choosing what they want internally and externally is https://material-ui.com/demos/pickers/.   Material UI's approach is smart because they recommend to people the libraries/functionality they want but they only allow solid choices in terms of accessibility and mobile devices inside their library.  Material UI's approach is also easy to implement because they have exercised caution about what they support

rickyp-uber:
> https://github.com/airbnb/react-dates/blob/096453d2d9405e61d085f9b0d0a44a548e1b0e87/src/utils/getInputHeight.js - we need to override a dozen of props to make sure it will work as we expect
Sure as expected because the styling is extremely different. The idea is though the overrides maybe many, it's still less than the actual cost of developing the component.

> https://github.com/airbnb/react-dates/blob/096453d2d9405e61d085f9b0d0a44a548e1b0e87/src/constants.js#L32 - is impossible to override, afaik
It's used as a defaultProp so to override it you would just pass in `verticalSpacing` as a prop to either the Single or DateRange pickers

It was more just a suggestion... date pickers are complicated and there is a lot of logic going into them. Us contributing to Airbnb's and customizing theirs means two large companies are helping support a single custom and complicated component instead of two large companies supporting two separate but equally custom and complicated components.

If it doesn't make sense then it doesn't make sense but from reading the RFC and the API, I didn't see any absolute blockers that would require us to create a whole new separate component other than ownership.

@benshope Creating a fork and making sure it's up-to-date works too. I would argue that material-ui's approach is missing quite a bit of key functionality that comes with moment.js; an example is (localized calendar support)[https://github.com/chingyawhao/material-ui-next-pickers/blob/master/src/calendar.tsx#L283]. To my knowledge, the most accurate locale date library that exists current is moment.js so unless part of this RFC is implementing our own date/time locale support library maybe that should be listed as a dependency in the RFC to make it clear as implementing our own would be a large undertaking.

I think the point is make sure we do a cost analysis (short and long term) of
1. Building the whole thing ourselves from scratch
1. Forking someone else's and using it as a base
1. Contributing to someone else's and utilizing those changes
For what generally looks like a simple control but ends up becoming extremely complicated when actually building one (mostly due to globalization)

gergelyke:
As we'd like to have the override injection for the datepicker as we have for the rest of the components, I don't think we can easily contribute that concept / fork & change. Is this statement correct @nadiia @blackswanny ?

blackswanny:
@gergelyke yes. We may try to apply styles to make it fit with our design, which is also not so trivial. However, applying functional parts like overrides seems to be even more complex task.

rfcs/datepicker-component.md
```diff
@@ -0,0 +1,159 @@
...
+### Advanced usage
+
+```javascript
+import * as React from 'react';
+import {DatePicker} from 'baseui/datepicker';
+
+export default () => {
+  return <DatePicker
+    initialState={{
+      value: [new Date(2018, 1, 15), new Date(2018, 8, 25)]
```
benshope: why is `value` inside `initialState` and not a root-level prop?

rfcs/datepicker-component.md
```diff
@@ -0,0 +1,159 @@
...
+    }}
+    isRange={true}
```
benshope: a range picker is essentially two individual pickers wired together as controlled components so it could be helpful to create the individual inputs first before considering range.  I just built a date range component so I can send an example if this comment is cryptic

rfcs/datepicker-component.md
```diff
@@ -0,0 +1,159 @@
...
+    dateFormat={'MM/DD/YY'}
```
benshope: d3's date format is also worth taking a glance at because there is potentially a reason why they decided to use a different format

rfcs/datepicker-component.md
```diff
@@ -0,0 +1,159 @@
...
+    minDate={new Date(2000, 1, 1)}
```
benshope: nit: perhaps `min` rather than `minDate`

rfcs/datepicker-component.md
```diff
@@ -0,0 +1,159 @@
...
+    maxDate={new Date(2020, 1, 1)}
+    monthRange={2}
+    onChange={(event, value)=>{console.log(value)}}
+    getMonthGridDay={props => <div style={{
```
benshope: is this inside an `overrides` prop or a root level prop?

rfcs/datepicker-component.md
```diff
@@ -0,0 +1,159 @@
...
+  Method, which returns current markupda for day in month. Default is number of month. Receives most useful properties, such as:
+  * `$inRange: boolean`
+    True if day is in range between min and max selected date value
+  * `$value: number`
+    `Date` of this day.
+* `disabled: boolean`:
+  Disable control from being changed  
+* `onChange: func(event, value)` - Optional
+  handler for events on trigger element, current selected date `value` (single or 2 dates)
+* `onMonthSelect: func(event, month)` - Optional
```
benshope: why are `onMonthSelect` and `onYearSelect` not covered by `onChange`?

rfcs/datepicker-component.md
```diff
@@ -0,0 +1,159 @@
...
+export default () => {
+  return <DatePicker
+    initialState={{
+      value: [new Date(2018, 1, 15), new Date(2018, 8, 25)]
+    }}
+    isRange={true}
+    dateFormat={'MM/DD/YY'}
+    minDate={new Date(2000, 1, 1)}
+    maxDate={new Date(2020, 1, 1)}
+    monthRange={2}
```
benshope: nit: might consider naming this something like `visibleMonths`

PR #221, opened 2018-09-11
fix select styles and extra features

blackswanny:
#### Minor: fix select styles and extra features


src/select/styled-components.js
```diff
@@ -33,6 +33,9 @@ export const Input = styled('input', props => {
   return {
     ...getInputStyles({...props, $size: SIZE.default}),
     cursor: 'pointer',
+    ':hover': {
+      cursor: 'pointer',
+    },
```
schnerd: Is there any reason to nest `cursor` rules inside of `:hover` anywhere in our codebase? It looks like here you're doing it to override a style from `Input`, but maybe we should just update the Input component styles to not use `:hover`.

blackswanny: another way is to override `$disabled` prop here, what I did right now

schnerd: Just curious–why is that a better choice? Wouldn't you still want some of the other disabled input styles (like text color and placeholder color)?

Is moving `cursor` rule out of `:hover` inside of `getInputStyles` not an option?

blackswanny: No, it's not. Now those two styles are compromised in Select. Changing code in original Input implementation just to comply with Select is a violation of inheritance. Input shouldn't know about needs of Select as long as Select can cope with it.

abmai: I don't think that is necessarily true. We developed most of these components in a silo and if the change makes reasonable sense when it comes to integrating with other components then it's worth making. Otherwise, our users will end up having to jump through the hoops just like we did now. That's speaking generally. But it is weird to see a `:hover` rule with `:cursor` inside of it. Maybe we can open a separate task for that?

schnerd: +1 abmai. 

We should be consistent about how we define `cursor` rules throughout baseui, and it seems pretty widely accepted to not nest them in `:hover`. In reality we're fixing Input, and the side effect of this is that Select (and probably many other components) also benefit.

src/select/styled-components.js
```diff
@@ -31,7 +31,7 @@ export const Input = styled('input', props => {
     sizing: {scale300},
   } = $theme;
   return {
-    ...getInputStyles({...props, $size: SIZE.default}),
+    ...getInputStyles({...props, $size: SIZE.default, $disabled: false}),
```
nadiia: The disabled is false by default. It looks weird that we use disabled input as a select in the first place and this change in addition to it can be confusing. I also added a change based on a valid comment from @schnerd and @abmai to move the cursor style on input out of `:hover`.

blackswanny: ok, I am removing this at all here.
However, Select needs Input in disabled state for read-only purpose. It totally complies with behavior of Input, functionally it is correct. Agree that Select is not in disabled state, but Input is not internal part of it, it's just reused by Select, so it can be adjusted to whatever needs by external Select consumer, which is read-only mode in this case. We already discussed amount of efforts and possible caveats to not set Input in disabled state and provide our own implementation instead. And seems like this effort is to be postponed for future improvements.

src/select/styled-components.js
blackswanny: ok, this code is no longer actual

src/select/styled-components.js
nadiia: I believe moving out the cursor style out of `:hover` hasn't fixed the issue here, so you still need to do the override for it below.

abmai:
The `TODO` thing will be fixed once you rebase with master

src/select/styled-components.js
```diff
@@ -37,6 +37,9 @@ export const Input = styled('input', props => {
     flexGrow: '1',
     paddingTop: scale300,
     paddingBottom: scale300,
+    ':hover': {
+      cursor: 'pointer',
+    },
```
schnerd: Did we add this back for some reason? seems like the `:hover` shouldn't be needed if we rebase with master.

blackswanny: ok, now i see it in master. removed again

PR #222, opened 2018-09-11
docs(accordion): add rfc for accordion

blackswanny:
#### Minor: add RFC for accordians

### DIFF
--- a/rfcs/accordion-component.md
+++ b/rfcs/accordion-component.md
@@ -1,0 +1,124 @@
+# Accordion Component
+
+## Usage
+
+```js
+import {Accordion, AccordionItem, Root, Toggle} from './index';
+
+export default () => {
+  return (
+    <div>
+      <Accordion>
+        <AccordionItem key={1} title={'First title'}><div>First content</div></AccordionItem>
+        <AccordionItem key={2} title={'Second title'}><div>Second content</div></AccordionItem>
+      </Accordion>
+    </div>
+  );
+};
+````
+
+```js
+import {Accordion, AccordionItem, Root, Toggle} from './index';
+import {styled} from '../styles/index';
+
+const CustomRoot = styled(Root, props => ({
+  textColor: 'red',
+}));
+
+const CustomToggle = styled(Toggle, props => ({
+  backgroundColor: props.$isExpanded ? 'green' : 'blue',
+}));
+
+export default () => {
+  return (
+    <div>
+      <Accordion
+        onToggle={(event, index, isExpanded) => {console.log('Accordian ' + index + ' is ' + (isExpanded ? 'expanded' : 'collapsed') )}}
+        overrides={{
+          Root: CustomRoot,
+          Toggle: CustomToggle,
+        }}
+        expanded={[false, true]}
+      >
+          <AccordionItem key={1} title={'First title'}><div>First content</div></AccordionItem>
+          <AccordionItem key={2} title={'Second title'}><div>Second content</div></AccordionItem>
+          <AccordionItem key={3} title={'Third title'}><div>Third content</div></AccordionItem>
+      </Accordion>
+    </div>
+  );
+};
+```
+
+## Exports
+
+* `Accordion`
+* `StatefulAccordion`
+* `StatefulAccordionContainer`
+* `AccordionItem`
+* `StyledRoot`
+* `StyledToggle`
+* `StyledTitle`
+* `StyledContent`
+
+## `Accordion` API
+
+* `children: Array<AccordionItem>` - Required
+  All accordion items in component. Each `AccordionItem` has:
+  * `title: React$Node` - title of accordion item
+  * `children: React$Node` - content of accordion item is represented as children of it correspondingly
+* `expanded: Array<boolean>` - Optional
+  Array of the same size as `children`(array of `AccordionItem`). If element is `true` corresponding accordion item is expanded. Default is `Array<false>`
+* `disabled: boolean`:
+  Disable control from being changed  
+* `overrides: {Root: (props: {[string]: any}) => React$Node, Toggle: (props: {[string]: any}) => React$Node, Title: (props: {[string]: any}) => React$Node, Content: (props: {[string]: any}) => React$Node}`
+  * `Root` container element to render.
+  * `Toggle` is the expand\collapse button for each accordion item.
+  * `Title` to render for each accordion item.
+  * `Content` to render for each accordion item.
+* `onToggle: func(event, index, isExpanded)` - Optional
+  handler for events on trigger element. `index` indicates number of accordion item toggled. `isExpanded` boolean to show if accordion item expands or collapses.
+
+## `StatefulAccordion` API
+
+* `initialState: {}`
+  Initial state of an uncontrolled accordion component.
+  * `expanded: Array<boolean>` initial state of array of the same size as `children`(array of `AccordionItem`). If element is `true` corresponding accordion item is expanded
+* `stateReducer: (type: text, nextState: {}, currentState: {}, e: any) => nextState`
+  A state change handler.
+  * `type` - state change type
+  * `nextState` - a new state changes that will be set
+  * `currentState` - current full state of the component
+* `children: Array<AccordionItem>` - Required
+  All accordion items in component. Each `AccordionItem` has:
+  * `title: React$Node` - title of accordion item
+  * `children: React$Node` - content of accordion item is represented as children of it correspondingly
+* `exclusive: boolean` - Optional
+  If set to true, it will exclusively expand only one accordion item. So next accordion item to be expanded right after current is collapsed. Default is `'false`  
+* `disabled: boolean` - Optional
+  Disable control from being changed
+* `onToggle: func(event, index, isExpanded)` - Optional
+  handler for events on trigger element. `index` indicates number of accordion item toggled. `isExpanded` boolean to show if accordion item expands or collapses.
+
+## `AccordionItem` API
+
+Represents content and title of each accordion item
+
+* `title: React$Node` - title of accordion item
+* `children: React$Node` - content of accordion item is represented as children of it correspondingly
+
+## Presentational components props API
+
+These properties are passed to every presentational (styled) component that is exported:
+
+* `$disabled: boolean`
+  Disable control from being changed
+* `$index: number`
+  For `Toggle`, `Title` and `Content` to indicate number of accordion item.
+* `$isExpanded: number`
+  For `Toggle`, `Title` and `Content` to indicate if accordion item is expanded.  
+
+## Accessibility
+
+Should support keyboard hotkeys: `tab` - to switch between accordion and expand the next one.
+Accessibility best practices for this component (`aria-expanded`, `aria-controls`, `aria-disabled`, `aria-level`, `role=presentation` for all component, `role=heading` and `role=region` for title and content of each accordion item)
+Please, see w3.org/TR/wai-aria-practices-1.1/examples/accordion/accordion.html


rfcs/accordians-component.md
```diff
@@ -0,0 +1,113 @@
+# Accordians Component
+
+## Usage
```
gergelyke: can you add a basic example section too?

blackswanny: ok

gergelyke:
@schnerd @nadiia @abmai can you take a look please?

rfcs/accordians-component.md
```diff
@@ -0,0 +1,141 @@
...
+* `StyledContent`
+
+## `Accordians` API
+
+* `accordians: Array<Item>` - Required
+  All accordians in component. Each `Item` has:
+  * `title: React$Node` - title of accordian
+  * `content: React$Node` - content of accordian
+* `autoExpand: boolean` - Optional
+  If set to true, it will auto expand next accordian right after current is collapsed. Default is `'false`
```
schnerd: I don't know if I've ever seen this behavior in an accordion before? I've see the case where opening the _next_ accordion closes the previous (which I called `exclusive` in the code example above), but not the case you're describing here.

rfcs/accordians-component.md
```diff
@@ -0,0 +1,141 @@
...
+            title: 'First title',
+            content: <div>First content</div>
+          },
+          {
+            title: 'Second title',
+            content: <div>Second content</div>
+          }
+        ]}
+      />
+    </div>
```
schnerd: What do you think about an API more like this:

```
<Accordion exclusive>
  <Panel title="This is panel header 1" key="1">
    <p>{text}</p>
  </Panel>
  <Panel title="This is panel header 2" key="2">
    <p>{text}</p>
  </Panel>
  <Panel title="This is panel header 3" key="3">
    <p>{text}</p>
  </Panel>
</Accordion>
```

Kind of similar to RadioGroup – accordion would communicate which panel is open through react context. The `exclusive: boolean` prop makes it so that only one panel can be open at once. Just feels little cleaner and more react-like than passing an array with react nodes inside of it.

@abmai @nadiia thoughts?

rfcs/accordians-component.md
```diff
@@ -0,0 +1,141 @@
...
+## Presentational components props API
+
+These properties are passed to every presentational (styled) component that is exported:
+
+* `$disabled: boolean`
+  Disable control from being changed
+* `$isHovered: boolean`
+  If component is hovered
+* `$isActive: boolean`
+  If component is active
```
schnerd: What does it mean for the Accordion to be "Active"?

rfcs/accordians-component.md
```diff
@@ -0,0 +1,141 @@
...
+  handler for events on trigger element. `index` indicates number of accordian toggled. `isExpand` boolean to show if accordian expands or collapses.
+
+## Presentational components props API
+
+These properties are passed to every presentational (styled) component that is exported:
+
+* `$disabled: boolean`
+  Disable control from being changed
+* `$isHovered: boolean`
+  If component is hovered
```
schnerd: I'm struggling to figure out a case when someone would want to use this. Should we just leave it out initially to keep things simple until a custom requests it?

rfcs/accordians-component.md
```diff
@@ -0,0 +1,141 @@
...
+  * `title: React$Node` - title of accordian
+  * `content: React$Node` - content of accordian
+* `autoExpand: boolean` - Optional
+  If set to true, it will auto expand next accordian right after current is collapsed. Default is `'false`  
+* `disabled: boolean` - Optional
+  Disable control from being changed
+* `expanded: Array<boolean>` - Optional
+  Array of the same size as `accordians`. If element is `true` corresponding accordian is expanded. Default is `Array<false>`
+* `onToggle: func(event, index, isExpand)` - Optional
+  handler for events on trigger element. `index` indicates number of accordian toggled. `isExpand` boolean to show if accordian expands or collapses.
```
schnerd: We're thinking about changing the signature of some of these handlers, feel free to weigh in here if you have thoughts: https://github.com/uber-web/baseui/issues/220

rfcs/accordians-component.md
```diff
@@ -0,0 +1,141 @@
+# Accordians Component
+
+## Usage
+
+```js
+import {Accordians, Root, Toggle} from './index';
```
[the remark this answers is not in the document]
blackswanny: I am confused by english meaning. `Accordian` appeared to comply with Figma naming.
A can rename it. For singular: does `Accordion` mean one section of component or the whole component. Afaik, accordion is only one of those in it.

rfcs/accordians-component.md
blackswanny: yes, this is what i meant

rfcs/accordians-component.md
blackswanny: I thought about it. But for Select we chose to provide it via attribute. For instance because content will not render by default. However, in this case we have to introduce one more abstract `Panel` and literally extract `title` to put it somewhere else, as title doesn't belong to content part. For me it looks as over-complication.

rfcs/accordians-component.md
blackswanny: removed. no meaning

rfcs/accordians-component.md
blackswanny: if we hover over title or action button, we may want to change colors ( this is what happens in Figma design right now)

rfcs/accordians-component.md
blackswanny: renamed to exclusive

rfcs/accordians-component.md
blackswanny: answered

rfcs/accordians-component.md
schnerd: Gotcha. I wonder if we can just do this via `:hover` rules for now?

blackswanny: that should work, removed

rfcs/accordians-component.md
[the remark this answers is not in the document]
schnerd: It's meant to resemble one of these things:

image

The accordion is the whole object, the panels sort of loosely correspond to the folds that collapse when you squeeze the accordion.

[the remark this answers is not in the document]
schnerd: @nguyensomniac should we change the name in the figma?

rfcs/accordians-component.md
schnerd: > But for Select we chose to provide it via attribute. For instance because content will not render by default.

We are passing it via attribute for Select, but there are some slight differences. Select could have hundreds or thousands of items, Accordion would probably have fewer. Also the select items are just POJOs, in this case you have react elements in the array.

> literally extract title to put it somewhere else

Assuming the markup for Panel looks like below:

```
  <Panel title="This is panel header 3" key="3">
    <p>{text}</p>
  </Panel>
```

It seems pretty trivial to map this to the following markup inside `Panel`

```js
<StyledPanel>
  <StyledPanelTitle>{title}</StyledPanelTitle>
  {isOpen && <StyledPanelContent>{children}</StyledPanelContent>}
</StyledPanel>
```

Though maybe I'm missing something.

rfcs/accordians-component.md
```diff
@@ -0,0 +1,137 @@
...
+* `expanded: Array<boolean>` - Optional
+  Array of the same size as `accordians`. If element is `true` corresponding accordian is expanded. Default is `Array<false>`
+* `disabled: boolean`:
+  Disable control from being changed  
+* `overrides: {Root: (props: {[string]: any}) => React$Node, Toggle: (props: {[string]: any}) => React$Node, Title: (props: {[string]: any}) => React$Node, Content: (props: {[string]: any}) => React$Node}`
+  * `Root` container element to render.
+  * `Toggle` is the expand\collapse button for each accordian.
+  * `Title` to render for each accordian.
+  * `Content` to render for each accordian.
+* `onToggle: func(event, index, isExpand)` - Optional
```
schnerd: `isExpanded` to match your code example?

rfcs/accordians-component.md
schnerd: Context certainly would complicate things though, I agree with that. `overrides` also becomes more tricky when we introduce another `Panel` component. I think there are trade-offs with both designs. Curious to get others opinions here.

blackswanny: Select items are not so POJO, they can even accept any React markup.
Although amount of items of accordion is small, but each of those can contain a huge piece of markup. For instance I made checkout flow with accordions, and each item contained a full experience of forms, cms content, etc.

rfcs/accordians-component.md
```diff
@@ -0,0 +1,137 @@
...
+        exclusive={true}
+      />
+    </div>
+  );
+};
+```
+
+## Exports
+
+* `Accordians`
```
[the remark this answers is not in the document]
nadiia: Seems like we do have stateful one, just need to include it in exports ^

rfcs/accordians-component.md
```diff
@@ -0,0 +1,137 @@
...
+  * `currentState` - current full state of the component
+* `accordians: Array<Item>` - Required
+  All accordians in component. Each `Item` has:
+  * `title: React$Node` - title of accordian
+  * `content: React$Node` - content of accordian
+* `exclusive: boolean` - Optional
+  If set to true, it will exclusively expand only one accordian. So next accordian to be expanded right after current is collapsed. Default is `'false`  
+* `disabled: boolean` - Optional
+  Disable control from being changed
+* `expanded: Array<boolean>` - Optional
```
nadiia: no need for this props since it is passed through the `initialState` for the StatefulAccordion

rfcs/accordians-component.md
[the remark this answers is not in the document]
blackswanny: renamed

rfcs/accordians-component.md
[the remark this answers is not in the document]
blackswanny: sure

rfcs/accordians-component.md
blackswanny: correct

rfcs/accordians-component.md
blackswanny: updated with items as children. Called `AccordionItem` to not mix with real `Panel` components

rfcs/accordians-component.md
```diff
@@ -0,0 +1,127 @@
...
+export default () => {
+  return (
+    <div>
+      <Accordion
+        onToggle={(event, index, isExpanded) => {console.log('Accordian ' + index + ' is ' + (isExpanded ? 'expanded' : 'collapsed') )}}
+        overrides={{
+          Root: CustomRoot,
+          Toggle: CustomToggle,
+        }}
+        expanded={[false, true]}
```
schnerd: For stateless Accordion, I'm wondering if switching to an `isOpen` prop on each `AccordionItem` is more straightforward. Internally AccordionItem could be implemented where the `isOpen` prop–if set–takes precedence over context from the accordion parent.

rfcs/accordians-component.md
```diff
@@ -0,0 +1,127 @@
...
+        exclusive={true}
```
schnerd: Seems like `exclusive` has no effect for stateless accordion – the user is always passing in which accordions are opened vs closed.

rfcs/accordians-component.md
schnerd: And actually, I'm almost wondering if its even worth breaking up into `Accordion` and `StatefulAccordion`. Maybe `Accordion` should just be stateful by default, and the `isOpen` prop on AccordionItem can be used for control if needed. 

I know this is kind of a break from our existing pattern in other components, but I'm struggling to think of a case where I'd want the stateless accordion that isn't solved by the solution above. Having a single `Accordion` export could make things conceptually simpler.

Thoughts @blackswanny @nadiia @abmai ?

rfcs/accordians-component.md
blackswanny: yes, correct. Fixed

rfcs/accordians-component.md
blackswanny: only one should have the state. I would rather gather full state of accordion in one place than to separate it between smaller AccordionItems. Export is simpler, but the data is scattered among more instances instead of one.

schnerd: Cool, works for me.

rfcs/accordion-component.md
```diff
@@ -0,0 +1,124 @@
...
+};
+```
+
+## Exports
+
+* `Accordion`
+* `StatefulAccordion`
+* `StatefulAccordionContainer`
+* `AccordionItem`
+* `StyledRoot`
```
nadiia: We'll probably also have a `StyledItem` here.

rfcs/accordion-component.md
```diff
@@ -0,0 +1,124 @@
...
+* `StyledTitle`
+* `StyledContent`
+
+## `Accordion` API
+
+* `children: Array<AccordionItem>` - Required
+  All accordion items in component. Each `AccordionItem` has:
+  * `title: React$Node` - title of accordion item
+  * `children: React$Node` - content of accordion item is represented as children of it correspondingly
+* `expanded: Array<boolean>` - Optional
```
nadiia: I think this prop should be on every accordion item, maybe in addition to the top level prop.

rfcs/accordion-component.md
blackswanny: `StyledContent` is the one

rfcs/accordion-component.md
blackswanny: ok

PR #225, opened 2018-09-12
feat: extract Tag into separate component

blackswanny:

### DIFF
--- a/rfcs/tag-component.md
+++ b/rfcs/tag-component.md
@@ -18,7 +18,7 @@ export default () => {
   return (
     <div>
       <Tag
-        onActionClick={event => {}}
+        onActionClick={(event, children) => {}}
         disabled={false}
         overrides={{
           Root: CustomRoot,
@@ -40,8 +40,8 @@ export default () => {
 
 ## `Tag` API
 
-* `color: STYLE.blue(theme.colors.primary470) | STYLE.orange(theme.colors.warning400) | STYLE.green(theme.colors.positive400) | STYLE.red(theme.colors.negative400)`:
-  the color theme to be applied to a Tag. Default is `STYLE.blue`.
+* `color: STYLE.primary(theme.colors.primary470) | STYLE.warning(theme.colors.warning400) | STYLE.positive(theme.colors.positive400) | STYLE.negative(theme.colors.negative400)`:
+  the color theme to be applied to a Tag. Default is `STYLE.primary`.
 * `children: React$Node`:
   Component or String value for label of tag. Default is empty string
 * `disabled: boolean`:
@@ -49,5 +49,5 @@ export default () => {
 * `overrides: {Root: (props: {[string]: any}) => React$Node, Action: (props: {[string]: any}) => React$Node}`
   * Custom `Action` button.
   * `Root` wrapper element for the whole tag control to apply styles
-* `onActionClick: func(evt => void)`:
-  handler for events on Action button element (default is Cross).
+* `onActionClick: func((evt, children) => void)`:
+  handler for events on Action button element (default is Cross). `children` provides which actual Tag was clicked
--- a/src/select/select.js
+++ b/src/select/select.js
@@ -12,11 +12,12 @@ import {
   Root as StyledRoot,
   Input as StyledInput,
   InputContainer as StyledInputContainer,
-  Tag as StyledTag,
+  SingleSelection as StyledSingleSelection,
   SearchIcon as StyledSearchIcon,
 } from './styled-components';
 
 import {Input as InputComponent} from '../input';
+import {Tag as StyledTag} from '../tag';
 import {ICON, TYPE, STATE_CHANGE_TYPE} from './constants';
 import SelectDropDown from './dropdown';
 import type {
@@ -290,12 +291,12 @@ class Select extends React.Component<PropsT, StatelessStateT> {
             }
           />
         )}
-        {selectedOptions.map(option => (
-          <Tag key={option.id} $multiple={multiple}>
-            {this.getSelectedOptionLabel(option)}
-            {multiple && (
-              <SearchIcon
-                onClick={e => {
+        {selectedOptions.map(
+          option =>
+            multiple ? (
+              <Tag
+                key={option.id}
+                onActionClick={e => {
                   this.setState({
                     selectedOptions: this.state.selectedOptions.filter(
                       selectedOption => selectedOption.id !== option.id,
@@ -303,14 +304,15 @@ class Select extends React.Component<PropsT, StatelessStateT> {
                   });
                   e.stopPropagation();
                 }}
-                $type={ICON.clearTag}
-                src={
-                  'data:image/svg+xml;utf8,<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg"><path fill-rule="evenodd" clip-rule="evenodd" d="M0.195262 0.195262C0.455612 -0.0650874 0.877722 -0.0650874 1.13807 0.195262L3.33333 2.39052L5.5286 0.195262C5.78895 -0.0650874 6.21106 -0.0650874 6.4714 0.195262C6.73175 0.455612 6.73175 0.877722 6.4714 1.13807L4.27614 3.33333L6.4714 5.5286C6.73175 5.78895 6.73175 6.21106 6.4714 6.4714C6.21106 6.73175 5.78895 6.73175 5.5286 6.4714L3.33333 4.27614L1.13807 6.4714C0.877722 6.73175 0.455612 6.73175 0.195262 6.4714C-0.0650874 6.21106 -0.0650874 5.78895 0.195262 5.5286L2.39052 3.33333L0.195262 1.13807C-0.0650874 0.877722 -0.0650874 0.455612 0.195262 0.195262Z" transform="translate(4.66675 4.6665)" fill="#276EF1"/></svg>'
-                }
-              />
-            )}
-          </Tag>
-        ))}
+              >
+                {this.getSelectedOptionLabel(option)}
+              </Tag>
+            ) : (
+              <StyledSingleSelection>
+                {this.getSelectedOptionLabel(option)}
+              </StyledSingleSelection>
+            ),
+        )}
       </React.Fragment>
     );
   }
--- a/src/select/styled-components.js
+++ b/src/select/styled-components.js
@@ -60,41 +60,18 @@ export const InputContainer = styled('div', props => {
   };
 });
 
-export const Tag = styled('span', props => {
-  const {$theme, $multiple} = props;
+export const SingleSelection = styled('span', props => {
+  const {$theme} = props;
   const {
-    colors: {primary400, mono400, mono1000},
-    sizing: {scale800, scale0},
-    typography: {font200},
+    colors: {mono1000},
   } = $theme;
-  return $multiple
-    ? {
-        ...font200,
-        height: scale800,
-        fontWeight: 'bold',
-        display: 'flex',
-        alignItems: 'center',
-        paddingTop: scale0,
-        paddingBottom: scale0,
-        paddingRight: '5px',
-        paddingLeft: '5px',
-        margin: '5px',
-        borderWidth: '1px',
-        borderColor: primary400,
-        color: primary400,
-        borderRadius: '7px',
-        backgroundColor: mono400,
-      }
-    : {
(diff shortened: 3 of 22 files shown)


src/select/styled-components.js
```diff
@@ -60,40 +60,16 @@ export const InputContainer = styled('div', props => {
...
-        },
-      };
+export const SingleSelection = styled('span', props => {
+  return {
+    ...getInputStyles({...props, $size: SIZE.default, $disabled: true}),
+    cursor: 'pointer',
+    width: 'auto',
+    flexGrow: '1',
+    ':hover': {
+      cursor: 'pointer',
```
kahwee: I'm curious why we need to specify `cursor` again in `:hover`.

src/select/select.js
```diff
@@ -209,6 +210,7 @@ class Select extends React.Component<PropsT, StatelessStateT> {
           overrides={{
             Root: Root,
             Input: Input,
+            InputContainer: StyledInputContainer,
```
abmai: This is putting a border around the drop-down opener (what do we call that?) component now; intended?

image

src/select/styled-components.js
blackswanny: pne of the ways to override the same style from original Input. This is the bug about

src/select/select.js
blackswanny: it was always a part of original Input component. It's very root one, which allows to arrange `input` field and other pieces of it in flexbox model. DropDown is outside of it

image

abmai: But it's not part of the design for Select, right? Can we disable it?

src/tag/tag.js
```diff
@@ -0,0 +1,64 @@
...
+      disabled,
+      isFocused,
+      isHovered,
+      isActive,
+      children,
+      onActionClick,
+      color,
+    } = this.props;
+    const {Root: RootOverride, Action: ActionOverride} = overrides;
+    const Root = getOverride(RootOverride) || StyledRoot;
```
abmai: Let's use `getOverrideObject` here instead, so we don't need to do `getOverride` + `getOverrideProps` separately.

rfcs/tag-component.md
```diff
@@ -49,5 +49,5 @@ export default () => {
 * `overrides: {Root: (props: {[string]: any}) => React$Node, Action: (props: {[string]: any}) => React$Node}`
   * Custom `Action` button.
   * `Root` wrapper element for the whole tag control to apply styles
-* `onActionClick: func(evt => void)`:
-  handler for events on Action button element (default is Cross).
+* `onActionClick: func((evt, children) => void)`:
+  handler for events on Action button element (default is Cross). `children` provides which actual Tag was clicked
```
schnerd: I'm trying to figure out how much value `children` has here. It seems like if it's an actual react element it has zero value. If its a string it has a little bit more value, though usually the tag has the human-readable label like "San Francisco" instead of the key like "san_francisco".

Maybe we should instead encourage users to do something like

```js
tags.map(() => (
  <Tag onActionClick={this.onClose.bind(this, tag.key)}>{tag.label}</Tag>
))
```

src/tag/types.js
```diff
@@ -0,0 +1,44 @@
...
+  disabled: boolean,
+  isFocused: boolean,
+  isHovered: boolean,
+  isActive: boolean,
+  color: string,
+  onActionClick: (
+    e: SyntheticEvent<HTMLInputElement>,
+    children: React$Node,
+  ) => void,
+};
```
schnerd: @abmai Just had a PR where he removed a `DefaultPropsT` object to get flow correctly identify required props #236 

Should we remove the DefaultPropsT here to stay consistent? (I'm guessing there may be other existing components we'll need to change in the future too)

src/tag/types.js
```diff
@@ -0,0 +1,44 @@
...
+
+export type PropsT = {
+  overrides?: OverridesT,
+  disabled?: boolean,
+  isFocused?: boolean,
+  isHovered?: boolean,
+  isActive?: boolean,
```
schnerd: What does it mean for a Tag to be "active"? As a customer of Tag, when would I pass in isFocused?

src/tag/tag.js
```diff
@@ -0,0 +1,64 @@
...
+
+class StatelessTag extends React.Component<PropsT, StatelessStateT> {
+  static defaultProps: DefaultPropsT = {
+    overrides: {},
+    onActionClick: () => {},
+    color: STYLE.blue,
+    disabled: false,
+    isFocused: false,
+    isHovered: false,
+    isActive: false,
```
schnerd: How do all of these work? It looks like you're not specifying them in `<Select>`.

Should we remove these for now and rely on CSS pseudo selector like `:focus` and `:hover`?

src/tag/styled-components.js
```diff
@@ -0,0 +1,51 @@
...
+    justifyContent: 'space-between',
+    paddingTop: scale100,
+    paddingBottom: scale100,
+    paddingLeft: scale300,
+    borderWidth: '1px',
+    borderColor: $color,
+    color: $color,
+    opacity: $disabled ? '.56' : null,
+    borderRadius: '2px',
+    cursor: $disabled ? 'not-allowed' : 'pointer',
```
schnerd: Should the `Root` have `pointer` by default? Technically clicking on the tag itself does nothing, and we don't expose `onClick` prop for tag besides action.

Maybe we should have `cursor: $disabled ? 'not-allowed' : 'auto'` for Root and `cursor: $disabled ? 'not-allowed' : 'pointer'` for Action.

rfcs/tag-component.md
blackswanny: yes, React element has no sense. String label has sense and was added only to avoid extra efforts by user to find himself, which tag has been clicked actually. I mean, Idk, what would be the most common case for example. Last time (in Select I believe) in similar case I was asked to add extra value in handler to help user not spend time to find Option clicked.

src/select/select.js
blackswanny: it is a part of Select styles still. That's why I used it and had some styles changed here

src/tag/tag.js
blackswanny: ok

src/tag/types.js
blackswanny: removed

src/tag/tag.js
blackswanny: ok

src/tag/styled-components.js
blackswanny: yes, you are right

src/tag/types.js
blackswanny: ok, still works

src/select/select.js
abmai: I'm slightly confused, I don't see a border in the design. Maybe it's just a matter of passing in extra props here to disable the border?

image

blackswanny: its present in Search mode, you look at Select mode, which's fix is not in master yet
image

blackswanny: @abmai oh, I see your point. Yes, it's not in design. Let me remove it.

src/select/styled-components.js
schnerd: Input has been updated on master, can we fix this?

src/tag/styled-components.js
```diff
@@ -0,0 +1,52 @@
...
+    alignItems: 'center',
+    justifyContent: 'space-between',
+    paddingTop: scale100,
+    paddingBottom: scale100,
+    paddingLeft: scale300,
+    borderWidth: '1px',
+    borderColor: $color,
+    color: $color,
+    opacity: $disabled ? '.56' : null,
+    borderRadius: '2px',
```
schnerd: Can we check $theme.borders.useRoundedCorners

src/tag/styled-components.js
```diff
@@ -0,0 +1,52 @@
...
+    cursor: $disabled ? 'not-allowed' : 'auto',
+    backgroundColor: $color + ($isFocused || $isActive ? '33' : '15'),
```
schnerd: It looks like support for 8-digit hexes is questionable: https://caniuse.com/#feat=css-rrggbbaa

Is there another approach we can use?

src/tag/tag.js
```diff
@@ -0,0 +1,66 @@
...
+LICENSE file in the root directory of this source tree.
+*/
+// @flow
+import React from 'react';
+import {getOverrideObject} from '../helpers/overrides';
+import type {PropsT, StatelessStateT} from './types';
+import {Action as StyledAction, Root as StyledRoot} from './styled-components';
+import {STYLE} from './constants';
+
+class StatelessTag extends React.Component<PropsT, StatelessStateT> {
```
schnerd: Can we just call this `Tag`? There's no stateful version so disambiguating is kind of pointless.

You can also leave out the StatelessStateT

src/select/styled-components.js
blackswanny: ok

src/tag/styled-components.js
blackswanny: ok

src/tag/styled-components.js
blackswanny: as we set only one color in theme and Figma design also shows only one color and trasparency change, we have to stick to this color or introduce one more( or more) colors props for other parts of tag. However, background color supports only this type of notation or RGBA() function. But I will have to add some code to convert HEX to plain RGB numbers

src/tag/tag.js
blackswanny: ok

src/tag/tag.js
```diff
@@ -0,0 +1,66 @@
...
+        <Action
+          {...sharedProps}
+          {...actionProps}
+          onClick={e => onActionClick(e, children)}
+          src={
+            'data:image/svg+xml;utf8,<svg width="8" height="8" viewBox="0 0 8 8" fill="none" xmlns="http://www.w3.org/2000/svg"><path fill-rule="evenodd" clip-rule="evenodd" d="M0.861278 0.862254C1.12163 0.601905 1.54374 0.601905 1.80409 0.862254L3.99935 3.05752L6.19461 0.862254C6.45496 0.601905 6.87707 0.601905 7.13742 0.862254C7.39777 1.1226 7.39777 1.54471 7.13742 1.80506L4.94216 4.00033L7.13742 6.19559C7.39777 6.45594 7.39777 6.87805 7.13742 7.1384C6.87707 7.39875 6.45496 7.39875 6.19461 7.1384L3.99935 4.94313L1.80409 7.1384C1.54374 7.39875 1.12163 7.39875 0.861278 7.1384C0.600928 6.87805 0.600928 6.45594 0.861278 6.19559L3.05654 4.00033L0.861278 1.80506C0.600928 1.54471 0.600928 1.1226 0.861278 0.862254Z" fill="' +
+            color +
+            '"/></svg>'
+          }
+        />
```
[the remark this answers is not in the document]
blackswanny: fixed

src/tag/styled-components.js
blackswanny: added rgba version 
https://www.w3schools.com/cssref/func_rgba.asp

src/tag/styled-components.js
```diff
@@ -0,0 +1,60 @@
...
+
+export const Root = styled('span', props => {
+  const {$theme, $color, $disabled, $isFocused, $isActive} = props;
+  const {
+    sizing: {scale800, scale100, scale300},
+    typography: {font200},
+  } = $theme;
+  return {
+    ...font200,
+    display: 'flex',
```
schnerd: Should we use inline-flex? Tags seem like more of an inline element than a block-level element (I'm assuming that's why they're a span too)

blackswanny: ok

src/tag/constants.js
```diff
@@ -0,0 +1,18 @@
...
+*/
+// @flow
+/* eslint-disable import/prefer-default-export */
+import DEFAULT_THEME from '../../src/themes/light-theme';
+
+export const STYLE = {
+  blue: DEFAULT_THEME.colors.primary470,
+  orange: DEFAULT_THEME.colors.warning400,
+  green: DEFAULT_THEME.colors.positive400,
+  red: DEFAULT_THEME.colors.negative400,
```
nadiia: I think we should name the props similar to the theme color naming: `primary, warning, positive, negative`, so it makes sence in the case, for example, when the primary color is set to `green`.

blackswanny: ok

PR #227, opened 2018-09-12
feat(pagination): initial implementation with Select integration (part 4 WIP)

abmai:
### Pagination Component

This is a WIP PR to integrate with our `Select` component. I think I got it as far as I could with our current implementations. There will be changes needed to `Select` to integrate further. We are pretty close though!!

Comparing with the Live Preview from the previous part 3 (no integration): [Ideal State](https://deploy-preview-214--baseui.netlify.com/?selectedKind=Pagination&selectedStory=Stateful%20Pagination&full=0&addons=1&stories=1&panelRight=0&addonPanel=REACT_STORYBOOK%2Freadme%2Fpanel)

This is the current state of the integration: [Current State](https://deploy-preview-227--baseui.netlify.com/?selectedKind=Pagination&selectedStory=Stateful%20Pagination&full=0&addons=1&stories=1&panelRight=0&addonPanel=REACT_STORYBOOK%2Freadme%2Fpanel)

Suggestions (from my perspective) on how to get to the ideal state and some open questions:

* How do we get the hover effect on the drop-down button?
* Should we separate `Select` and `Autocomplete` into 2 internal components, so that the logic does not have to overlap between the two?
* Instead of overriding sub-components of `Menu`, we should allow full override of the entire `Menu` component. This way the drop-down look and feel is entirely up to the user to modify if they want.

Some other issues that I observed:

* `Root` is rendered twice (at least for Select mode). Once in the main `render()` function and another by the `InputComponent`. This caused some confusion when I tried to override `Root`.
* `InputContainer` isn't overridable in Select mode, but it is when in Search/Autocomplete mode. This also caused some confusion for me.

Would love some suggestions and insight here as I continue to work on the integration!

### DIFF
--- a/src/pagination/select.js
+++ b/src/pagination/select.js
@@ -1,0 +1,57 @@
+/*
+Copyright (c) 2018 Uber Technologies, Inc.
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+// @flow
+import * as React from 'react';
+import {Select as DefaultSelect, TYPE} from '../select';
+import {StyledListItem} from '../menu';
+import {SelectInput, SelectTag} from './styled-components';
+import type {SelectMenuItemT} from './types';
+
+const noop = () => {};
+
+const getOptionLabel = item => item.label;
+
+// This is used internally only, not meant to be exported
+
+export default function Select({
+  currentPage,
+  options,
+  ...restProps
+}: {
+  currentPage: number,
+  options: Array<SelectMenuItemT>,
+}) {
+  return (
+    <DefaultSelect
+      // Required to re-initialize Select component so that selectedOptions work
+      key={currentPage}
+      options={options}
+      selectedOptions={[options[currentPage - 1]]}
+      rows={5}
+      type={TYPE.select}
+      getOptionLabel={getOptionLabel}
+      overrides={{
+        Input: SelectInput,
+        Tag: SelectTag,
+        DropDownItem: StyledListItem,
+      }}
+      // None of these props should be required
+      filterOption={() => false}
+      filterable={false}
+      multiple={false}
+      error={false}
+      textValue={''}
+      onBlur={noop}
+      onFocus={noop}
+      onMouseEnter={noop}
+      onMouseLeave={noop}
+      onMouseDown={noop}
+      onMouseUp={noop}
+      {...restProps}
+    />
+  );
+}
--- a/src/pagination/styled-components.js
+++ b/src/pagination/styled-components.js
@@ -8,6 +8,8 @@ LICENSE file in the root directory of this source tree.
 import {styled} from '../styles';
 import {StyledList} from '../menu';
 import {StyledBaseButton} from '../button';
+import {StyledInput} from '../input';
+import {StyledTag} from '../select';
 
 export const Root = styled('div', {
   display: 'flex',
@@ -22,8 +24,10 @@ export const MaxLabel = styled('span', ({$theme}) => ({
 
 export const DropdownContainer = styled('div', ({$theme}) => ({
   position: 'relative',
+  cursor: 'pointer',
   marginLeft: $theme.sizing.scale600,
   marginRight: $theme.sizing.scale300,
+  minWidth: `calc(${$theme.sizing.scale1600} + ${$theme.sizing.scale400})`,
 }));
 
 export const DropdownMenu = styled(StyledList, ({$theme}) => ({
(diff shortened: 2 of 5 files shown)


src/pagination/select.js
```diff
@@ -0,0 +1,57 @@
...
+      filterable={false}
+      multiple={false}
+      error={false}
+      textValue={''}
+      onBlur={noop}
+      onFocus={noop}
+      onMouseEnter={noop}
+      onMouseLeave={noop}
+      onMouseDown={noop}
+      onMouseUp={noop}
```
schnerd: I think flow is smart enough to know that a prop is not required if a defaultProp is set for it, even if PropsT has marked it as required ([reference](https://flow.org/en/docs/react/components/#toc-using-default-props)). It looks like Select sets defaultProps for most of these, were you still seeing flow errors when omitting them?

abmai: Yep. Maybe it's just with my particular installation? I think some consolidation is needed here regardless. We should just *not* type `defaultProps` if it's causing us so much headache (we're already skipping that step in other components as well).

abmai: Expanding on that, I mean removing the typedef here: https://github.com/uber-web/baseui/blob/master/src/select/select.js#L33 and properly mark `PropsT` with what's required/what's not.

schnerd:
Nice, looks like its close!

> How do we get the hover effect on the drop-down button?

I agree the hover effect is nice... maybe it should be added to select? cc @nguyensomniac 

> Should we separate Select and Autocomplete into 2 internal components, so that the logic does not have to overlap between the two?

Splitting into Select and Autocomplete could certainly make sense. But ideally we'd support some level of keyboard search in both. Even native select supports a rudimentary search when you type the first few characters of an option. I think this is especially important for pagination where you could have 500 pages and may want to jump to a specific one – it would be annoying to scroll when you could just type "88" and press enter.

> Instead of overriding sub-components of Menu, we should allow full override of the entire Menu component. This way the drop-down look and feel is entirely up to the user to modify if they want.

Seems reasonable.

src/pagination/styled-components.js
```diff
@@ -38,5 +42,17 @@ export const DropdownMenu = styled(StyledList, ({$theme}) => ({
...
-  minWidth: `calc(${$theme.sizing.scale1600} + ${$theme.sizing.scale400})`,
+}));
+
+// The following are meant to be used internallt and not exported to external customers
+
+export const SelectInput = styled(StyledInput, {display: 'none'});
+export const SelectTag = styled(StyledTag, ({$theme}) => ({
+  ...$theme.typography.font300,
+  color: $theme.colors.black,
+  // Hacking to get around not being able to override borders
```
abmai: This is a side effect of not being able to override `InputContainer` in Select mode.

abmai:
> Splitting into Select and Autocomplete could certainly make sense. But ideally we'd support some level of keyboard search in both. Even native select supports a rudimentary search when you type the first few characters of an option. I think this is especially important for pagination where you could have 500 pages and may want to jump to a specific one – it would be annoying to scroll when you could just type "88" and press enter.

This is a good point. Maybe this is a design update on the Pagination component to support a Search/Autocomplete paradigm instead then, cc @nguyensomniac.

src/pagination/select.js
schnerd: Yeah that probably makes sense.

abmai: You were right, flow recognizes `defaultProps` properly. The issue here was that we were typing `defaultProps` explicitly to something else. Removing that typedef for `defaultProps` produced the expected behavior.

abmai:
Going to start putting up separate PRs for some of the changes discussed in this PR.

abmai:
It looks like this might require additional work and changes (that might be an API change). How does this interim solution sound:

1. Pull `Select`-like functionalities of #213 into a separate internal `Select` component, and expose `Select` as an API to users. So the final `overrides` for `Pagination` will be:

```js
export type OverridesT = {
  Root?: OverrideT<*>,
  PrevButton?: OverrideT<*>,
  NextButton?: OverrideT<*>,
  MaxLabel?: OverrideT<*>,
  Select?: OverrideT<*>,
};
```

Users can opt to override the entire `Select` component or default to the internal `Select` component.

2. Work on updating `Select`'s so that it's easier to override certain things (i.e. the drop-down option, or change API to allow full `Menu` override). We can update the internal `Select` to the full `Select` component once that's done. User impact is just that they won't be able to override internals of the initial `Select` (that we implemented inside `Pagination`).

Thoughts? @schnerd

schnerd:
> OPTION B Leave Pagination implementation as-is with reliance on Button and Menu directly. This will result in a breaking API change once we do move to use Select directly.

This seems fine to me for now – using Select in its current state seems more burdensome than initially intended (hopefully our customers don't run into these issues too :/ ). Maybe we can land pagination as-is and then follow up with an issue describing features we should eventually support and the changes required to get there. Your choice ultimately.

abmai:
Kewl, let's go with the current implementation then and use this PR as a supplement for `Select` changes discussions.

abmai:
Thank you!!

PR #231, opened 2018-09-12
fix(package.json): use the same naming convention for scripts

gergelyke:


nadiia:
Like the updated consistent script names! Thanks for the change!

PR #241, opened 2018-09-13
fix(input): remove default prop for value in BaseInput

abmai:
Fixes https://github.com/uber-web/baseui/issues/237


nadiia:
Added a comment

PR #242, opened 2018-09-13
fix(flow): include flow files in the npm dist

gergelyke:
Once merged, the tarball that we publish to npm will include `.flow` files.


scripts/flow-copy-src.js
```diff
@@ -1,10 +1,12 @@
+#!/usr/bin/env node
+
 /* eslint-env node */
 // @flow
 
 const flowCopySource = require('flow-copy-source');
 
 async function run() {
-  await flowCopySource([`${__dirname}/../src`], `${__dirname}/../dist/src`, {
+  await flowCopySource([`${__dirname}/../src`], `${__dirname}/../dist`, {
```
rtsao: nit: `path.resolve` these instead

PR #246, opened 2018-09-14
feat(button): add loading style; refactor tests

abmai:
### Button Component

Refactored tests to get more concise snapshots. Added `isLoading` style, mainly to take care of this case:

image

The spinner icon should be up to the users to add.

### DIFF
--- a/src/button/__tests__/button.test.js
+++ b/src/button/__tests__/button.test.js
@@ -1,4 +1,10 @@
-/* eslint-disable flowtype/require-valid-file-annotation, react/prop-types */
+/*
+Copyright (c) 2018 Uber Technologies, Inc.
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+// @flow
 import React from 'react';
 import {mount} from 'enzyme';
 import {StartEnhancer, EndEnhancer} from '../styled-components';
@@ -24,7 +30,8 @@ describe('Button Component', () => {
     const NewStartEnhancer = () => <div />;
 
     const props = {
-      startEnhancer: 'start',
+      /* eslint-disable-next-line */
+      startEnhancer: () => <div />,
       overrides: {
         StartEnhancer: NewStartEnhancer,
       },
--- a/src/button/styled-components.js
+++ b/src/button/styled-components.js
@@ -7,10 +7,19 @@ LICENSE file in the root directory of this source tree.
 // @flow
 import {styled} from '../styles';
 import {KIND, SIZE, SHAPE} from './constants';
+import type {ThemeT} from '../styles';
+
+type StylePropsT = {
+  $theme: ThemeT,
+  $size?: $Keys<typeof SIZE>,
+  $kind?: $Keys<typeof KIND>,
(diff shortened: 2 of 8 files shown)


gergelyke:
@abmai would it make sense to split it into two PRs? first, to refactor the tests, then the second to add the spinner?

src/button/__tests__/button.test.js
```diff
@@ -24,7 +30,8 @@ describe('Button Component', () => {
     const NewStartEnhancer = () => <div />;
 
     const props = {
-      startEnhancer: 'start',
+      /* eslint-disable-next-line */
+      startEnhancer: () => <div />,
```
blackswanny: i think null should work here as well as default value

schnerd:
So I might be missing something, but does this prop actually add the spinner in? I can't find that part in the code. Or is it up to the user to still add the end enhancer? Ideally this is done automatically for them.

src/button/styled-components.js
```diff
@@ -29,52 +38,62 @@ export const BaseButton = styled(
...
 }));
 
-function getStyleForKind({$theme, $kind}) {
+export function getStyleForShape({$theme, $shape, $size}: StylePropsT) {
+  switch ($shape) {
+    case SHAPE.round:
+    case SHAPE.square:
+      return {
+        paddingTop:
+          $theme.sizing[$size === SIZE.compact ? 'scale400' : 'scale500'],
```
blackswanny: this notation is better for some IDEs as allows indexing and search better
```js
$size === SIZE.compact ? $theme.sizing.scale400 : $theme.sizing.scale500
```

gergelyke:
Agreed with David - shouldn't this PR include the spinner too? I know we haven't implemented the component for now, but it seems misleading

abmai:
Hmm originally I thought users might want to switch icons depending on the state of the button - i.e. I normally have a cloud icon but I want it to be a spinner when in a loading state. So it'll be more straightforward to just let the users manage their own spinner icon.

I think we can also do this by attaching our own spinner as a default `endEnhancer` if `isLoading` is true (and users can still override). But what if I want the spinner on the left, rather than on the right?

abmai:
Going to also split this PR into 2 as requested, stay tuned!

src/button/styled-components.js
abmai: didn't think about that, happy to switch!

schnerd:
> I think we can also do this by attaching our own spinner as a default endEnhancer if isLoading is true (and users can still override). But what if I want the spinner on the left, rather than on the right?

Could set endEnhancer explicitly to `null` and pass your own spinner to startEnhancer. Or we could eventually split this into two props `isLoading` and `isLoadingStyle` – `isLoadingStyle` would automatically default to the value of `isLoading`, but if you wanted to provide your own spinner you could enable `isLoadingStyle` and pass in your own enhancer.

All that said, I don't know if the "custom spinner" is a case we need to immediately handle. I'd vote to just get spinner as a default endEnhancer for now and we can revisit in the future if customers ask about it.

abmai:
Sounds like a plan to me.

abmai:
Closing this in favor of the new individual PRs.

PR #250, opened 2018-09-14
docs(rfc): button group

gergelyke:


rfcs/button-group-component.md
```diff
@@ -0,0 +1,45 @@
...
+* `disabled: boolean` - Optional:
+  Defines if the button group disabled.
+* `size: 'default' | 'compact'`
+  Defines the size of a button
+* `onChange: func` - Optinal:
+  Called when a toggleable Button is selected
+* `onClick: func` - Optional:
+  Called with the click event
+* `checked: numer | Array<number>` - Optional:
+  Index or array of indices of the selected
```
nadiia: What is this prop for? Seems to be not related.

rfcs/button-group-component.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Exports
+
+* `ButtonGroup`
+
+## `ButtonGroup` API
+
+* `children: (props: Props) => React.Node` - Required
+  This can be an SVG icon as well.
+* `ariaLabel: string` - Required:
```
abmai: Do we need to enforce aria labels?

rfcs/button-group-component.md
```diff
@@ -0,0 +1,45 @@
...
+  Accessible label
+* `disabled: boolean` - Optional:
+  Defines if the button group disabled.
+* `size: 'default' | 'compact'`
```
abmai: How do you see this prop working with the underlying `Button`, via this composition approach?

rfcs/button-group-component.md
```diff
@@ -0,0 +1,43 @@
...
+### Basic usage
+
+```javascript
+import * as React from 'react';
+import {Button} from 'baseui/button';
+import {ButtonGroup} from 'baseui/button-group';
+
+export default () => {
+  return
+    <ButtonGroup>
```
blackswanny: button group can change styles around button, so we need a root component to be exposed for that

rfcs/button-group-component.md
```diff
@@ -0,0 +1,43 @@
...
+
+## `ButtonGroup` API
+
+* `children: (props: Props) => React.Node` - Required
+  This can be an SVG icon as well.
+* `ariaLabel: string` - Optional:
+  Accessible label
+* `disabled: boolean` - Optional:
+  Defines if the button group disabled.
+* `size: 'default' | 'compact'`
```
abmai: Not sure why the other conversation got collapsed :(. But Do you think it's worth providing all the `Button` config API here and just say that it'll automatically be applied to all children `Button`? i.e. `KIND`, `SIZE`, `SHAPE`

gergelyke: I think that makes sense - thoughts @nadiia @schnerd @blackswanny ?

gergelyke:
> Will the same onClick and onChange from the ButtonGroup props be attached to every child button?

not sure - can we think of any use-cases when we would want a separate event handler for the group and the button instances? 🤔 if not, probably it worth passing them in

rfcs/button-group-component.md
blackswanny: Yes, the same happens in other components. Props fall down to nested components.
For instance, Select sends props to each Option

rfcs/button-group-component.md
```diff
@@ -0,0 +1,43 @@
...
+  Defines the size of a button
+* `onChange: func` - Optinal:
+  Called when a toggleable Button is selected
+* `onClick: func` - Optional:
+  Called with the click event
```
schnerd: So just to clarify, is this component meant to behave almost like a radio group? Or is it just meant to visually group the buttons together?

If just the latter, then maybe we don't need `onChange` and `onClick` handlers here. I'm trying to figure out what I would do with those as a customer – seems like it makes more sense to just attach onClick to each individual buttons.

But if it's the former (more like a radio group, where a button can be "selected"), then I think we may want to align more closely with the radio group API. This is what antd does: https://codesandbox.io/s/6xoqyy57o3

Which one were you thinking?

cc @nadiia @abmai @blackswanny

rfcs/button-group-component.md
schnerd: I like the idea of propagating the `kind`, `size`, `shape`, and related props down to buttons, though we might want to explicitly pass them rather than just passing `{...restProps}` down to each button.

rfcs/button-group-component.md
gergelyke: oh, I see - [mineral-ui](https://mineral-ui.com/components/button-group) does something very similar.

I think based on this discussion, we should follow up with design to see what behavior they had in mind - does that sound reasonable?

schnerd: Yeah that's probably a good idea.

rfcs/button-group-component.md
nadiia: Totally makes sense since we already moving the `size` up

rfcs/button-group-component.md
nadiia: Yes, let's extend the question to the design team. But I'd assume every button would need an individual click or change handler.

rfcs/button-group-component.md
```diff
@@ -0,0 +1,77 @@
...
+* `mode: string` - Optional:
+  Use the `mode` prop to render toggleable Buttons:
+  * the value `radio` will cause Buttons to behave like radio buttons,
+  * the value `checkbox` will cause Buttons to behave like checkboxes.
+* `defaultChecked: number | Array<number>`
+  Index or array of indices of the selected Button(s).
+  Primarily for use with uncontrolled components with a `mode` prop defined.
+* `checked: number | Array<number>` - Optional
+  Index or array of indices of the selected Button(s).
+  Primarily for use with controlled components with a `mode` prop defined.
```
schnerd: So the `defaultChecked`/`checked` props sort of allow you to use this component as both a controlled (stateless) and uncontrolled (stateful) component. If we want to be more consistent with the rest of our BaseUI components, it may be preferable to have a `StatefulButtonGroup` with an `initialState prop` and a `ButtonGroup` with the `checked` prop.

rfcs/button-group-component.md
```diff
@@ -0,0 +1,77 @@
+# Button Group Component
+
+## Usage
+
+### Basic usage
+
+```javascript
+import * as React from 'react';
+import {Button} from 'baseui/button';
```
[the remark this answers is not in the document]
gergelyke: what do you think @nadiia @abmai @blackswanny?

rfcs/button-group-component.md
gergelyke: ah, you are right, I'll update it

rfcs/button-group-component.md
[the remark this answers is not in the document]
blackswanny: The task of integrating already existing components (radio\checkbox) becomes non-trivial.
In fact we need to repeat the same interface we have for them. I think, just visual appearance shouldn't lead us to make things more complex.
I vote for not adding radio\checkbox modes in here, but just use `override` there or add some default styles in those components for these cases.

rfcs/button-group-component.md
blackswanny: did we add extra root element to be able to change padding\margin, etc around buttons?

rfcs/button-group-component.md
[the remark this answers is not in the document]
abmai: Agreed with David's second option as well, new `Button` that's meant to be used inside this `ButtonGroup` but built on top of `Button` and takes the same overrides.

rfcs/button-group-component.md
```diff
@@ -0,0 +1,89 @@
...
+* `onChange: func` - Optinal:
+  Called when a toggleable Button is selected
+* `onClick: func` - Optional:
+  Called with the click event
+
+## `StatefulButtonGroup` API
+
+Extends the `ButtonGroup` API with this additional props:
+
+* `initalState: number | Array<number>`
```
schnerd: Most of our initialState props are objects, meant to represent the normal state object. Maybe we want to do that here too? `initialState: {checked: number | Array<number>}`

PR #254, opened 2018-09-15
feat(button): add isLoading prop and loading spinner

abmai:
### Button Component

Adding `isLoading` prop; spinner component

#### Updates

Updated with all the changes talked about here and in chat. I'll just let the previews do the talking.

## [Live Preview](https://deploy-preview-254--baseui.netlify.com/?selectedKind=Button&selectedStory=Buttons&full=0&addons=1&stories=1&panelRight=0&addonPanel=REACT_STORYBOOK%2Freadme%2Fpanel)


abmai:
Thanks for the insight, Nadiia! I think your second option to just render the Loading Spinner instead of children makes sense for that case (and if you don't want it then you just disable `endEnhancer` like normal).

Let's see if others have any thoughts.

schnerd:
Agree with Nadiia! And PR looks great.

src/button/styled-components.js
```diff
@@ -54,43 +57,92 @@ export const StartEnhancer = styled('div', ({$theme}: StylePropsT) => ({
...
+      ? 'rgba(255, 255, 255, 0.32)'
+      : 'rgba(39, 110, 241, 0.32)',
+  borderBottomColor:
+    $kind === KIND.primary
+      ? 'rgba(255, 255, 255, 0.32)'
+      : 'rgba(39, 110, 241, 0.32)',
+  borderRightColor:
+    $kind === KIND.primary
+      ? 'rgba(255, 255, 255, 0.32)'
+      : 'rgba(39, 110, 241, 0.32)',
```
schnerd: Maybe disabled button should have grey loader instead of white? White is a tad hard to see.

image

abmai: @nguyensomniac do you have any specific recommendations for this case in terms of colors?

nguyensomniac: +1 for grey—let's start with the same color as text

abmai:
Got design sign off, gonna land!

PR #260, opened 2018-09-17
WIP - (ci) add unsafe-perm configuration to package.json

KevinGrandon:
This is necessary in order to run scripts as root while using npm publish.

Additional context: https://stackoverflow.com/a/19132229/9078881


KevinGrandon:
Actually, because I'm still getting permissions error in docker, let's try this another way.

PR #262, opened 2018-09-17
chore(overrides): better merging of overrides objects

schnerd:
This PR updates our `overrides` logic to have more intelligent merging capabilities.

#### Motivation

We'll use an example of ModalButton, which is just a `Button` with some default margin applied to it:

```js
const overrides = {
  BaseButton: {
    style: ({$theme}) => ({
      marginLeft: $theme.sizing.scale500,
    }),
  },
};

export default class ModalButton extends React.Component<ButtonPropsT> {
  static defaultProps = Button.defaultProps;

  render() {
    return (
      <Button
        {...this.props}
        overrides={mergeOverrides(overrides, this.props.overrides)}
      >
        {this.props.children}
      </Button>
    );
  }
}
```

With the existing logic on master, someone passing in a style override to ModalButton would _replace_ the margin styles here. The same goes for props (although we don't have any props in this example).

Ideally we can apply our intermediate overrides, and also accept overrides from the parent component that get _merged_ into our intermediate overrides. That's what this PR introduces, and it should make style overrides work much more predictably for components that reuse other base ui components.

### DIFF
--- a/src/utils/deep-merge.js
+++ b/src/utils/deep-merge.js
@@ -1,0 +1,29 @@
+// @flow
+
+export default function deepMerge(
+  target?: ?{},
+  ...sources: Array<null | ?{}>
+): {} {
+  target = target || {};
+  const len = sources.length;
+  let obj;
+  let value;
+  for (let i = 0; i < len; i++) {
+    obj = sources[i] || {};
+    for (let key in obj) {
+      if (obj.hasOwnProperty(key)) {
+        value = obj[key];
+        if (isCloneable(value)) {
+          target[key] = deepMerge(target[key] || {}, value);
+        } else {
+          target[key] = value;
+        }
+      }
+    }
+  }
+  return target;
+}
+
+function isCloneable(obj: mixed) {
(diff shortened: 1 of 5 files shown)


src/utils/deep-merge.js
```diff
@@ -0,0 +1,29 @@
...
+        if (isCloneable(value)) {
+          target[key] = deepMerge(target[key] || {}, value);
+        } else {
+          target[key] = value;
+        }
+      }
+    }
+  }
+  return target;
+}
```
schnerd: Debated whether to use a dependency for this, let me know if anyone has a strong opinion on that.

schnerd:
> if this is a breaking change, should we add a description of what users are affected in what way, and how can they migrate?

Actually did a bit more digging and it looks like none of our components are using `mergeOverrides` with with styles/props (only components), and this PR doesn't change the component for semantics. So I take back my statement – this should probably just be a patch bump.

gergelyke:
cool, thanks David 👍

PR #263, opened 2018-09-17
chore(modal): use new button component

schnerd:
Switch `ModalButton` to use new `Button` component now that it exists.


schnerd:
@nadiia good catch, just updated the base branch.

PR #269, opened 2018-09-18
feat(icon): implement icon component

schnerd:
Initial implementation for the Icon component.

### [Live Preview](https://deploy-preview-269--baseui.netlify.com/?knob-Size=48px&knob-Color=0&knob-Button%20Kind=primary&selectedKind=Icon&selectedStory=Icon&full=0&addons=1&stories=1&panelRight=0&addonPanel=REACT_STORYBOOK%2Freadme%2Fpanel)

**Progress**
- [x] Basic implementation
- [x] Stories
- [x] Implement all icons (23 / 23)
- [x] Unit Tests

**Questions**

- Should we use names that describe the iconography itself or its semantic meaning? For example, the X icon can be used for both `Delete` and `Close` use cases, so maybe `X` is more appropriate? In theory we could have aliases too.


src/icon/delete-alt.js
```diff
@@ -0,0 +1,22 @@
...
+LICENSE file in the root directory of this source tree.
+*/
+// @flow
+import * as React from 'react';
+import Icon from './icon';
+import type {IconPropsT} from './types';
+
+export default function IconDeleteAlt(props: IconPropsT) {
+  return (
+    <Icon viewBox="0 0 16 16" title="Delete" {...props}>
```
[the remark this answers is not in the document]
schnerd: Conclusion – yes everything should have viewbox="0 0 24 24" matching the white rectangle in the docs:

image

side effect of this is that the a prop like `size="12"` doesn't necessarily mean your icon will be exactly 12px tall/wide. But the icons are designed in a way to be cohesive from a design perspective without actually all being the same size.

src/icon/icon-template.txt
```diff
@@ -0,0 +1,19 @@
+/*
+Copyright (c) 2018 Uber Technologies, Inc.
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+// @flow
+// BASEUI-GENERATED-REACT-ICON
```
gergelyke: should we also explicitly state that one should not modify it manually?

rfcs/icon-component.md
```diff
@@ -48,7 +48,7 @@ From left-to-right based on the image above:
 ## `Icon` API
 
 * `size: number|string` - Optional, Defaults to `scale500` (12px)
-  Size of element, will be passed to the svg width/height style. Can also be a baseui `theme.sizing` constant like `sizing500`.
+  Size of element, will be passed to the svg width/height style. Can also be a baseui `theme.sizing` constant like `scale500`.
```
gergelyke: i know it is not part of this PR, but should we move the usage example up, so it follows the order we have in the rest of the rfcs?

gergelyke: also, should we include an example with buttons too?

src/icon/icon-template.txt
schnerd: good call, will update.

rfcs/icon-component.md
schnerd: Good call, will add.

kahwee: Wow I really like how the icons looks in your [live preview](https://deploy-preview-269--baseui.netlify.com/?knob-Size=48px&knob-Color=0&knob-Button%20Kind=primary&selectedKind=Icon&selectedStory=Icons%20in%20Button&full=0&addons=1&stories=1&panelRight=0&addonPanel=REACT_STORYBOOK%2Freadme%2Fpanel).

Maybe it's just me but the start and end enhancers in the button seems to have too much whitespace. What do you think @nguyensomniac ?

PR #279, opened 2018-09-19
Feature/select

blackswanny:
Keyboard support

- [x] Single-select should have tabindex so it can be focused (multi-select is okay since it has input)
- [x] When TYPE.select is focused, down arrow and spacebar should open dropdown
- [x] Backspace should delete an item in multi-select
- [x] Pressing enter on highlighted item should select/deselect it.
- [x] Escape should close the dropdown for both single and multi-select


src/select/select.js
```diff
@@ -152,21 +154,24 @@ class Select extends React.Component<PropsT, StatelessStateT> {
...
-          this.setState({filteredOptions});
+        if (!this.handledHotKeys(e)) {
+          // $FlowFixMe
+          const newTextValue = e.target.value;
+          this.setState({textValue: newTextValue});
+          if (this.props.filterable) {
+            let filteredOptions = this.props.options.filter(option =>
+              this.props.filterOption(option, newTextValue),
+            );
+            filteredOptions = filteredOptions.length
```
gergelyke: maybe it is just me, but I have hard time understanding the double ternary operator here - can we refactor this?

src/select/select.js
```diff
@@ -350,6 +359,40 @@ class Select extends React.Component<PropsT, StatelessStateT> {
     const {type, multiple} = this.props;
     return type === TYPE.search ? true : multiple;
   }
+
+  handledHotKeys(
```
gergelyke: it seems that this function has side-effects - should we split it into two functions, one that translates `e.key` into a state like `isDropDownOpen: true`, and one that actually sets it?

src/select/select.js
```diff
@@ -350,6 +359,40 @@ class Select extends React.Component<PropsT, StatelessStateT> {
...
+          this.onChange(e, STATE_CHANGE_TYPE.select, option.id, option.label);
+        }
+        return;
+      case KEY_STRINGS.Backspace:
+        if (this.isMultiple() && !this.state.textValue) {
+          const selectedOptions = this.state.selectedOptions.slice();
+          selectedOptions.pop();
+          this.setState({selectedOptions});
+          return true;
+        }
```
schnerd: It looks like it still deletes the previous tag when I delete the last character in the input:

kapture 2018-09-19 at 10 29 06

Can we make it so that it only deletes when the input is empty?

src/select/select.js
```diff
@@ -350,6 +359,40 @@ class Select extends React.Component<PropsT, StatelessStateT> {
...
+
+  handledHotKeys(
+    // $FlowFixMe
+    e: SyntheticEvent<EventTarget> | KeyboardEvent,
+    option?: ?OptionT,
+  ) {
+    switch (e.key) {
+      case KEY_STRINGS.ArrowDown:
+      case KEY_STRINGS.Space:
+        if (e.key === KEY_STRINGS.Space && this.props.type === TYPE.search) {
```
schnerd: the `e.key === KEY_STRINGS.Space` seems redundant here since we're already in the case statement?

schnerd:
Can we add some unit or e2e tests too?

src/select/select.js
blackswanny: ok

src/select/select.js
blackswanny: no side effects. Function does one single action and notifies outside if this action has happened

src/select/select.js
schnerd: oh wow, didn't notice the `case` fall-through, you're right. my bad.

blackswanny:
@schnerd unit tests are added

src/select/dropdown.js
```diff
@@ -95,8 +83,26 @@ export default function SelectDropDown(props: DropDownPropsT) {
               );
             },
           },
+          component: (optionProps: OptionListPropsT) => {
```
abmai: This is fine for now but can we sync up about why such custom overrides are needed here. What behavior is this `DropDown` trying to capture? Are we just trying not to change the external API here?

blackswanny: еthis is the way to add click handler not to inner custom label, but to the full body of menu item. I couldn't provide it via static props as they are same for all items, but we have different logic per each item

abmai: But, that's probably a separate conversation not in the scope of this PR. I remember looking into this before and couldn't figure out how to do it without changing the API for Select overrides.

blackswanny: `getRequiredItemProps` doesn't work as it is defined in `menu/stateful-container.js` and the one I provide in props just doesn't come through. Cause I use stateful menu since I have to( we had discussion on that before), otherwise it doesn't change state properly.

And below doesn't work as well as I said props are the same for all Options, which I can't use, I need dynamic props for each of them
```<StatefulMenu
     overrides={{
        List: { /* this is where you're doing per-item override now */ },
        Option: { /* this should be where you're doing per-item override */},
     }}
/>```

abmai: Yeah I realized `getRequiredItemProps` wouldn't work in this case. Maybe the fix there is to merge user's `getRequiredItemProps` with the internal `getRequiredItemProps` in `StatefulContainer`.

But what I meant with the second option was to define your own `Option` component that would take in the `item` as a prop, and you can encapsulate all your logic in there. That's what the API is meant for (which seems like what you're doing here through using all the other APIs separately).

blackswanny: Second option is no difference for me than what I provided so far.

I think adding `getRequiredItemProps` from props to container seems reasonable so far and requires just one line of code

blackswanny: I updated with the use of method from props

PR #282, opened 2018-09-19
feat(testing): integrate saucelabs

gergelyke:
This initial PR just integrates Sauce Labs with the project, but does not address a few important things, that will be covered in follow up PRs:

* run tests in different browsers
* proper instructions on running Sauce Labs tests from local dev envs
* migrating over to Nightwatch
* link sauce lab builds to PRs
* "thank you" to sauce labs in the README

https://saucelabs.com/beta/tests/b14ca805d1174333bb8949c5a2715389/commands

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -5,4 +5,3 @@ storybook-static
 coverage
 docs
 webpack.e2e.config.js
-src/e2e-test-utils.js
--- a/.eslintignore
+++ b/.eslintignore
@@ -5,3 +5,4 @@ storybook-static
 coverage
 docs
 webpack.e2e.config.js
+src/e2e-test-utils.js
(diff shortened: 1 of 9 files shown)


.eslintignore
```diff
@@ -5,3 +5,4 @@ storybook-static
 coverage
 docs
 webpack.e2e.config.js
+src/e2e-test-utils.js
```
blackswanny: why is this one ignored? if it contains some local settings, we may need to split it

gergelyke: Yeah, the reason was because I did not want to write sauce labs capabilities specific flow type definitions, if we are going to remove this file with Nightwatch anyways. Do you think this is a blocker?

KevinGrandon: Might want to consider ignoring inline in the file in that case, with either `/* eslint-disable */`, or a `/* no-flow */` comment.

gergelyke: that makes perfect sense

blackswanny: ok, seems like the whole is little early to ignore so far

PR #284, opened 2018-09-20
docs(progress-steps): progress steps component RFC

nadiia:
screen shot 2018-09-20 at 9 10 10 am

### DIFF
--- a/rfcs/progress-steps-component.md
+++ b/rfcs/progress-steps-component.md
@@ -1,0 +1,90 @@
+# Progress Steps Component
+
+Progress steps are used for multi-step flows where indicating the number of steps in a process is important.
+
+## Usage
+
+### Basic usage
+
+```javascript
+import * as React from 'react';
+import {ProgressSteps} from 'baseui/progress-steps';
+
+export default () => {
+  return (
+    <ProgressSteps current={2}>
+      <Step title="Step 1" />
+      <Step title="Step 2" />
+      <Step title="Step 3" />
+    </ProgressSteps>
+  );
+}
+```
+
+### Advanced usage
+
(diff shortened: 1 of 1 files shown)


rfcs/progress-steps-component.md
```diff
@@ -0,0 +1,88 @@
...
+import {ProgressSteps, StyledNumberedStep} from 'baseui/progress-steps';
+
+export default () => {
+  const overrides ={
+    Step: props => <StyledNumberedStep {...props}/>,
+    StepTitle: props => <CustomTitle {...props} />,
+  };
+  return
+    <ProgressSteps
+      onClick={()=>{console.log('step is clicked')}}
```
schnerd: will onclick be called with the step index?

nadiia: yes! updated

abmai:
Thanks for knocking this out!!

How do you see this supporting the last case, would additional internal items just be children of `Step`?

image

nadiia:
@abmai I was thinking those would be just children of a `Step`. Do you think we should rather define these type of children through the props or maybe just provide additional styled components like `StyledStepSubtitle`, `StyledStepAction`?

abmai:
> @abmai I was thinking those would be just children of a `Step`. Do you think we should rather define these type of children through the props or maybe just provide additional styled components like `StyledStepSubtitle`, `StyledStepAction`?

I think providing more styled components is nice, maybe not the action since it's just a button but `<StyledStepSubtitle>` and `<StyledStepText>` would be good to have. What do you think?

nadiia:
Added the two mentioned to the exports list ^

PR #287, opened 2018-09-20
fix(fonts): use UberMove and fallback to system fonts

gergelyke:


kahwee:
The default font should be `UberMoveText` instead of `UberMove`. Can @nguyensomniac chime in on this?

PR #292, opened 2018-09-21
feature: fix bugs for select and change styles

blackswanny:
- [x] Dropdown begins in an open state
- [x] For `TYPE.search` dropdown does not re-open after closing
- [x] Selecting an object should return entire object not just `id` and `label`
- [x]  Search doesn't seem to work reliably in the first multi-select storybook example. Maybe we should do a fixed set of options in first story, and have a separate story for demonstrating async results?
- [x] Images and labels are overflowing tag in custom label story
- [x] Make sure disabled state works as per #198
- [ ] For `TYPE.search` selecting an option does not clear current input value
- [ ] Single-mode `TYPE.search` not supported? We should have something similar to default behavior of react-select Single mode


rfcs/select-component.md
```diff
@@ -154,15 +154,17 @@ export default () => {
...
   * `Option: ?React.ComponentType` component to use for options in dropdown list
   * `Root: ?React.ComponentType` component to use for most top of the select component
   * `Input: ?React.ComponentType` component for Input showing current selected value(s). See `Input` Control of this framework for reference to override it's functionality.
   * `SearchIcon: ?React.ComponentType` component for all icons appearing in Select component. It's provided `$type: ICON.loop | ICON.clearTag | ICON.clearAll | ICON.selected` to setup corresponding icon of Select component
   * `Tag: ?React.ComponentType` component for selected options Tags shown in Input for multiple mode selection
-* `onChange: func(e: SyntheticInputEvent, params: Object)`:
-  handler for events on trigger element when option are changing selection or text of search input (in Search mode has changed). `params` has `type` indicating which action is performed, `id` and `label` of selected\unselected option and `selectedOptions` array of all of selected, has new `textValue` set in input (for `keyUp`)
+* `onChange: func(e: SyntheticInputEvent, params: Object) => Promise`:
```
abmai: Why does this need to be a Promise?

src/menu/stateful-container.js
```diff
@@ -117,7 +118,7 @@ export default class MenuStatefulContainer extends React.Component<
 
   getRequiredItemProps: GetRequiredItemPropsFnT = (item, index) => {
     const {highlightedIndex} = this.state;
-    const {onItemSelect} = this.props;
+    const {onItemSelect, getRequiredItemProps} = this.props;
```
abmai: Nice!

rfcs/select-component.md
blackswanny: because when we want to update Options if text value has changed (in Search mode for instance), than it might be async call to request options from server.

blackswanny:
for the next two items:
1) For TYPE.search selecting an option does not clear current input value
2) Single-mode TYPE.search not supported? We should have something similar to default behavior of react-select Single mode

Answers:
1) I don't think we should clear input once we picked option. I had examples, where user wanted to pick couple options with similar search text and clearing it will force them to input it again
2) This is not supported as there is no design for it. Look at the same design for TYPE.select mode. Currently I don't see , how to design it so far and where to show picked value.
cc @nguyensomniac

rfcs/select-component.md
schnerd: Talked with Oleg and we came up with some proposed changes that reduce the complexity of this change handler:
https://github.com/uber-web/baseui/issues/171#issuecomment-424884154

src/select/select.js
```diff
@@ -192,37 +201,42 @@ class Select extends React.Component<PropsT, StatelessStateT> {
...
-        }}
-      >
+      <div tabIndex={this.props.tabIndex} {...events}>
         <InputComponent
           disabled={true}
           placeholder={!selectedOptions.length ? placeholder : ''}
           overrides={{
-            Root: Root,
-            Input: Input,
+            Root: {component: Root, props: rootProps},
```
nadiia: We should probably merge overrides here with those passed by customers in props.

src/select/select.js
```diff
@@ -312,13 +331,18 @@ class Select extends React.Component<PropsT, StatelessStateT> {
...
                 }}
+                {...tagProps}
               >
                 {this.getSelectedOptionLabel(option)}
               </Tag>
             ) : (
-              <StyledSingleSelection key={option.id}>
+              <SingleSelection
+                key={option.id}
+                $disabled={disabled}
```
nadiia: seems like the `$disabled` prop should be passed to all underlying styled components

src/tag/tag.js
```diff
@@ -53,14 +53,15 @@ class Tag extends React.Component<PropsT, {}> {
...
+        };
     return (
       <Root {...sharedProps} {...rootProps}>
         {children}
-        <Action
-          {...sharedProps}
-          {...actionProps}
-          onClick={e => onActionClick(e, children)}
-        >
+        <Action {...sharedProps} {...actionProps} {...events}>
```
nadiia: The props overrides seems to be missing here

nadiia: And on the line 62 ^

blackswanny: this code is above and should work well to override props if present
```js
const {component: Root, props: rootProps} = getOverrideObject(
      overrides.Root,
      StyledRoot,
    );
    const {component: Action, props: actionProps} = getOverrideObject(
      overrides.Action,
      StyledAction,
    );
```

src/select/constants.js
```diff
@@ -28,4 +28,5 @@ export const STATE_CHANGE_TYPE = {
   unselect: 'unselect',
   clearAll: 'clear_all',
   keyDown: 'key_down',
+  textChange: 'text_change',
```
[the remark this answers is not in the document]
blackswanny: when we change text in input. We will probably refactor this later to make a separate handler instead of type.

src/select/__tests__/select.test.js
```diff
@@ -348,7 +382,9 @@ describe('Stateless select', function() {
         .find(StyledInput)
         .first()
         .simulate('click');
-      expect(wrapper.instance().state.isDropDownOpen).toBeTruthy();
+      setTimeout(() => {
+        expect(wrapper.instance().state.isDropDownOpen).toBeTruthy();
+      }, 0);
```
schnerd: It looks like all these tests are async now (due to promise / setTimeout). Should we be using a done callback or async await to make sure the assertions are being called before test cleanup? https://jestjs.io/docs/en/asynchronous

blackswanny: thanks, it works well

PR #295, opened 2018-09-21
chore: add a component template

nadiia:


schnerd:
This is awesome!

As a follow up I wonder if we could also add a `yarn run scaffold:component Pagination` that automatically copies the template directory and replaces all the places all instances of `Component` with the specified component name. Ideally such a script could even be interactive kind of like the uber-web scaffold, where it asks you questions–one of those questions being whether your component is stateful or not, which would affect the files that are copied. I could also be totally over-thinking this 😛. Either way this PR is a big step forward as-is.

abmai:
NICE 🚀🚀🚀🚀!!

nadiia:
> This is awesome!
> 
> As a follow up I wonder if we could also add a `yarn run scaffold:component Pagination` that automatically copies the template directory and replaces all instances of the string `Component` with the specified component name. Ideally such a script could even be interactive kind of like the uber-web scaffold, where it asks you questions–one of those questions being whether your component is stateful or not, which would affect the files that are copied. I could also be totally over-thinking this 😛. Either way this PR is a big step forward as-is.

Yes, that would be ideal.

PR #302, opened 2018-09-25
feat(e2e): add a11y checks

gergelyke:


schnerd:
Regarding the modal a11y issues

1. It seems like aria-modal="true" is only valid for role="dialog". I had read in the bootstrap github that [role="document" is sometimes preferable.](https://github.com/twbs/bootstrap/issues/15875), but it seems like it's not compatible with aria-modal. Let's just revert to using `role="dialog"` as the default. This can be changed here in modal.js:



Let me know if you'd prefer that I open a PR instead,

2. For the contrast error, #666 on top of #fff does not seem like a big contrast problem in my opinion, but it looks like it is required for AAA compliance:


https://dequeuniversity.com/rules/axe/2.6/color-contrast

I'm inclined to ignore that error but maybe we should check with design and let them know that this is being reported as an issue?

PR #305, opened 2018-09-25
fix(modal): use role=dialog

gergelyke:


schnerd:
Seeing what some others do:
- [react-aria-modal](http://davidtheclark.github.io/react-aria-modal/demo/) uses role="dialog" on dialog body and aria-hidden="true" on body
- [Antd](https://ant.design/components/modal/) uses role="dialog" on mask and role="document" on modal body, no aria-hidden/aria-modal 
- [Atlaskit](https://atlaskit.atlassian.com/packages/core/modal-dialog) uses role="dialog" on dialog body, no aria-hidden/aria-modal
- [Material](https://material-ui.com/utils/modal/) uses role="document", no aria-hidden/aria-modal
- [Blueprint](https://blueprintjs.com/docs/#core/components/dialog) has no roles or aria attributes

[Latest w3c aria spec for modals](https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html) recommends role="dialog" and aria-modal="true", so I'll stick with that for now given the inconsistency among all the other implementations.

[Some more details on aria-modal](http://www.maxability.co.in/2018/04/aria-modal-property/)

PR #309, opened 2018-09-25
[RFC] Component for quick layout, spacing, and typography

schnerd:
Not trying to merge right now, just trying to gather feedback as this is still just an experimental idea. Opened as a PR to make inline comments easier.

### DIFF
--- a/rfcs/block-component.md
+++ b/rfcs/block-component.md
@@ -1,0 +1,78 @@
+# Block Component
+
+One of the most common needs when building web apps is creating layout elements that have certain layout and typography styles. 
+
+For more traditional apps that use global stylesheets, many of the popular CSS frameworks have helper classes to quickly create styled elements (see [Bootstrap](https://getbootstrap.com/docs/4.1/utilities/sizing/), [Tachyons](http://tachyons.io/docs/layout/spacing/)). Uber's superfine css framework also had this and it was used very widely–just search something like `push-large` in sourcegraph.
+
+The move to CSS-in-JS sort of eliminates these classes in favor of encapsulating styles within components. This is a generally a good thing because it promotes reusability, however the reality is that there are dozens or hundreds of different layout/spacing/typography combinations in any app, and there's not always a clear semantic meaning or potential for reuse. In practice you end up just defining and naming a bunch of things at the top of the file:
+
+```js
+const PricingPageHero = styled('div', {
+  margin: '10px 15px',
+  color: '#ccc',
+  display: 'flex',
+});
+
+const PricingPageHeroText = styled('h2', {
+  margin: '10px 0',
+});
+
+// ... and many more
+```
+
+---
+
+## Proposal
+
+### `<Block/>` Component
+
+The `Block` component is a helper component that lets you write these common less-reusable elements more succinctly inline.
+
+The above example might turn into:
+
+```js
+<Block marginSides="scale200" marginEnds="scale150" color="grey200" display="flex" flexDirection="column">
+  <Block as="h2" margin="scale150" justifySelf="left">
+    Simple Pricing
+  </Block>
+  {/* ... etc */}
+</Block>
+```
+
+Props would be exposed for:
+- Changing base element
+- Spacing (margin/padding)
+- Typography (color, font, font-size, font-style, etc)
+- Display, Position, Flexbox
+- Box shadows
+- `$style` prop for further style overrides
+
+[Pinterest's Gestalt has a similar Box component](https://pinterest.github.io/gestalt/#/Box), and [other](http://mineral-ui.com/components/text) [frameworks](https://evergreen.surge.sh/components/typography) have Text components that focus primarily on quick typography helpers. The [jsxstyle](https://github.com/jsxstyle/jsxstyle) css-in-js library uses this pattern as well. I've also created similar helper components internally at Uber and they've been helpful. 
+
+> **Note** that we would likely _not_ use this component internally to build other baseui components. It's primarily a helper component for customers when building their apps.
+
+### Why not separate `Spacing`,  `Color`, `Text`, `Shadow` etc components?
+
+The reality is you often want to apply multiple of these concerns to one element, but instead you'd end up with something like this:
+
+```js
+<Shadow type="overlay150">
+  <Spacing sides="scale150" ends="scale200">
+    <Color color="primary">Hello world</Color>
+  </Spacing>
+</Shadow>
+```
+
+Having multiple elements can complicate things like layout (flexbox children, etc), as well as just needlessly increasing the amount of DOM the client has to deal with.
+
+Combining most things into a single `Block` component allows you to keep things as a single element, and generally has less cognitive overload due to fewer react elements. (maybe there's a happy middle-ground though, for example a `Block` and `Text` component?)
+
+### Downsides
+
+It seems like there are two primary downsides here:
+- **Styles cluttering your markup** This is the price of being able to conveniently define things inline–one could argue the tradeoff is worth it here.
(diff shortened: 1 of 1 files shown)


kahwee:
I really like this and I did the same for my UI via [styled-system](https://github.com/jxnblk/styled-system). I'm not clear what is a `side` and what is an `end`.

I did some shorthands as well to make it look like:

```
<Block my={2} color="grey200">
```

Compare it with your original proposal:

```
<Block marginSides="scale200" color="grey200">
```

I find it frustrating to be verbose on composing margins and paddings.

rfcs/block-component.md
```diff
@@ -0,0 +1,78 @@
...
+
+### Downsides
+
+It seems like there are two primary downsides here:
+- **Styles cluttering your markup** This is the price of being able to conveniently define things inline–one could argue the tradeoff is worth it here.
+- **Performance** This component would need to convert its props into a style object during render, which in theory may be slower than `styled('div', ...)` with a static object. It's hard to say whether this would be a problem in practice, and there are probably ways to optimize this if needed like memoization or a babel transform.
+
+---
+
+It'd be great to get peoples reactions to this – would a component like this be helpful? are there any other benefits or downsides?
```
Warbring3r: Please take a look at Box, this worked very well for web developers at Pinterest as a div substitute for intelligent, flexible, and accessible layouts - https://github.com/pinterest/gestalt/blob/master/packages/gestalt/src/Box.js

Usage of would look like: 
`<Box marginTop={6} padding={3} shape="rounded" width={300} justifyContent="center" alignItems="center">
{props.children}
</Box>`
Box documentation: https://pinterest.github.io/gestalt/#/Box

kahwee:
[styled-system](https://github.com/jxnblk/styled-system) has a neat way too for attributes shorthand

* `m` means `margin`
* `mt` means `marginTop`
* `my` means `marginTop; marginBottom`
* `mx` means `marginLeft; marginRight`.

The same convention for padding is supported as well.

rfcs/block-component.md
```diff
@@ -0,0 +1,78 @@
...
+## Proposal
+
+### `<Block/>` Component
+
+The `Block` component is a helper component that lets you write these common less-reusable elements more succinctly inline.
+
+The above example might turn into:
+
+```js
+<Block marginSides="scale200" marginEnds="scale150" color="grey200" display="flex" flexDirection="column">
```
gergelyke: should `scale200` and these values rather came from the `theme` object? or that was the plane, you just wanted to keep the example simple?

rfcs/block-component.md
```diff
@@ -0,0 +1,78 @@
...
+- Typography (color, font, font-size, font-style, etc)
+- Display, Position, Flexbox
+- Box shadows
+- `$style` prop for further style overrides
+
+[Pinterest's Gestalt has a similar Box component](https://pinterest.github.io/gestalt/#/Box), and [other](http://mineral-ui.com/components/text) [frameworks](https://evergreen.surge.sh/components/typography) have Text components that focus primarily on quick typography helpers. The [jsxstyle](https://github.com/jsxstyle/jsxstyle) css-in-js library uses this pattern as well. I've also created similar helper components internally at Uber and they've been helpful. 
+
+> **Note** that we would likely _not_ use this component internally to build other baseui components. It's primarily a helper component for customers when building their apps.
+
+### Why not separate `Spacing`,  `Color`, `Text`, `Shadow` etc components?
```
Warbring3r: I think `<Text>` should be separated into its own component. Text has alignment, bold, italic, inline, overflow, and size properties.

rfcs/block-component.md
schnerd: The idea was that internally `Block` would first check if the a prop is a theme constant (like `scale200`), and if so it would pull it from the theme object. But in theory a user could also pass it from theme themselves:

```js
import {LightTheme} from 'baseui';
export default () => <Block marginSides={$theme.sizing.scale200} />;
```

just requires importing and a few more keystrokes to access the constant from theme yourself. I'm sort of indifferent to which approach is used.

rfcs/block-component.md
```diff
@@ -0,0 +1,78 @@
...
+  <Block as="h2" margin="scale150" justifySelf="left">
+    Simple Pricing
+  </Block>
+  {/* ... etc */}
+</Block>
+```
+
+Props would be exposed for:
+- Changing base element
+- Spacing (margin/padding)
```
Warbring3r: Thoughts: margin should be separated into marginTop, marginRight, marginLeft, and marginBottom, whereas padding should only have symmetric (padding, paddingX, and paddingY) attributes, to emphasize the differences between the two properties and proper usage.

rfcs/block-component.md
tusharsoni: I would recommend taking a look at https://github.com/jxnblk/reflexbox. It has just two primitives, Flex and Box. The API surface seems fairly small and only tackles the layout piece.

rfcs/block-component.md
```diff
@@ -0,0 +1,78 @@
...
+```js
+<Block marginSides="scale200" marginEnds="scale150" color="grey200" display="flex" flexDirection="column">
+  <Block as="h2" margin="scale150" justifySelf="left">
+    Simple Pricing
+  </Block>
+  {/* ... etc */}
+</Block>
+```
+
+Props would be exposed for:
```
tusharsoni: We would need props for responsive spacing, sizing, and layout as well with some default (but customizable) breakpoints

rfcs/block-component.md
```diff
@@ -0,0 +1,78 @@
...
+```
+
+Having multiple elements can complicate things like layout (flexbox children, etc), as well as just needlessly increasing the amount of DOM the client has to deal with.
+
+Combining most things into a single `Block` component allows you to keep things as a single element, and generally has less cognitive overload due to fewer react elements. (maybe there's a happy middle-ground though, for example a `Block` and `Text` component?)
+
+### Downsides
+
+It seems like there are two primary downsides here:
+- **Styles cluttering your markup** This is the price of being able to conveniently define things inline–one could argue the tradeoff is worth it here.
```
tusharsoni: Definitely want to stress this downside as we would be introducing a loaded component which could have a growing list of props as the there is no clear boundary of its role.
Additionally, it requires devs to lookup docs for something that most people would already know CSS props for. With this, I wouldn't know how to add something as simple as margins and need docs for it.

rfcs/block-component.md
Warbring3r: Padding doesn't have to be symmetric, it's not in the rules. It's more about making sure there's only one way to do things. If you want an asymmetric div in your box model, use margins (esp. since they can be negative), otherwise use padding. It maintains a "consistent visual rhythm" - from http://basscss.com/v7/docs/white-space/

abmai: I feel like good practice should be documented and not enforced via code. We can give guidance on what's the best approach to using the component and doing layout, but if people really want to roll with their own ideas they should be able to do it.

abmai:
Definitely sounds like a good idea; but perhaps we should highlight and align on the goals very clearly? Are we trying to save developers time, and/or enforce best practices, and/or just building syntactic sugar on top of `styletron`?

schnerd:
> Definitely sounds like a good idea; but perhaps we should highlight and align on the goals very clearly? Are we trying to save developers time, and/or enforce best practices, and/or just building syntactic sugar on top of styletron?

Agree we should be clear about goals.

In my mind the primary goal is saving developers time by eliminating the need to define and name styled components at the top of the file. The same 4 or 5 lines of code at the top of the file (or in a separate file) can sometimes be more quickly defined as a single inline element.

rfcs/block-component.md
schnerd: This is a great point. If you have any ideas of what such an API would look like, please share. I'm thinking we might be able to publish an initial version and then add responsive behavior later.

PR #316, opened 2018-09-27
refactor(e2e): create goToUrl

gergelyke:


gergelyke:
I agree, that it can be tedious  - what I am not sure of now is how turning into a class solves the issue? Can you drop a quick example here?

gergelyke:
will merge it now as is, and will refactor later

PR #318, opened 2018-09-27
fix(LightTheme): update 'primary' color and remove 'primary450' and 'primary470' values

nadiia:


rfcs/spinner-component.md
```diff
@@ -44,7 +44,7 @@ export default () => {
 
 * `size: number|string` - Optional, Defaults to 44
   The size of the circle. Is passed to the underlying svg width/height style. Can also be a baseui `theme.sizing` constant like `scale500`
-* `color: string` - Optional, Defaults to `theme.primary450`
+* `color: string` - Optional, Defaults to `theme.primary400`
```
abmai: Why not just call these `theme.primary`?

nadiia: good point! will update

gergelyke:
just for my information - where does the name for the docs file come from `docs/migrations/00318.md`? is that just the PR number?

nadiia:
yes, it's the [PR number](https://github.com/uber-workflow/probot-app-migrations/blob/master/index.js#L44)

PR #319, opened 2018-09-28
test(vrt): add for button

gergelyke:


TESTING.md
```diff
@@ -13,3 +13,25 @@ In a new terminal window start the tests using:
 ```sh
 yarn e2e:test
 ```
+
+## Running end to end tests against Sauce Labs
+
+```sh
+yarn e2e:build && yarn e2e:serve
```
blackswanny: can we combine these into one command?

e2e/helpers/index.js
```diff
@@ -11,7 +11,25 @@ function goToUrl({suite, test, client}) {
   return client.url(getUrl({launchUrl: client.launchUrl, suite, test}));
 }
 
+function formatFileName(testName) {
+  return testName.toLowerCase().replace(/ /g, '-');
+}
+
+function assertVisuals({client, id}) {
+  const isChrome = client.options.desiredCapabilities.browserName === 'chrome';
+  const isMac = client.options.desiredCapabilities.platform === 'macOS 10.12';
```
blackswanny: this is very exact version, may not be the same for other machines. Btw, what the Sauce Labs runs on?

gergelyke: what do you mean by `may not be the same for other machines`? here we request a chrome under macOS 10.12 - or are you referring to virtualization? if so, I don't know, but I guess that does not matter, does it?

TESTING.md
KevinGrandon: My recommendation: document how to do this with the docker command, so we can run tests as they are run in CI.

PR #326, opened 2018-10-01
Change to options as a function Promise

blackswanny:
- [x] Currently `props.onChange` is called for both _selected value_ change and _text input value_ change. We should refactor so that `onChange` is only called when _selected value_ changes, and we create another `onTextInputChange` for changes to the _text input_ value
- [x] Rather than having `onTextInputChange` return a promise that says when options have finished loading, we should just overload the `options` prop to be `Option[] | () => Promise<Option[]>`. This is more flexible as we can call the async options function when the dropdown is focused but still empty to get the initial list of options, as well as when the text input value changes.


src/select/stateful-select-container.js
```diff
@@ -41,6 +42,15 @@ class StatefulSelectContainer extends React.Component<
     return onChange(e, params);
   };
 
+  onTextInputChange = (e: SyntheticInputEvent<HTMLInputElement>) => {
+    this.stateReducer(STATE_CHANGE_TYPE.textChange, e, {
```
nadiia: Would it make sense to use `internalSetState` name here to align with [other components](https://github.com/uber-web/baseui/blob/master/src/popover/stateful-container.js#L85)?

src/select/select.js
```diff
@@ -384,6 +387,12 @@ class Select extends React.Component<PropsT, StatelessStateT> {
     return type === TYPE.search ? true : multiple;
   }
 
+  openDropDown(newTextValue?: string, callback?: () => void) {
+    this.loadOptions(newTextValue).then(() => {
+      this.setState({isDropDownOpen: true}, callback);
```
schnerd: There will probably be cases where async results take 1 or 2 seconds to return, and with this logic the dropdown may seem broken or laggy.

We should probably open the dropdown without waiting for results and show a spinner in the dropdown until they're finished loading.

src/select/stateful-select-container.js
```diff
@@ -26,7 +26,8 @@ class StatefulSelectContainer extends React.Component<
       selectedOptions: [],
     },
     stateReducer: defaultStateReducer,
-    onChange: () => Promise.resolve(),
+    onTextInputChange: () => Promise.resolve(),
```
schnerd: I think this should just be `() => {}`?

src/select/select.js
```diff
@@ -107,82 +96,95 @@ class Select extends React.Component<PropsT, StatelessStateT> {
...
-              this.setState({filteredOptions});
-            }
-          });
-        break;
-      }
-      case STATE_CHANGE_TYPE.keyDown: {
-        this.handledHotKeys(e);
-        break;
+        if (this.props.type === TYPE.select && !multiple) {
+          this.setState({isDropDownOpen: false});
```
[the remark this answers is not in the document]
schnerd: I'm struggling to think of a use case for giving the user full control over when the dropdown closes. It seems like the logic here may be sufficient for now.

src/select/select.js
```diff
@@ -107,82 +96,95 @@ class Select extends React.Component<PropsT, StatelessStateT> {
...
+    });
+    this.props.onTextInputChange(e);
+    this.openDropDown(newTextValue, () => {
+      if (this.props.filterable) {
+        let filteredOptions = this.state.options.filter(option =>
+          this.props.filterOption(option, newTextValue),
+        );
+        // reset filtered options for new search
+        if (!filteredOptions.length) {
+          filteredOptions = newTextValue ? [] : null;
```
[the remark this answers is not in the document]
blackswanny: empty array is still a filtered value with empty results. Whereas null identifies that we flushed previous filtered results and will start to check for new ones

src/select/select.js
```diff
@@ -107,82 +96,95 @@ class Select extends React.Component<PropsT, StatelessStateT> {
...
+      this.setState({selectedOptions: selectedOptions});
+      this.props.onChange(e, {
+        type: STATE_CHANGE_TYPE.unselect,
+        selectedOptions,
+      });
+    } else {
+      const selected = this.state.selectedOptions.find(
+        tag => tag.id === pendingOption.id,
+      );
+      const isSelectOrUnselect = !selected;
```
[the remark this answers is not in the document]
blackswanny: renamed to `isSelect`

src/select/select.js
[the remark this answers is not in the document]
blackswanny: right now `isDropDownOpen` sets by many rules, but not from `initialState` from customer. I am not sure that we need to provide it, at least it may be controvertial in Search mode, where no results yet in DropDown

src/select/select.js
```diff
@@ -280,7 +283,7 @@ class Select extends React.Component<PropsT, StatelessStateT> {
           After: () => (
             <SearchIcon
               $disabled={disabled}
-              onClick={e => this.onChange(e, STATE_CHANGE_TYPE.clearAll)}
+              onClick={e => this.onChange(e, undefined, true)}
```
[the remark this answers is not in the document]
blackswanny: `SearchIcon` is the name of all icons for this components, including `clear all` one. I was going to rename it, but after some comment left it as is.

src/select/stateful-select-container.js
blackswanny: fixed

src/select/select.js
```diff
@@ -215,7 +217,10 @@ class Select extends React.Component<PropsT, StatelessStateT> {
       : {
           onKeyDown: e => this.handledHotKeys(e),
           onClick: () => {
-            this.setState({isDropDownOpen: !this.state.isDropDownOpen});
+            const newValue = !this.state.isDropDownOpen;
+            newValue
+              ? this.openDropDown()
+              : this.setState({isDropDownOpen: newValue});
```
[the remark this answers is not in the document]
blackswanny: I was trying to be more descriptive

src/select/select.js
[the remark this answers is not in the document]
nadiia: It seems like the dropdown won't be closed when an optioned selected and it's a multiselect component.

src/select/select.js
blackswanny: so different products may have full loading mask or smaller one for particular region\component. We don't know the exact use case. Would you like me to use our standard spinner?

src/select/select.js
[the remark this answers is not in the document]
blackswanny: Yes, in multiple options they usually pick many options at once, so if we close it after each selection, that might be not the expected behavior
@nguyensomniac

src/select/select.js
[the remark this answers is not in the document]
nadiia: Maybe we need to rename the icon to have some general name, like `Icon`, `AppendIcon` 🤷‍♀️

src/select/select.js
nadiia: +1 previous conversation can be [found here](https://github.com/uber-web/baseui/pull/323#discussion_r221406151)

blackswanny: ok, fixed, added spinner

src/select/select.js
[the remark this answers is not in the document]
blackswanny: renamed to `SelectComponentIcon`

src/select/stateful-select-container.js
```diff
@@ -70,12 +80,11 @@ class StatefulSelectContainer extends React.Component<
     switch (type) {
       case STATE_CHANGE_TYPE.select:
       case STATE_CHANGE_TYPE.unselect:
-      case STATE_CHANGE_TYPE.clearAll:
         nextState = {
           selectedOptions: params.selectedOptions,
         };
         break;
-      case STATE_CHANGE_TYPE.keyDown: {
+      case STATE_CHANGE_TYPE.textInputChange: {
```
schnerd: Continuing the conversion from here: https://github.com/uber-web/baseui/pull/323#discussion_r221400458

It seems problematic to offer `textValue` as a prop in the stateless select if it's also stored in state internally. If a user tries to change the `textValue` prop after the component is constructed, it won't actually do anything. `textValue` should probably be removed from stateful component and only kept in `state` of stateless component. Can't think of a good use case for a user controlling this from outside.

`selectedOptions`, on the other hand, is the primary form control value and should probably live in stateful component and only be passed as a prop to stateless component (not stored in `state` there).

These changes might make more sense of a follow-up PR, but I think it's relevant to this PR because there's no point in renaming and offering `STATE_CHANGE_TYPE .textInputChange` if we're just going to remove it in a future PR.

blackswanny: ok, changed

src/select/types.js
```diff
@@ -29,25 +29,25 @@ export type ParamsT = {
...
 
 export type OverridesDropDownT = {
   Option?: OverrideT<*>,
   DropDown?: OverrideT<*>,
-  SearchIcon?: OverrideT<*>,
+  SelectComponentIcon?: OverrideT<*>,
   DropDownItem?: OverrideT<*>,
 };
 
 export type PropsT = {
```
schnerd: I think you also want to remove `textValue` from PropsT?

src/select/stateful-select-container.js
nadiia: Do we clear the text input's value when an option is selected? When the input's value is cleared in the case of multiselect?

src/select/types.js
blackswanny: ok

src/select/stateful-select-container.js
blackswanny: no, we do not. This request from Ahn seems to be controversial. I explained that commenting it in previous PR https://github.com/uber-web/baseui/pull/292#issuecomment-423381247

It can happen only in Search mode, however:
- we do not have design for single select mode. And I can't imagine how it should look like
- In multiple select (and may be in single) once we clear it, it will trigger close of DropDown (not sure if it's what expected) and will cause options to be reloaded as text query has changed

src/select/dropdown.js
```diff
@@ -88,13 +91,17 @@ export default function SelectDropDown(props: DropDownPropsT) {
                   )}
                   {getOptionLabel(option)}
                 </Option>
+              ) : (
+                <div style={{paddingLeft: '50%'}}>
```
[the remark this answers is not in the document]
blackswanny: It was just one style.
I have added this element to overrides sd `SelectSpinner`. However, `Spinner` itself is external dependency, we shouldn't provide override for it as well. If someone needs to override `SelectSpinner` they can do it.

PR #329, opened 2018-10-01
fix(license): backfill licenses

gergelyke:


src/form-control/types.js
```diff
@@ -1,5 +1,11 @@
 /*
 Copyright (c) 2018 Uber Technologies, Inc.
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+/*
+Copyright (c) 2018 Uber Technologies, Inc.
```
nadiia: seems like all `form-control` files have the license message doubled

PR #330, opened 2018-10-02
[RFC] Component for Navigation Bar

ywu151:


rfcs/navigation-component.md
```diff
@@ -0,0 +1,103 @@
...
+* `StyledLi`
+* `StyledLogoLink`
+* `StyledNav`
+* `StyledNavLink`
+* `StyledUl`
+
+## `Navigation` API
+
+* `navItems: Array<Object>`- Required:
+  All navigatioin items. Every item object has `id: string`, `url: string` and `text: string`. Optional `disabled: boolean` for item to be disabled from navigation.
```
blackswanny: navigation is not only about URL, or URL may be not static. I propose to have a `action` function here, which does, what needs to be done itself (go to url, activate menu, etc), with Promise return (when action is done)

rfcs/navigation-component.md
```diff
@@ -0,0 +1,103 @@
...
+  );
+};
+```
+
+## Exports
+
+* `Navigation`
+* `StyledActiveNavLink`
+* `StyledHeader`
+* `StyledLi`
```
blackswanny: `li` and `ul` are not semantic here. What part of navigation they refer to?

schnerd:
Some higher level questions I think we should answer before moving forward with implementation:
- How is the currently "Active" nav item determined?
  - Relying directly on the URL (`window.location.href`) alone likely wouldn't work for server-side rendering, adds more url pattern matching logic to the component, and doesn't necessarily work for nav items that don't change the URL.
- How does this integrate with something like react-router?
  - The nav item can't just be a naive `<a href="">`, we probably want to a strategy that works well with react-router or other router libraries. We also don't want to directly depend on react-router. What this means is we may want to offer an API that allow users to use their own react-router [`<NavLink>`](https://reacttraining.com/react-router/web/api/NavLink) or something similar. We should create some code examples of how this might work.
- How do we give users more flexibility in terms of navigation layout? 
  - The current API in the PR seems to assume one set of navigation items, but in many applications you often have multiple navigation sections. One for main page navigation and then one for authentication and settings. Uber.com is a perfect example of this. We should have a solution here.

rfcs/navigation-component.md
```diff
@@ -0,0 +1,103 @@
...
+* `StyledUl`
+
+## `Navigation` API
+
+* `navItems: Array<Object>`- Required:
+  All navigatioin items. Every item object has `id: string`, `url: string` and `text: string`. Optional `disabled: boolean` for item to be disabled from navigation.
+* `projectName: string` - Required:
+  Project name.
+* `isWithLoge: boolean` - Optional:
+  With `Uber` logo at the front of project name.
```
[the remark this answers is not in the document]
schnerd: +1, components ideally shouldn't have anything uber-specific.

schnerd:
Following up on my last comment, it seems like a more composable API would solve a lot of these problems:

```js
import {Navigation, NavigationList, NavigationItem, Button} from 'baseui/navigation';
import {NavLink} from 'react-router';

// ... 

render() {
  return (
    <Navigation>

      <NavigationList align="left">
        <NavigationItem>
          <NavLink to="/" title="Home">
            <MyCustomLogo />
          </NavLink>
        </NavigationItem>
        <NavigationItem>
          <NavLink to="/about">About</NavLink>
        </NavigationItem>
        <NavigationItem>
          <NavLink to="/products">Products</NavLink>
        </NavigationItem>
        <NavigationItem>
          <NavLink to="/faqs">FAQs</NavLink>
        </NavigationItem>
      </NavigationList>

      <NavigationList align="right">
        <NavigationItem>
          <Button kind="tertiary" onClick={this.showSignInModal()}>Sign In</Button>
        </NavigationItem>
        <NavigationItem>
          <Button kind="primary" onClick={this.showSignUpModal()}>Sign Up</Button>
        </NavigationItem>
      </NavigationList>

    </Navigation>
  );
}
```

ywu151:
> Some higher level questions I think we should answer before moving forward with implementation:
> 
> * How is the currently "Active" nav item determined?
>   
>   * Relying directly on the URL (`window.location.href`) alone likely wouldn't work for server-side rendering, adds more url pattern matching logic to the component, and doesn't necessarily work for nav items that don't change the URL.
> * How does this integrate with something like react-router?
>   
>   * The nav item can't just be a naive `<a href="">`, we probably want to a strategy that works well with react-router or other router libraries. We also don't want to directly depend on react-router. What this means is we may want to offer an API that allow users to use their own react-router [`<NavLink>`](https://reacttraining.com/react-router/web/api/NavLink) or something similar. We should create some code examples of how this might work.
> * How do we give users more flexibility in terms of navigation layout?
>   
>   * The current API in the PR seems to assume one set of navigation items, but in many applications you often have multiple navigation sections. One for main page navigation and then one for authentication and settings. Uber.com is a perfect example of this. We should have a solution here.

- For my code, currently I use `withRouter` to pass `location.pathname` into props. It doesn't work for nav items that don't change the URL.
- I think it would be a complex design work for the layout, since we need to consider the dropdown list, and sublists of some items.

gergelyke:
@ywu151 do you think you'll have time to address the comments on this PR?

ywu151:
Sorry, I don’t have enough time for this component

gergelyke:
I am closing it, we will take this on

PR #331, opened 2018-10-02
test(cards): add end to end tests

gergelyke:
* moves to mocha-style tests from the nighwatch test runner, as that had issues when we had multiple tests in a single file
* adds the vrt tests for the card component


package.json
```diff
@@ -28,8 +28,7 @@
     "build:es": "cross-env NODE_ENV=production BABEL_ENV=es babel ./src --out-dir ./dist/es --ignore *.test.js",
     "build": "yarn build:es2015 && yarn build:es2015modules && yarn build:es && yarn build:copy-files && yarn build:copy-flow-files",
     "unit-test": "yarn jest --coverage",
-    "localtunnel": "localtunnel",
-    "e2e": "node ./e2e/e2e.js",
+    "localtunnel": "lt",
```
nadiia: what does this do?

gergelyke: it just creates a shortcut to the localtunnel package (so you don't have to do `./node_modules/lt`, but just `yarn localtunnel`), which exposes the web app externally

PR #332, opened 2018-10-02
feat: add implementation of stateless slider

blackswanny:
add implementation of stateless slider


gergelyke:
@blackswanny should we fix the flow / lint errors before reviewing?

blackswanny:
@gergelyke they are fixed in child branch as a part of implementation

src/slider/styled-components.js
```diff
@@ -0,0 +1,89 @@
+/*
+Copyright (c) 2018 Uber Technologies, Inc.
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+// @flow
+import {styled} from '../styles/index';
+import {} from './constants';
+import {getToggleCheckMarkStyles} from '../checkbox/styled-components';
```
schnerd: This really may be a situation where it's better to copy the styles rather than trying to be overly DRY. Even though this looks similar to the toggle, we're now significantly complicating our checkbox logic by adding `$isStart`, `$isEnd`, etc.

src/slider/styled-components.js
```diff
@@ -0,0 +1,89 @@
...
+      background += ' 0% ';
+    } else {
+      background += i % 2 === 0 ? fillColor : emptyColor;
+      background += ' ' + values[i] + '%, ';
+      background += i % 2 === 0 ? emptyColor : fillColor;
+      background += ' 0% ';
+    }
+  }
+  background += ', ' + emptyColor + ' 0%';
+  background += ')';
```
schnerd: Seems like there are cases where the blue fill can be past the handle?
image

schnerd: It also feels like using two elements here instead of a linear gradient would simplify a lot of this logic. Also makes it easier for people to customize the fill and background color.

src/slider/styled-components.js
```diff
@@ -0,0 +1,89 @@
...
+      position: 'absolute',
+      content: `"${Math.round(props.$value[props.$index])}"`,
+      display: 'block',
+      top: '-24px',
+      marginLeft: '4px',
+      width: '16px',
+      height: '16px',
+    },
+  };
+});
```
schnerd: Should the thumb have `cursor: pointer`?

src/slider/types.js
```diff
@@ -0,0 +1,72 @@
...
+
+export type PropsT = {
+  value: Array<*>,
+  range: Array<*>,
+  overrides?: OverridesT,
+  tabIndex: number,
+  error: boolean,
+  autoFocus: boolean,
+  disabled?: boolean,
+  $theme?: *,
```
schnerd: Why not `ThemeT` here?

schnerd: Also ThemeT isn't really part of the public props interface.

src/slider/types.js
```diff
@@ -0,0 +1,72 @@
...
+  Root?: OverrideT<*>,
+  Axis?: OverrideT<*>,
+  Tick?: OverrideT<*>,
+  TickBar?: OverrideT<*>,
+  Thumb?: OverrideT<*>,
+};
+
+export type PropsT = {
+  value: Array<*>,
+  range: Array<*>,
```
schnerd: Can we be more descriptive than `*` here? Is it `number | {value: number, label: string}`

src/slider/types.js
```diff
@@ -0,0 +1,72 @@
...
+  autoFocus: boolean,
+  disabled?: boolean,
+  $theme?: *,
+  onChange: (e: SyntheticEvent<HTMLInputElement>, params: ParamsT) => void,
+  onMouseEnter: (e: SyntheticEvent<HTMLInputElement>) => void,
+  onMouseLeave: (e: SyntheticEvent<HTMLInputElement>) => void,
+  onMouseDown: (e: SyntheticEvent<HTMLInputElement>) => void,
+  onMouseUp: (e: SyntheticEvent<HTMLInputElement>) => void,
+  onFocus: (e: SyntheticEvent<HTMLInputElement>) => void,
+  onBlur: (e: SyntheticEvent<HTMLInputElement>) => void,
```
schnerd: Can we leave off all these extra handlers besides `onChange`? From the discussion in #372 it sounds like @abmai is also on board with this change. Unless there is some special behavior that we provide beyond what the user would receive by using `overrides={{Root: {props: {onMouseEnter: ...}}}}`.

src/slider/slider.js
```diff
@@ -0,0 +1,170 @@
...
+    const sharedProps = {
+      $value: value,
+      $range: range,
+      $max: range[range.length - 1],
+      $min: range[0],
+      $currentThumb: currentThumb,
+    };
+    return (
+      <Root {...rootProps} {...sharedProps}>
+        <Axis {...axisProps} {...sharedProps}>
```
schnerd: Most other sliders support clicking on the bar to set the value

kapture 2018-10-09 at 10 02 42

Do we plan to support this?

src/slider/styled-components.js
schnerd: Hmmm i guess native doesn't have it. Suppose

src/slider/styled-components.js
blackswanny: ok, removed extra code from `checkbox` and left it in slider so far

src/slider/styled-components.js
blackswanny: fixed this. Also applied approach used in progress bar with couple divs for that

src/slider/styled-components.js
blackswanny: fixed

src/slider/types.js
blackswanny: fixed

src/slider/types.js
blackswanny: fixed

src/slider/slider.js
blackswanny: I have added this feature for only 1 thumb

src/slider/slider.js
```diff
@@ -0,0 +1,170 @@
...
+    );
+    const sharedProps = {
+      $value: value,
+      $range: range,
+      $max: range[range.length - 1],
+      $min: range[0],
+      $currentThumb: currentThumb,
+    };
+    return (
+      <Root {...rootProps} {...sharedProps}>
```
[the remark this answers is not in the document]
blackswanny: yes. fixed

blackswanny:
In src/slider/styled-components.js:

>     backgroundColor: null,
    position: 'relative',
    top: '-10px',
    left: offset + '%',
    marginLeft: props.$index % 2 ? '-32px' : '-16px',
    ':before': {
      position: 'absolute',
      content: `"${Math.round(props.$value[props.$index])}"`,
      display: 'block',
      top: '-24px',
      marginLeft: '4px',
      width: '16px',
      height: '16px',
    },
  };
});
> Hmmm i guess native doesn't have it. Suppose

@schnerd do you mean label?

blackswanny:
@nadiia I checked new version and it seems to address these issues, except in Range mode we can overlap one thumb with another. I am to fix this as well

src/slider/examples.js
```diff
@@ -0,0 +1,67 @@
...
+*/
+// @flow
+/* eslint-disable react/display-name*/
+
+import * as React from 'react';
+// Styled elements
+import {StatefulSlider as Slider} from './index';
+
+export const suite = 'Slider Test Suite';
+export const tests = {
```
gergelyke: can you move this out to an `example-list.js` file, just like how the modal / chechbox has it?

src/slider/slider.js
```diff
@@ -0,0 +1,225 @@
...
+    const move = currentMove + movementX;
+    const {$value, $max, $min, $isRange} = this.getSharedProps();
+    const value = $value.slice();
+    let {onChange, step} = this.props;
+    const axisSizeInPixels = parseInt(
+      window.getComputedStyle(this.domNode).width,
+    );
+    const axisSize = $max - $min;
+    step = step ? Slider.scale(step, axisSize, axisSizeInPixels) : step;
+    const scaledStep = step ? (move >= 0 ? step : -step) : move;
```
gergelyke: maybe this is just me, but can we break it down into a more easily understandable format?

src/slider/slider.js
```diff
@@ -0,0 +1,225 @@
...
+    const {value} = this.props;
+    const {thumbRefs} = this.state;
+    // only for one thumb feature
+    //$FlowFixMe
+    if (value.length === 1 && e.currentTarget.parentNode) {
+      const thumb = thumbRefs[0].current;
+      const thumbWidth = 32;
+      const axisOffset = e.clientX - e.currentTarget.parentNode.offsetLeft;
+      //$FlowFixMe
+      let xPos = axisOffset - thumb.offsetLeft - thumbWidth;
```
gergelyke: can reason for not having `xPos` as a `const`?

src/slider/slider.js
```diff
@@ -0,0 +1,225 @@
...
+  }
+
+  onAxisClick(e: MouseEvent) {
+    const {value} = this.props;
+    const {thumbRefs} = this.state;
+    // only for one thumb feature
+    //$FlowFixMe
+    if (value.length === 1 && e.currentTarget.parentNode) {
+      const thumb = thumbRefs[0].current;
+      const thumbWidth = 32;
```
gergelyke: where does this constant comes from? should we move it out to a contants file?

src/checkbox/styled-components.js
```diff
@@ -143,11 +143,13 @@ function getToggleThumbColor(props) {
 const getToggleCheckMarkStyles = props => {
   const {$checked, $theme, $disabled} = props;
   const {animation, colors} = $theme;
+  const backGroundColor = $disabled ? colors.mono500 : 'white';
```
nadiia: Should probably be consistent here with the "background" camelcase - `backgroundColor`

src/checkbox/styled-components.js
```diff
@@ -143,11 +143,13 @@ function getToggleThumbColor(props) {
...
+  const toggleThumbColor = getToggleThumbColor(props);
   const toggleSVG =
     `<svg width="32" height="32" viewBox="0 0 32 32" fill="none" xmlns="http://www.w3.org/2000/svg"><g filter="url(#filter0_d)"><rect x="4" y="3" width="24" height="24" rx="4" fill="` +
-    ($disabled ? colors.mono500 : 'white') +
+    `${backGroundColor}` +
```
nadiia: same here

src/slider/slider.js
```diff
@@ -0,0 +1,225 @@
...
+  };
+
+  state = {
+    isThumbMoving: false,
+    currentThumb: -1,
+    currentMove: 0,
+    thumbRefs: this.props.value.map(() => React.createRef()),
+  };
+
+  constructor(props: PropsT) {
```
nadiia: no need for the constructor since it doesn't do anything in here

src/slider/slider.js
```diff
@@ -0,0 +1,225 @@
...
+    super(props);
+  }
+
+  onFocus = (e: SyntheticEvent<HTMLInputElement>) => {};
+
+  onBlur = (e: SyntheticEvent<HTMLInputElement>) => {};
+
+  onMouseEnter = (e: SyntheticEvent<HTMLInputElement>) => {};
+
+  onMouseLeave = (e: SyntheticEvent<HTMLInputElement>) => {};
```
nadiia: what's the purpose of these noops above?

src/slider/slider.js
```diff
@@ -0,0 +1,225 @@
...
+
+  onChange = (e: SyntheticEvent<HTMLInputElement>, type: ChangeActionT) => {};
```
nadiia: For this event handler we should probably adopt the `{}` arg approach we intended to move to. @gergelyke what do you think?

src/slider/slider.js
```diff
@@ -0,0 +1,225 @@
...
+
+  componentDidMount() {
+    /* eslint-disable-next-line react/no-find-dom-node */
+    this.domNode = findDOMNode(this);
```
nadiia: Why don't we pass and use refs instead of `findDOMNode(this)`?

src/slider/slider.js
```diff
@@ -0,0 +1,225 @@
...
+    this.addDocumentMouseEvents();
+  }
+
+  componentWillUnmount() {
+    if (super.componentWillUnmount) super.componentWillUnmount();
```
nadiia: Not sure why we need this line, could you collaborate in this?

src/slider/slider.js
```diff
@@ -0,0 +1,225 @@
...
+    );
+    const [Thumb, thumbProps] = getOverrides(overrides.Thumb, StyledThumb);
+    const [Tick, tickProps] = getOverrides(overrides.Tick, StyledTick);
+    const [TickBar, tickBarProps] = getOverrides(
+      overrides.TickBar,
+      StyledTickBar,
+    );
+    const sharedProps = this.getSharedProps();
+    return (
+      <Root {...sharedProps} onClick={e => this.onAxisClick(e)} {...rootProps}>
```
nadiia: Wonder if we should we move the `onClick` handler to go after `rootProps` but call a customer's `onClick` handler within it. What do you think?

src/slider/stateful-slider-container.js
```diff
@@ -0,0 +1,68 @@
...
+    onChange: () => {},
+    onMouseEnter: () => {},
+    onMouseLeave: () => {},
+    onFocus: () => {},
+    onBlur: () => {},
+  };
+
+  state = {...this.props.initialState};
+
+  onChange = (e: SyntheticInputEvent<HTMLInputElement>, params: ParamsT) => {
```
nadiia: Change to the `{}` type arg for this handler

src/slider/styled-components.js
```diff
@@ -0,0 +1,147 @@
...
+    position: 'relative',
+  };
+});
+
+export const Axis = styled('div', props => {
+  const {$theme} = props;
+  const {colors} = $theme;
+  return {
+    position: 'relative',
+    margin: '34px 10px 10px',
```
nadiia: Use longhand props instead of a shorthand prop.

src/slider/styled-components.js
```diff
@@ -0,0 +1,147 @@
...
+    borderRadius: $theme.borders.useRoundedCorners ? '2px' : '0px',
```
nadiia: Should we use values from `$theme` for the border radius px value?

src/slider/examples.js
blackswanny: ok

src/slider/styled-components.js
```diff
@@ -0,0 +1,147 @@
...
+    background: colors.mono400,
+    height: '4px',
+  };
+});
+
+export const AxisRange = styled(StyledBarProgress, props => {
```
nadiia: Does it make sense to create a separate styled component for the slider instead of reusing `StyledBarProgress`?

src/slider/styled-components.js
```diff
@@ -0,0 +1,147 @@
...
+  const {$max, $min, $index, $isRange, $value, $theme} = props;
+  const {colors} = $theme;
+  const fillColor = colors.primary400;
+  const emptyColor = colors.mono400;
+  let $color;
+  if ($isRange) {
+    $color = $index % 2 === 0 ? emptyColor : fillColor;
```
nadiia: How about we name this var `barColor` for example, since usage of the `$` may bring some confusion around if this color comes from props.

src/slider/styled-components.js
```diff
@@ -0,0 +1,147 @@
...
+    $max,
+    $index,
+    $currentThumb,
+    $disabled,
+    $isRange,
+  } = props;
+  const $isActive = $index === $currentThumb;
+  const offset = Math.round(($value[$index] / $max) * 100);
+  const {animation, colors} = $theme;
+  const backGroundColor = $disabled ? colors.mono500 : 'white';
```
nadiia: `backgroundColor`

src/slider/slider.js
blackswanny: changed to be more readable

src/slider/slider.js
blackswanny: moved to constants and shown link to Thumb itself. I don't like magic numbers myself.

src/checkbox/styled-components.js
blackswanny: ok

src/checkbox/styled-components.js
blackswanny: ok

src/slider/slider.js
blackswanny: ok

src/slider/slider.js
blackswanny: removed

src/slider/slider.js
```diff
@@ -0,0 +1,225 @@
...
+      );
+    }
+  }
+
+  removeDocumentEvents() {
+    this.onMouseMoveListener && this.onMouseMoveListener.remove();
+    this.onMouseUpListener && this.onMouseUpListener.remove();
+  }
+
+  render() {
```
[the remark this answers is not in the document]
blackswanny: ok

src/slider/slider.js
```diff
@@ -0,0 +1,225 @@
...
+      <Root {...sharedProps} onClick={e => this.onAxisClick(e)} {...rootProps}>
+        <Axis {...sharedProps} {...axisProps}>
+          {value.map((thumb, $index) => (
+            <React.Fragment key={$index}>
+              <AxisRange $index={$index} {...sharedProps} {...axisRangeProps} />
+              <Thumb
+                $ref={thumbRefs[$index]}
+                key={$index}
+                $index={$index}
+                {...thumbProps}
```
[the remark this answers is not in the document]
blackswanny: ok

src/slider/slider.js
```diff
@@ -0,0 +1,225 @@
...
+                {...sharedProps}
+                onMouseDown={e => this.onMouseDown(e, $index)}
```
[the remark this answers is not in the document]
blackswanny: added, and other handlers as well

src/slider/stateful-slider-container.js
```diff
@@ -0,0 +1,68 @@
...
+  StateT,
+> {
+  static defaultProps = {
+    initialState: {},
+    stateReducer: defaultStateReducer,
+    onChange: () => {},
+    onMouseEnter: () => {},
+    onMouseLeave: () => {},
+    onFocus: () => {},
+    onBlur: () => {},
```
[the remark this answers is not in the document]
blackswanny: removed

src/slider/stateful-slider-container.js
blackswanny: ok

src/slider/styled-components.js
blackswanny: ok

src/slider/styled-components.js
blackswanny: ok

src/slider/styled-components.js
blackswanny: ok

src/slider/styled-components.js
blackswanny: ok

src/slider/styled-components.js
blackswanny: ok

src/slider/styled-components.js
```diff
@@ -0,0 +1,147 @@
...
+  } else {
+    toggleSVG = `<svg width="32" height="32" viewBox="0 0 32 32" fill="none" xmlns="http://www.w3.org/2000/svg"><g filter="url(#filter0_d)"><rect x="4" y="3" width="24" height="24" rx="4" fill="${backGroundColor}"/><rect width="2" height="8"  x="15" y="11" rx="1" fill="${toggleThumbColor}"/></g><defs><filter id="filter0_d" x="0" y="0" width="32" height="32" filterUnits="userSpaceOnUse" color-interpolation-filters="sRGB"><feFlood flood-opacity="0" result="BackgroundImageFix"/><feColorMatrix in="SourceAlpha" type="matrix" values="0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 127 0"/><feOffset dy="1"/><feGaussianBlur stdDeviation="2"/><feColorMatrix type="matrix" values="0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0.32 0"/><feBlend mode="normal" in2="BackgroundImageFix" result="effect1_dropShadow"/><feBlend mode="normal" in="SourceGraphic" in2="effect1_dropShadow" result="shape"/></filter></defs></svg>`;
+  }
+  return {
+    transitionDuration: animation.timing100,
+    transitionTimingFunction: animation.easeOutCurve,
+    transitionProperty: 'background-color',
+    ':after': {
+      position: 'absolute',
+      content: `url('data:image/svg+xml;utf8,` + toggleSVG + `')`,
```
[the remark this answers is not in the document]
blackswanny: ok

src/slider/slider.js
blackswanny: ok

src/slider/slider.js
```diff
@@ -0,0 +1,225 @@
...
+    const {currentThumb, currentMove} = this.state;
+    const move = currentMove + movementX;
+    const {$value, $max, $min, $isRange} = this.getSharedProps();
+    const value = $value.slice();
+    let {onChange, step} = this.props;
+    const axisSizeInPixels = parseInt(
+      window.getComputedStyle(this.domNode).width,
+    );
+    const axisSize = $max - $min;
+    step = step ? Slider.scale(step, axisSize, axisSizeInPixels) : step;
```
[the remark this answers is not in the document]
blackswanny: my Lint checker highlighted it as warning and proposed me to make it static

src/slider/slider.js
```diff
@@ -0,0 +1,225 @@
...
+    const scaledStep = step ? (move >= 0 ? step : -step) : move;
+    if (Math.abs(move) - Math.abs(scaledStep) >= 0) {
+      const scaledMove = Slider.scale(scaledStep, axisSizeInPixels, axisSize);
+      const max = !currentThumb && $isRange ? value[1] : $max;
+      const min = currentThumb && $isRange ? value[0] : $min;
+      value[currentThumb] = Slider.clampIt(
```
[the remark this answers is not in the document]
blackswanny: my Lint checker highlighted it as warning and proposed me to make it static

src/slider/slider.js
```diff
@@ -0,0 +1,225 @@
...
+    } else {
+      this.setState({currentMove: move});
+    }
+    onChange(e, {type: STATE_CHANGE_TYPE.value, value});
+  }
+
+  onAxisClick(e: MouseEvent) {
+    const {value} = this.props;
+    const {thumbRefs} = this.state;
+    // only for one thumb feature
```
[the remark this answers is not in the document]
blackswanny: the main confusion is to detect, which of the thumbs to consider active here, left or right. Maybe user would expect left thumb to be moved, when we choose right one, even if right was closer to click

blackswanny:
>+      const thumbWidth = 32;
Same question. Is this the width of a thumb that can be customized?

@nadiia . Yes, it's SVG icon and can be customized. However, it's not provided in any measurements I found for thumb( may be it is somewhere deep in DOM), so I don't have bulletproof way to avoid some deviation in coordinates

blackswanny:
> Maybe add some comments on this part of the code to help understand what's happening and what all these calculations are for?


added

blackswanny:
> Let's move this SVGs into a separate file, it will help keep the styled-components file cleaner

moved

blackswanny:
> +    const {thumbRefs} = this.state;
> +    // only for one thumb feature
> We will only need to decide which of the thumbs to move in the case a click is inside a current range. It's a good idea of moving the one that closer to the click in this case.
> For the case, a click is outside of the current range there wouldn't be any confusion since the thumb in a range can't go past the second one. Also in a later case, the distance approach would still work the same.

Yes, outside seems clear case. However, inside is not yet. As a user of sliders I always have expectation of the one thumb to move that I stare at, and it's not always the closest to click.

src/checkbox/styled-components.js
```diff
@@ -143,11 +143,13 @@ function getToggleThumbColor(props) {
 const getToggleCheckMarkStyles = props => {
   const {$checked, $theme, $disabled} = props;
   const {animation, colors} = $theme;
+  const backgroundColor = $disabled ? colors.mono500 : 'white';
```
nadiia: use `colors.white` instead of a `'white'` string

src/slider/slider.js
```diff
@@ -0,0 +1,233 @@
...
+    const {isThumbMoving} = this.state;
+    if (isThumbMoving) {
+      this.setState(
+        {
+          isThumbMoving: false,
+          currentThumb: -1,
+          currentMove: 0,
+        },
+        () =>
+          this.props.onMouseUp({
```
nadiia: Maybe we should name this prop with more context on what it's for. Something like `onThumbDropped`. Also when a thumb is dropped it means the slider's value changed, so it's basically a change event and we can name it `onChange`. What do you think?

src/slider/slider.js
```diff
@@ -0,0 +1,233 @@
...
+            event,
+            isThumbMoving: this.state.isThumbMoving,
```
nadiia: What is the value of `isThumbMoving` prop to be passed?

src/slider/slider.js
```diff
@@ -0,0 +1,233 @@
...
+          }),
+      );
+    } else {
+      this.props.onMouseUp({event, isThumbMoving});
```
nadiia: It seems like we don't need to call the user's handler here since a thumb wasn't active/moving.

src/slider/slider.js
```diff
@@ -0,0 +1,233 @@
...
+    }
+  };
+
+  onMouseMove = (event: MouseEvent) => {
+    const {isThumbMoving} = this.state;
+    if (isThumbMoving) {
+      this.onMove(event.movementX, event);
+    }
+    this.props.onMouseMove({event, isThumbMoving});
```
nadiia: Same here, no need to call this handler when `move` event is not related to the Slider component.

src/slider/stateful-slider-container.js
```diff
@@ -0,0 +1,67 @@
...
+  };
+
+  state = {...this.props.initialState};
+
+  onChange = (config: {event: *, params: ParamsT}) => {
+    const {event, ...params} = config;
+    //$FlowFixMe
+    this.internalSetState(STATE_CHANGE_TYPE.value, event, params);
+    const {onChange} = this.props;
+    return onChange({event, ...params});
```
nadiia: Isn't it going to produce args that look like next:
```
{
  event,
  params: {
    value: 'some value'
  }
}
```
Based on [this change](https://github.com/uber-web/baseui/pull/372/files) we don't next extra arguments under `params` prop, but it should look like next:
```
{
  event,
  value: 'some value'
}
```

src/slider/stateful-slider-container.js
```diff
@@ -0,0 +1,67 @@
...
+  StatefulContainerPropsT,
+  StateT,
+> {
+  static defaultProps = {
+    initialState: {},
+    stateReducer: defaultStateReducer,
+    onChange: () => {},
+  };
+
+  state = {...this.props.initialState};
```
nadiia: Should there be any initial value state?

src/slider/stateful-slider-container.js
```diff
@@ -0,0 +1,67 @@
...
+  ) => {
+    const nextState = {value};
+    const {stateReducer} = this.props;
+    const newState = stateReducer(type, nextState, this.state, e);
+    this.setState(newState);
+  };
+
+  render() {
+    const {
+      children = (childProps: {}) => null, // eslint-disable-line no-unused-vars
```
nadiia: Why don't we make `children` a required prop instead of providing a default value here.

src/slider/styled-components.js
```diff
@@ -0,0 +1,156 @@
...
+  const {$theme} = props;
+  const {colors, borders, sizing} = $theme;
+  return {
+    position: 'relative',
+    marginTop: sizing.scale900,
+    marginBottom: sizing.scale900,
+    marginRight: sizing.scale400,
+    marginLeft: sizing.scale400,
+    borderRadius: $theme.borders.useRoundedCorners ? borders.radius100 : '0px',
+    background: colors.mono400,
```
nadiia: `backgroundColor`

src/slider/styled-components.js
```diff
@@ -0,0 +1,156 @@
...
+  const emptyColor = colors.mono400;
+  let barColor;
+  if ($isRange) {
+    barColor = $index % 2 === 0 ? emptyColor : fillColor;
+  } else {
+    barColor = $index % 2 === 0 ? fillColor : emptyColor;
+  }
+  const value =
+    $isRange && $index ? $value[$index] - $value[$index - 1] : $value[$index];
+  const $successValue = $max - $min;
```
nadiia: What is `successValue`?

src/slider/styled-components.js
```diff
@@ -0,0 +1,156 @@
...
+    $max,
+    $index,
+    $currentThumb,
+    $disabled,
+    $isRange,
+  } = props;
+  const $isActive = $index === $currentThumb;
+  const offset = Math.round(($value[$index] / $max) * 100);
+  const {animation, colors, sizing} = $theme;
+  const backgroundColor = $disabled ? colors.mono500 : 'white';
```
nadiia: `colors.white` instead of a `'white'` string

src/slider/styled-components.js
```diff
@@ -0,0 +1,156 @@
...
+    transitionProperty: 'background-color',
+    ':after': {
+      position: 'absolute',
+      content: `url('data:image/svg+xml;utf8,${toggleSVG}')`,
+    },
+    position: 'relative',
+    top: '-14px',
+    cursor: 'pointer',
+    zIndex: '1',
+    left: offset + '%',
```
nadiia: string template instead of a string concatenation

src/slider/constants.js
```diff
@@ -0,0 +1,7 @@
+// @flow
+
+export const STATE_CHANGE_TYPE = {
+  value: 'value',
```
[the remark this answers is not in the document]
blackswanny: ok

src/slider/examples-list.js
```diff
@@ -0,0 +1,14 @@
...
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+/* eslint-env node */
+/* eslint-disable flowtype/require-valid-file-annotation */
+module.exports = {
+  AS_SIMPLE_SLIDER: 'Simple slider example',
+  AS_SIMPLE_RANGE_SLIDER: 'Simple range slider',
+  AS_SIMPLE_RANGE_SLIDER_WITH_STEP: 'Simple range slider with step=20',
+  AS_RANGE_SLIDER_WITH_MANY_TICKS: 'Range slider with many ticks',
```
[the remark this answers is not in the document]
blackswanny: added

src/slider/index.js
```diff
@@ -0,0 +1,18 @@
...
+*/
+// @flow
+export {default as Slider} from './slider';
+export {default as StatefulSlider} from './stateful-slider';
+export {default as StatefulSliderContainer} from './stateful-slider-container';
+// Styled elements
+export {
+  Root as StyledRoot,
+  Axis as StyledAxis,
+  Tick as StyledTick,
```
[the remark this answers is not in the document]
blackswanny: ok

src/slider/index.js
```diff
@@ -0,0 +1,18 @@
+/*
+Copyright (c) 2018 Uber Technologies, Inc.
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+// @flow
+export {default as Slider} from './slider';
+export {default as StatefulSlider} from './stateful-slider';
+export {default as StatefulSliderContainer} from './stateful-slider-container';
```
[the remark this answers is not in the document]
blackswanny: ok

src/slider/index.js
```diff
@@ -0,0 +1,18 @@
...
+// Styled elements
+export {
+  Root as StyledRoot,
+  Axis as StyledAxis,
+  Tick as StyledTick,
+  Thumb as StyledThumb,
+} from './styled-components';
+export {} from './constants';
```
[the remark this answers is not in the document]
blackswanny: ok

src/slider/stateful-slider-container.js
blackswanny: fixed

src/slider/styled-components.js
blackswanny: ok

src/slider/styled-components.js
blackswanny: renamed to `totalRangeValue`, previous was confused from other code

src/slider/styled-components.js
blackswanny: ok

src/slider/styled-components.js
blackswanny: ok

src/slider/slider.js
```diff
@@ -0,0 +1,233 @@
...
+  };
+
+  componentDidMount() {
+    /* eslint-disable-next-line react/no-find-dom-node */
+    this.domNode = findDOMNode(this);
+    this.addDocumentMouseEvents();
+  }
+
+  componentWillUnmount() {
+    this.removeDocumentEvents();
```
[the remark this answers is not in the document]
blackswanny: renamed removeDocumentMouseEvents

src/slider/slider.js
```diff
@@ -0,0 +1,233 @@
...
+  }
+
+  onMouseDown = (event: MouseEvent, thumbIndex: number) => {
+    this.setState(
+      {
+        isThumbMoving: true,
+        currentThumb: thumbIndex,
+      },
+      () =>
+        this.props.onMouseDown({
```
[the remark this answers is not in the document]
blackswanny: ok

src/slider/slider.js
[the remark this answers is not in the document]
blackswanny: if we need to pass `isThumbMoving` value, than we need to wait for it to be switched after `setState`. Otherwise if call immediately, the value will be inconsistent.

src/slider/slider.js
```diff
@@ -0,0 +1,233 @@
...
+    const {isThumbMoving} = this.state;
+    if (isThumbMoving) {
+      this.onMove(event.movementX, event);
+    }
+    this.props.onMouseMove({event, isThumbMoving});
+  };
+
+  addDocumentMouseEvents() {
+    if (__BROWSER__) {
+      this.onMouseMoveListener = document.addEventListener(
```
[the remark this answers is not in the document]
blackswanny: removed

src/slider/slider.js
```diff
@@ -0,0 +1,233 @@
...
+        'mousemove',
+        this.onMouseMove,
+      );
+      this.onMouseUpListener = document.addEventListener(
```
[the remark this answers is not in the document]
blackswanny: removed

src/slider/slider.js
```diff
@@ -0,0 +1,233 @@
...
+        'mouseup',
+        this.onMouseUp,
+      );
+    }
+  }
+
+  removeDocumentEvents() {
+    this.onMouseMoveListener && this.onMouseMoveListener.remove();
+    this.onMouseUpListener && this.onMouseUpListener.remove();
```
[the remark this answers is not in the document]
blackswanny: removed

src/slider/slider.js
```diff
@@ -0,0 +1,233 @@
...
+  }
+
+  onMove(movementX: number, event: SyntheticEvent<HTMLElement> | MouseEvent) {
+    const {currentThumb, currentMove} = this.state;
+    let {onChange, step} = this.props;
+    const {$value, $max, $min, $isRange} = this.getSharedProps();
+    const newMove = currentMove + movementX;
+    const moveDirection = newMove / Math.abs(newMove);
+    const value = $value.slice();
```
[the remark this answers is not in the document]
blackswanny: because we change this value later here to send to statefull comps
Changing props inside of comp is forbidden

src/slider/slider.js
```diff
@@ -0,0 +1,233 @@
...
+  removeDocumentEvents() {
+    this.onMouseMoveListener && this.onMouseMoveListener.remove();
+    this.onMouseUpListener && this.onMouseUpListener.remove();
+  }
+
+  onMove(movementX: number, event: SyntheticEvent<HTMLElement> | MouseEvent) {
+    const {currentThumb, currentMove} = this.state;
+    let {onChange, step} = this.props;
+    const {$value, $max, $min, $isRange} = this.getSharedProps();
+    const newMove = currentMove + movementX;
```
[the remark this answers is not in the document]
blackswanny: added

src/slider/slider.js
blackswanny: so it is called regardless thumb is dropped or not. For when value is changed there is main `onChange` handler

src/slider/slider.js
blackswanny: boolean. Actually `true` in this case

src/slider/slider.js
blackswanny: it's still mouseUp event happened. If we don't want to call it all the time, we need to remove it from handlers or rename it

src/slider/slider.js
blackswanny: you asked to add `onMouseDown` handler. For consistency I added other handlers of the same nature

src/slider/stateful-slider-container.js
blackswanny: I checked 
image

src/slider/stateful-slider-container.js
blackswanny: yes. It has to be `value` here. fixed it

nadiia: The fix seems to be lost as the line remained unchanged, so we still don't set the initial value here.

src/slider/stateful-slider-container.js
nadiia: Well, I would `console.log` the arguments the `onChange` handler gets instead of creating a custom case like in the screenshot you posted above.
screen shot 2018-10-25 at 4 09 49 pm

nadiia:
Left a few minor comments to address. But the biggest concern here is the document event listeners we add for the whole lifecycle of a slider component. If there are a few sliders added on a page we'll end up with a lot of those listeners. The events I'm talking about are `onMouseMove` and `onMouseUp`.
We need to subscribe to those in the `onMouseDown` handler and unsubscribe from both the above mentioned in the 'onMouseUp' handler.

blackswanny:
> Renaming sounds like a great idea

renamed

blackswanny:
> The fix seems to be lost as the line remained unchanged, so we still don't set the initial value here.

The fix was made for other parts of code. We usually have something like below for this case. However, `value` has no default value and is a required field, so it always comes in `initialState` 
image

blackswanny:
> Well, I would console.log the arguments the onChange handler gets instead of creating a custom case like in the screenshot you posted above.

I tested with console.log in my default handler, and still see no such result
I think in example above it already calls `onChangeTst` with wrong argument a,d it should be
`onChangeTst({ event: {}, value: {} })`

image

image

blackswanny:
> It looks like those two methods (scale, clampIt) can be extracted into helpers functions outside of this class. Otherwise, could you make sure to cover this case in tests? Specifically if keeping it here instead of extracting into helper functions.

yes, will be addressed in PR for unit-tests

blackswanny:
> Left a few minor comments to address. But the biggest concern here is the document event listeners we add for the whole lifecycle of a slider component. If there are a few sliders added on a page we'll end up with a lot of those listeners. The events I'm talking about are onMouseMove and onMouseUp.
> We need to subscribe to those in the onMouseDown handler and unsubscribe from both the above mentioned in the 'onMouseUp' handler.

thanks. moved, it works as well.

src/slider/slider.js
```diff
@@ -0,0 +1,209 @@
...
+    if (isThumbMoving) {
+      this.setState({
+        isThumbMoving: false,
+        currentThumb: -1,
+        currentMove: 0,
+      });
+    }
+    this.props.onThumbUp({
+      event,
+      isThumbMoving,
```
nadiia: If we want to pass this value in here its value should be set to `false` explicitly since the state's value won't be changed yet: `isThumbMoving: false`

src/slider/stateful-slider-container.js
nadiia: I added a comment based on the type on line 32, but your console.log clearifies that the args look  fine. It looks like just the type of the config on line 32 is incorrect `onChange = (config: {event: *, params: ParamsT}) => {`. Should be `onChange = (config: {event: *, value: some_value_type}) => {`. Unless I'm missing something.

src/slider/slider.js
blackswanny: removed from API and cleaned it up

blackswanny:
> 
> 
> Thanks for addressing all the comments!
> One more question around the handlers
> 
>     onThumbDown: () => {},
>     onMouseMove: () => {},
>     onThumbUp: () => {},
> 
> is there a use case for these to be exposed in the top-level API? Wonder if we should omit exposing those three.
> 

removed from API

blackswanny:
> I added a comment based on the type on line 32, but your console.log clearifies that the args look fine. It looks like just the type of the config on line 32 is incorrect onChange = (config: {event: *, params: ParamsT}) => {. Should be onChange = (config: {event: *, value: some_value_type}) => {. Unless I'm missing something.

fixed

PR #337, opened 2018-10-03
chore(select): update examples

schnerd:
Simplifies `Select` stories.

**Current**
image

**This PR**
image

Many of the original stories had similar behavior that could be combined, for example:
image

This PR also introduces what I think is better sample data for examples - a list of 100 or so colors. These make easier to demonstrate search functionality compared to generated labels:


src/select/__tests__/select.test.js
```diff
@@ -302,28 +308,29 @@ describe('Stateless select', function() {
...
+      wrapper.update();
+      expect(wrapper.find(StyledOption)).toHaveLength(1);
 
       input.simulate('change', {target: {value: 'aaaa'}});
-      onTextInputChangePromise.then(() =>
-        expect(wrapper.find(StyledOption)).toHaveLength(0),
-      );
+      await sleep(20);
+      wrapper.update();
+      expect(wrapper.find(StyledOption)).toHaveLength(0);
```
schnerd: These tests were essentially broken on master. They used promises without using jest's helpers for async tests, so the test would "succeed" but then later have failures:

image

src/select/examples.js
```diff
@@ -11,315 +11,164 @@ LICENSE file in the root directory of this source tree.
...
-                // eslint-disable-next-line no-console
-                console.log('Unselected id:' + option.id);
-                break;
-            }
-          }}
-        />
-      </React.Fragment>
-    );
-  }
-}
```
schnerd: Thee `Parent*` were helpful when we were auto-generating options but are less helpful after the refactor. The new examples just define their Selects directly (reusing change handlers where applicable), and the source code for each story still ends up being quite short (~10 or so lines).

These ParenSelect's also led to sort of misleading examples in the Storybook "Show Info" plugin
image

src/select/select.js
```diff
@@ -43,12 +43,6 @@ class Select extends React.Component<PropsT, StatelessStateT> {
     multiple: false,
     tabIndex: 0,
     textValue: '',
-    filterOption: (option: OptionT, query: string) => {
-      return (
-        typeof option.label === 'string' &&
-        option.label.toLowerCase().indexOf(query.toLowerCase()) >= 0
-      );
-    },
```
schnerd: Moving this out of defaultProps and into a method allows us to utilize `this.getOptionLabel` for our default filtering instead of only filtering on `option.label`.

src/select/__tests__/select.test.js
gergelyke: did you fixed this in your PR, or should we address it in a new one?

gergelyke: oh I see, never mind

src/select/__tests__/select.test.js
```diff
@@ -302,28 +308,29 @@ describe('Stateless select', function() {
         .find(StyledInput)
         .first()
         .find('input');
+
       input.simulate('change', {target: {value: 'a'}});
-      onTextInputChangePromise.then(() =>
-        expect(wrapper.find(StyledOption)).toHaveLength(3),
-      );
+      await sleep(20);
```
blackswanny: currently tests were asynced, but only by 0 to 1 tick of processor and only because JS released thread while  Promise works in the [next tick](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises#Timing). But now they are slower by 20ms, multiplied by their quantity.
Tests should be as fast as possible to pass [quickly for many reasons](https://github.com/ghsukumar/SFDC_Best_Practices/wiki/F.I.R.S.T-Principles-of-Unit-Testing)

src/select/select.js
```diff
@@ -394,6 +388,18 @@ class Select extends React.Component<PropsT, StatelessStateT> {
...
 
+  filterOption(option: OptionT, query: string) {
+    if (this.props.filterOption) {
+      return this.props.filterOption(option, query);
+    }
+
+    const label = this.getOptionLabel(option);
+    return (
+      typeof label === 'string' &&
+      label.toLowerCase().indexOf(query.toLowerCase()) >= 0
```
gergelyke: can we simply use `String.includes`? https://caniuse.com/#search=includes

src/select/__tests__/select.test.js
blackswanny: thanks for fixing it

src/select/__tests__/select.test.js
schnerd: Yeah I agree–it didn't feel great adding these sleep statements here but I was struggling to figure out a way to do this that wouldn't be flaky. `await onTextInputChangePromise` didn't work (same as `.then` logic that existed before)

Will try to remove this sleep or lower the number of milliseconds.

src/select/select.js
schnerd: I suppose it depends if we're trying to support IE11. Did we document our browser support anywhere? I forget.

blackswanny: we have babel-polyfill
it should fallback here
https://github.com/zloirock/core-js#ecmascript-string-and-regexp

src/select/__tests__/select.test.js
schnerd: Got it down to `sleep(1)`

src/select/select.js
schnerd: awesome, changed.

Issue #252, opened 2018-09-14
[Theming] create a theme factory?


jckr:
* [x] I have searched the [issues](https://github.com/uber-web/baseui/issues) of this repository and believe that this is not a duplicate.

## Expected Behavior
TL;DR - I want to be able to create variations of existing themes. Right now in the LightTheme, a lot of the options come from constants in the file which are not exposed. What if in addition to existing themes we added a function that could take parameters (such as fonts, colors etc.) and generated a new theme? 
For instance: 
```
import {LightTheme, createTheme, ThemeProvider} from 'baseui';
const myTheme = createTheme({template: LightTheme, primaryFontFamily: 'Times New Roman'});
const myComponent = <ThemeProvider theme={myTheme}>
... 
</ThemeProvider>
```
## Current Behavior
If I want to create a full theme I can either create it from scratch or at the very least override all the items I need to change even if they contain similar information. 

## Context
I'd like to set my typography choices once and for all and probably not use Helvetica.

schnerd:
There are some related ideas in #48

jckr:
also I just worded the issue. The idea came from @schnerd as an answer to how this could be done. 
So I'm new to this. What's the approach for creating proofs of concept?

abmai:
Similar ideas indeed. Do we want to first proceed with a short RFC on how the theme creation/extension could be improved? @jckr I don't think we have a particular RFC format for non-component RFC so it's pretty much free-form at this point.

schnerd:
I think this exists now, see docs here:
https://github.com/uber-web/baseui/tree/master/src/themes

Feel free to re-open if there's still functionality that's missing.

PR #338, opened 2018-10-03
feat(fonts): font can be overwritten; adding move font

gergelyke:


src/tag/constants.js
```diff
@@ -6,13 +6,13 @@ LICENSE file in the root directory of this source tree.
...
 
 export const STYLE = {
-  primary: DEFAULT_THEME.colors.primary400,
-  warning: DEFAULT_THEME.colors.warning400,
-  positive: DEFAULT_THEME.colors.positive400,
-  negative: DEFAULT_THEME.colors.negative400,
+  primary: LightTheme.colors.primary400,
+  warning: LightTheme.colors.warning400,
+  positive: LightTheme.colors.positive400,
+  negative: LightTheme.colors.negative400,
```
nadiia: Not related to this PR, but it needs to be changed to a string like `primary400 `, `warning400 `, etc, and then get the values from a `$theme` in a styled component. This way the LightTheme wouldn't be the default but whatever theme is passed to the Provider.

schnerd:
Does this resolve #252?

gergelyke:
yeah, I think so, if we extend it, so all values can be overwritten - I can do that, should we?

schnerd:
I think it's fine to just resolve that issue now since fonts were the primary concern, and then if customers ask to override other things in the future we can re-evaluate how that fits into the current API.

gergelyke:
makes sense

PR #341, opened 2018-10-03
docs(move): create a move version of the docs site

gergelyke:


.storybook-move/addons.js
```diff
@@ -0,0 +1,7 @@
+import 'storybook-readme/register';
```
blackswanny: I think we can put the same files in separate folder in import from those folders in here, so we don't repeat these configs twice

PR #343, opened 2018-10-04
feat(toast): initial implementation of the toast component (part1)

nadiia:


src/toast/styled-components.js
```diff
@@ -0,0 +1,57 @@
...
+    backgroundColor:
+      getBackgroundColor($kind, $theme) || $theme.colors.primary500,
+    borderRadius: $theme.borders.useRoundedCorners
+      ? $theme.borders.radius200
+      : '0px',
+    boxShadow: $theme.lighting.shadow600,
+    opacity: $isHidden ? 0 : 1,
+    transitionProperty: 'opacity, height',
+    transitionDuration: $theme.animation.timing100,
+    transitionTimingFunction: $theme.animation.easeInOutCurve,
```
schnerd: Based on the storybooks it seems like the height isn't actually animating? I also don't think you can use css animations to animate from `height: 0` to `height: auto`–you instead have to do some nonsense where you measure the height of the content within an element and then explicitly animate to that height.

src/toast/styled-components.js
```diff
@@ -0,0 +1,57 @@
...
+import type {SharedStylePropsT, KindTypeT} from './types';
+
+function getBackgroundColor(kind: KindTypeT, theme) {
+  return {
+    [KIND.info]: theme.colors.primary500,
+    [KIND.positive]: theme.colors.positive500,
+    [KIND.warning]: theme.colors.warning500,
+    [KIND.negative]: theme.colors.negative500,
+  }[kind];
+}
```
schnerd: For some reason these feel very dark? The orange almost looks brown:

image

(they match the "active" button color, not the default button color)

nadiia: Yeah, in the design they use darker (`500`) colors. But I'll double check with Jeff.

src/toast/styled-components.js
nadiia: Good catch! And on the fade-in animation. Will fix.

src/toast/toast.js
```diff
@@ -0,0 +1,186 @@
...
+    this.setState({isAnimating: true});
+  };
+
+  animateOut = (callback: () => void = () => {}) => {
+    this.setState({isAnimating: true});
+    // Remove the toast from the DOM after animation finishes
+    this.animateOutCompleteTimer = setTimeout(() => {
+      this.setState({isAnimating: false});
+      callback();
+    }, 600);
```
gergelyke: should we move this constant to the Constants file?

PR #344, opened 2018-10-04
feat(toast): add toast storybook examples (part2)

nadiia:


schnerd:
image

it's.... it's beautiful 😭

nadiia:
> image
> 
> it's.... it's beautiful 😭

IG's gradient 😆

src/toast/examples.js
```diff
@@ -0,0 +1,207 @@
...
+LICENSE file in the root directory of this source tree.
+*/
+// @flow
+import * as React from 'react';
+import {styled} from '../styles';
+import {Toast, KIND} from './index';
+import {Button, KIND as ButtonKind, SIZE} from '../button';
+
+export const suite = 'Toast Test Suite';
+export const tests = {
```
gergelyke: can we move these out to an `example-list.js` file, so the Toast component is ready for e2e tests? We did the same for modal / cards.

nadiia: Sure, I'll add a follow up PR for it

src/toast/examples.js
```diff
@@ -0,0 +1,207 @@
...
+      </div>
+    );
+  }
+}
+
+export default {
+  [tests.SIMPLE_EXAMPLE]: function Story1() {
+    return (
+      <Centered>
+        <Toast>Default info notification</Toast>
```
blackswanny: every story contain many examples. Afaik, we still use stories for e2e tests as basic. If we start writing e2e tests , it may be not easy to test story as black-boxed Selenium will not be able to pick a single example and work only with it without providing extra props and other identifiers

nadiia: I think we need to optimize examples for better stories and documentation in the first place and figure out tests as a second. For instance, maybe create additional examples for e2e tests only.

blackswanny: yes, this is reasonable. However, the main idea of shared examples file is to re-use it for both stories and e2e tests. Having different examples may duplicate our code. There is not single approach on how to optimize examples for customer better. Some provide examples for every single property. Others gather some examples under "one roof"
https://mineral-ui.com/components/text-input
https://material-ui.com/demos/text-fields/
https://ant.design/components/input/
https://react.semantic-ui.com/elements/input/
https://grommet.io/docs/search

PR #345, opened 2018-10-04
feat(toast): add toast tests (part3)

nadiia:


src/toast/__tests__/toast.test.js
```diff
@@ -0,0 +1,200 @@
...
+// @flow
+
+import React from 'react';
+import {mount} from 'enzyme';
+import {Toast, StyledBody, StyledCloseIcon, KIND} from '../index';
+
+jest.useFakeTimers();
+
+describe('Toast', () => {
+  test('basic functionality', () => {
```
blackswanny: tests names seems to be very generic and do not describe their meaning and "under-test" functionality. It's harder to maintain and fix tests without more descriptive naming.

src/toast/__tests__/toast.test.js
```diff
@@ -0,0 +1,200 @@
...
+    expect(wrapper.find(StyledCloseIcon).first()).toExist();
+  });
+
+  test('Body component props', () => {
+    let renderedRoot;
+    const wrapper = mount(<Toast>Notification</Toast>);
+
+    renderedRoot = wrapper.find(StyledBody).first();
+    expect(renderedRoot).toExist();
+    expect(renderedRoot.props().$kind).toBe(KIND.info);
```
blackswanny: you can use https://jestjs.io/docs/en/expect#tomatchobjectobject to check a subset of object

blackswanny:
do we have tests for `async` part of this component yet? For instance, we have `setTimeout` based methods, which may be tested with `async, await`\`done` approach.

src/toast/__tests__/toast.test.js
nadiia: @blackswanny do you have any suggestions here?

src/toast/__tests__/toast.test.js
nadiia: Here I want specifically check the `$kind` prop. I think breaking down the check works well and easy to read.

src/toast/__tests__/toast.test.js
blackswanny: this seems to be `should render Toast with default settings` test.
I try to avoid usage of `toMatchSnapshot`, but it works well for initial render check and styled components check.
```js

  test.each([
    [TYPE.search, false],
    [TYPE.select, false],
    [TYPE.search, true],
    [TYPE.select, true],
  ])(
    'should render component in %s mode and %s for multiple choice',
    (type, multiple) => {
      allProps.type = type;
      allProps.multiple = multiple;
      wrapper = mount(<Select {...allProps} />);
      expect(wrapper).toMatchSnapshot(
        'Component has correct render in ' +
          type +
          ' mode and mutiple choice equal ' +
          multiple,
      );
    },
  );
```

src/toast/__tests__/toast.test.js
```diff
@@ -0,0 +1,200 @@
...
+    expect(renderedRoot.props().$closeable).toBe(true);
+    expect(renderedRoot.props().$isHidden).toBe(false);
+    expect(renderedRoot.props().$isAnimating).toBe(true);
+
+    // pass new kind value set to KIND.positive
+    // and closeable set to false
+    wrapper.setProps({kind: KIND.positive, closeable: false});
```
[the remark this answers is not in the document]
nadiia: What do you mean by the `call of setProps may activate a whole different logic of component`? There are specific changes and expectations for every set of props.

nadiia:
@blackswanny not sure what ou mean here. I don't think we need to test a `setTimeout`.

src/toast/__tests__/toast.test.js
blackswanny: yes. However, if we would like to emphasize on particular logic, I would suggest to do it like this:
```js
  describe('Body component', () => {
     let renderedRoot;
     beforeEach(() => {
         const wrapper = mount(<Toast>Notification</Toast>);
        renderedRoot = wrapper.find(StyledBody).first();
     });
    test('should render Body of Toast immediately', () => {
       expect(renderedRoot).toExist();
    });

    test('should set styles of Toast Body to be of information by default', () => {
       expect(renderedRoot.props().$kind).toBe(KIND.info);
    })
```

src/toast/__tests__/toast.test.js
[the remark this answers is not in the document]
blackswanny: correct. When we set new props, we re-render React component, which may change everything, even functionality of it.
https://airbnb.io/enzyme/docs/api/ReactWrapper/setProps.html
so for example, if we call
`wrapper.setProps({kind: KIND.positive, closeable: false}); `
we expect component to
- look like a `positive` Toast styled
- to not be able to closed via X button, so the whole `DeleteAltIcon` with `onClick` handler no longer exists.
If I test only props of particular subcomponent I may not be sure that component works as expected.

blackswanny:
> @blackswanny not sure what ou mean here. I don't think we need to test a setTimeout.
sure, we do not test `setTimeout`, but we would like to test methods, which use it. For instance, animateOut\In to make sure that `timeout` feature works well and animation happens exactly for this period of time

src/toast/__tests__/toast.test.js
nadiia: I think it's a good approach to write tests in the way you mentioned. Currently, we don't follow a consistent pattern in how we structure our unit tests and tests' code. How about opening a new discussion on the unit test writing best practices in order to get feedback from others and try to come to an agreement on the consistency of our unit tests across the library? @blackswanny Would you mind starting this discussion?

nadiia:
> > @blackswanny not sure what ou mean here. I don't think we need to test a setTimeout.
> 
> sure, we do not test `setTimeout`, but we would like to test methods, which use it. For instance, animateOut\In to make sure that `timeout` feature works well and animation happens exactly for this period of time

The tests do check that the state is set correctly. We mock the timeouts and assume `setTimeout` works as expected. Not sure what else you mean here.

blackswanny:
> The tests do check that the state is set correctly. We mock the timeouts and assume setTimeout works as expected. Not sure what else you mean here.


if I change method
```js
animateOut = (callback: () => void = () => {}) => {
    this.setState({isAnimating: true});
    // Remove the toast from the DOM after animation finishes
    this.animateOutCompleteTimer = setTimeout(() => {
      this.setState({isAnimating: false});
      callback();
    }, 600);
}
```
to
```js
animateOut = (callback: () => void = () => {}) => {
    this.setState({isAnimating: true});
    // Remove the toast from the DOM after animation finishes
     this.setState({isAnimating: false});
     callback();
}
```
the tests checking for `isAnimating===false` will still be false positive(green) despite logic started to work improperly. Jest has examples of how to make sure timeout did work as expected

https://jestjs.io/docs/en/timer-mocks.html

src/toast/__tests__/toast.test.js
blackswanny: Sure, thanks. I can provide some slides and ideas on the next Base UI meeting

nadiia:
> if I change method
...
> ```js
> animateOut = (callback: () => void = () => {}) => {
>     this.setState({isAnimating: true});
>     // Remove the toast from the DOM after animation finishes
>      this.setState({isAnimating: false});
>      callback();
> }
> ```
> 
> the tests checking for `isAnimating===false` will still be false positive(green) despite logic started to work improperly. Jest has examples of how to make sure timeout did work as expected


Good catch on this one! Added a check for the delay time [here](https://github.com/uber-web/baseui/pull/358/files#diff-666ca91dbe7e74f7b61e7179ab6a315cR109)

PR #346, opened 2018-10-04
feat(theme): add theme factory

gergelyke:


src/themes/light-theme.js
```diff
@@ -5,245 +5,244 @@ This source code is licensed under the MIT license found in the
...
-    primaryFontFamily,
-  }: {
-    primaryFontFamily: string,
-  } = {
-    primaryFontFamily: '"Helvetica Neue", arial, sans-serif',
-  },
-) {
+// $FlowFixMe
+export default function(overrides) {
+  const theme = Object.assign({}, defaultTheme, overrides);
```
nadiia: This should probably be a deep merge

gergelyke: Why is that? we only a single level of properties

src/themes/light-theme.js
```diff
@@ -5,245 +5,244 @@ This source code is licensed under the MIT license found in the
...
-const mono300 = '#F0F0F0';
-const mono400 = '#E5E5E5';
-const mono500 = '#CCCCCC';
-const mono600 = '#B3B3B3';
-const mono700 = '#999999';
-const mono800 = '#666666';
-const mono900 = '#333333';
-const mono1000 = '#000000';
+  primaryFontFamily: '"Helvetica Neue", arial, sans-serif',
+};
```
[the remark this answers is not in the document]
gergelyke: Sounds great! Let me update the PR

src/themes/light-theme.js
abmai: Typography is two level, i.e. I think this might end up overriding all of `typography`:
```js
Object.assign({}, defaultTheme, {
  typography: {
   myNewFontStyles: {...}
  }
}
```

src/themes/index.js
```diff
@@ -6,11 +6,15 @@ LICENSE file in the root directory of this source tree.
...
 // @flow
 
-import {default as LightThemeCreator} from './light-theme';
+import createTheme from './creator';
+import {primitives as lightThemePrimitives} from './light-theme';
 
-export const LightTheme = LightThemeCreator();
+export const LightTheme = createTheme({
+  ...lightThemePrimitives,
+});
```
schnerd: Probably don't need the spread here since we're just using `lightThemePrimitives` directly and createTheme is pure.

src/themes/creator.js
```diff
@@ -0,0 +1,286 @@
...
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+// @flow
+
+import type {ThemeT} from '../styles/types';
+import deepMerge from '../utils/deep-merge';
+
+// $FlowFixMe
```
schnerd: What's this for? Hoping we can fix this since Theme is a pretty important thing to flowtype properly.

At the very least maybe we can add a comment to the `$FlowFixMe`

PR #347, opened 2018-10-04
feat(buildkite): list vrt differences

gergelyke:


gergelyke:
closing in favor of moving these images to be printed inline

PR #349, opened 2018-10-04
fix(tag): fixes #339

### LINKED ISSUE #339: Tag component should use $theme
Instead of directly depending on values from the [light theme](https://github.com/uber-web/baseui/blob/master/src/tag/constants.js#L12), tag should depend on the [$theme prop](https://github.com/uber-web/baseui/pull/338#discussion_r222471860).

MarkReeder:
#### Fixed Issues?

Fixes #339 

#### Major: Breaking Change?

This is a breaking change as it reworks publicly available constants, which was necessary to remove the references to `LightTheme` from the Tags constants. Aside from the constants changing, the component interface is unchanged.

#### Tests Added + Pass?

Yes

#### License

MIT


gergelyke:
Thanks for taking this on, @MarkReeder !

Do you think users actually imported that file? If so, can we provide a migration guide?

src/tag/tag.js
```diff
@@ -47,38 +48,47 @@ class Tag extends React.Component<PropsT, {}> {
...
-              fillRule="evenodd"
-              clipRule="evenodd"
-              d="M0.861278 0.862254C1.12163 0.601905 1.54374 0.601905 1.80409 0.862254L3.99935 3.05752L6.19461 0.862254C6.45496 0.601905 6.87707 0.601905 7.13742 0.862254C7.39777 1.1226 7.39777 1.54471 7.13742 1.80506L4.94216 4.00033L7.13742 6.19559C7.39777 6.45594 7.39777 6.87805 7.13742 7.1384C6.87707 7.39875 6.45496 7.39875 6.19461 7.1384L3.99935 4.94313L1.80409 7.1384C1.54374 7.39875 1.12163 7.39875 0.861278 7.1384C0.600928 6.87805 0.600928 6.45594 0.861278 6.19559L3.05654 4.00033L0.861278 1.80506C0.600928 1.54471 0.600928 1.1226 0.861278 0.862254Z"
-              fill={color}
-            />
-          </ActionIcon>
-        </Action>
-      </Root>
+      <ThemeContext.Consumer>
+        {({colors}) => {
```
nadiia: I was thinking to do something like we have in an [`icon` component](https://github.com/uber-web/baseui/blob/master/src/icon/styled-components.js#L13). So there wouldn't be a need to wrap it in a ThemeContext consumer.

schnerd: Ideally we should replace `ActionIcon` to be `baseui/icons/Delete`. By default these icons use CSS `currentColor`, so in theory if we set the `color` in action using an approach like Nadiia just linked, then we don't need to set the path fill.

nadiia: We can set `color` on icon's container and use `currentColor` as a fill value. +1 to David's comment on replacing the icon with the one from `/icon` but it would be outside of the scope of this change.

gergelyke:
@nadiia @schnerd do we want to change this PR, or should we merge as is?

src/tag/examples.js
```diff
@@ -26,26 +27,40 @@ const colorIcons = {
...
+              if (typeof tag === 'string') {
+                // eslint-disable-next-line no-console
+                console.log('Tag is clicked:' + tag);
+              }
+            }}
+          >
+            Default Color
+          </Tag>
+          {tagStyleKeys.map(colorKey => (
+            <ThemeContext.Consumer key={colorKey}>
```
nadiia: We can remove the `ThemeContext.Consumer` usage in the examples now, correct?

src/tag/examples.js
```diff
@@ -106,20 +121,24 @@ export default {
...
-                if (typeof tag === 'string') {
-                  // eslint-disable-next-line no-console
-                  console.log('Tag is clicked:' + tag);
-                }
-              }}
-            >
-              Color {color}
-            </Tag>
+          {tagStyleKeys.map(colorKey => (
+            <ThemeContext.Consumer key={colorKey}>
```
nadiia: same here

src/tag/styled-components.js
```diff
@@ -5,10 +5,24 @@ This source code is licensed under the MIT license found in the
...
+import {COLOR_STYLE_KEYS} from './constants';
+import {styled, hexToRgb, type ThemeT} from '../styles/index';
+
+export function getColor(
+  $theme: ThemeT,
+  $color: string,
+  defaultColor: string = '#000',
+) {
+  let themeColor;
+  if ($theme && $theme.colors && $theme.colors[COLOR_STYLE_KEYS.primary]) {
```
nadiia: I think it should be something like 
```
if ($theme && $theme.colors && $theme.colors[$color]) {
  themeColor = $theme.colors[$color];
} 
return $theme.colors[$color] || $color || defaultColor;
```

src/tag/tag.js
```diff
@@ -13,13 +13,12 @@ import {
...
   ActionIcon as StyledActionIcon,
 } from './styled-components';
-import {STYLE} from './constants';
 
 class Tag extends React.Component<PropsT, {}> {
   static defaultProps = {
     overrides: {},
     onActionClick: () => {},
-    color: STYLE.primary,
+    color: undefined,
```
nadiia: we should still set the default value to be the `COLOR_STYLE_KEYS.primary`

src/tag/examples.js
MarkReeder: No- since we only have the color keys and not the values (the component needs the color hex values, not the color keys that are in the `tagStyleKeys`), we need a reference to the theme for the examples if we want to pull the values from the theme. Alternatively, I could just import the light theme directly for the demo, but that doesn't seem like something we want to do.

src/tag/styled-components.js
MarkReeder: It feels weird to me to overload `color` with named color values from the theme and css color values. Is that a pattern we have anywhere else?

src/tag/tag.js
MarkReeder: If we overload color to take theme keys, then yes, but that seems potentially problematic. At a minimum, if we took that approach, we'd want to add checks to ensure that we never had namespace conflicts between the theme color values and css named colors.

src/tag/examples.js
nadiia: We use this pattern in the [icon](https://github.com/uber-web/baseui/blob/master/src/icon/styled-components.js#L13) component.

src/tag/styled-components.js
nadiia: We use this pattern in the [icon](https://github.com/uber-web/baseui/blob/master/src/icon/styled-components.js#L13) component.

src/tag/tag.js
nadiia: That's a good point.

nadiia:
My only concern here is that it requires extra code and imports in order to add a tag of a different type. We provide the `COLOR_STYLE` constant but it seems to have no much value since what customers still need to pass a `theme.colors[anyColorFromTheme]` (by wrapping in the theme consumer or importing the theme directly) or  pass a valid css color value. 
@schnerd what do you think?

MarkReeder:
@nadiia @schnerd What about a separate prop for the named color styles? `kind`; `tagType`; something else?

nadiia:
> @nadiia @schnerd What about a separate prop for the named color styles? `kind`; `tagType`; something else?

I like it, and we'll have a fixed set of types to choose from. Then it will follow the theme's overrides (in cases where a custom primary, etc., colors set) and for one off(s) they'd just need to override styles in a regural way through `overrides` API. +1 for `kind`.

src/tag/tag.js
abmai: Do we even need to set this default, then?

schnerd:
> > @nadiia @schnerd What about a separate prop for the named color styles? `kind`; `tagType`; something else?
> 
> I like it, and we'll have a fixed set of types to choose from. Then it will follow the theme's overrides (in cases where a custom primary, etc., colors set) and for one off(s) they'd just need to override styles in a regural way through `overrides` API. +1 for `kind`.

I like this.

src/tag/constants.js
```diff
@@ -5,14 +5,12 @@ This source code is licensed under the MIT license found in the
...
 // @flow
-/* eslint-disable import/prefer-default-export */
-import {LightTheme} from '../themes';
 
-export const STYLE = {
-  primary: LightTheme.colors.primary400,
-  warning: LightTheme.colors.warning400,
-  positive: LightTheme.colors.positive400,
-  negative: LightTheme.colors.negative400,
+export const COLOR_STYLE_KEYS = {
```
nadiia: Not sure if we still need this constant, but we should add/export the `KIND` constant as well.
```
export const KIND = {
  primary: 'primary',
  positive: 'positive',
  warning: 'warning',
  negative: 'negative',
};
```

src/tag/examples.js
```diff
@@ -59,12 +70,18 @@ export default {
             <Tag
               key={color}
               color={color}
```
nadiia: I thought we're getting rid of the `color` prop in favor of `kind`. But we can delete this prop in a follow-up update if everyone's on board. What do you think?

src/tag/examples.js
```diff
@@ -59,12 +70,18 @@ export default {
...
+              kind="custom"
```
nadiia: Is it possible to avoid an explicit `custom` kind?

src/tag/examples.js
MarkReeder: That was my plan, but getting rid of color makes basic overrides more complicated than necessary.

src/tag/examples.js
MarkReeder: I could remove the `custom` kind, but I felt it was confusing to leave `color` and allow `kind` to be set. I wanted a more explicit mechanism for enabling the color prop.

nadiia: I see, sounds good

PR #351, opened 2018-10-04
feat: add implementation of slider onChange and step feature

blackswanny:
add implementation of slider onChange and step feature


blackswanny:
I close this one and reuse the work from it after addressing feedback in #332

PR #353, opened 2018-10-05
fix(overrides): propose new interface for getOverrideObject

abmai:
Wanted to get feedback on this before I make further changes. Right now, using `getOverrideObject` is cumbersome because of the redundant `component` and `props` destructuring. I want to change the return value to an array, which really simplifies that.

I don't see any drawback in this approach, plus it's an internal helper we use. We get the convenience with less code. I modified `pagination.js` to show an example of how it's used. Thoughts?

### DIFF
--- a/src/helpers/__tests__/overrides.test.js
+++ b/src/helpers/__tests__/overrides.test.js
@@ -12,7 +12,7 @@ import {
   toObjectOverride,
   mergeOverrides,
   mergeOverride,
-  getOverrideObject,
+  getOverrides,
 } from '../overrides';
 
 function getMockComponent<T>(): React.ComponentType<T> {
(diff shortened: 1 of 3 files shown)


gergelyke:
This makes sense to me - what do you think @schnerd @blackswanny @nadiia ?

blackswanny:
it does save some few lines of code, but not much. However in case of Array it's easier to make a mistake and assign them vice versa than having explicit props. For `overrides` approach it also works vice versa. Previously we could write:
```js
overrides={{
    someOverride: getOverrideObject(....)
}}
```
and now it has to be
```js
const [Comp, compProps] = getOverrideObject(....)
overrides={{
    someOverride: {component: Comp, props: compProps}
}}
```
is it correct?

abmai:
> it does save some few lines of code, but not much. However in case of Array it's easier to make a mistake and assign them vice versa than having explicit props. For `overrides` approach it also works vice versa. Previously we could write:
> 
> ```js
> overrides={{
>     someOverride: getOverrideObject(....)
> }}
> ```
> 
> and now it has to be
> 
> ```js
> const [Comp, compProps] = getOverrideObject(....)
> overrides={{
>     someOverride: {component: Comp, props: compProps}
> }}
> ```
> 
> is it correct?

In the case that you mentioned, yes. But when do we need to call `getOverrideObject` that deep inside the `overrides`?

abmai:
With regards to the easier to make a mistake portion - I thought the same, actually. But I think this is exactly where `flow` comes in to help protect against. The tuple is strongly typed.

blackswanny:
> In the case that you mentioned, yes. But when do we need to call getOverrideObject that deep inside the overrides? It doesn't seem like an intended use-case.

We started to actively use it, when we started to re-use components in other component.
src/select/dropdown.js
src/select/select.js
src/tooltip/tooltip.js

abmai:
> > In the case that you mentioned, yes. But when do we need to call getOverrideObject that deep inside the overrides? It doesn't seem like an intended use-case.
> 
> We started to actively use it, when we started to re-use components in other component.
> src/select/dropdown.js
> src/select/select.js
> src/tooltip/tooltip.js

Interesting. I'll go through and address those use cases for now. However, I don't think we should be doing this at that granularity (`mergeOverrides` might be more appropriate for that).

blackswanny:
@abmai  ok. let's try new approach

gergelyke:
One question re logistics - @abmai do you think we should do it in one go for all the components, so the code base stays consistent? Or do you want to do this on a per component base?

PR #355, opened 2018-10-05
docs(themes): documenting create theme

gergelyke:


README.md
```diff
@@ -46,6 +46,8 @@ export default function Hello() {
 
 ## Docs
 
+* [Creating custom themes](src/themes/README.md)
+
 [Docs by Storybook](https://baseui.netlify.com/)
```
nadiia: Could you add the `*` here as well, so it's a part of the unordered list of docs.

src/themes/README.md
```diff
@@ -0,0 +1,36 @@
+# Creating custom themes
+
+Base UI ships by default with the `Light` theme. We understand that in some cases
+you want to change the look and feel of the components. To do so, follow the steps
+outlined here
+
+## Define your primitives
+
+Primitives are used define the colors and fonts for the components. An example for
```
nadiia: `used define` => `used to define`

src/themes/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+In practice, you can define your custom theme using this approach:
+
+```js
+import createTheme from 'baseui/themes/creator'
+
+import {primitives} from 'baseui/themes/light-theme';
+
+const yourTheme = createTheme({
+  ...primitives,
```
blackswanny: Currently we have only primitives, but if we may have other elements, they might not be covered with this example.
And seems like it should be
```js
const yourTheme = createTheme({
  ...{ 
         ...primitives,
         primaryFontFamily: '"Comic Sans MS", cursive, sans-serif'
      }
})
```

gergelyke: What do you mean by other elements? I guess if we break this API, we'll have to update the readme anyways - what am I missing?

blackswanny: I thought the whole theme object is
```js
{
   primitives,
   // other properties
}
```
according to this file src/themes/light-theme.js
If so, we override here only primitives features

gergelyke: that's not the case - checkout out the theme object too, that's the second param of the function

blackswanny: ok

PR #357, opened 2018-10-05
fix(e2e): add retry logic

gergelyke:
Sometimes the sauce lab connection from buildkite to suacelab seems flaky. It causes tests to fail sometimes - this will try any failed tests cases up to two more times.


blackswanny:
can you show some example logs of such failing? I just ould like to check that it is due to some sauslabs issue with infratructure, but not because our e2e tests written improperly and have undetermined not-repeatable results

gergelyke:
@blackswanny 
* https://buildkite.com/uberopensource/baseui/builds/709#19714e8d-c612-459a-9b9b-044980ad8083/105-775
* https://github.com/uber-web/baseui/blob/master/src/modal/e2e.js#L40

PR #358, opened 2018-10-05
chore(tests): some minor tests related clean up

nadiia:


src/spinner/__tests__/component.test.js
```diff
@@ -29,9 +29,7 @@ describe('Spinner', () => {
 
   test('component overrides', () => {
     const overrides = {
-      Svg: jest
-        .fn()
-        .mockImplementation(({children}) => <span>{children}</span>),
+      Svg: jest.fn().mockImplementation(({children}) => <svg>{children}</svg>),
```
nadiia: the `span` usage caused a warning in tests

blackswanny: probably those children are allowed only inside `svg` tag

PR #359, opened 2018-10-05
fix(button): disable pointer events when in loading state

abmai:
### Button Component

Right now when the button is in `isLoading` state, `onClick` would still get triggered. I think we shouldn't allow that.


src/button/styled-components.js
```diff
@@ -21,6 +21,7 @@ export const BaseButton = styled(
     ...($size === SIZE.compact
       ? $theme.typography.font200
       : $theme.typography.font300),
+    pointerEvents: $isLoading ? 'none' : 'initial',
```
schnerd: Would it be more robust to set the `onClick` to null? In theory `pointer-events` wouldn't affect people pressing button with keyboard.

blackswanny: the only confident way is to apply `disabled` while in loading as this is `button`, it should work. But it also changes styles of element

abmai: Hm that’s true. I thought this was a clever way of achieving this without having to modify the button code but I guess not :p.

I was thinking to have an internal onClick that can decide to call the user’s onClick or not. But that only covers one event and not all of them. Any ideas?

PR #366, opened 2018-10-08
feat(handler-consistency): use the same handler style across the code base

gergelyke:
Once I get the green light on the proposed changes, I'll add them to this PR.

Implementation done:

- [ ] button
- [ ] checkbox
- [ ] input
- [ ] menu
- [ ] pagination
- [ ] popover
- [ ] radiogroup
- [ ] select
- [ ] tag
- [ ] textarea
- [ ] toast


rfcs/select-component.md
```diff
@@ -177,17 +177,17 @@ export default () => {
   * `SelectComponentIcon: ?React.ComponentType` component for all icons appearing in Select component. It's provided `$type: ICON.loop | ICON.clearTag | ICON.clearAll | ICON.selected` to setup corresponding icon of Select component
   * `Tag: ?React.ComponentType` component for selected options Tags shown in Input for multiple mode selection
   * `SelectSpinner: ?React.ComponentType` component for dropdown spinner if async options are loading
-* `onChange: func(e: SyntheticInputEvent, params: Object)`:
+* `onChange: func({event, params: Object})`:
```
schnerd: We should merge these – everything that's now in params should become a top-level property in this object.

rfcs/radiogroup-component.md
```diff
@@ -138,15 +138,15 @@ export default () => {
   How to position radio buttons in the group. Default is `vertical`
 * `labelPlacement: 'top' | 'right' | 'bottom' | 'left'`:
   How to position the label relative to the radio itself. Default is `right`
-* `onChange: (evt) => {}`:
+* `onChange: ({event}) => void`:
```
schnerd: `{event, value}` here too?

rfcs/select-component.md
```diff
@@ -177,17 +177,17 @@ export default () => {
   * `SelectComponentIcon: ?React.ComponentType` component for all icons appearing in Select component. It's provided `$type: ICON.loop | ICON.clearTag | ICON.clearAll | ICON.selected` to setup corresponding icon of Select component
   * `Tag: ?React.ComponentType` component for selected options Tags shown in Input for multiple mode selection
   * `SelectSpinner: ?React.ComponentType` component for dropdown spinner if async options are loading
-* `onChange: func(e: SyntheticInputEvent, params: Object)`:
+* `onChange: func({event, type, selectedOptions})`:
```
schnerd: I think this needs `option` too:
image

gergelyke: thanks, I've totally missed that! I am not sure if I 100% understand that option though - is that the option the user currently hovers over / navigated to, but did not selected yet? if so, shouldn't we use the `pendingOption` name externally too? cc @blackswanny

blackswanny: it's the one , which is Selected\Unselected. OnChange does work only when we change selections

gergelyke:
@schnerd yeah, that's what I am planning on, I'll add the code changes to this PR

abmai:
Also have some concerns around blasting this change to the other event handlers i.e. `onMouseEnter`- since they already don't do much except propagate the normal even. Let me know if this isn't the best place to discuss that.

1) Why even include them in the public APIs in the first place, if they don't do much (and users could propagate them down themselves)?
2) How do we ensure users won't be confused between these and other event handlers they could pass in?

gergelyke:
@abmai I think that's a very valid question, and we can definitely discuss here 

to start with addressing your second question - as I see, if we provide the same change handler interface to all components we can eliminate any confusion (hopefully). and that I think leads to the answer to your first question  - if we wouldn't include them in the public API, we could not provide the same change handler interface

what do you think?

schnerd:
> Why even include them in the public APIs in the first place, if they don't do much (and users could propagate them down themselves)?
I agree, in some places we are exposing these handlers when we probably don't need to (since people could just pass in via overrides). Removing those handlers in this PR might be outside its scope though.

> How do we ensure users won't be confused between these and other event handlers they could pass in?
I'm also slightly concerned about this. That said, it doesn't seem like we're passing `...props` into the root element in any of our components, so I don't think people will be able to pass an extra `onMouseLeave` etc as a top-level prop. They'll be forced to use overrides, which should probably make them think twice and realize they need a normal event handler.

One option is we could just start with the onChange–style handlers (ones that fire when some state changes), and leave the other `onMouseEnter`, `onFocus`, etc DOM-related handlers as-is for now?

gergelyke:
> One option is we could just start with the onChange–style handlers (ones that fire when some state changes), and leave the other onMouseEnter, onFocus, etc DOM-related handlers as-is for now?

it we'd do that, won't we end up with the same inconsistent API? for some of the change handlers one would use the `onChange({event, value})` syntax, and for others it would be `onChange(event)

kahwee:
I have a similar concern with @abmai.

There's too much code bloat if we provide wrap around all the handlers in the name of consistency throughout our API. 

Why don't we just change `onChange` to do this:

```
onChange={(event, value) => {
}}
```

The second parameter is always the value or values (for multiple select).

gergelyke:
that makes sense - also it seems like most other react component libraries do the that

I am happy to change this PR in favor pf that. Thoughts David, Nadiia, Oleg, Anh?

abmai:
I still think there's value in doing `onSomethingChange({event, value})` in cases where we implement our own logic inside the handlers. It's that when we're just overriding a handler that should otherwise be pass-thru, it can be confusing.

Again, worth calling out that from looking at the surface area of change, there are two categories of handlers we're documenting in our public API:

1) Custom handlers, or handler overrides with additional logic (i.e. `onChange(event, params)`)
2) Pass-thru handlers, (many cases of `onChange`, all cases of `onMouseEnter`, etc.)

Having an object-style params for the first case further differentiates that these are custom handlers, which is a good thing in my opinion. In the second case, I don't think we should even document it, but should properly propagate those handlers through. Therefore, we wouldn't even need to change them. Another argument for doing proper pass-thru is that the full list of events that a component should support is pretty big: https://reactjs.org/docs/events.html and it's odd if we choose to document only some of them even though we don't do anything on top of the raw handler.

With that said, those changes are out of scope of this PR. My suggestions for going forward:

1) Keep this PR, but scope the changes to only custom and overriden handlers.
2) Open a separate PR that will remove *all* pass-thru handlers, and implement correct props pass-thru for those components.

Thoughts?

schnerd:
I'm on board with what Anh suggested.

Seems like that would mean only updating the following handlers in this PR:
Accordion - onToggle
File Uploader - onChange
Menu - onItemSelect
Pagination - onPageChange
ProgressSteps - onClick
Select - onChange
Slider - onChange
Tag - onActionClick

I'm still unclear about the following:
Button Group - onClick - needs `index` to be useful

Everywhere else we'd just use `onChange((event) = > {})`?

@gergelyke thank you for making the RFC changes, I think that helped visualize how this would look across the codebase.

gergelyke:
I like that too - let me update the PR tomorrow, and let’s continue from there

abmai:
> @gergelyke thank you for making the RFC changes, I think that helped visualize how this would look across the codebase.

Agreed, this helped a lot! Thanks, @gergelyke!

kahwee:
> Seems like that would mean only updating the following handlers in this PR:
> Accordion - onToggle
> File Uploader - onChange
> Menu - onItemSelect
> Pagination - onPageChange
> ProgressSteps - onClick
> Select - onChange
> Slider - onChange
> Tag - onActionClick
> 
> I'm still unclear about the following:
> Button Group - onClick - needs `index` to be useful
> 
> Everywhere else we'd just use `onChange((event) = > {})`?

Yes, thanks for surfacing these @gergelyke. This is a really helpful PR!

Generally I think using `onChange` to do our custom events is good so long as it doesn't create ambiguity to the existing DOM element's `onChange`. I think terms like `RadioGroup` and `Pagination` are not part of DOM and I don't expect it to have the same signature as the DOM's `onChange`. In this case, I wouldn't overwrite `Input`'s `onChange` due to DOM also having an `input`. We should document our custom `onChange` with an extra icon or label, I have a feeling it would confuse a new comer.

gergelyke:
closing in favor of https://github.com/uber-web/baseui/pull/372

nadiia:
I'm more inclined to make the args consistent across all handlers on the top level props. And avoid passing through the props - anything extra should be passed through overrides. There would still be a confusion when a component reused as a subcomponent of another one. That would create a mix of handlers on the overrides. 
A good documentation would be required for any approach we choose to clear all the differences handlers args.

PR #368, opened 2018-10-08
rfc(input) Propsed changes to flow typing for SyntheticEvent

kahwee:
It's more accurate to use [`SyntheticFocusEvent`](https://github.com/facebook/flow/blob/v0.82.0/lib/react-dom.js#L170) and [`SyntheticInputEvent`](https://github.com/facebook/flow/blob/v0.82.0/lib/react-dom.js#L170) instead of the generic `SyntheticEvent` for `onBlur`/`onFocus` and `onChange` respectively, they have more properties such as `relatedTarget`.

If this RFC is approved, I'll open another PR to include this work for `input`.

Other components are also using `SyntheticEvent`, if it makes sense, I'll include the RFC changes all in here.

### DIFF
--- a/rfcs/input-component.md
+++ b/rfcs/input-component.md
@@ -73,8 +73,12 @@ export default () => {
   * `nextState` - a new state value to be set
   * `currentState` - current state value
   * `stateToSet` - a return value that the state will be updated with
-* `onChange: (e: SyntheticEvent<HTMLInputElement>) => void` - Optional
+* `onChange: (e: SyntheticInputEvent<HTMLInputElement>) => void` - Optional
   onChange event handler.
+* `onBlur: (e: SyntheticFocusEvent<HTMLInputElement>) => void` - Optional
+  onBlur event handler.
+* `onFocus: (e: SyntheticFocusEvent<HTMLInputElement>) => void` - Optional
+  onFocus event handler.
 
 ## Input component API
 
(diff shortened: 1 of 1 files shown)


gergelyke:
should we merge this effort with https://github.com/uber-web/baseui/pull/366? as I understand this would be a breaking change around change handlers, would be nice to limit the breaking changes on the same concern to only one

what do you think?

rfcs/input-component.md
```diff
@@ -125,11 +129,11 @@ export default () => {
   Defines styles for inputs that are grouped with other controls.
 * `size: 'default' | 'compact'`
   Defines the size of an input control.
-* `onChange: Function` - Optional
+* `onChange: SyntheticInputEvent` - Optional
```
nadiia: this should still be a `Function` but maybe `(e: SyntheticInputEvent<HTMLInputElement>) => void`

rfcs/input-component.md
```diff
@@ -125,11 +129,11 @@ export default () => {
...
   onChange event handler.
-* `onFocus: Function` - Optional
+* `onFocus: SyntheticFocusEvent` - Optional
```
nadiia: same for `onFocus` and `onBlur` below

kahwee:
Thanks for the feedback! After chatting with @gergelyke, we determined it is best to have this effort be as part of a larger effort in #366. I'm closing this to avoid confusion.

PR #372, opened 2018-10-09
docs(rfcs): change handler consistency

gergelyke:


rfcs/select-component.md
```diff
@@ -177,18 +177,8 @@ export default () => {
...
-  handler for events on trigger element when text of search input (in Search mode has changed).
-* `onMouseEnter: func`:
-  handler for events on trigger element
-* `onMouseLeave: func`:
-  handler for events on trigger element
-* `onFocus: func`:
-  handler for events on trigger element
-* `onBlur: func`:
-  handler for events on trigger element
+* `onSelect: func({event: SyntheticInputEvent, type, option, selectedOptions})`:
```
gergelyke: if we'd change from `onChange` to `onSelect`, we could reserve `onChange` for the input event - what do you think?

rfcs/checkbox-component.md
```diff
@@ -112,16 +112,6 @@ export default () => {
   * `Input` may replace original functionality of checkbox. If provided should follow the same interface and logic or may break the whole component
 * `labelPlacement: 'top' | 'right' | 'bottom' | 'left'`:
   How to position the label relative to the checkbox itself. Default is `right`
-* `onChange: func`:
```
nadiia: I don't think we should remove these handlers from the components API, since we override them and call the user's handlers within it. Could you give more details on why it's being removed?

rfcs/select-component.md
nadiia: Yes, like the idea.

rfcs/checkbox-component.md
gergelyke: I am not sure  - to me it seems we do not override them at all, just passing them through as is: https://github.com/uber-web/baseui/blob/master/src/checkbox/checkbox.js#L168

What am I missing?

nadiia: Ok, we do anything with the `onChange` handler, but we do some state changes on the mouse events.

gergelyke: Oh, found it - but even in that case, from a user's point of view there is no change in the API, right? We are calling anything they pass in with the original input event, right?

rfcs/input-component.md
```diff
@@ -73,8 +73,6 @@ export default () => {
   * `nextState` - a new state value to be set
   * `currentState` - current state value
   * `stateToSet` - a return value that the state will be updated with
-* `onChange: (e: SyntheticEvent<HTMLInputElement>) => void` - Optional
```
nadiia: We override the `onChange` handler in the stateful component. Should keep it.

rfcs/input-component.md
```diff
@@ -125,12 +123,6 @@ export default () => {
   Defines styles for inputs that are grouped with other controls.
 * `size: 'default' | 'compact'`
   Defines the size of an input control.
-* `onChange: Function` - Optional
```
nadiia: We don't do any overrides for the `onChange` on the stateless input but wouldn't it be confusing when we don't say how the `onChange` handler supposed to be passed in? For example, is it on the top level or is it in `overrides` for the BaseInput? Currently, we do pass through all the top-level props to the BaseInput (maybe we need to change it and make it consistent for all components), but we don't pass through any undocumented props to the subcomponents in others, like `Modal` for example.

rfcs/input-component.md
```diff
@@ -125,12 +123,6 @@ export default () => {
...
-  onChange event handler.
-* `onFocus: Function` - Optional
-  onFocus event handler.
-* `onBlur: Function` - Optional
```
nadiia: The `onFocus` and `onBlur` are overwritten to handle internal `isFocused` state.

rfcs/checkbox-component.md
nadiia: That's right. But I think it introduces the issue I mentioned in my [other comment here](https://github.com/uber-web/baseui/pull/372#discussion_r223795765)

rfcs/input-component.md
gergelyke: I see - what are your thoughts on this @schnerd @blackswanny @abmai?

blackswanny: customer shouldn't know, what components we re-use under the hood. I would rather have onChange passed through parent component, cause RFC represents interface only for the very exact component. So may be experienced customers will find the way around via overrides, but it's not for veryone's usage

rfcs/checkbox-component.md
schnerd: Even if it's a pass through, I think the `onChange` handler is important enough to document since it's a required prop. The other handlers are optional and rarely used.

rfcs/select-component.md
```diff
@@ -177,10 +177,10 @@ export default () => {
...
   * `Tag: ?React.ComponentType` component for selected options Tags shown in Input for multiple mode selection
   * `SelectSpinner: ?React.ComponentType` component for dropdown spinner if async options are loading
-* `onChange: func(e: SyntheticInputEvent, params: Object)`:
-  handler for events on trigger element when option are changing selection.  `params` has `type` indicating which action is performed, current `Option` with `id` and `label` of selected\unselected option and `selectedOptions` array of all of selected.
-* `onTextInputChange: func(e: SyntheticInputEvent)`:
-  handler for events on trigger element when text of search input (in Search mode has changed).
+* `onSelect: func({event: SyntheticInputEvent, type, option, selectedOptions})`:
+  handler for events on trigger element when option are changing selection.  The `type` indicating which action is performed, current `Option` with `id` and `label` of selected\unselected option and `selectedOptions` array of all of selected.
+* `onChange: func`:
+  handler for events on trigger element
```
nadiia: It may not be clear what is a `trigger element`, should we say something like`onChange handler for an input element`?

rfcs/checkbox-component.md
```diff
@@ -113,7 +113,7 @@ export default () => {
 * `labelPlacement: 'top' | 'right' | 'bottom' | 'left'`:
   How to position the label relative to the checkbox itself. Default is `right`
 * `onChange: func`:
-  handler for events on trigger element
+    handler for events on trigger element
```
abmai: Some weird spacing here

rfcs/pagination-component.md
```diff
@@ -38,7 +38,7 @@ export default () => <StatefulPagination numPages={6} />;
   Callback for prev button click
 * `onNextClick: (event) => void` - Optional
```
abmai: Should this and `onPrevClick` go to the new API too, since they're technically custom handlers (custom name, not expected in DOM)?

rfcs/checkbox-component.md
gergelyke: on it

rfcs/select-component.md
blackswanny: I remember `onTextInputChange` was renamed to it's current name due to ambiguous name of `onChange`, because `onChange` is used for main state change, whereas text input is a secondary change yet. So in case of select it's not obvious which change is meant
@schnerd @abmai

abmai: Yep, `onChange` is a bit vague here I think. Any reason why we don't want to keep `onTextInputChange` or `onInputChange`?

docs/event-handlers.md
```diff
@@ -0,0 +1,6 @@
+# Event Handler Usage in Base UI
+
+In Base UI we follow this event handler pattern:
+
+- custom handlers, or handler overrides with custom logic follow use the `onChange({event, value})` method signature,
+- pass-through handlers, such as `onBlur` or `onMouseEnter` are not changed.
```
nadiia: 🎉

docs/migrations/00372.md
```diff
@@ -0,0 +1,53 @@
+# Purpose
+
+The purpose of this change was to converge on a standardized way of using change handlers.
+
+In Base UI we follow this event handler pattern:
+
+- custom handlers, or handler overrides with custom logic follow use the `onChange({event, value})` method signature,
```
nadiia: `with custom logic follow use the` => `with custom logic follow the` | `with custom logic use the`

rfcs/select-component.md
```diff
@@ -202,15 +202,15 @@ export default () => {
...
   * `params` may contain `id` and `label` of selected option and `selectedOptions` array of all of selected, as well as new `textValue` set in input.
 * `children: func` should return `Select` instance with standard or customized inner elements.
-* `onChange: func(e: SyntheticInputEvent, params: Object)`:
-  handler for events on trigger element when option are changing selection.  `params` has `type` indicating which action is performed, current `Option` with `id` and `label` of selected\unselected option and `selectedOptions` array of all of selected.
-* `onTextInputChange: func(e: SyntheticInputEvent)`:
-  handler for events on trigger element when text of search input (in Search mode has changed).
+* `onChange: func({event: SyntheticInputEvent, type, option, selectedOptions})`:
+  handler for events on trigger element when option are changing selection.  The `type` indicating which action is performed, current `Option` with `id` and `label` `label` of selected\unselected option and `selectedOptions` array of all of selected.
+* `onTextInputChange: func`:
+  event handler for the input element
```
nadiia: 👍

Issue #312, opened 2018-09-25
[select] Revisit/refactor the select component


nadiia:
## Expected Behavior

Provide a stateless select component where a selected values can be controlled.

## Current Behavior

Currently `Select` handles selected values within its internal state and doesn't provide any hook to manage the value without actually selecting/deselecting those.

schnerd:
Just merged #373 – which I think addresses this. Feel free to re-open otherwise.

PR #373, opened 2018-10-09
fix(select): remove selectedOptions from controlled select state

schnerd:
Fix for #312 

Since `selectedOptions` is the value of the form control, it should _not_ be part of the state. Since `Select` is intended to be a controlled component, it should always reflect the value passed in through props.

### DIFF
--- a/src/select/__tests__/select.test.js
+++ b/src/select/__tests__/select.test.js
@@ -10,7 +10,6 @@ import {mount, shallow} from 'enzyme';
 import {
   Select,
   StyledSelectComponentIcon,
-  SelectDropDown,
   StyledInput,
   StyledOption,
 } from '../index';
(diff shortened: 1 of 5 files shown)


src/select/__tests__/select.test.js
```diff
@@ -139,74 +125,18 @@ describe('Stateless select', function() {
...
-          };
-          const event =
-            type === STATE_CHANGE_TYPE.unselect
-              ? [e, undefined, true]
-              : [e, selectedOption];
-          wrapper.instance().onChange(...event);
-          expect(allProps.onChange).toHaveBeenCalledWith(
-            e,
-            Object.assign({}, expectedResult, {type}),
-          );
```
schnerd: This was testing the internal implementation – we should just be testing the public interface of the component.

blackswanny: if this is internal , there has to be the way to test it. For instance to emulate `'click` event here. We shoulnd't just remove tests

src/select/__tests__/select.test.js
```diff
@@ -406,48 +372,31 @@ describe('Stateless select', function() {
           {isDropDownOpen: mode === TYPE.select},
         ],
         [KEY_STRINGS.Escape, 'close dropdown', {isDropDownOpen: false}],
-        [
```
blackswanny: why is this test moved to another place. We re-used more code in here and it was still readable

src/select/__tests__/select.test.js
schnerd: yeah I guess I assumed this behavior was covered by another test that actually simulated click events, but I see now that it was not. i'll add a separate test for that 👍

src/select/__tests__/select.test.js
schnerd: backspace no longer changes internal component state, it emits an `onChange` event, and all of these test were just asserting on component state and not checking for calls to `onChange`. keeping the backspace test here would have resulted in a few more conditional statements that make this code more convoluted.

PR #374, opened 2018-10-09
feat: add unit tests for progress bar

blackswanny:
add unit tests for progress bar


src/progressbar/__tests__/progressbar.test.js
```diff
@@ -0,0 +1,60 @@
...
+  test('should render component', function() {
+    wrapper = mount(<ProgressBar {...allProps} />);
+    expect(wrapper).toMatchSnapshot('Component has correct render');
+  });
+
+  test('should not render label', function() {
+    allProps.showLabel = false;
+    wrapper = mount(<ProgressBar {...allProps} />);
+    const label = wrapper.find(StyledLabel);
+    expect(label.length).toBe(0);
```
nadiia: Could you add a test where we check that the label is actually rendered when the `showLabel=true`?

src/progressbar/__tests__/progressbar.test.js
```diff
@@ -0,0 +1,60 @@
...
+    beforeEach(function() {
+      allProps.getProgressLabel = jest
+        .fn()
+        .mockImplementation(() => <div {...{[customData]: true}} />);
+      wrapper = mount(<ProgressBar {...allProps} />);
+    });
+
+    test('should render custom label provided', function() {
+      const customLabel = wrapper.find('[' + customData + ']');
+      expect(customLabel.length).toBe(1);
```
nadiia: Could you also add tests for actual overrides?

src/progressbar/__tests__/progressbar.test.js
```diff
@@ -0,0 +1,60 @@
...
+    const customData = 'data-custom-label-for-progress-bar';
+    beforeEach(function() {
+      allProps.getProgressLabel = jest
+        .fn()
+        .mockImplementation(() => <div {...{[customData]: true}} />);
+      wrapper = mount(<ProgressBar {...allProps} />);
+    });
+
+    test('should render custom label provided', function() {
+      const customLabel = wrapper.find('[' + customData + ']');
```
nadiia: Please use a string template instead concatenation.

src/progressbar/__tests__/styled-components.test.js
```diff
@@ -0,0 +1,35 @@
...
+        $value: 40,
+        $successValue: 100,
+      };
+      const component = shallow(
+        <Component {...props}>
+          <div />
+        </Component>,
+      );
+      expect(component.instance().getStyles()).toMatchSnapshot(
+        compName + ' has correct styles',
```
nadiia: string concatenation `=>` string template

src/progressbar/__tests__/progressbar.test.js
blackswanny: this is included in default render test. If someone changes it, that test is to be red.

nadiia: I think a snapshot with `MockStyledComponent`s in it is not as useful for testing if the `StyledLabel` is rendered, and is not as readable or easy to understand what's the issue if the test fails

blackswanny: I am not a fan of snapshots as well, but when we discussed, why should we use `jest` I was told about snapshots as powerful tool absent in jasmine or other similar frameworks. I should agree that it's handy despite it may not be readable for those, who are not so familiar with it. So if we would like to go back from snapshots here, we should have a test for every single default property for all components. This way we will make it readable as it should be.

nadiia: I did not say that I am not a fan of it. What I'm saying is there are different use cases where a snapshot is not the best approach for checking if a specific component rendered.

src/progressbar/__tests__/progressbar.test.js
blackswanny: added

src/progressbar/__tests__/styled-components.test.js
blackswanny: ok

PR #376, opened 2018-10-11
feat: improve deep merge function

blackswanny:
improve deep merge function  to avoid stack overflow and possible circular dependency

### DIFF
--- a/src/utils/__tests__/deep-merge.test.js
+++ b/src/utils/__tests__/deep-merge.test.js
@@ -2,6 +2,13 @@
 import deepMerge from '../deep-merge';
 
 describe('deepMerge', () => {
+  let consoleError;
+  beforeEach(function() {
+    consoleError = jest.spyOn(console, 'error').mockImplementation(() => {});
+  });
+  afterEach(function() {
+    consoleError.mockRestore();
+  });
   test('performs deep merge on target', () => {
     const target = {foo: {bar: {baz: true, quux: true}}};
     const source = {foo: {bar: {baz: false}}};
@@ -16,5 +23,27 @@ describe('deepMerge', () => {
     const source2 = {foo: 3, quux: 4};
     const ret = deepMerge({}, target, source1, source2);
     expect(ret).toEqual({foo: 3, bar: 2, baz: 3, quux: 4});
+  });
+
+  test('should forcefully merge very right source for heavy objects to avoid stack overflow and possible circular dependency', () => {
+    const target = {foo: 1, bar: 2};
+    const source1 = {foo: 2, baz: 3};
+    const source2 = {
+      foo: 3,
+      quux: {
+        quaz: {
+          puuz: {
+            gearz: {
+              guanz: {marz: {breez: {fleez: {looz: {weerz: {poolz: {}}}}}}},
+            },
+          },
+        },
+      },
+    };
+    const ret = deepMerge({}, target, source1, source2);
+    expect(ret).toEqual({bar: 2, baz: 3, ...source2});
+    expect(consoleError).toHaveBeenCalledWith(
+      `Object force-merged to target to avoid stack overflow`,
+    );
   });
 });
--- a/src/utils/deep-merge.js
+++ b/src/utils/deep-merge.js
@@ -1,20 +1,33 @@
 // @flow
 
-export default function deepMerge(
-  target?: ?{},
-  ...sources: Array<null | ?{}>
-): {} {
+const MAX_DEPTH_OF_CHECK = 10;
+let depthOfCheck;
+export default function deepMerge(target?: ?{}, ...sources: Array<null | ?{}>) {
+  depthOfCheck = 0;
+  return _deepMerge.apply(this, arguments);
+}
+
+function _deepMerge(target?: ?{}, ...sources: Array<null | ?{}>): {} {
   target = target || {};
   const len = sources.length;
   let obj;
+  if (++depthOfCheck >= MAX_DEPTH_OF_CHECK) {
+    obj = sources[len - 1] || {};
+    // eslint-disable-next-line no-console
+    console.error(`Object force-merged to target to avoid stack overflow`);
+    return obj;
+  }
   let value;
   for (let i = 0; i < len; i++) {
     obj = sources[i] || {};
+    if (obj === target) {
+      continue;
+    }
     for (let key in obj) {
       if (obj.hasOwnProperty(key)) {
         value = obj[key];
         if (isCloneable(value)) {
-          target[key] = deepMerge(target[key] || {}, value);
+          target[key] = _deepMerge.apply(this, [target[key] || {}, value]);
         } else {
           target[key] = value;
         }


src/utils/__tests__/deep-merge.test.js
```diff
@@ -17,4 +24,26 @@ describe('deepMerge', () => {
     const ret = deepMerge({}, target, source1, source2);
     expect(ret).toEqual({foo: 3, bar: 2, baz: 3, quux: 4});
   });
+
+  test.only('should forcefully merge very right source for heavy objects to avoid stack overflow and possible circular dependency', () => {
```
schnerd: remove test.only?

src/utils/deep-merge.js
```diff
@@ -25,5 +35,5 @@ export default function deepMerge(
 }
 
 function isCloneable(obj: mixed) {
-  return Array.isArray(obj) || {}.toString.call(obj) == '[object Object]';
+  return typeof obj === 'object' || typeof obj === 'function';
```
schnerd: why `typeof obj === 'function'`? seems like you wouldn't want to call deepMerge with a function as one of the arguments.

nadiia: Why do you consider functions and `null` cloneable?

src/utils/deep-merge.js
```diff
@@ -1,20 +1,30 @@
 // @flow
 
-export default function deepMerge(
-  target?: ?{},
-  ...sources: Array<null | ?{}>
-): {} {
+const MAX_DEPTH_OF_CHECK = 10;
```
nadiia: Where the number `10` comes from?

src/utils/__tests__/deep-merge.test.js
blackswanny: ok

src/utils/deep-merge.js
blackswanny: may be I didn't get it, but actually function is object and I want it to be called in deepMerge as a normal object

blackswanny: i did exclude `null`, but function is an object. Lots of frameworks starting with jQuery use it as object with props, especially it's common if it's a function-constructor

src/utils/deep-merge.js
blackswanny: we can replace it with any other likable number. When I did research about redux data-objects in some project, the most optimal size was 6, deeper there was "trash" that sometimes appears in data-object inaccurately mixed in. Other number is 40 as it's a const for stack overflow in JS runtime. I can put here any reasonable number if any

src/utils/deep-merge.js
nadiia: Could you point out where you exclude the `null`? How are you intended to `deepMerge` functions?

blackswanny: Just pushed fresh code. Function merges the same way as any other object. We go through it's props and add them to target function, which is the prop itself of parent object

gergelyke:
can we simply use https://www.npmjs.com/package/lodash.merge or https://www.npmjs.com/package/just-extend instead of managing our own?

gergelyke:
Yeah, I know it's been there - my question is if it is worth maintaining it, while we could just pull something on.

blackswanny:
@gergelyke that depends on strategy we follow. If we try to avoid extra dependencies, and take into account those drawbacks I mentioned, than we should maintain our own.

gergelyke:
what are those drawbacks? how many KBs of extra we are talking about in comparison to out implementation?

blackswanny:
@gergelyke size increased by 50KB. For others need more analysis, but this merge may be slower by performance due to more complex logic

gergelyke:
Can you help me understand how could the size increase by 50Ks, if the whole file of just-extend is 1.5K?

nadiia:
We can uses [`just-clone`](https://www.npmjs.com/package/just-clone) by @angus-c

blackswanny:
@gergelyke  this is for `lodash.merge` size.
`just.extend` is only 2KB in webpack. However it misses some smart checks and literally blindly extends, not merges

angus-c:
@blackswanny what do you mean by blindly extends ? (also use just-clone as @nadiia says)

blackswanny:
@angus-c lodash has some safe way of merging. For instance, it checks if objects are the same and does not proceed. Also in this version we added stack overflow prevention, which might be common case in misuse of redux or other data frameworks. 
`just-clone` may work, but It seems like it doesn't clone functions as objects, also it takes time to clone it, whereas we just need to merge\assign the very right source to target.

angus-c:
@blackswanny I don't see lodash merge doing that check
```
const a = b = {a: 3};
a === b; // true
_.merge(a, b); // {a: 3};
```
it also doesn't prevent property clobbering
```
_.merge({a: 3}, {a: 4, b: 2}); // {a: 4, b: 2};
```
It does prevent circular reference errors although e.g. the following example also works in just-extend (maybe some circular references don't) 
```
const extend = require('just-extend')
const x = {a: 2};
const y = {b: 3};
y.f = x;

extend(x, y);
``` 
Full tradeoffs are here https://github.com/angus-c/just/blob/master/TRADEOFFS.md

schnerd:
I added this deepMerge implementation and basically copied the code directly out of `just-extend` as you can probably tell. I copied because it's just a simple function and I didn't want to add a dependency, but I'm indifferent to whether or not we use a dependency instead.

Regarding `just-clone`, maybe I'm missing something but I don't get how that helps here. We need to be able to deep merge two objects, it seems like `just-clone` is for cloning objects, not merging them.

While having protection against stack overflow due to cyclical references is nice, I'm wondering if we need to add it immediately–maybe we should wait until a customer _actually_ runs into this issue and reports it. Also, if we flow type `overrides` and other inputs well, hopefully this should catch any cases where a user would be passing something cyclical.

src/utils/deep-merge.js
schnerd: if you check where `isClonable` is used, it's doing this:

```
if (isCloneable(value)) {
  target[key] = _deepMerge.apply(this, [target[key] || {}, value])
}
```

Imagine that `value` is a function like `() => true`, this means you're trying to make a call like:

```
deepMerge(obj, () => true)
```

this makes no sense. you should only be passing object/null/undefined to deepMerge.

What was wrong with the initial implementation of this method? It's what was used in `just-extend` which is battle tested in countless production apps. We should just keep that.

blackswanny:
@angus-c I just debugged this example and if `a===b` it comes back, which is safe fallback
image

for circular dependencies the next example causes stack overflow. Lodash has smarter way to prevent it, but it's more complex.
```js
  const b = {};
  const c = {d: b};
  b.a = c;
  extend(true, b, c);
```
image

@schnerd yes, I may exclude it, but may at least leave error, so customer will find the bug easier. However, i remember this is pretty common case.

angus-c:
@blackswanny Ok, I can add a patch for circular dependency to just-extend for that over the weekend. I couldn't reproduce the object equality check in lodash.merge

@schnerd BTW just-extend is 369 bytes bundled and minified so I don't think rolling our own will help much with size :) https://bundlephobia.com/result?p=just-extend@3.0.0

nadiia:
> I added this deepMerge implementation and basically copied the code directly out of `just-extend` as you can probably tell. I copied because it's just a simple function and I didn't want to add a dependency, but I'm indifferent to whether or not we use a dependency instead.
> 
> Regarding `just-clone`, maybe I'm missing something but I don't get how that helps here. We need to be able to deep merge two objects, it seems like `just-clone` is for cloning objects, not merging them.
> 
> While having protection against stack overflow due to cyclical references is nice, I'm wondering if we need to add it immediately–maybe we should wait until a customer _actually_ runs into this issue and reports it. Also, if we flow type `overrides` and other inputs well, hopefully this should catch any cases where a user would be passing something cyclical.

My bad, about mentioning the `just-clone` here, got lost in the bunch of comments and arguments  
My biggest concern here as I mentioned earlier in several of my comments is `merging functions` 😳

blackswanny:
@schnerd Here is example. Old merge and new merge. As you see. Property 'bar' is not saved in old version as all props are eliminated by new function. We may consider that it's ok to override all props of function, but I am not sure that it's true for all frameworks and object using function as high-order objects
```js
    const A = () => {};
    A.bar = 'bar';
    const B = () => {};
    B.foo = 'foo';
    const merged = deepMerge({override: A}, {override: B});
```
image
image

schnerd:
@blackswanny For your example, i'd argue that the new merge semantics in this PR are incorrect. If A and B are components, you would want the deepMerge to return B, not A with the properties of B.

While you can assign properties to a function in javascript, I'm not sure that means we have to treat it like an object. This deepMerge helper should only be concerned with merging POJOs.

blackswanny:
@schnerd I just fixed that. Now it returns `B`. However, we do have such examples, for instance all components in React are functions.

blackswanny:
I removed check for functions

blackswanny:
@schnerd yes, we can still use below code for custom height
```js
        overrides={{
          DropDown: {
            style: {
              maxHeight: '100px',
            }
          }
        }}
```

PR #382, opened 2018-10-12
fix: fix styles for rows in dropdown

### LINKED ISSUE #377: [Bug] Select various bugs
* [x ] I have searched the [issues](https://github.com/uber-web/baseui/issues) of this repository and believe that this is not a duplicate.

## Expected Behavior
1. `multiple` prop should default to false
2. `rows` prop should indicate number of options shown in the dropdown list
3. each option needs an `id` in order for the selected styling to be applied appropriately
4. Keyboard shortcuts expose which `Option` override is focused so that customer styling can be added. (This may be a feature enhancement rather than a bug).

## Current Behavior
1. When the `multiple` prop is specified and when it is set (`multiple={false}`) I am still able to select several options
2. While using a custom `Option` component, the number of rows does not correlate to the number or custom options. For example, when `rows={3}` my dropdown only showed half of one option. It seems like this is configured for use with a very specific row height. 
3. When using a list of items, every time I selected one every single item in the dropdown would update with a checkmark. It was not until I looked at the source code for the dropdown until I realized that it determined this styling by checking for `selected.id === item.id`. It was unclear from the docs that each item needs an `id` to behave appropriately. Ideally this should be stated in the docs. 
4. When a custom option is used, the keyboard arrows will accurately select an item. However, the style of the focused item is overridden by the custom option. There is no prop that indicates which index is focussed from keyboard shortcuts, so I cannot apply styles to that option. 

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v2.2.1  |
| React   |    16.5.2     |
| browser |    Chrome     |
| node.js |         |
| etc.    |         |

## Steps to Reproduce

Link:

1.

2.

3.

4.

blackswanny:
Fixes #377 only bug 2

Expected Behavior
    rows prop should indicate number of options shown in the dropdown list

Current Behavior

    While using a custom Option component, the number of rows does not correlate to the number of custom options. For example, when rows={3} my dropdown only showed half of one option. It seems like this is configured for use with a very specific row height


blackswanny:
@schnerd I prefer more natural way to show scroll bar always when it's needed rather then playing with partially visible elements. Unfortunately we can do it only in Firefox and Webkit browsers, for instance:
```js
      '::-webkit-scrollbar': {
        '-webkit-appearance': 'none',
        width: '10px',
      },
      '::-webkit-scrollbar-thumb': {
        borderRadius: '5px',
        backgroundColor: 'rgba(0,0,0,.5)',
        '-webkit-box-shadow': '0 0 1px rgba(255,255,255,.5)',
      },
```

src/select/styled-components.js
```diff
@@ -107,18 +107,27 @@ export const SelectComponentIcon = styled('img', props => {
...
-  listStyle: 'none',
-  borderRadius: $theme.sizing.scale300,
-  boxShadow: $theme.lighting.shadow600,
-}));
+export const DropDown = styled(
+  MenuList,
+  ({$theme, $isOpen, $type, $rows, $itemRef}) => {
+    let maxHeight = null;
+    if ($rows && __BROWSER__ && $itemRef && $itemRef.current) {
+      maxHeight = $itemRef.current.clientHeight * $rows + 'px';
```
nadiia: We shouldn't set a maxHeight based on the height of a single dropdown option.

nadiia:
Overriding a scroll bar appearance doesn't seem to be a good or needed way to go with. Setting a max height for a dropdown menu is totally fine and the right thing to do. We don't necessarily need to expose the `maxHeight` in the toplevel props, just let it override through a normal overrides flow.

src/select/styled-components.js
blackswanny: this is actually the main fix itself. So we need to set the height (max height) of the list to the amount of rows, where each row's height is unknown. I have searched possible solution. So far there is no way to do it via css, as this value is dynamic and changing over the rendering cycle. If you have working solution, I can change it accordding to it.

blackswanny:
> We don't necessarily need to expose the maxHeight in the toplevel props, just let it override through a normal overrides flow.

so we do not expose it. Actually, I didn't get the concern here. We have property `rows`, but no `maxHeight` exposed.

src/select/dropdown.js
```diff
@@ -68,12 +69,14 @@ export default function SelectDropDown(props: DropDownPropsT) {
...
             $rows: rows,
             $type: type,
+            $itemRef: itemRef,
             $isOpen: isDropDownOpen,
             ...dropDownProps,
           },
         },
         Option: {
           props: {
+            $ref: itemRef,
```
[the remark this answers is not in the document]
blackswanny: the ref is needed to measure height of one element in list. Yes, it could have been assigned many times, but to avoid this I will have to change inner code of Menu. The menu itself works with only one as it has only one cycle of rendering

blackswanny:
@nadiia Please, clarify, which exact changes are requested? This is a fix of bug of `rows` props to work with custom elements in dropdown. If you would like to change `rows` to `maxHeight` or remove it, I can do it in another scheduled PR for further improvements, but this is just a fix of bug, which our customers already use.

nadiia:
> > We don't necessarily need to expose the maxHeight in the toplevel props, just let it override through a normal overrides flow.
> 
> so we do not expose it. Actually, I didn't get the concern here. We have property `rows`, but no `maxHeight` exposed.

@blackswanny you skipped the initial and related part of my comment in your quoted reply. And the first part talks about `maxHeight` as a reply to @schnerd'a comment. I did not say that we currently `expose it` (`it` = `maxHeight`):

> Setting a max height for a dropdown menu is totally fine and the right thing to do. We don't necessarily need to expose the maxHeight in the toplevel props, just let it override through a normal overrides flow.

nadiia:
> @nadiia Please, clarify, which exact changes are requested? This is a fix of bug of `rows` props to work with custom elements in dropdown. If you would like to change `rows` to `maxHeight` or remove it, I can do it in another scheduled PR for further improvements, but this is just a fix of bug, which our customers already use.

@blackswanny I would argue that the current solution fully fixes the reported bug. Please see an example in the screenshot below where the same issue will pop up again. In the example the options have different heights. The current fix of setting the `maxHeight` based on the last rendered option's height will give the same outcome where an earlier rendered option with a larger height value takes the max height space.
screen shot 2018-10-15 at 10 28 46 am

Replacing the `rows` prop with `maxHeight` prop is a good way to solve this height issue. I don't have a preference if it's done in this PR or a separate. Also replacing `rows` with `maxHeight` will introduce a breaking change.

@schnerd do you think it makes sense to merge the current change in its current state?

blackswanny:
@nadiia ok, let me fix this custom case as well

schnerd:
> @schnerd do you think it makes sense to merge the current change in its current state?

If I'm following correctly, sounds like we're gravitating towards getting rid of `rows`, and using a default `maxHeight` style of 250px or so – then people can use overrides to change maxHeight style rule if they need to. This seems like the easiest solution to me and we should just update this PR.

blackswanny:
@schnerd maxHeight and rows serve different things. I just fixed it for very custom example provided by @nadiia , so now we always guarantee amount of visible items, not height.

gergelyke:
quick update - I've just talked with Jeff, and the conclusion is: `maximum height should be capped at 900px or 90% of the viewport height which ever is appropriate to the size of the expanded viewport.`Based on this, I'd recommend picking that as the default value, and expose the `maxHeight` property

blackswanny:
closed in favor of https://github.com/uber-web/baseui/pull/397

PR #383, opened 2018-10-12
fix(button): propagate event by spreading args

schnerd:
Minor regression from #359 reported by an engineer on chat.

### DIFF
--- a/src/button/__tests__/button.test.js
+++ b/src/button/__tests__/button.test.js
@@ -50,6 +50,18 @@ describe('Button Component', () => {
     expect(component.find(LoadingSpinner)).toExist();
   });
 
+  test('onClick called with event', () => {
+    const onClick = jest.fn();
+    const component = mount(<Button onClick={onClick} />);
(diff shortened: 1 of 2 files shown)


src/button/__tests__/button.test.js
```diff
@@ -50,6 +50,18 @@ describe('Button Component', () => {
...
   });
 
+  test('onClick called with event', () => {
+    const onClick = jest.fn();
+    const component = mount(<Button onClick={onClick} />);
+    component.simulate('click', {
+      target: 'foo',
+    });
+    expect(onClick).toHaveBeenCalledTimes(1);
+    expect(onClick.mock.calls[0][0]).toMatchObject({
```
blackswanny: in this case we can use `toEqual` as nothing is mixed in

schnerd: Tried that, React/enzyme still wraps the simulated event in a wrapper object (SyntheticEvent?) so they don't match.

PR #389, opened 2018-10-16
feat(toast): add toaster tests (part 2)

nadiia:


src/toast/__tests__/toaster.test.js
```diff
@@ -0,0 +1,361 @@
...
+  describe('toaster methods', () => {
+    let renderedToaster;
+
+    beforeEach(() => {
+      wrapper = mount(toaster.create());
+      renderedToaster = wrapper.find(Toaster).first();
+      toaster.getRef = jest.fn().mockReturnValue(renderedToaster.instance());
+    });
+
+    test('toaster[show | update | clear]', () => {
```
gergelyke: do you think we can split this into multiple test cases? it seems to be super long

nadiia: Only the show part could be separated, but the update/clear still need some toasts to be shown first so it would be included anyway.

PR #397, opened 2018-10-17
fix: add Max Height for Select Component

### LINKED ISSUE #377: [Bug] Select various bugs

blackswanny:
Fixes #377 only bug 2

Expected Behavior
rows prop should indicate number of options shown in the dropdown list

Current Behavior

While using a custom Option component, the number of rows does not correlate to the number of custom options. For example, when rows={3} my dropdown only showed half of one option. It seems like this is configured for use with a very specific row height


rfcs/select-component.md
```diff
@@ -153,6 +153,8 @@ export default () => {
   Whether the control should be in an error state. Default is `false`
 * `multiple: ?boolean`:
   Sets if multiple choices are allowed in Select component. Default is `false`
+* `maxHeight: ?string`:
```
nadiia: We should also delete `rows` (line 146) from the API doc here

src/select/dropdown.js
```diff
@@ -65,6 +66,7 @@ export default function SelectDropDown(props: DropDownPropsT) {
       overrides={{
         List: {
           component: DropDown,
+          style: {maxHeight},
           props: {
             $rows: rows,
```
nadiia: Is this one still needed?

src/select/select.js
```diff
@@ -334,11 +336,17 @@ class Select extends React.Component<PropsT, StatelessStateT> {
 
   getDropDown() {
     const {overrides, type, rows, selectedOptions} = this.props;
+    let maxHeight = this.props.maxHeight;
+    if (__BROWSER__ && parseInt(maxHeight) > window.innerHeight) {
```
nadiia: Should we have some check to make sure we're dealing here with `maxHeight` pixels value?

src/select/select.js
```diff
@@ -334,11 +336,17 @@ class Select extends React.Component<PropsT, StatelessStateT> {
...
+      // only for pixel-sized maxHeight
+      maxHeight = '80vh';
+    }
     const options = this.getOptions();
     const {isDropDownOpen, optionsLoaded} = this.state;
     const dropDownProps = {
       rows,
```
nadiia: We should remove `rows` is it's now used.

rfcs/select-component.md
nadiia: Since this property is exposed on the select, how about we name it something like `maxMenuHeight` or `maxDropdownHeight` so there is no confusion that it can be related to the select's height?

gergelyke: I'd go with `maxMenuHeight` as https://react-select.com/props

src/select/dropdown.js
blackswanny: removed

src/select/select.js
blackswanny: removed

gergelyke:
are we looking at a breaking change with this one?

src/select/select.js
blackswanny: added

rfcs/select-component.md
gergelyke: makes sense

blackswanny:
@gergelyke added migration guide

docs/migrations/00397.md
```diff
@@ -0,0 +1,8 @@
+#### In `Select` component use `maxDropdownHeight` instead of `rows` to setup max height of Select DropDown
+
+##### `maxDropdownHeight` accepts any height style as a string. For `px` value it can prevent Dropdown from being hidden and always set to no more 90% of screen
```
gergelyke: I am not sure if I understand this part - can you rephrase? also may worth mentioning the default values?

blackswanny: rephrased

src/select/__tests__/styled-components.test.js
```diff
@@ -61,23 +61,21 @@ describe('Select styled components', () => {
...
+      (prop, value) => {
+        const props = {};
+        props[prop] = value;
+        const component = shallow(
+          <StyledDropDown {...props}>
+            <div />
+          </StyledDropDown>,
+        );
+        expect(component.instance().getStyles()).toMatchSnapshot(
+          'StyledDropDown has correct styles when ' + prop + ' is ' + value,
```
nadiia: Use string template instead concatenation.

blackswanny: ok

PR #403, opened 2018-10-18
test(tooltip): add tooltip end-to-end tests

gergelyke:


src/tooltip/e2e.js
```diff
@@ -0,0 +1,49 @@
...
+    goToUrl({
+      suite,
+      test: scenarios.SIMPLE_EXAMPLE,
+      browser,
+    })
+      .waitForElementVisible('body', 1000)
+      .moveToElement('span', 10, 10)
+      .waitForElementPresent(selectors.tooltip, 1000)
+      .moveToElement('body', 10, 10)
+      .waitForElementNotPresent(selectors.tooltip, 1000);
```
blackswanny: nightwatch\selenium are not assert frameworks, they just help to reach the point, when we can use `jest` asserts\expects. Assuming this will fail, but not as normal reported jest test, but due to exception raised by nightwatch. So every test should probably contain exact `expect`s or `assertVisuals`, which we stated to use recently

PR #407, opened 2018-10-19
test(select): add e2e tests to Select

blackswanny:
add e2e tests to Select

### DIFF
--- a/src/select/e2e.js
+++ b/src/select/e2e.js
@@ -1,0 +1,141 @@
+/*
+Copyright (c) 2018 Uber Technologies, Inc.
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+/* eslint-env node */
+/* eslint-disable flowtype/require-valid-file-annotation */
+/* global after */
+
+const scenarios = require('./examples-list');
+const {goToUrl} = require('../../e2e/helpers');
+
+const suite = 'Select Test Suite';
+
+const selectors = {
+  selectInput: '[role="combobox"]',
+  selectDropDown: '[role="listbox"]',
+  dropDownOption: '[role="option"]',
+  selectedList: '[role="list"]',
+  selectedOption: '[role="listitem"]',
+  searchType: '[aria-autocomplete="list"]',
+  selectType: '[aria-autocomplete="none"]',
+  expandedDropDown: '[aria-expanded="true"]',
+};
+
+describe('The select component', () => {
+  after((browser, done) => {
+    browser.end(() => done());
+  });
+
+  // TODO(#424)
+
+  it('opens dropdown menu when click on select input', browser => {
+    goToUrl({
+      suite,
+      test: scenarios.SELECT,
+      browser,
+    })
+      .waitForElementVisible(selectors.selectInput, 1000)
+      .click(selectors.selectInput)
+      .waitForElementPresent(selectors.selectDropDown, 1000);
+  });
+
+  it('opened dropdown can be closed with ESC', browser => {
+    goToUrl({
+      suite,
+      test: scenarios.SELECT,
+      browser,
+    })
+      .waitForElementVisible(selectors.selectInput, 1000)
+      .click(selectors.selectInput)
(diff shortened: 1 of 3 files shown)


src/select/e2e.js
```diff
@@ -0,0 +1,76 @@
...
+  selectDropDown: '[role="listbox"]',
+  dropDownOption: '[role="option"]',
+  selectedList: '[role="list"]',
+  selectedOption: '[role="listitem"]',
+  searchType: '[aria-autocomplete="list"]',
+  selectType: '[aria-autocomplete="none"]',
+  expandedDropDown: '[aria-expanded="true"]',
+};
+
+describe('The select component', () => {
```
gergelyke: can you add the accessibility checks?

src/select/select.js
```diff
@@ -457,6 +467,16 @@ class Select extends React.Component<PropsT, StatelessStateT> {
       ),
     };
   }
+
+  getAccessibilityProps() {
```
gergelyke: shouldn't we address these in a separate PR?

src/select/select.js
```diff
@@ -296,7 +303,7 @@ class Select extends React.Component<PropsT, StatelessStateT> {
     const {type, disabled, selectedOptions} = this.props;
     const multiple = this.isMultiple();
     return (
-      <React.Fragment>
+      <div role="list">
```
nadiia: With adding an additional element we need to add it to the overrides API to be replaceable

src/select/select.js
blackswanny: https://github.com/uber-web/baseui/pull/409

src/select/select.js
blackswanny: added here #409

src/select/e2e.js
```diff
@@ -0,0 +1,76 @@
...
+    browser.end(() => done());
+  });
+
+  it('opens dropdown menu when click on select input', browser => {
+    goToUrl({
+      suite,
+      test: scenarios.SELECT,
+      browser,
+    })
+      .waitForElementVisible('body', 1000)
```
nadiia: We should wait on the `selectors.selectInput` to be visible instead

src/select/e2e.js
```diff
@@ -0,0 +1,76 @@
...
+      .click(selectors.selectInput)
+      .waitForElementPresent(selectors.selectDropDown, 1000)
+      .click(
+        selectors.selectDropDown +
+          ' ' +
+          selectors.dropDownOption +
+          ':first-child',
```
nadiia: string template instead of string concatenation

src/select/e2e.js
blackswanny: ok

src/select/e2e.js
blackswanny: ok

blackswanny:
> Left a few comments but other than that looks great!
> Could you add a few tests for the multiselect?

@nadiia added

src/select/e2e.js
```diff
@@ -0,0 +1,151 @@
...
+    goToUrl({
+      suite,
+      test: scenarios.MULTI_SELECT_SEARCH,
+      browser,
+    })
+      .waitForElementVisible(selectors.selectInput, 1000)
+      .setValue(selectors.selectInput, 'dark', () => {
+        browser
+          .waitForElementPresent(selectors.selectDropDown, 100)
+          .pause(1000)
```
nadiia: Can we increase the wait time to `1000` on the line 139 instead of `pause` here?

nadiia:
Oh, the tests on CI fail, please make sure it passes before merging.

src/select/e2e.js
blackswanny: that will not work as `pause`, which works to wait for async call, which we emulate here. Cause DropDown is shown immediately with loading spinner, but no data yet there

nadiia: Oh, I see. Makes sense. Another approach would be to wait until a specific option is visible. If we keep `pause` versus `wait for an option`, could you add a comment to it about the async call and a loading state here?

PR #408, opened 2018-10-22
feat(input): use more accurate Flow types

### LINKED ISSUE #381: [Input] Flow typing for onChange's emitted event should be SyntheticEvent<HTMLInputElement>
Instead, it is SyntheticEvent<HTMLElement> which runs into issues when trying to access event.target.value.

phou87:
Fixes #381 

These more specific Flow types allow clients to access properties such as event.target.value without needing a runtime check.


nadiia:
Talked to Paul about this change effecting the textarea component. A proposed update will follow here.

PR #409, opened 2018-10-22
feat(select): add accessibility to select component

blackswanny:
add accessibility to select component

### DIFF
--- a/rfcs/select-component.md
+++ b/rfcs/select-component.md
@@ -43,6 +43,7 @@ import {
   StyledSelectComponentIcon,
   StyledDropDown,
   StyledOption,
+  StyledSelectionContainer,
   ICON,
   OPTIONS,
   TYPE,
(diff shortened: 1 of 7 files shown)


src/select/styled-components.js
```diff
@@ -152,3 +152,7 @@ export const SelectSpinner = styled('div', () => {
     paddingLeft: '50%',
   };
 });
+
+export const SelectionList = styled('div', props => {
```
nadiia: How about we call it `SelectionContainer`? Since it's not a list.

rfcs/select-component.md
```diff
@@ -177,6 +178,7 @@ export default () => {
   * `SelectComponentIcon: ?React.ComponentType` component for all icons appearing in Select component. It's provided `$type: ICON.loop | ICON.clearTag | ICON.clearAll | ICON.selected` to setup corresponding icon of Select component
   * `Tag: ?React.ComponentType` component for selected options Tags shown in Input for multiple mode selection
   * `SelectSpinner: ?React.ComponentType` component for dropdown spinner if async options are loading
+  * `SelectionList: ?React.ComponentType` component for the wrapper around selected options\tags
```
nadiia: Commented below regarding the name.

src/select/styled-components.js
blackswanny: ok

rfcs/select-component.md
blackswanny: ok

PR #416, opened 2018-10-24
chore(package): move commit hooks to husky.hooks to avoid warnings

leyanlo:
#### Minor: move commit hooks to husky.hooks to avoid warnings

Otherwise will see the following warning on commit
```
Warning: Setting pre-commit script in package.json > scripts will be deprecated
Please move it to husky.hooks in package.json, a .huskyrc file, or a husky.config.js file
Or run ./node_modules/.bin/husky-upgrade for automatic update

See https://github.com/typicode/husky for usage
```


leyanlo:
Ah thanks for catching that! Updated the commit

PR #418, opened 2018-10-24

blackswanny:
add unit tests to Slider


blackswanny:
close in favor of #423 due to broken build code here and forced rebase

PR #420, opened 2018-10-25
chore(input): prep for e2e tests

nadiia:


src/input/base-input.js
```diff
@@ -99,30 +101,20 @@ class BaseInput extends React.Component<BaseInputPropsT, InternalStateT> {
...
-    const Input = getOverride(InputOverride) || StyledInput;
-    const Before = getOverride(BeforeOverride) || null;
-    const After = getOverride(AfterOverride) || null;
-
+    const [InputContainer, inputContainerProps] = getOverrides(
+      InputContainerOverride,
+      StyledInputContainer,
+    );
+    const [Input, inputProps] = getOverrides(InputOverride, StyledInput);
+    const [Before, beforeProps] = getOverrides(BeforeOverride, NullComponent);
```
blackswanny: I think it's better to change `getOverrides` or change logic of input itself to accept `null` rather than maintain this `NullComponent`

blackswanny: `null` value returned for render instead of component seems to be valid for React to not render at all. I used it in some cases

PR #421, opened 2018-10-25
test(e2e): temporarily disable visual tests

nadiia:


e2e/helpers/index.js
```diff
@@ -26,9 +27,9 @@ function assertVisuals({browser, id}) {
   }
 }
 
+// Disables the visual tests for now while waiting for the baseline approved
 module.exports = {
   getUrl,
   goToUrl,
   formatFileName,
-  assertVisuals,
```
blackswanny: if we don't export, consuming code will fail
try changing this flag to false

image

nadiia: Good catch! Updated.

PR #425, opened 2018-10-26

blackswanny:
add unit tests to Slider


src/slider/__tests__/stateful-slider-container.test.js
```diff
@@ -0,0 +1,86 @@
...
+      allProps.stateReducer = stateReducerMock;
+      wrapper = mount(<StatefulSliderContainer {...allProps} />);
+      instance = wrapper.instance();
+    });
+
+    test.each(handlers)(
+      'should call state reducer to apply new state for %s event with %s type',
+      (eventHandler, type, newState) => {
+        const handler = instance[eventHandler];
+        const params = Object.assign({}, newState);
```
nadiia: `const params = {...newState};`

src/slider/__tests__/stateful-slider-container.test.js
```diff
@@ -0,0 +1,86 @@
...
+        handler({event, ...params});
+        expect(stateReducerMock).toHaveBeenCalledWith(
+          type,
+          newState,
+          {},
+          event,
+        );
+        expect(events[eventHandler]).toHaveBeenCalledWith({event, ...params});
```
nadiia: Could you add a test to check that StatefulContainer's `onChange` method is called when Slider's change event happens?

src/slider/__tests__/stateful-slider-container.test.js
blackswanny: fixed in #429

src/slider/__tests__/stateful-slider-container.test.js
blackswanny: stateless Slider? as we mock stateless slider due to be an external dependency so I can't simulate 'onChange' of that Slider. If I don't mock it, the test may be not pure and isolated. I believe, this test should be in stateless Slider code

blackswanny:
> Will you add tests for the Slider component?

adding in #429

PR #429, opened 2018-10-26

blackswanny:
add unit tests to Slider

### DIFF
--- a/src/slider/__tests__/slider.test.js
+++ b/src/slider/__tests__/slider.test.js
@@ -1,0 +1,239 @@
+/*
+Copyright (c) 2018 Uber Technologies, Inc.
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+// @flow
+import React from 'react';
+import {mount} from 'enzyme';
+import {Slider, StyledThumb} from '../index';
+
+/* global document */
+/* global window */
+
+const originalDocumentAddListener = document.addEventListener;
+const originalDocumentRemoveListener = document.removeEventListener;
+
+describe('Stateless slider', function() {
+  let wrapper;
+  let allProps: {
+      range: Array<number>,
+      value: Array<number>,
+      step: ?number,
+      onChange: () => void,
+    },
+    mockFn;
+
+  beforeAll(function() {
+    //$FlowFixMe
+    document.addEventListener = jest.fn();
+    //$FlowFixMe
+    document.removeEventListener = jest.fn();
+  });
+
+  beforeEach(function() {
+    mockFn = jest.fn();
+    allProps = {
+      onChange: mockFn,
+      range: [0, 100],
+      value: [60],
+      step: 0,
+    };
+  });
+
+  afterEach(function() {
+    jest.restoreAllMocks();
+    document.addEventListener.mockClear();
+    document.removeEventListener.mockClear();
+    wrapper && wrapper.unmount();
+  });
+
+  afterAll(function() {
+    //$FlowFixMe
+    document.addEventListener = originalDocumentAddListener;
+    //$FlowFixMe
+    document.removeEventListener = originalDocumentRemoveListener;
+  });
+
+  it('should render slider component', function() {
+    wrapper = mount(<Slider {...allProps} />);
+    expect(wrapper).toMatchSnapshot('Component has correctly rendered');
+  });
+
+  describe('events', function() {
+    let event, thumb, step, axisSizeInPixels;
+    function emulateThumbDown(thumbIndex = 0) {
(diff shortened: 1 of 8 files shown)


src/slider/__tests__/slider.test.js
```diff
@@ -0,0 +1,186 @@
...
+      step: 0,
+    };
+  });
+
+  afterEach(function() {
+    jest.restoreAllMocks();
+    document.addEventListener.mockClear();
+    wrapper && wrapper.unmount();
+    //$FlowFixMe
+    document.addEventListener = originalDocumentAddListener;
```
nadiia: We don't need to restore it after every single test but can do it in the `afterAll` instead

src/slider/__tests__/slider.test.js
```diff
@@ -0,0 +1,186 @@
...
+      range: Array<number>,
+      value: Array<number>,
+      step: ?number,
+      onChange: () => void,
+    },
+    mockFn;
+
+  beforeEach(function() {
+    //$FlowFixMe
+    document.addEventListener = jest.fn();
```
nadiia: We clear mock after each test therefore we can do this assignment only once in the `beforeAll`

src/slider/__tests__/slider.test.js
```diff
@@ -0,0 +1,186 @@
...
+      step = 10;
+      axisSizeInPixels = 300;
+      jest
+        .spyOn(window, 'getComputedStyle')
+        .mockImplementation(() => ({width: axisSizeInPixels}));
+      allProps.step = step;
+    });
+
+    it('should start tracking thumb move when user clicks on it', function() {
+      emulateThumbDown();
```
nadiia: here we should also check that it subscribed for the `mousemove` and `mouseup` events

src/slider/__tests__/slider.test.js
```diff
@@ -0,0 +1,186 @@
...
+      it('should reset current move progress when thumb moved to X pixels', function() {
+        emulateThumbDown();
+        event.movementX = 50;
+        fireMouseMove(event);
+        expect(wrapper.instance().state).toMatchObject({
+          currentMove: 0,
+        });
+      });
+      it('should note move thumb to X pixels if mouse moved less then step', function() {
+        allProps.step = 50;
```
nadiia: It seems strange that we modify the shared `allProps` object here. I can see that we override it before each test, but the code looks messy. How about making `emulateThumbDown` accept props override, for example?

src/slider/__tests__/slider.test.js
```diff
@@ -0,0 +1,186 @@
...
+        emulateThumbDown();
+        event.movementX = 20;
+        fireMouseMove(event);
+        expect(wrapper.instance().state).toMatchObject({
+          currentMove: 20,
```
nadiia: Could you explain why currentMove is expected to be set in this case but not in the test case above with almost the same setup?

src/slider/__tests__/stateful-slider.test.js
```diff
@@ -0,0 +1,70 @@
...
+// @flow
+import React from 'react';
+import {mount} from 'enzyme';
+
+describe('Stateful slider', function() {
+  let allProps: any, wrapper;
+
+  beforeEach(function() {
+    allProps = {};
+    jest.mock('../slider', () => jest.fn(() => <div>test</div>));
```
nadiia: why do we need to mock the Slider here?

src/slider/__tests__/slider.test.js
blackswanny: restore it in `afterAll` is not safe to preserve isolation of unit-test and may cause conflicts in environments. We already have some code in other places, which does similar

blackswanny: ok, moved to afterAll

src/slider/__tests__/slider.test.js
blackswanny: ok

src/slider/__tests__/slider.test.js
blackswanny: the test name explains it `should not move thumb to X pixels if mouse moved less then step`
so mouse move is 20 and it is less than step=50. I added extra text there

src/slider/__tests__/stateful-slider.test.js
blackswanny: we need to mock it right before we require it in test, so we are able to track mocks for each test. If we bubble require to `imports` jest will start mocking it for the whole file once and forever, so we can't know which test receives which mock call

blackswanny:
> Would be nice to add the check for document.removeEventListener calls. Other than that looks good.


added

PR #442, opened 2018-10-31
test(pagination): add e2e tests

gergelyke:


src/pagination/e2e.js
```diff
@@ -0,0 +1,69 @@
...
+  });
+
+  it('passes basic accessiblity tests', browser => {
+    goToUrl({
+      suite,
+      test: scenarios.STATEFUL_PAGINATION,
+      browser,
+    })
+      .initAccessibility()
+      .waitForElementVisible('body')
```
nadiia: We should wait for an actual element we want to test instead of the`body` element. In a case component is not rendered for some reason this test will pass successfully.

PR #448, opened 2018-11-03
Update README.md to handle dependencies

jonsadka:
Baseui requires peer dependencies, which should be installed but are not listed in the `Usage` section. Additionally, update the example with loading of the Styletron Client and Provider (seems required?)

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@
 Add `baseui` and it's peer dependencies to your project:
 
 ```bash
-# using yarn 
+# using yarn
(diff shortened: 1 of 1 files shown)


gergelyke:
@jonsadka thanks for this contribution! it seems like the markdown linter fails - can you fix that please? you can run it using `yarn lint:markdown`

jonsadka:
> @jonsadka thanks for this contribution! it seems like the markdown linter fails - can you fix that please? you can run it using `yarn lint:markdown`

Sure thing. Updated to clean lint error

PR #449, opened 2018-11-05
feat(modal): prevent background scroll when modal opens

### LINKED ISSUE #414: Modal Background Scroll Bar Lock
Want to have scroll bar lock when modal open,
but tried with Root overflow none does not work.

What dost "Root" point to and how to lock the background if possible? 

Ref: overrides: {Root, Backdrop, Dialog, DialogContainer, Close}

schnerd:
#### Fixed Issues?

Fixes #414

#### Patch: Bug Fix?

Yes

#### Major: Breaking Change?

No

#### Minor: New Feature?

No

#### Tests Added + Pass?

Yes

#### Any Dependency Changes?

No

#### License

MIT

### DIFF
--- a/src/modal/__tests__/modal.test.js
+++ b/src/modal/__tests__/modal.test.js
@@ -122,6 +122,22 @@ describe('Modal', () => {
     expect(onClose).not.toHaveBeenCalled();
   });
 
+  test('prevents scroll on mount node', () => {
+    const onClose = jest.fn();
+    wrapper = mount(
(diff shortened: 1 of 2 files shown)


gergelyke:
@schnerd should it be a `patch` if this is a bugfix?

schnerd:
> @schnerd should it be a `patch` if this is a bugfix?

Yes, i think it should be a patch. i'd consider it a bug fix.

PR #450, opened 2018-11-05
fix(textarea): remove types related condition in the input stateful container that was breaking stateful textarea

nadiia:


src/input/stateful-container.js
```diff
@@ -32,20 +32,16 @@ class StatefulContainer<T: EventTarget> extends React.Component<
...
     type: StateTypeT,
     e: SyntheticInputEvent<T>,
   ) => {
     let nextState = {};
-    if (
-      type === STATE_CHANGE_TYPE.change &&
-      // eslint-disable-next-line cup/no-undef
-      e.target instanceof window.HTMLInputElement
-    ) {
+    if (type === STATE_CHANGE_TYPE.change) {
```
gergelyke: should we add a unit test to cover this?

nadiia: good point, added

PR #454, opened 2018-11-06
feat(accordion): add accordion component (part1)

nadiia:
screen shot 2018-11-06 at 11 31 30 am

### DIFF
--- a/src/accordion/accordion.js
+++ b/src/accordion/accordion.js
@@ -1,0 +1,115 @@
+/*
+Copyright (c) 2018 Uber Technologies, Inc.
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+// @flow
+import * as React from 'react';
+import {getOverrides} from '../helpers/overrides';
+import {Root as StyledRoot} from './styled-components';
+import {STATE_CHANGE_TYPE} from './constants';
+import type {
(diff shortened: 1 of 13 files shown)


src/accordion/accordion.js
```diff
@@ -0,0 +1,110 @@
...
+    this.setState(newState);
+    onChange(newState);
+  }
+
+  getItems() {
+    const {expanded} = this.state;
+    const {accordion, disabled, children} = this.props;
+    const newChildren = [];
+
+    React.Children.forEach(children, (child, index) => {
```
gergelyke: can we use `React.Children.map` here so we can get rid of the side effect of the iterator function?

gergelyke:
should we add the `aria-controls` attribute to the implementation? http://web-accessibility.carnegiemuseums.org/code/accordions/

src/accordion/accordion.js
```diff
@@ -0,0 +1,109 @@
+// @flow
```
austin94: [nit] Does this file need the copyright header?

nadiia: it does! let me fix it

nadiia:
> should we add the `aria-controls` attribute to the implementation? http://web-accessibility.carnegiemuseums.org/code/accordions/

Created an issue for myself to add it in a follow-up PR #473

PR #458, opened 2018-11-06
Table rfc

dswho:
This is a technical proposal for baseUI regarding the API design of a basic table, which handles the canonical table format of one header row and many rows of data. Any other table formats (e.g. having rows or columns that may span multiple rows or columns can be handled using a [renderProp](https://reactjs.org/docs/render-props.html))

Also, there will be support for sorting.

I had a question about what was the best way to determine how many rows are displayed initially in the table. Let's say the user has 5000 rows to load into a table. I was thinking about arbitrarily setting a `pageSize` of 10, such that 10 rows are displayed initially, and then on user scroll or click of a pagination element, more elements are displayed. This lends the question: should a table come with a scrollbar and/or pagination element out of the box? I currently don't have this documented in the proposal but needed further discussion.


rfcs/table.md
```diff
@@ -0,0 +1,321 @@
...
+         );
+    }
+}
+```
+
+
+## Exports
+
+* `Table`
+* `StatefulTable`
```
nadiia: What state does the `StatefulTable` handles?

rfcs/table.md
```diff
@@ -0,0 +1,321 @@
...
+        const {currentPage, pageSize} = this.state;
+        this.fetch(currentPage, pageSize);
+    }
+    onPageChange = (nextPage, prevPage) => {
+        const {pageSize} = this.state;
+        this.fetch(nextPage, pageSize);
+    }
+    render = () => {
+         return (
+             <Table
```
nadiia: Seems like there should be some loading state in the table with remote data

rfcs/table.md
```diff
@@ -0,0 +1,321 @@
...
+  * `Thead` - Container for the table header related elements.
+  * `TheadTr` - A header row. `Tr` for the `Thead` element.
+  * `Th` - A header cell. Corresponds with the `th` html element.
+  * `Td` - A data cell in the table. Corresponds with the `td` html element.
+  * `Tr` - A data row that contains `Td` components.
+
+
+## `Column` API
+One of the Table columns prop for describing the table's columns, Column has the same API.
+* `title: string` - Required. Title of the column to display in the header.
```
nadiia: How about `React.Node` instead of a `string`?

rfcs/table.md
dswho: okay, I'll add an `isLoading` prop to the table.

rfcs/table.md
```diff
@@ -0,0 +1,321 @@
...
+- Expandable Row
+
+## Usage
+
+### Basic usage
+#### Render static tabular data.
+```javascript
+import * as React from 'react';
+import {Table} from 'baseui/table';
+const data = {
```
CharlieMartell: I take it this is supposed to be an array of objects? And should it not be an [[], []] for rows?

rfcs/table.md
```diff
@@ -0,0 +1,321 @@
+# Table Component
+
+Represents control to render tabular data.
+
+## Features
+- Sorting
+- Basic mobile responsiveness (overflow scrollbar)
+
+
+## Unsupported
```
CharlieMartell: Is this a list of *Unsupported Currently* but with plans to or *We do not plan on supporting* ?

rfcs/table.md
dswho: I think your suggestion of using the nested array data structure  `[[],[]]` is another possibility and is commonly seen for handling csv data where the first row is a row of headers and the following rows are rows of text data.

An example would be:
```
[
  [ 'Emp_ID', 'First', 'Last']
  [1, 'John', 'Smith']
  [2, 'Jane', 'Doe']
]
```

The design I have suggested is inspired by ant design's table https://ant.design/components/table/#How-To-Use and one of our projects also has a similar implementation

CharlieMartell: Gotcha yeah I see, but this should still be 
```
[{}, {}]
```

rfcs/table.md
```diff
@@ -0,0 +1,321 @@
...
+```
+
+#### Custom Table Rendering
+There may be a case where more complex layouts are needed,
+such as columns that with colSpan > 1 and/or rowspan > 1
+```
+export default() => {
+    <Table
+     render: (dataSource, columns) => {
+        return (<Tbody>
```
ericsoco: I'm not sure we actually want to render using HTML tables, given the conversation about using react-virtualized for out-of-the-box performance in this table component. Will defer to @CharlieMartell on this one.

Or is this just an example of completely custom rendering?

ericsoco: Ohh, I see -- these are the names of the components you're exporting from the lib. K nevermind :)
Though I would be careful about using names so close to HTML element names, as consumers will expect the same functionality. If you intend to deviate from the HTML `<table>` spec, you might consider different (and more semantic) names, e.g. `<Header>`, `<Row>`, `<Cell>`.

rfcs/table.md
```diff
@@ -0,0 +1,321 @@
...
+* `data:Array<T>` - Required
+  Description of prop
+* `columns: Column` - Required
+  Description of prop
+* `render: (dataSource, columns) => $ReactNode` - Optional
+    Custom render table body.
+* `overrides - Object:` - Optional
+
+* `overrides(continued):`
+  * `Tbody` - Body of the table. Corresponds.
```
ericsoco: See my comment above about naming these sub-components.

rfcs/table.md
```diff
@@ -0,0 +1,321 @@
...
+  * `TheadTr` - A header row. `Tr` for the `Thead` element.
+  * `Th` - A header cell. Corresponds with the `th` html element.
+  * `Td` - A data cell in the table. Corresponds with the `td` html element.
+  * `Tr` - A data row that contains `Td` components.
+
+
+## `Column` API
+One of the Table columns prop for describing the table's columns, Column has the same API.
+* `title: string` - Required. Title of the column to display in the header.
+* `dataIndex: string` - Optional. Attribute at which to index the row data.
```
ericsoco: Sorry, I don't understand? What do you mean by "index the row data"?

rfcs/table.md
```diff
@@ -0,0 +1,321 @@
...
+* `render: <T>(<T>, index, Array<T>) => $ReactNode` - Optional.	Renderer of the table cell. The return value should be a ReactNode.
+* `sorter:Function`- Optional  Sort function for [local sort](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort)
+* `defaultSortOrder:'ascend'|'descend'`: Only to be declared on one of the columns, this will be define the default sort on the initial set of data.
```
ericsoco: Why not a prop of the Table component? To avoid the ambiguity of `defaultSortOrder` defined on multiple columns.

rfcs/table.md
```diff
@@ -0,0 +1,321 @@
...
+
+## Presentation Components
```
ericsoco: I'm not convinced we want to hew strictly to the HTML table spec, per my comments above. @CharlieMartell thoughts from the `react-virtualized` world?

rfcs/table.md
```diff
@@ -0,0 +1,321 @@
...
+
+
+### Dependencies
+
+Does this component depend on any 3rd party packages or other internal components?
+
+## Accessibility
+
+How can this component be used via keyboard controls?
+No keyboard shortcuts will be enabled.
```
ericsoco: Might be nice to enable highlighting via keyboard? e.g. click in a cell, table gains keyboard focus, move the highlight with arrow keys. I'm ambivalent...also, this would add local state, which we might want to avoid if possible.

rfcs/table.md
```diff
@@ -0,0 +1,321 @@
...
+
+What are the accessibility best practices for this component (aria-\*, role, etc.)
+
+We are making the assuming that in most use cases, a [table with one header](https://www.w3.org/WAI/tutorials/tables/one-header/)
+will be rendered.
+
+The [W3 guideline for table states](https://www.w3.org/WAI/tutorials/tables/):
+> Header cells must be marked up with `<th>`, and data cells with `<td>` to make tables accessible. For more complex tables, explicit associations may be needed using `scope`, `id`, and `headers` attributes.
```
ericsoco: Ahhh...this is certainly a benefit of sticking to the HTML table spec...hm. I do see ARIA roles on `react-virtualized`, at least: https://github.com/bvaughn/react-virtualized/issues/606

CharlieMartell: react-virtualized does add aria-* roles & sets tabIndex to assist with keyboard nav. 
It also provides some additional components ie: ArrowKeyStepper to help with navigating virtualized lists

rfcs/table.md
CharlieMartell: https://bvaughn.github.io/react-virtualized/#/components/ArrowKeyStepper
Provided out of the box here

rfcs/table.md
CharlieMartell: Agree, depends on if @dswho goes down react-virtualized path

kahwee: Consider https://github.com/bvaughn/react-window too :)

rfcs/table.md
```diff
@@ -0,0 +1,321 @@
+# Table Component
```
blackswanny: all RFC files have `-component` suffix

rfcs/table.md
dswho: A row of data would be an object literal such as `{ firstName: 'John', lastName: 'Smith', id: 1}`.
`dataIndex` in this would be one of `firstName`, `lastName` or `id`.
For `index the row data`, I mean access the row object with a certain attribute.

rfcs/table.md
dswho: Thanks for the suggestion. I'll add it to the RFC

rfcs/table.md
dswho: Yeah that's a good suggestion, in someone wants to display custom html in the header.

rfcs/table.md
dswho: I originally included StatefulTable, but since I wasn't able to find a need for it yet, I'll remove it until there is a need.

rfcs/table.md
dswho: I would change this to _Unsupported Currently_ because I don't want to overpromise but underdeliver any features. Some features like Fixed Header and  Fixed Column which might be tricky to implement and my team doesn't need this.

rfcs/table.md
chasestarr: landed on semantic naming based on conversation today 👍

rfcs/table.md
chasestarr: we landed on tabular format

chasestarr:
closing this to continue conversation in https://github.com/uber-web/baseui/pull/632

PR #460, opened 2018-11-07
fix(fonts): ubermove scale updates

gergelyke:


src/themes/light-theme-with-move.js
```diff
@@ -0,0 +1,101 @@
...
+        fontFamily: primaryFontFamily,
+        fontSize: '12px',
+        fontWeight: 'normal',
+        lineHeight: '20px',
+      },
+      font250: {
+        fontFamily: primaryFontFamily,
+        fontSize: '12px',
+        fontWeight: 'bold',
+        lineHeight: '20px',
```
nadiia: these are the same as in the default theme, no need to override

src/themes/light-theme-with-move.js
```diff
@@ -0,0 +1,101 @@
...
+      },
+      font400: {
+        fontFamily: primaryFontFamily,
+        fontSize: '16px',
+        fontWeight: 'normal',
+        lineHeight: '24px',
+      },
+      font425: {
```
nadiia: seems like this one should be the 450 font, and 20/28 - 500

src/themes/light-theme-with-move.js
gergelyke: yeah, I agree - the reason I still added them here is for clarity, to make sure who wants to quickly understand what fonts the UberMoveTheme is using will know instantly

does it make sense?

nadiia: yeah, makes sense

src/themes/light-theme-with-move.js
```diff
@@ -0,0 +1,104 @@
...
+      },
+      font900: {
+        fontFamily: primaryFontFamily,
+        fontSize: '52px',
+        fontWeight: 'bold',
+        lineHeight: '68px',
+      },
+      font1000: {
+        fontFamily: primaryFontFamily,
+        fontSize: '92px',
```
nadiia: this should be `72px`

src/themes/light-theme-with-move.js
```diff
@@ -0,0 +1,104 @@
...
+const primaryFontFamily =
+  'UberMoveText, "Open Sans", "Helvetica Neue", Helvetica, sans-serif';
+
+const secondaryFontFamily =
+  'UberMove, UberMoveText, "Open Sans", "Helvetica Neue", Helvetica, sans-serif';
+
+export const LightThemeMove = createTheme(
+  {
+    ...lightThemePrimitives,
+    primaryFontFamily,
```
nadiia: do we need to pass `secondaryFontFamily ` in here?

gergelyke: we don't - the factory has no concept of the `secondaryFontFamily`

PR #461, opened 2018-11-07
feat(card): add StyledActionLink and improved margin for the body

gergelyke:
as per design review:

- updated the margin for the body
- introduced StyledActionLink to implement design's ask on the look and feel for the CTA for cards in a non-breaking semver manner

### DIFF
--- a/rfcs/card-component.md
+++ b/rfcs/card-component.md
@@ -15,11 +15,11 @@ Props based usage:
 
 ```javascript
 import * as React from 'react';
-import {Card} from 'baseui/card';
+import {Card, StyledActionLink} from 'baseui/card';
 
 export default () => (
   <Card
-   action={<a href="#">Link to a Place&nbsp;&nbsp;&nbsp;&gt;</a>}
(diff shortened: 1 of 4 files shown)


src/card/styled-components.js
```diff
@@ -4,6 +4,15 @@ import {styled} from '../styles/index';
 
 export const Action = styled('div', () => ({}));
 
+export const ActionLink = styled('a', ({$theme}) => {
```
nadiia: How about we create a separate `Link` component for it and use it in the Card example? And meanwhile just create a styled link element for an example only purpose instead of a componet's export. Since later getting rid of the link from a card would be a breaking change and people may start using this styled link outside of the card component.

rfcs/card-component.md
```diff
@@ -16,10 +16,11 @@ Props based usage:
 ```javascript
 import {Card} from 'baseui/card';
+import {StyledLink} from 'baseui/link';
```
nadiia: `StyledLink  as StyledActionLink ` or use `StyledLink` name below

PR #463, opened 2018-11-07
fix(tag): use a border-radius of 12px

gergelyke:


src/tag/styled-components.js
```diff
@@ -77,7 +81,7 @@ export const Root = styled('span', props => {
     color: color,
     opacity: $disabled ? '.56' : null,
     borderRadius: $theme.borders.useRoundedCorners
-      ? $theme.borders.radius100
+      ? $theme.borders.radius400
```
nadiia: Could you also add to it
```    
paddingLeft: '12px',
boxSizing: 'border-box',
```
And reduce padding on the `Action` element to make it fit into the container.
The `delete` icon's size could also be adjusted to `10px`.
So with the above it will be close to the Figma design:
screen shot 2018-11-07 at 12 11 47 pm

nadiia: I would also change `Action`'s element type to `span` since the `Root` is an inline element.

gergelyke: @nadiia updated - can you take a look?

nadiia: Looks great, thanks!

PR #467, opened 2018-11-08
chore: add license header eslint rule

austin94:
#### Fixed Issues

There are many files that are currently missing our MIT license header.  This PR introduces a new eslint plugin to assist with header linting and automated fixes.

#### Reasoning

The current license check mechanism, [license-check-and-add](https://www.npmjs.com/package/license-check-and-add), is manually triggered. When ran it seems some of the logic is wiping out some necessary flow annotations and `#!/usr/bin/env node` usages.

Due to the aggressive comment removal from the current mechanism, I would suggest using the more popular, [eslint-plugin-header](https://www.npmjs.com/package/eslint-plugin-header) plugin.

The license check is now ran as a part of the standard `yarn lint` command. It also supports the `eslint --fix` functionality which I have used to apply the missing license to all files with a missing license.

#### Any Dependency Changes

Removed: `license-check-and-add`
Added: `eslint-plugin-header` [MIT LICENSED]

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,6 +1,11 @@
 module.exports = {
   parser: 'babel-eslint',
-  plugins: ['flowtype', 'eslint-plugin-react', 'eslint-plugin-import'],
+  plugins: [
+    'flowtype',
+    'eslint-plugin-react',
+    'eslint-plugin-import',
+    'header',
+  ],
   env: {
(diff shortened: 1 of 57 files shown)


gergelyke:
that makes perfect sense, and yeah, a new PR sounds great for that

PR #468, opened 2018-11-08
feat(accordion): add unit tests for accordion (part3)

nadiia:


src/accordion/__tests__/panel.test.js
```diff
@@ -0,0 +1,89 @@
...
+    };
+    const event = {key: 'Enter'};
+    const wrapper = mount(<Panel {...props}>Content</Panel>);
+    const header = wrapper.find(StyledHeader);
+    header.simulate('keydown', event);
+    expect(props.onChange).toHaveBeenCalledWith({expanded: true});
+    expect(props.onKeyDown).toHaveBeenCalledWith(event);
+  });
+
+  test.only('onChange is called on a Space keydown event', () => {
```
gergelyke: I think you've accidentally left this `only` here

src/accordion/__tests__/panel.test.js
```diff
@@ -0,0 +1,89 @@
...
+    };
+    // $FlowFixMe
+    const wrapper = mount(<Panel {...props}>Content</Panel>);
+    const header = wrapper.find(StyledHeader);
+    header.simulate('click');
+    expect(props.onChange).toHaveBeenCalledWith({expanded: true});
+    expect(props.onClick).toHaveBeenCalled();
+  });
+
+  test('onChange is called on a Ender keydown event', () => {
```
gergelyke: typo, `Enter`?

src/accordion/__tests__/panel.test.js
nadiia: good catch! removed

src/accordion/__tests__/panel.test.js
nadiia: yep, fixed

PR #470, opened 2018-11-08
fix(icons): update to new styles

gergelyke:


package.json
```diff
@@ -33,7 +33,8 @@
     "e2e:test": "nightwatch --filter e2e.js",
     "e2e:test:ci": "nightwatch --filter e2e.js --env chrome-mac,chrome-windows --retries 2",
     "prepublish": "npm run build",
-    "license": "license-check-and-add"
+    "license": "license-check-and-add",
```
austin94: In #467 I removed the `license-check-and-add` dependency, but forgot to remove this NPM script. Could you remove it here?

gergelyke: I'll send a new PR here, just to keep a clean git history

PR #472, opened 2018-11-08
fix(toast): button is left-aligned

gergelyke:


src/toast/examples.js
```diff
@@ -181,17 +181,15 @@ export default {
               <React.Fragment>
                 Info notification with no close button and children as a
                 function.
-                <Centered>
-                  <Space>
```
nadiia: I guess we still want to keep spacing here

PR #474, opened 2018-11-08
fix(icons): size to scale 600

gergelyke:


src/icon/styled-components.js
```diff
@@ -16,7 +16,7 @@ export function getSvgStyles({$theme, $size, $color}: StyledComponentParamsT) {
       $size = `${$size}px`;
     }
   } else {
-    $size = $theme.sizing.scale800;
+    $size = $theme.sizing.scale550;
```
nadiia: How about the default set to `16px`?

gergelyke: updated

PR #481, opened 2018-11-09
feat(block): introduce new Block layout and typography component

austin94:
#### Minor: New Feature

This PR creates a new `Block` component. This component is primarily meant for easy access to themeable values, including: `font`, `color`, `spacing`.

By default it will check the available theming values to see if there is a match, if none is found it will pass the value through as it's new style.

I have also added in some additional positioning and layout properties relating to flexbox to allow some simple layouts.

The `as` and `$style` props allow for more complicated styles to be created.

#### Tests Added + Pass

Without snapshot serialization that includes applied styles, the testing added is limited.

#### License

MIT


package.json
```diff
@@ -33,7 +33,6 @@
     "e2e:test": "nightwatch --filter e2e.js",
     "e2e:test:ci": "nightwatch --filter e2e.js --env chrome-mac,chrome-windows --retries 2",
     "prepublish": "npm run build",
-    "license": "license-check-and-add",
```
austin94: Forgot to remove this in #467

rfcs/block-component.md
```diff
@@ -67,14 +120,3 @@ The reality is you often want to apply multiple of these concerns to one element
 Having multiple elements can complicate things like layout (flexbox children, etc), as well as just needlessly increasing the amount of DOM the client has to deal with.
 
 Combining most things into a single `Block` component allows you to keep things as a single element, and generally has less cognitive overload due to fewer react elements. (maybe there's a happy middle-ground though, for example a `Block` and `Text` component?)
-
```
austin94: Do we normally want to leave sections like this in an RFC after implementation? Willing to leave it if it's normal.

src/block/examples.js
```diff
@@ -0,0 +1,87 @@
...
+      500,
+      600,
+      700,
+      800,
+      900,
+      1000,
+    ];
+    const elements = [];
+
+    for (const size of sizes) {
```
gergelyke: can we simply do a `const elements = sizes.map(...)`?

src/block/styled-components.js
```diff
@@ -0,0 +1,199 @@
...
+
+    if ($font !== undefined) {
+      const font: Font = typography[$font];
+      styles.fontFamily = font.fontFamily;
+      styles.fontWeight = font.fontWeight;
+      styles.fontSize = font.fontSize;
+      styles.lineHeight = font.lineHeight;
+    }
+
+    if ($alignContent) {
```
austin94: @nadiia Open styletron question...

Is it better to just apply these values blindly and let styletron remove them or is it preferred to do an `undefined` as shown here?

src/block/examples.js
austin94: Good catch. I was able to just directly return the `.map()`

rfcs/block-component.md
```diff
@@ -52,6 +52,59 @@ Props would be exposed for:
 
 > **Note** that we would likely _not_ use this component internally to build other baseui components. It's primarily a helper component for customers when building their apps.
```
schnerd: The above content can probably be moved down into some sort of "Motivation" section now, I'd imagine the API docs / usage examples are probably most useful.

rfcs/block-component.md
```diff
@@ -52,6 +52,59 @@ Props would be exposed for:
...
 
+### `Block` API
+
+* `children: (props: Props) => React.Node` - Optional
+* `as: string | React.Node` - Optional
```
schnerd: Should this be React.ElementType instead of React.Node?

rfcs/block-component.md
```diff
@@ -52,6 +52,59 @@ Props would be exposed for:
...
+* `$style: Object` - Optional
+  * Applies provided styles to Styletron
+* `color: String` - Optional
+  * Accepts all themeable color properties (`primary200`, etc.)
+* `font: String` - Optional
+  * Accepts all themeable font properties (`font200`, etc.)
+* `alignContent: 'start' | 'end' | 'center' | 'between' | 'around' | 'stretch'` - Optional
+* `alignItems: 'start' | 'end' | 'center' | 'baseline' | 'stretch'` - Optional
+* `alignSelf: 'auto' | 'start' | 'end' | 'center' | 'baseline' | 'stretch'` - Optional
+* `direction: 'row' | 'column'` - Optional
```
schnerd: Despite it being a couple more keystrokes, I wonder if we should use `flexDirection` to match the css property name as closely as possible (reduces ambiguity and the need to look up Block API to verify)

rfcs/block-component.md
```diff
@@ -52,6 +52,59 @@ Props would be exposed for:
...
+  * Accepts all themeable sizing properties (`scale100`, etc.)
+* `paddingTop: string` - Optional
+  * Accepts all themeable sizing properties (`scale100`, etc.)
+* `paddingRight: string` - Optional
+  * Accepts all themeable sizing properties (`scale100`, etc.)
+* `paddingBottom: string` - Optional
+  * Accepts all themeable sizing properties (`scale100`, etc.)
+* `paddingLeft: string` - Optional
+  * Accepts all themeable sizing properties (`scale100`, etc.)
+* `wrap: boolean` - Optional
```
schnerd: Same here, maybe use `flexWrap`?

rfcs/block-component.md
schnerd: Probably fine to Nuke this, I'd imagine we can also go back to the original PR if we want to look at the related discussion.

src/block/__tests__/block.spec.js
```diff
@@ -0,0 +1,25 @@
...
+*/
+
+// @flow
+import * as React from 'react';
+import {mount} from 'enzyme';
+import Block from '../block';
+
+describe('Block', () => {
+  it('renders correctly', () => {
+    expect(mount(<Block color="primary200">test</Block>)).toMatchSnapshot();
```
schnerd: So technically we do have some ways to do style snapshots, `styled` is mocked in tests:
https://github.com/uber-web/baseui/blob/master/src/styles/__mocks__/styled.js

and we have a `toHaveStyleRule` custom just matcher that leverages this:
https://github.com/uber-web/baseui/blob/master/src/test/expect-to-have-style-rule.js

that said, it's meant to be called directly on a styled component, whereas what you have here is a wrapper around a styled component. You might have to assert styles on the first child (StyledBlock), or try using enzyme shallow with dive().

rfcs/block-component.md
```diff
@@ -52,6 +52,59 @@ Props would be exposed for:
...
+
+* `children: (props: Props) => React.Node` - Optional
+* `as: string | React.Node` - Optional
+  * Modifies the base element used to render the block
+* `$style: Object` - Optional
+  * Applies provided styles to Styletron
+* `color: String` - Optional
+  * Accepts all themeable color properties (`primary200`, etc.)
+* `font: String` - Optional
+  * Accepts all themeable font properties (`font200`, etc.)
```
schnerd: At the expense of adding more to this API, I wonder if we also want to expose props for setting fontSize, fontStyle, fontWeight, lineHeight independently. I suppose people can always do this using `$style`, but I'd imagine it'll be a pretty common use case.

rfcs/block-component.md
austin94: I've reordered the page and removed some of the unnecessary content.

rfcs/block-component.md
austin94: From the [Flow & React docs](https://github.com/facebook/flow/blob/master/website/en/docs/react/types.md#reactnode-) it looks like it's the common approach for typing `children`.  It also looks like this is the common approach in the other BaseUI components.

I'm more familiar with TypeScript honestly, so maybe @nadiia or @gergelyke could provide some Flow insight here?

schnerd: Yeah, React.Node is usually what you want for `children` but I think what we have here is a different than children. `<div>` (or its transpiled version `React.createElement('div')`) is a React.Node, but styletron wouldn't be able to do anything with that. It just needs the string or the class, which it will pass to React.createElement internally.

Looks like styletron also uses ElementType internally:
https://github.com/styletron/styletron/blob/master/packages/styletron-react-core/src/index.js#L181

schnerd:
> Looks great. One concern around $style and as props - seems to be very Styletron specific. 

They do resemble styletron, though hopefully if we were ever to switch to some other library we'd be able to remap them internally. 

> One thought for additional styles is we can use same overrides approach but may seem a little too verbose. @schnerd what do you think about passing additional styles through overrides?

I'm on the fence, consistency with other components is nice but I agree it would be quite verbose in this case. This component is a bit of an edge case since it's meant to just represent a single element, so maybe it's okay to diverge a bit here and go with the simpler `style` API.

src/block/styled-components.js
```diff
@@ -0,0 +1,199 @@
...
+    if ($overflow === 'scrollX') {
+      styles.overflowX = 'scroll';
+    } else if ($overflow === 'scrollY') {
+      styles.overflowY = 'scroll';
+    } else if ($overflow !== undefined) {
+      styles.overflow = $overflow;
+    }
+
+    if ($margin) {
+      styles.margin = sizing[$margin] || $margin;
```
nadiia: Should probably map the shorthand `margin`| `padding` to its longhand props since shorthand will not override longhand with Styletron and vice versa

nadiia:
> > Looks great. One concern around $style and as props - seems to be very Styletron specific.
> 
> They do resemble styletron, though hopefully if we were ever to switch to some other library we'd be able to remap them internally.
> 
> > One thought for additional styles is we can use same overrides approach but may seem a little too verbose. @schnerd what do you think about passing additional styles through overrides?
> 
> I'm on the fence, consistency with other components is nice but I agree it would be quite verbose in this case. This component is a bit of an edge case since it's meant to just represent a single element, so maybe it's okay to diverge a bit here and go with the simpler `style` API.

Still have a doubt regarding `inconsistent API` but I'm on board with `style`.

rfcs/block-component.md
austin94: 🤦‍♂️ I was only looking at `children`. Just updated `as` to ElementType

rfcs/block-component.md
austin94: Updated to `flexDirection`

rfcs/block-component.md
austin94: Updated to `flexWrap`

rfcs/block-component.md
austin94: I would lean toward allowing people to use `$style` if they need individual overrides.

Since the `fontXXX` theming variables combine all of those values into a single token I think we might want to keep them together for consistency.

src/block/__tests__/block.spec.js
```diff
@@ -0,0 +1,25 @@
+/*
```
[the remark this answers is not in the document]
austin94: Fixed.

src/block/__tests__/block.spec.js
austin94: Thanks for the references. I was able to use `getStyle()` to reach 100% coverage

Should be good now.

rfcs/block-component.md
```diff
@@ -31,22 +27,68 @@ The `Block` component is a helper component that lets you write these common les
...
-* Display, Position, Flexbox
-* Box shadows
-* `$style` prop for further style overrides
+### `Block` API
+
+* `children: (props: Props) => React.Node` - Optional
+* `as: string | React.Node` - Optional
+  * Modifies the base element used to render the block
+* `overrides: {Block}` - Optional
+  * `Block: ReactComponent` - Optional
```
nadiia: * `Block: {style: {} | () => {}}` - Optional

src/block/styled-components.js
```diff
@@ -0,0 +1,199 @@
...
+
+    if ($justifySelf) {
+      styles.justifySelf = $justifySelf;
+    }
+
+    if ($position) {
+      styles.position = $position;
+    }
+
+    if ($width) {
```
nadiia: `width`, `min|maxWidth` and `height`, `min|maxHeight` can be set to `0`

src/block/styled-components.js
```diff
@@ -0,0 +1,199 @@
...
+
+    if ($paddingRight) {
+      styles.paddingRight = sizing[$paddingRight] || $paddingRight;
+    }
+
+    if ($paddingBottom) {
+      styles.paddingBottom = sizing[$paddingBottom] || $paddingBottom;
+    }
+
+    if ($paddingLeft) {
```
nadiia: same with margins and padding - can be set to `0` and won't pass the condition

src/block/styled-components.js
```diff
@@ -0,0 +1,199 @@
...
+
+    if ($top) {
+      styles.top = $top;
+    }
+
+    if ($right) {
+      styles.right = $right;
+    }
+
+    if ($bottom) {
```
nadiia: same for the positioning values

rfcs/block-component.md
austin94: Fixed

src/block/styled-components.js
austin94: I was actually checking `undefined` above, but stopped for some reason. I'm now doing that for all values.

src/block/styled-components.js
austin94: Resolved above

PR #483, opened 2018-11-09
docs(readme): fix themes code example

AEgan:
The file where these primitives are defined changed in #460. This updates the themes README to point to the correct file, and updates the code example to correct usage


gergelyke:
thanks for the contribution @AEgan 👏

PR #484, opened 2018-11-09
feat(vrt): add basic screener integration

gergelyke:


screener.config.js
```diff
@@ -0,0 +1,39 @@
...
+    /Input with tags/,
+    /Stateless Menu/,
+    /Modal$/,
+    /Stateful Pagination/,
+    /stateless popover/,
+    /Radio Group example/,
+    /Multi-Select/,
+    /Simple slider example/,
+    /Textarea state/,
+    /Toast notifications/,
```
nadiia: Are these names of stories? Can we use `examples-list` object here?

gergelyke: that's a great call, sure we can - let me update it

PR #494, opened 2018-11-13
chore: upgrade flow versions

austin94:
#### Fixed Issues

This PR upgrades `flow-bin` from version `0.84` to `0.86`. There were several flow related changes necessary to pass the flow checker.

#### Tests Pass

Yes

#### Any Dependency Changes

* Flow `v0.84` -> `v0.86`

#### License

MIT


src/styles/styled.js
```diff
@@ -35,6 +35,8 @@ export default function styledWrapper(...args: any) {
   // Ex: <StyledDiv $style={{color: 'red'}} />
   // Issue for supporting this natively in styletron:
   // https://github.com/rtsao/styletron/issues/221
+
+  // $FlowFixMe
```
austin94: The `$FlowFixMe` isn't preferable here, but with the complexity of the nested structure I wasn't able to correct the typings here.

PR #514, opened 2018-11-15
add mini RFC for header navigation

blackswanny:
add mini RFC for header navigation
image

### DIFF
--- a/rfcs/header-navigation-component.md
+++ b/rfcs/header-navigation-component.md
@@ -1,0 +1,82 @@
+# Header Navigation Component
+
+## Usage
+
+### Basic usage
+
+```javascript
+import * as React from 'react';
+import {HeaderNavigation, NavigationItem, NavigationList} from 'baseui/header-navigation';
+import {Button, KIND} from 'baseui/button';
+
+export default () => <HeaderNavigation>
+  <NavigationList align="left">
+    <NavigationItem><span>Uber</span></NavigationItem>
+    <NavigationItem><a href="/a">Link One</a></NavigationItem>
+    <NavigationItem><a href="/b">Link Two</a></NavigationItem>
+  </NavigationList>
+  <NavigationList align="right">
+    <NavigationItem><Button onClick={() => console.log('clicked')} kind={KIND.primary}>Get started</Button></NavigationItem>
+  </NavigationList>
+</HeaderNavigation>;
+```
+
+### Advanced usage
+
+```javascript
+import * as React from 'react';
+import {HeaderNavigation, NavigationItem, NavigationList} from 'baseui/header-navigation';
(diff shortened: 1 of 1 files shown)


rfcs/header-navigation-component.md
```diff
@@ -0,0 +1,56 @@
...
+```
+
+## Exports
+
+* `HeaderNavigation`
+* `StyledRoot`
+
+## `HeaderNavigation` API
+
+* `mainElement: React$Node` - Optional
```
austin94: Is `mainElement` too generic? If it's always going to be a logo/wordmark maybe just naming it `logo`?

blackswanny: I thought about it. By default it's just a logo. However makes sense to put here some hamburger menu or other active element of header as it usually happens
@nguyensomniac what do you think?

blackswanny: homeElement sounds better

rfcs/header-navigation-component.md
```diff
@@ -0,0 +1,56 @@
...
+### Advanced usage
+
+```javascript
+import * as React from 'react';
+import {HeaderNavigation} from 'baseui/header-navigation';
+import {Button, KIND} from 'baseui/button';
+import {StatefulMenu as Menu} from 'baseui/menu';
+const ITEMS = [{label: 'menu item 1'}, {label: 'menu item 2'}];
+
+export default () => <HeaderNavigation homeElement={<Menu items={ITEMS} />}
```
schnerd: Using `Menu` here directly seems like it would just render a broken or janky header (unless I'm missing something), maybe we should use another example so users aren't confused?

rfcs/header-navigation-component.md
```diff
@@ -0,0 +1,56 @@
...
+
+* `HeaderNavigation`
+* `StyledRoot`
+
+## `HeaderNavigation` API
+
+* `homeElement: React$Node` - Optional
+  If present, is shown left docked by default styles.
+* `children: React$Node` - Required.
+  All the children (except `homeElement`) of header navigation. By default docked to the right side.
```
schnerd: Do we want to try to support more than just navigation elements docked to the right? This PR comment from a while back has a possible alternative API:
https://github.com/uber-web/baseui/pull/330#issuecomment-426836888

Offering `NavigationItem` makes it easy to use the right margin between navigation elements. In the example code for this PR, it seems like all the nav elements would be packed together with no spacing, so customers will need to write their own spacing styles anyway.

blackswanny: I like it

rfcs/header-navigation-component.md
blackswanny: Yes, it won't work like this. Need another example

PR #519, opened 2018-11-16
fix(toast): fix animate in

schnerd:
As of right now, `Toast`s do not animate in – they only animate out. This PR updates the logic so that they animate in as well.

kapture 2018-11-16 at 15 09 17

#### Patch: Bug Fix?

Yes

#### Major: Breaking Change?

No

#### Minor: New Feature?

No

#### Tests Added + Pass?

Yes

### DIFF
--- a/src/toast/styled-components.js
+++ b/src/toast/styled-components.js
@@ -82,12 +82,12 @@ export const Root = styled('div', (props: ToasterSharedStylePropsT) => {
 });
 
 export const Body = styled('div', (props: SharedStylePropsT) => {
-  const {$isHidden, $isAnimating, $kind, $theme} = props;
+  const {$isVisible, $kind, $theme} = props;
   return {
     ...$theme.typography.font300,
     pointerEvents: 'auto',
     color: $theme.colors.white,
-    height: $isHidden && !$isAnimating ? 0 : 'auto',
+    height: 'auto',
     width: '288px',
     paddingTop: $theme.sizing.scale600,
     paddingRight: $theme.sizing.scale600,
@@ -101,7 +101,7 @@ export const Body = styled('div', (props: SharedStylePropsT) => {
       ? $theme.borders.radius200
       : '0px',
     boxShadow: $theme.lighting.shadow600,
-    opacity: $isHidden ? 0 : 1,
+    opacity: $isVisible ? 1 : 0,
     transitionProperty: 'all',
     transitionDuration: $theme.animation.timing100,
     transitionTimingFunction: $theme.animation.easeInOutCurve,
--- a/src/toast/toast.js
+++ b/src/toast/toast.js
@@ -37,11 +37,12 @@ class Toast extends React.Component<ToastPropsT, ToastPrivateStateT> {
   };
 
   autoHideTimeout: ?TimeoutID;
+  animateInTimer: ?TimeoutID;
   animateOutCompleteTimer: ?TimeoutID;
 
   state = {
-    isAnimating: false,
-    isHidden: true,
+    isVisible: false,
+    isRendered: true,
   };
 
   componentDidMount() {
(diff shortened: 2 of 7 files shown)


src/toast/toast.js
```diff
@@ -37,11 +37,12 @@ class Toast extends React.Component<ToastPropsT, ToastPrivateStateT> {
...
 
   autoHideTimeout: ?TimeoutID;
+  animateInTimer: ?TimeoutID;
   animateOutCompleteTimer: ?TimeoutID;
 
   state = {
-    isAnimating: false,
-    isHidden: true,
+    isVisible: false,
+    isRendered: true,
```
schnerd: These names felt a little bit clearer. `isRendered` controls whether or not it's actually mounted in the DOM, `isVisible` controls whether it's shown.

src/toast/toast.js
```diff
@@ -63,31 +64,35 @@ class Toast extends React.Component<ToastPropsT, ToastPrivateStateT> {
...
       }
     });
   }
 
   animateIn = () => {
-    this.setState({isHidden: false, isAnimating: true});
+    // Defer to next event loop
+    this.animateInTimer = setTimeout(() => {
+      this.setState({isVisible: true});
+    }, 0);
```
schnerd: Strangely this wasn't working without this setTimeout, even though animateIn should only be called after component is already mounting.

src/toast/styled-components.js
```diff
@@ -82,12 +82,12 @@ export const Root = styled('div', (props: ToasterSharedStylePropsT) => {
...
 
 export const Body = styled('div', (props: SharedStylePropsT) => {
-  const {$isHidden, $isAnimating, $kind, $theme} = props;
+  const {$isVisible, $kind, $theme} = props;
   return {
     ...$theme.typography.font300,
     pointerEvents: 'auto',
     color: $theme.colors.white,
-    height: $isHidden && !$isAnimating ? 0 : 'auto',
+    height: 'auto',
```
schnerd: Animating between `0` and `auto` doesn't actually work, you can only animate between two literal values., so this wasn't actually do anything.

blackswanny: but `opacity` animation is different from `height`. Which one is supposed to be? I think, for toast it has to be `height`, so should you try '100%' instead of `auto`, but in this case we need to reconsider outer container of it

src/toast/toast.js
blackswanny: animate in and animate out do pretty much the same, except operate with opposite values of state. Should we combine them into one animate function and even remove one of timers ` this.animateInTimer, this.animateOutCompleteTimer` ?

src/toast/toast.js
nadiia: `isHidden` and `isAnimating` were introduced in order to handle height animation in addition to the opacity so the rest of notifications shifted in a smooth way when a single notification from a list of stacked notifications is dismissed

nadiia: Still seems to be slightly confusing since we set `isRendered` to true as an initial state that doen't mean component is mounted.

schnerd: > isHidden and isAnimating were introduced in order to handle height animation in addition to the opacity so the rest of notifications shifted in a smooth way when a single notification from a list of stacked notifications is dismissed

Maybe I'm missing something, but height animation didn't work at all so I sort of just nuked that functionality (see later comment)

> Still seems to be slightly confusing since we set isRendered to true as an initial state that doesn't mean component is mounted.

Yeah that's sort of why I chose the name `isRendered` instead of `isMounted`, it's meant to signify that render method should return something (not `null`).

src/toast/styled-components.js
schnerd: `100%` would work, but as you pointed out you need to also have a container with a fixed height for that to work, so I'm not sure that's a viable solution here. 

If I recall correctly the only reiable way to animate an element from 0 height to its "auto" height is to first render it hidden, use `ref.current.clientHeight` or `ref.current.getBoundingClientRect().height`, and then make it visible and transition from 0 to that explicit height value.

gergelyke:
@schnerd I've just added you to the screener project - if it breaks, and it feels like an issue with screener, ping me (like a state that's already on master, but for some reason fails for your branch - as I see, this happened for you), or if it is a change in a component or a new component, you can ping Jeff on the design team

src/toast/styled-components.js
blackswanny: so we have to render it to know exact height. So some day I sued the next trick: I rendered it inside of fake container to get height and than used it in real DOM. Look at this below. It shouldn't take much of performance.

https://www.npmjs.com/package/react-rendered-size
https://github.com/planttheidea/react-rendered-size/blob/master/src/element.js#L16

PR #520, opened 2018-11-17
feat(select): rethink select component implementation

nadiia:


schnerd:
Left a few minor comments. Behavior in storybook feels much smoother, nice work there!

src/select/dropdown.js
```diff
@@ -6,121 +6,116 @@ LICENSE file in the root directory of this source tree.
...
+    const [OptionContent, optionContentProps] = getOverrides(
+      overrides.OptionContent,
+      StyledOptionContent,
+    );
+
+    let $selected;
+    if (Array.isArray(value)) {
+      $selected = value.find(
+        selected => selected && selected[valueKey] === option[valueKey],
+      );
```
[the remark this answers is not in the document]
nadiia: It will, but on the line `61` it's turned into a `boolean`. But I do agree that it's not that nice to have different types assigned to the `$selected`. Changed it here.

src/select/dropdown.js
```diff
@@ -6,121 +6,116 @@ LICENSE file in the root directory of this source tree.
...
+            List: {
+              style: ({$theme: {sizing}}) => ({
+                maxHeight: maxDropdownHeight,
+                marginTop: sizing.scale300,
+                marginBottom: '0',
+                marginLeft: sizing.scale300,
+                marginRight: sizing.scale300,
+              }),
+              props: {
+                role: 'listbox',
```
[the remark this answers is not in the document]
nadiia: Yep, good catch!

src/select/utils/default-filter-options.js
```diff
@@ -0,0 +1,89 @@
...
+  trimFilter: true,
+  valueKey: 'value',
+};
+
+const filterOptions = (
+  options: ValueT,
+  filterValue: string,
+  excludeOptions: ValueT,
+  newProps: $Shape<typeof defaultProps>,
+) => {
```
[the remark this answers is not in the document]
nadiia: Forgot to export it for reuse. But it's used as the default filter function. 
Thought that providing a way to override some of those options would make it easier to tweak without recreating the filter function.

src/menu/styled-components.js
```diff
@@ -36,6 +42,7 @@ export const ListItem = styled(
     display: 'block',
     color: $isHighlighted ? $theme.colors.primary : $theme.colors.black,
     cursor: 'pointer',
+    backgroundColor: $isHighlighted ? $theme.colors.mono300 : 'transparent',
```
blackswanny: by default `initial` is transparent, but if user overrides it, this could cause wrong styling
https://www.w3schools.com/cssref/pr_background-color.asp

src/select/autosize-input.js
```diff
@@ -0,0 +1,86 @@
...
+  };
+  updateInputWidth() {
+    if (
+      !this.mounted ||
+      !this.sizer ||
+      typeof this.sizer.scrollWidth === 'undefined'
+    ) {
+      return;
+    }
+    const newInputWidth = this.sizer.scrollWidth + 2;
```
blackswanny: what is `2` stands for? Just to invoke update?

src/select/examples.js
```diff
@@ -11,159 +11,222 @@ LICENSE file in the root directory of this source tree.
...
+  [tests.SELECT_MULTI]: () => {
     return (
       <StatefulSelect
-        options={COLORS}
-        getOptionLabel={option => option.id}
-        multiple
+        {...options}
         initialState={{
-          selectedOptions: [COLORS[1], COLORS[3]],
+          value: [{...COLORS[0], clearableValue: false}, COLORS[1], COLORS[3]],
```
blackswanny: Do we use it somewhere?
`clearableValue`

src/select/types.js
```diff
@@ -12,64 +12,65 @@ import type {OnItemSelectFnT} from '../menu/types';
...
 
-export type ParamsT = {
-  option?: OptionT,
-  selectedOptions?: Array<OptionT>,
-  textValue?: string,
-  type: ChangeActionT,
-};
+export type ValueT = Array<OptionT> | $Shape<OptionT>;
+
+// export type ParamsT = {
```
blackswanny: these commented probably should be removed

src/select/types.js
```diff
@@ -12,64 +12,65 @@ import type {OnItemSelectFnT} from '../menu/types';
...
+  DropDownOption?: OverrideT<*>,
+  OptionValue?: OverrideT<*>,
 };
 
 export type PropsT = {
-  options: Array<OptionT> | ((query?: *) => Promise<Array<OptionT>>),
+  autoFocus: boolean,
+  options: Array<OptionT>,
   overrides?: OverridesT,
   selectedOptions: Array<OptionT>,
```
blackswanny: is this still needed?

src/select/dropdown.js
```diff
@@ -6,121 +6,116 @@ LICENSE file in the root directory of this source tree.
...
+  getItemLabel = (option: {[string]: any}) => {
+    const {getOptionLabel, overrides = {}, value, valueKey} = this.props;
+
+    const [OptionContent, optionContentProps] = getOverrides(
+      overrides.OptionContent,
+      StyledOptionContent,
+    );
+
+    let $selected;
+    if (Array.isArray(value)) {
```
blackswanny: can we have either array or object for `value` ?

src/select/autosize-input.js
nadiia: No, it's just some extra little space around input text.

src/select/examples.js
nadiia: Yes, check out the selected value functionality.

src/select/dropdown.js
nadiia: Yes, an object can be passed in a single select case.

blackswanny: previously it was just one type of Array. Now we have to support logic to check if it's Array or not. I found a few places, which we need to maintain to make sure it works. I would rather leave Array. What is the benefit of having multiple types of selected value?

src/select/autosize-input.js
```diff
@@ -0,0 +1,86 @@
...
+      },
+    );
+    const componentInputProps = {
+      ...rest,
+      $width: `${this.state.inputWidth}px`,
+    };
+    return (
+      <React.Fragment>
+        <Input {...componentInputProps} $ref={this.inputRef} {...inputProps} />
+        {/* a hidden helper element to calculate the size os the input */}
```
blackswanny: of

src/select/dropdown.js
```diff
@@ -37,7 +37,7 @@ export default class SelectDropDown extends React.Component<DropDownPropsT> {
     const {getOptionLabel, overrides = {}, value, valueKey} = this.props;
 
     const [OptionContent, optionContentProps] = getOverrides(
-      overrides.OptionContent,
+      overrides.OptionContent || {},
```
blackswanny: we usually don't need this with default props set to `{}`

src/select/examples.js
```diff
@@ -64,7 +64,8 @@ function CustomOptionLabel({
   option,
   showColor = false,
 }: {
-  option: {id: string, color: string},
+  // eslint-disable-next-line flowtype/no-weak-types
+  option: any,
```
blackswanny: can we describe all properties of `option` or it became too customizable?

PR #525, opened 2018-11-20
feat(header-navigation): add implementation

blackswanny:
add header navigation implementation

Header navigation is highly customizable component. All examples provide different ways customers can address their design and concerns. They mainly show flexibility of Base UI rather than component itself.

### DIFF
--- a/src/header-navigation/examples-list.js
+++ b/src/header-navigation/examples-list.js
@@ -13,4 +13,5 @@ module.exports = {
   HEADER_NAVIGATION_WITH_OVERRIDES: 'With changed styles',
   HEADER_NAVIGATION_DIFFERENT_ALIGNMENTS: 'With alignments',
   HEADER_NAVIGATION_SEARCH_BAR: 'With search bar',
+  HEADER_NAVIGATION_STYLED_NAV_LIST: 'With different styled Navigstion List',
 };
--- a/src/header-navigation/examples-list.js
+++ b/src/header-navigation/examples-list.js
@@ -1,0 +1,16 @@
+/*
+Copyright (c) 2018 Uber Technologies, Inc.
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+
+/* eslint-env node */
+/* eslint-disable flowtype/require-valid-file-annotation */
+module.exports = {
+  HEADER_NAVIGATION_WITH_MENU_ELEMENT: 'With Dropdown Menu',
+  HEADER_NAVIGATION_WITH_RIGHT_MENU: 'With Right Menu',
+  HEADER_NAVIGATION_WITH_OVERRIDES: 'With changed styles',
+  HEADER_NAVIGATION_DIFFERENT_ALIGNMENTS: 'With alignments',
+  HEADER_NAVIGATION_SEARCH_BAR: 'With search bar',
+};
--- a/src/header-navigation/examples.js
+++ b/src/header-navigation/examples.js
@@ -16,7 +16,12 @@ import {styled} from '../styles';
 import COLORS from '../select/examples-colors';
 import {HamburgerIcon} from './examples-icons';
 
-import {HeaderNavigation, ALIGN, NavigationItem, NavigationList} from './index';
+import {
+  HeaderNavigation,
+  ALIGN,
+  StyledNavigationItem as NavigationItem,
+  StyledNavigationList as NavigationList,
+} from './index';
 import tests from './examples-list';
 import {TYPE} from '../select';
 
@@ -26,13 +31,33 @@ const Link = withProps(
   })),
   {tabIndex: '0'},
 );
+// $FlowFixMe
+Link.displayName = 'Link';
+
 const Hamburger = styled('div', props => ({
   lineHeight: 'initial',
 }));
+Hamburger.displayName = 'Hamburger';
+
+const StyledHamburgerIcon = styled('span', props => ({
+  ':after': {
+    content: `url('data:image/svg+xml;utf8,${HamburgerIcon(
+      props.$theme.colors.foreground,
+    )}');`,
+  },
+}));
+
 let ITEMS = [];
 for (let i = 0; i < 100; i++) {
   ITEMS.push({label: `Item ${i}`});
 }
+const options = {
+  options: COLORS,
+  labelKey: 'id',
+  valueKey: 'color',
+  placeholder: 'Choose a color',
+  maxDropdownHeight: '300px',
+};
 class MenuContainer extends React.Component<
   {isOpen: boolean, align: string},
   {isOpen: boolean},
@@ -66,7 +91,7 @@ class MenuContainer extends React.Component<
           }}
           onClick={() => this.toggleMenu()}
         >
-          {HamburgerIcon()}
+          <StyledHamburgerIcon />
         </Hamburger>
         {this.state.isOpen && (
           <Menu
@@ -94,14 +119,14 @@ export default {
       <React.Fragment>
         <div style={{marginTop: '50px'}} />
         <HeaderNavigation>
-          <NavigationList $align={ALIGN.left}>
+          <NavigationList align={ALIGN.left}>
             <NavigationItem>
               <MenuContainer />
             </NavigationItem>
             <NavigationItem>Uber</NavigationItem>
           </NavigationList>
-          <NavigationList $align={ALIGN.center} />
-          <NavigationList $align={ALIGN.right}>
+          <NavigationList align={ALIGN.center} />
+          <NavigationList align={ALIGN.right}>
             <NavigationItem>
               <Link>Tab Link One</Link>
             </NavigationItem>
@@ -112,7 +137,7 @@ export default {
               <Link>Tab Link Three</Link>
             </NavigationItem>
           </NavigationList>
-          <NavigationList $align={ALIGN.right}>
+          <NavigationList align={ALIGN.right}>
             <NavigationItem>
               <Button>Get started</Button>
             </NavigationItem>
@@ -126,7 +151,7 @@ export default {
       <React.Fragment>
         <div style={{marginTop: '50px'}} />
         <HeaderNavigation>
-          <NavigationList $align={ALIGN.center}>
+          <NavigationList align={ALIGN.center}>
             <NavigationItem>
               <Link>Tab Link One</Link>
             </NavigationItem>
@@ -137,7 +162,7 @@ export default {
               <Link>Tab Link Three</Link>
             </NavigationItem>
           </NavigationList>
-          <NavigationList $align={ALIGN.right}>
+          <NavigationList align={ALIGN.right}>
             <NavigationItem>Uber</NavigationItem>
             <NavigationItem>
               <MenuContainer align={ALIGN.right} />
@@ -154,13 +179,13 @@ export default {
         <HeaderNavigation
           overrides={{Root: {style: {backgroundColor: 'lightblue'}}}}
         >
-          <NavigationList $align={ALIGN.left}>
+          <NavigationList align={ALIGN.left}>
             <NavigationItem>
               <MenuContainer />
             </NavigationItem>
             <NavigationItem>Uber</NavigationItem>
           </NavigationList>
-          <NavigationList $align={ALIGN.right}>
+          <NavigationList align={ALIGN.right}>
             <NavigationItem>
               <Link>Tab Link One</Link>
             </NavigationItem>
@@ -171,7 +196,7 @@ export default {
               <Link>Tab Link Three</Link>
             </NavigationItem>
           </NavigationList>
-          <NavigationList $align={ALIGN.right}>
+          <NavigationList align={ALIGN.right}>
             <NavigationItem>
               <Button>Get started</Button>
             </NavigationItem>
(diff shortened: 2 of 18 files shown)


src/header-navigation/header-navigation.js
```diff
@@ -0,0 +1,25 @@
...
+import type {PropsT} from './types';
+import {Root as StyledRoot} from './styled-components';
+
+class HeaderNavigation extends React.Component<PropsT, {}> {
+  static defaultProps = {
+    overrides: {},
+  };
+
+  render() {
+    const {overrides = {}} = this.props;
```
austin94: Is there a reason to have `{}` as a fallback value if we are defaulting the prop as well?

src/header-navigation/index.js
```diff
@@ -0,0 +1,16 @@
...
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+// @flow
+export {default as HeaderNavigation} from './header-navigation';
+// Styled elements
+export {
+  Root as StyledRoot,
+  NavigationItem,
```
austin94: Should we add `Styled` to the `NavigationItem` and `NavigationList` components for consistency?

gergelyke:
screen shot 2018-11-20 at 9 45 55 am

it seems like the hamburger icon is not in the middle

src/header-navigation/examples-list.js
```diff
@@ -0,0 +1,16 @@
...
+Copyright (c) 2018 Uber Technologies, Inc.
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+
+/* eslint-env node */
+/* eslint-disable flowtype/require-valid-file-annotation */
+module.exports = {
+  WITH_MENU_ELEMENT: 'With Dropdown Menu',
```
gergelyke: can you please add `Header navigation` (or something similar) to the names, so for the visual regression tests it is simpler to include in `screener.config.js`?

src/header-navigation/examples.js
```diff
@@ -0,0 +1,265 @@
...
+    return (
+      <React.Fragment>
+        <span
+          onClick={() =>
+            this.setState({
+              isOpen: !this.state.isOpen,
+            })
+          }
+        >
+          <svg
```
gergelyke: can we move the svg to a file and import it, just to make this example more readable?

gergelyke:
also, for the menu should we use `point` cursor instead of the text?

src/header-navigation/examples-list.js
blackswanny: done

src/header-navigation/examples.js
```diff
@@ -0,0 +1,265 @@
...
+    isOpen: false,
+    align: '',
+  };
+  state = {
+    isOpen: this.props.isOpen,
+  };
+  render() {
+    return (
+      <React.Fragment>
+        <span
```
austin94: For any icon that triggers a menu we should probably have it within the tab order `tabIndex=0` and provide it a human readable label `aria-label="Navigation menu"`

src/header-navigation/examples.js
blackswanny: ok

src/header-navigation/examples.js
austin94: There are also some accessibility issues with this. I see we have the `jsx-a11y` plugin added as a dependency, but it isn't enabled in `eslint`.

@gergelyke is there any specific reason this isn't enabled right now?

blackswanny:
> it seems like the hamburger icon is not in the middle
@gergelyke 
fixed

src/header-navigation/header-navigation.js
blackswanny: I fixed flow types. No need for that now

blackswanny:
> also, for the menu should we use point cursor instead of the text?
@gergelyke 
Menu is used with it's own default styles, however I added cursor for better customization

blackswanny:
> are you planning to add tests in a separate PR?


yes, next commit

src/header-navigation/index.js
blackswanny: As per @schnerd original proposal current names look more semantically right. I can rename it if needed

https://github.com/uber-web/baseui/pull/514#discussion_r234272393

src/header-navigation/examples.js
gergelyke: not that I am aware of - would you mind sending a PR @austin94 that enables it?

blackswanny: @austin94 added code

nadiia:
Tab links in the example are not keyboard navigatable. Could you fix it please?

src/header-navigation/index.js
```diff
@@ -0,0 +1,16 @@
...
+  NavigationList,
```
nadiia: I wonder if we should create container components for both `NavigationItem` and `NavigationList` that will take overrides and render `StyledNavigationItem` and `StyledNavigationList` accordingly. Since this falls outside the pattern we used in other components so far. Same as we do for the `HeaderNavigation`. 
@schnerd what do you think about it?

blackswanny:
> Tab links in the example are not keyboard navigatable. Could you fix it please?

@nadiia fixed

src/header-navigation/examples.js
```diff
@@ -0,0 +1,258 @@
...
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+// @flow
+/* eslint-disable react/display-name*/
+
+import * as React from 'react';
+import {Button} from '../button';
+import {StatefulMenu as Menu, KEY_STRINGS} from '../menu';
```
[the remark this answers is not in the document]
blackswanny: fixed

src/header-navigation/examples.js
```diff
@@ -0,0 +1,258 @@
...
+      isOpen: !this.state.isOpen,
+    });
+  }
+  render() {
+    return (
+      <React.Fragment>
+        <div
+          tabIndex="0"
+          aria-label="Navigation menu"
+          style={{lineHeight: 'initial'}}
```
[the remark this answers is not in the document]
blackswanny: fixed

src/header-navigation/examples.js
```diff
@@ -0,0 +1,258 @@
...
+      <React.Fragment>
+        <div style={{marginTop: '50px'}} />
+        <HeaderNavigation>
+          <NavigationList $align={ALIGN.left}>
+            <NavigationItem>
+              <MenuContainer />
+            </NavigationItem>
+            <NavigationItem>Uber</NavigationItem>
+          </NavigationList>
+          <NavigationList $align={ALIGN.center} />
```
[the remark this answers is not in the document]
blackswanny: So in my flexbox styles if all element are `left` or `right` docked flexbox still docks them together. I had to add this `center` to avoid it. I checked flexbox implementation and didn't find other solution. So there has to be at least one element, that is able to `flexGrow`.
image

image

image

gergelyke:
Can you update the cursor to use the `pointer` style instead of `text`?

gergelyke:
also, can you please add the warning that says that this component is in a beta state, and may change without notice in the near future?

blackswanny:
> Can you update the cursor to use the pointer style instead of text?

updated for the whole header navigation

blackswanny:
> also, can you please add the warning that says that this component is in a beta state, and may change without notice in the near future?


added in `componentDidMount` method for client-side warnings

src/header-navigation/styled-components.js
```diff
@@ -0,0 +1,58 @@
...
+// @flow
+import {styled} from '../styles/index';
+import {ALIGN} from './index';
+
+export const Root = styled('nav', props => {
+  const {$theme} = props;
+  const {
+    sizing: {scale500},
+    typography: {font400},
+    colors: {mono500},
```
gergelyke: instead of depending on exact colors, can you use theme variables like `colors.foreground` to make sure the navigation component will work with the dark theme too?

blackswanny: ok

src/header-navigation/index.js
nadiia: I agree with @austin94 comment, we should be consistent. So far we export any styled component with `Styled` prefix. We should be consistent.

src/header-navigation/index.js
nadiia: If we want to export it as a container we should add a container component to it

src/header-navigation/header-navigation.js
```diff
@@ -0,0 +1,30 @@
...
+import type {PropsT} from './types';
+import {Root as StyledRoot} from './styled-components';
+
+class HeaderNavigation extends React.Component<PropsT, {}> {
+  static defaultProps = {
+    overrides: {},
+  };
+  componentDidMount() {
+    // eslint-disable-next-line no-console
+    console.warn(
```
nadiia: it should only warn in a DEV env

src/header-navigation/examples.js
```diff
@@ -0,0 +1,271 @@
...
+  }
+}
+
+export default {
+  [tests.HEADER_NAVIGATION_WITH_MENU_ELEMENT]: () => {
+    return (
+      <React.Fragment>
+        <div style={{marginTop: '50px'}} />
+        <HeaderNavigation>
+          <NavigationList $align={ALIGN.left}>
```
nadiia: Sorry, I must've missed this one on a previous look through. Why is this prop `$` prefixed? It seems like there should be a container that would do the `$` mapping

blackswanny: this is a styled component. If we add container, it would add complexity to maintain things.

src/header-navigation/header-navigation.js
blackswanny: fixed

gergelyke:
can you please use the theme variables for the hamburger icon too?

screen shot 2018-11-29 at 9 41 46 am

src/header-navigation/examples.js
schnerd: I'm pretty torn on this. It could be confusing to users if some of the primary exports for a component require `$` prefixes. On the other hand I agree creating wrappers every time we use this pattern could be annoying.

A couple ideas:
1. Write an HOC that wraps styled components and maps the $ props:
```js
import {asPrimaryExport} from '../styled/as-primary-export-hoc';

export const StyledNavigationList = styled('div', {...});

export const NavigationList = asPrimaryExport(StyledNavigationList, ['align', 'prop2', 'prop3']);
```
This would be far less tedious than maintaining a wrapper component for every time we use this pattern.

2. Add logic to our `styled` implementation that supports similar prop mapping
```js
export const StyledNavigationList = styled('div', {...}, {
  mapProps: ['align', 'prop2', 'prop3']
});
```
Internally I think this could use styletron's [withWrapper](https://github.com/styletron/styletron/tree/master/packages/styletron-react#withwrapper)? Still not 100% sure. Also adds more complexity to our styled implementation.

--

Obviously those are outside the scope of this specific PR, just trying to think of the best long term solution. Maybe short term we bite the bullet and build a container manually?

blackswanny:
> can you please use the theme variables for the hamburger icon too?

fixed. @gergelyke

src/header-navigation/header-navigation.js
```diff
@@ -0,0 +1,25 @@
...
+
+class HeaderNavigation extends React.Component<PropsT, {}> {
+  static defaultProps = {
+    overrides: {},
+  };
+
+  render() {
+    const {overrides = {}} = this.props;
+    const [Root, rootProps] = getOverrides(overrides.Root, StyledRoot);
+    return <Root {...this.props} {...rootProps} />;
```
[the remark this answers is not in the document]
blackswanny: I think, current code fixed it 
image

chasestarr:

Built with commit 79438de18d4672bc3489d789dcb25a9f41aaa9e9

https://deploy-preview-525--baseui.netlify.com

src/styles/as-primary-export-hoc.js
```diff
@@ -0,0 +1,26 @@
...
+      if (key[0] === '$' || propsTransformNames.indexOf(key) < 0) {
+        acc[key] = props[key];
+      } else if (propsTransformNames.indexOf(key) >= 0) {
+        acc['$' + key] = props[key];
+      }
+      return acc;
+    }, {});
+    return <StyledComponent {...styledProps} />;
+  };
+}
```
schnerd: This is awesome.

src/header-navigation/styled-components.js
```diff
@@ -0,0 +1,61 @@
...
+      },
+      flex: $align === ALIGN.right || $align === ALIGN.left ? 'none' : '1',
+      paddingLeft: scale800,
+      paddingRight: scale800,
+      justifySelf: $align,
+      justifyContent: $align,
+    };
+  }),
+  ['align'],
+);
```
schnerd: It might be nice to also flowtype this component since it's a primary export. 

Flow, might complain that this is incompatible with return type of asPrimaryExport, but you could always cast through any to fix that.

```js
export const NavigationList: React.ComponentType<
  {align: string, children: React.Node}
> = (asPrimaryExport( ... ): any);
```

src/header-navigation/examples.js
nadiia: apparently, my last comment wasn't submitted when I replied

blackswanny: I applied 1) approach. Currently need to fix the test using it

src/header-navigation/index.js
blackswanny: fixed

src/header-navigation/index.js
blackswanny: fixed

src/styles/as-primary-export-hoc.js
blackswanny: cool

src/header-navigation/styled-components.js
blackswanny: fixed. I wonder why flow didn't complain about it

blackswanny:
> Before merging please make sure that the Styled component wrapped in HOC behaves as a styled component in cases of restyling

added example for that

PR #528, opened 2018-11-20
fix: enable jsx-a11y checks in eslint

austin94:
#### Fixed Issues

Even though we had the `jsx-a11y` dependency installed, the eslint rule wasn't enabled for some reason.

This PR enables the recommended linting rules and corrects the failing areas (most of which are in our examples).

Here is an example output for an accessibility issue:

screen shot 2018-11-20 at 2 04 23 pm

#### License

MIT

### DIFF
--- a/src/card/__tests__/card.test.js
+++ b/src/card/__tests__/card.test.js
@@ -12,7 +12,7 @@ import {header as headerImg, thumbnail as thumbnailImg} from '../images';
 
 test('Card - basic functionality', () => {
   const props = {
-    action: <a href="#">Link to a Place&nbsp;&nbsp;&nbsp;&gt;</a>,
+    action: <a href="#test">Link to a Place&nbsp;&nbsp;&nbsp;&gt;</a>,
     headerImage: headerImg,
     thumbnail: thumbnailImg,
     title: 'Card title',
@@ -26,7 +26,7 @@ test('Card - basic functionality', () => {
 
 test('Card - no images', () => {
   const props = {
-    action: <a href="#">Link to a Place&nbsp;&nbsp;&nbsp;&gt;</a>,
+    action: <a href="#test">Link to a Place&nbsp;&nbsp;&nbsp;&gt;</a>,
     title: 'Card title',
   };
 
--- a/src/checkbox/examples.js
+++ b/src/checkbox/examples.js
@@ -263,7 +263,7 @@ export default {
             }
           }}
         />
-        <Checkbox onChange={onChange} inputRef={inputRef} autoFocus={isFocused}>
+        <Checkbox onChange={onChange} inputRef={inputRef}>
           Focused checkbox
         </Checkbox>
       </div>
--- a/src/modal/__tests__/modal.test.js
+++ b/src/modal/__tests__/modal.test.js
@@ -175,6 +175,7 @@ describe('Modal', () => {
 
   test('role', () => {
     wrapper = mount(
+      // eslint-disable-next-line jsx-a11y/aria-role
       <Modal role="mycustomrole" isOpen>
         <ModalBody>Modal Body</ModalBody>
       </Modal>,
--- a/src/textarea/examples.js
+++ b/src/textarea/examples.js
@@ -92,7 +92,6 @@ export default {
         <br />
         <Textarea
           initialState={{value: 'uber'}}
-          autoFocus
           size={SIZE.compact}
           placeholder="Placeholder"
         />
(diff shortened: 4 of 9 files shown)


austin94:
@gergelyke The screener failure is due to the `autoFocus` removals. Should be good to continue.

src/checkbox/examples.js
```diff
@@ -263,7 +263,7 @@ export default {
             }
           }}
         />
-        <Checkbox onChange={onChange} inputRef={inputRef} autoFocus={isFocused}>
+        <Checkbox onChange={onChange} inputRef={inputRef}>
```
blackswanny: so after removing `autoFocus` it doesnt' work as expected in this example

src/modal/__tests__/modal.test.js
```diff
@@ -175,6 +175,7 @@ describe('Modal', () => {
 
   test('role', () => {
     wrapper = mount(
+      // eslint-disable-next-line jsx-a11y/aria-role
```
blackswanny: what's the issue in here?

blackswanny:
we have accessibility check in e2e tests, which is based on `aXe` core rules.
https://github.com/dequelabs/axe-core
But to make it work we need to write test for e2e test, which not always happens.
`aXe` rules considered more standard than those in this react plugin. So as we see this plugin complains on somethins, which aXe allows to pass through and maybe we should just remove it instead
@gergelyke @nadiia @austin94

src/modal/__tests__/modal.test.js
austin94: The provided `role` isn't a valid ARIA type. Since we are testing that something is getting passed through I disabled the check since it's a valid test.

src/checkbox/examples.js
austin94: The "Click to focus on checkbox" still seems to be working. The focus should be handled by the `inputRef` since autoFocus only happens on initial render.

austin94:
> aXe rules considered more standard than those in this react plugin. So as we see this plugin complains on somethins, which aXe allows to pass through and maybe we should just remove it instead

Axe (Deque engine) and jsx-a11y test different sides of the same coin. 

The jsx-a11y plugin is only able to statically look at our code and warns of best practices for React specifically. 

 - i.e. If an element has an onClick event, it should probably have an onKeyDown event as well

Axe-core has a much wider area of focus, specifically [these rule definitions](https://github.com/dequelabs/axe-core/blob/develop/doc/rule-descriptions.md).

- i.e. Color contrast, specific aria role combinations, as well as some basic overlap with jsx-a11y

I think keeping both around covers all of our bases.

src/checkbox/examples.js
blackswanny: does this piece still work?
https://github.com/uber-web/baseui/blob/master/src/checkbox/checkbox.js#L47

austin94: It still works, but it's an accessibility anti-pattern. It's also a native `<input />` attribute so not sure what the custom implementation is for 🤷‍♂️

nadiia: We should still support the `autoFocus` case, that sets focus on the initial render.

src/card/__tests__/card.test.js
```diff
@@ -12,7 +12,7 @@ import {header as headerImg, thumbnail as thumbnailImg} from '../images';
 
 test('Card - basic functionality', () => {
   const props = {
-    action: <a href="#">Link to a Place&nbsp;&nbsp;&nbsp;&gt;</a>,
+    action: <a href="#test">Link to a Place&nbsp;&nbsp;&nbsp;&gt;</a>,
```
nadiia: The `<a href="#">` links should still be allowed. Maybe we can override some rules for the plugin?

src/modal/__tests__/modal.test.js
nadiia: Same here. Is there a way to override/add some rules?

src/textarea/examples.js
```diff
@@ -90,9 +90,11 @@ export default {
       <React.Fragment>
         <Textarea size={SIZE.compact} placeholder="Placeholder" />
         <br />
+        {/* eslint-disable-next-line jsx-a11y/no-autofocus */}
+        <Textarea size={SIZE.compact} placeholder="Focused" autoFocus />
+        <br />
         <Textarea
           initialState={{value: 'uber'}}
-          autoFocus
```
nadiia: `autoFocus` is a part of functionality form controls provide, we should be able to keep these examples

PR #532, opened 2018-11-21
feat: add tabs RFC

anantkamath:
Add rfc for Tabs component


rfcs/tabs-component.md
```diff
@@ -0,0 +1,150 @@
...
+
+```js
+import * as React from 'react';
+import {Tabs, TabPanel} from 'baseui/tabs';
+
+export default () => {
+  return (
+    <div>
+      <Tabs initialState={{activeKey: '1'}}>
+        <TabPanel key="1" title="Tab 1">Tab 1 Content</Tab>
```
gergelyke: should we keep the basic usage absolute basic? so for example, leaving out the `key` prop, as that's optional?

rfcs/tabs-component.md
```diff
@@ -0,0 +1,150 @@
...
+  Overrides for presentational components. See "Presentational Components Props API" below
+  * `[ComponentName]: ReactComponent | {props: {}, style: {}, component: ReactComponent}` - Optional
+* `onChange: ({activeKey: string>})` - Optional.
+  onChange handler that is called every time a new tab is selected
+* `stateReducer: (type: string, nextState: {activeKey: string}, currentState: {activeKey: string}) => stateToSet: {activeKey: string}` - Optional.
+  A state change handler.
+  * `type` - a state change type
+  * `nextState` - a new state value to be set
+  * `currentState` - current state value
+  * `stateToSet` - a return value that the state will be updated with
```
austin94: [nit] spacing

gergelyke:
looks great, thanks for contributing! 👍

rfcs/tabs-component.md
```diff
@@ -0,0 +1,150 @@
...
+
+* `Tabs`
+* `TabPanel`
+* `StyledTab`
+* `StyledTabBar`
+* `StyledTabContent`
+* `ORIENTATION`
+* `STATE_CHANGE_TYPE`
+
+## `Tabs` API
```
nadiia: We should split it into `Tabs` with `activeKey: string` prop and `StatefulTabs` with `initialState: {activeKey: string}` prop

rfcs/tabs-component.md
```diff
@@ -0,0 +1,150 @@
...
+* `children: Array<TabPanel>` - Required.
+  An array of TabPanel items.
+* `initialState: {activeKey: string}` - Optional.
```
nadiia: Should add what it would default to if no `activeKey` passed.

rfcs/tabs-component.md
```diff
@@ -0,0 +1,150 @@
...
+  Initial state of the component where `activeKey` is the key of the tab to be selected by default
+* `disabled: boolean` - Optional. Default is `false`
+  If set to `true` all its tabs will be disabled
+* `orientation` - Optional. Defaults to ORIENTATION.horizontal
```
nadiia: `orientation: $Values<typeof ORIENTATION>`

rfcs/tabs-component.md
```diff
@@ -0,0 +1,150 @@
...
+* `overrides: {Root, Content, TabBar}` - Optional.
```
nadiia: Above in the example we override `TabContent` should probably update example or the API description.

rfcs/tabs-component.md
```diff
@@ -0,0 +1,150 @@
...
+
+These properties are passed to presentational (styled) components that are exported:
+
+* `$disabled: boolean` If the Tabs component (or an individual tab) is disabled
+* `$active: boolean` If a particular Tab is active
+* `$orientation: string` Orientation of the tab bar (either ORIENTATION.horizontal or ORIENTATION.vertical)
+
+## STATE_CHANGE_TYPE Constant
+
+* `activeTabChanged` - The active tab is changed (a new tab has been selected)
```
nadiia: Maybe a shorter `change` would work here

chasestarr:
*naming bikeshed, feel free to ignore*

the designs in figma imply that the 'tabs' component should be used for 'subnavigation'. is it valuable to name these components something explicit like `SubnaviationTabs` and `SubnavigationTab` in an attempt to suggest design best-practices? 

i can imagine a down-side to this approach would be tight coupling to current design conventions could cause inconsistencies in future iterations.

anantkamath:
> _naming bikeshed, feel free to ignore_
> 
> the designs in figma imply that the 'tabs' component should be used for 'subnavigation'. is it valuable to name these components something explicit like `SubnaviationTabs` and `SubnavigationTab` in an attempt to suggest design best-practices?
> 
> i can imagine a down-side to this approach would be tight coupling to current design conventions could cause inconsistencies in future iterations.

My main gripe would be folks are used to looking for a simple Tabs component in  any component library. Calling it SubnavigationTabs would be be needless confusing as to what it does (is it supposed to be used only within a Navbar?), besides being perhaps (annoyingly?) too verbose.

chasestarr:
Thanks for the contribution @anantkamath !

PR #533, opened 2018-11-21
fix: remove radio animation jitter in firefox

chasestarr:
#### Patch: Bug Fix

image-2018-11-12-10-37-14-705

Selecting a radio element in firefox previously caused a 'jitter' due to animating both the element `width/height` and `border-width` at the same time. Isolating either the `width/height` or `border-width` animations resulted in smooth transitions. [minimal reproducing example](https://codesandbox.io/s/v387xqmq0y)

Approach to fix was to break the single element into two nested elements so that only the inner circle needs to animate.

### DIFF
--- a/src/radio/radiogroup.js
+++ b/src/radio/radiogroup.js
@@ -7,7 +7,7 @@ LICENSE file in the root directory of this source tree.
 // @flow
 import React from 'react';
 
-import {default as StyledRadioMark} from './radiomark';
+import StyledRadioMark from './radiomark';
 import {
   RadioGroupRoot as StyledRadioGroupRoot,
   Label as StyledLabel,
--- a/src/radio/radiogroup.js
+++ b/src/radio/radiogroup.js
@@ -6,14 +6,15 @@ LICENSE file in the root directory of this source tree.
 */
 // @flow
 import React from 'react';
-import type {PropsT, DefaultPropsT, StatelessStateT} from './types';
+
+import {default as StyledRadioMark} from './radiomark';
 import {
   RadioGroupRoot as StyledRadioGroupRoot,
-  RadioMark as StyledRadioMark,
   Label as StyledLabel,
   Input as StyledInput,
   Root as StyledRoot,
 } from './styled-components';
+import type {PropsT, DefaultPropsT, StatelessStateT} from './types';
 
 class StatelessRadioGroup extends React.Component<PropsT, StatelessStateT> {
   static defaultProps: DefaultPropsT = {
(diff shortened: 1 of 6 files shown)


src/radio/radiogroup.js
```diff
@@ -6,14 +6,15 @@ LICENSE file in the root directory of this source tree.
 */
 // @flow
 import React from 'react';
-import type {PropsT, DefaultPropsT, StatelessStateT} from './types';
+
+import {default as StyledRadioMark} from './radiomark';
```
austin94: Could this just be `import StyledRadioMark from './radiomark'`?

src/radio/radiomark.js
```diff
@@ -0,0 +1,21 @@
...
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+// @flow
+
+import React from 'react';
+
+import {RadioMarkOuter, RadioMarkInner} from './styled-components';
+
+function StyledRadioMark(props: *) {
```
nadiia: nit: more accurate would be `RadioMark` (-`Styled` prefix)

src/radio/styled-components.js
```diff
@@ -73,31 +72,35 @@ export const RadioMark = styled('span', props => {
...
+  const {sizing} = props.$theme;
+
+  return {
+    alignItems: 'center',
+    backgroundColor: getBorderColor(props),
+    borderRadius: '50%',
+    display: 'flex',
+    height: sizing.scale700,
+    justifyContent: 'center',
+    margin: sizing.scale200,
```
nadiia: use longhand style props instead:
```
    marginTop: $theme.sizing.scale200,	
    marginBottom: $theme.sizing.scale200,	
    marginLeft: $theme.sizing.scale200,	
    marginRight: $theme.sizing.scale200,
```

gergelyke: just for me information, can you help me why we use longhands styles?

chasestarr: ah, right. thanks!

MarkReeder: > just for my information, can you help me why we use longhands styles?

We use longhand styles because mixing shorthand and longhand styles can cause issues (https://github.com/styletron/styletron/issues/68 for example) and standardizing on longhand styles allows us to be more granular with overrides and not have to recreate the full shorthand style that's being overridden.

gergelyke: makes sense, thanks Mark!

PR #534, opened 2018-11-21
feat: avatar component rfc

chasestarr:
i made some assumptions here in regards to a fallback state since design input is to be determined.

added a `StatefulAvatarContainer` to contain behavior related to supplying the `img` element either the primary or fallback `src`.

```js
<Avatar name="Jane Doe" src="https://url-of-img.jpg" />
```

### DIFF
--- a/rfcs/avatar-component.md
+++ b/rfcs/avatar-component.md
@@ -1,0 +1,65 @@
+# Avatar Component
+
+## Usage
+
+### Basic usage
+
+```javascript
+import * as React from 'react';
+import {Avatar} from 'baseui/avatar';
+
+export default () => <Avatar name="Jane Doe" src="https://url-of-img.jpg" />;
+```
+
+### Advanced usage
+
+```javascript
+import * as React from 'react';
+import {Avatar} from 'baseui/avatar';
+
+export default () => (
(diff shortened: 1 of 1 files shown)


chasestarr:
default image is fine with me, or even a simple color may be most appropriate while waiting for direction from design

austin94:
> Looks great! I am wondering if instead of the srcFallback property, we would simply ship a default image that would be used?

Do we have a "profile" SVG icon anywhere? Might be more preferable than an image

rfcs/avatar-component.md
```diff
@@ -0,0 +1,62 @@
...
+## `StatefulAvatarContainer` API
+
+* All props of a `Avatar` component. Handles displaying fallback color if provided src image fails to load.
+
+## Presentational components props API
+
+These properties are passed to every presentational (styled) component that is exported:
+
+* `$size: string`
+  Sets img width. Provided value should map to an option in $theme.sizing. If not found, defaults to `scale1000`.
```
nadiia: We can also do `string` that is either option in the `$theme.sizing` or if not found considered a valid width value, for example `50px` or `25%`, etc.

PR #540, opened 2018-11-26
feat: introduce Rating component

austin94:
### Minor: New Feature

This PR introduces the `Rating` component.

#### Star

starsexample

#### Emoticon

emoticonexample

### Tests Added + Pass

Yes

### License

MIT

### DIFF
--- a/src/rating/examples.js
+++ b/src/rating/examples.js
@@ -9,6 +9,7 @@ LICENSE file in the root directory of this source tree.
 
 import React from 'react';
 import {Rating, KIND} from './';
+import type {StyledRatingItemPropsT} from './';
 import examples from './examples-list';
 
 type ExampleState = {
@@ -40,6 +41,45 @@ export default {
         return (
           <Rating
             kind={KIND.emoticon}
+            value={this.state.value}
+            onChange={({value}) => this.setState({value})}
+          />
+        );
+      }
+    }
+
+    return <Example />;
+  },
+  [examples.OVERRIDES]: function Story2() {
+    const CustomRatingItem = ({
+      $index,
+      $isActive,
+      $isSelected,
+      ...props
+    }: StyledRatingItemPropsT) => (
+      <div
+        {...props}
+        style={{
+          display: 'inline-block',
+          marginRight: 8,
+          color: $isSelected ? 'blue' : $isActive && 'red',
+        }}
+      >
+        {$index}
+      </div>
+    );
+
+    class Example extends React.Component<{}, ExampleState> {
+      state = {};
+
+      render() {
+        return (
+          <Rating
+            overrides={{
+              Star: {
+                component: CustomRatingItem,
+              },
+            }}
             value={this.state.value}
             onChange={({value}) => this.setState({value})}
           />
--- a/src/rating/examples.js
+++ b/src/rating/examples.js
@@ -1,0 +1,52 @@
+/*
+Copyright (c) 2018 Uber Technologies, Inc.
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+
+// @flow
+
+import React from 'react';
+import {Rating, KIND} from './';
+import examples from './examples-list';
+
+type ExampleState = {
+  value?: number,
+};
+
+export default {
+  [examples.DEFAULT]: function Story1() {
+    class Example extends React.Component<{}, ExampleState> {
+      state = {};
+
+      render() {
+        return (
+          <Rating
+            value={this.state.value}
+            onChange={({value}) => this.setState({value})}
+          />
+        );
+      }
+    }
+
+    return <Example />;
+  },
+  [examples.EMOTICON]: function Story2() {
+    class Example extends React.Component<{}, ExampleState> {
+      state = {};
+
+      render() {
+        return (
+          <Rating
+            kind={KIND.emoticon}
+            value={this.state.value}
+            onChange={({value}) => this.setState({value})}
+          />
+        );
+      }
+    }
+
+    return <Example />;
+  },
+};
--- a/src/rating/rating.js
+++ b/src/rating/rating.js
@@ -1,0 +1,98 @@
+/*
+Copyright (c) 2018 Uber Technologies, Inc.
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+
+// @flow
+
+import React from 'react';
+import {KIND} from './constants';
+import type {RatingPropsT, RatingStateT} from './types';
+import {StyledRoot, StyledStar, StyledEmoticon} from './styled-components';
+import {getOverrides} from '../helpers/overrides';
+
+const ENTER_KEY_CODE = 13;
+const SPACE_KEY_CODE = 32;
+
+class Rating extends React.Component<RatingPropsT, RatingStateT> {
+  static defaultProps = {
+    overrides: {},
+    kind: KIND.star,
+  };
+
+  state = {};
+
+  renderRatingContents = () => {
+    const {overrides = {}, value = -1, onChange} = this.props;
+    const {previewIndex} = this.state;
+
+    const [Star, starProps] = getOverrides(overrides.Star, StyledStar);
+    const [Emoticon, emoticonProps] = getOverrides(
+      overrides.Emoticon,
+      StyledEmoticon,
+    );
+
+    const ratings = [];
+    const RatingItem = this.props.kind === KIND.star ? Star : Emoticon;
+    const overrideProps =
+      this.props.kind === KIND.star ? starProps : emoticonProps;
+
+    for (let x = 1; x <= 5; x++) {
+      ratings.push(
+        <RatingItem
+          key={x}
+          role="radio"
+          tabIndex={0}
+          aria-setsize={5}
+          aria-checked={x <= value}
+          aria-posinset={x}
+          $index={x}
+          $isActive={
+            previewIndex !== undefined ? x <= previewIndex : x <= value
+          }
+          $isSelected={x === previewIndex}
+          onClick={() => {
+            onChange && onChange({value: x});
+            this.setState({previewIndex: undefined});
+          }}
+          onKeyDown={e => {
+            if (e.keyCode === SPACE_KEY_CODE || e.keyCode === ENTER_KEY_CODE) {
+              onChange && onChange({value: x});
+              this.setState({previewIndex: undefined});
+            }
+          }}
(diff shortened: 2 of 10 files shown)


blackswanny:
tests should have `src/rating/__tests__/rating.test.js` suffix
also we usually have styled-components.test.js which checks for styles in every state for them

src/rating/__tests__/rating.test.js
```diff
@@ -0,0 +1,79 @@
...
+  describe('Root', () => {
+    it('applies correct accessibility attributes to the Root element', () => {
+      expect(example).toHaveProp('tabIndex', 0);
+      expect(example).toHaveProp('role', 'radiogroup');
+    });
+  });
+
+  describe('RatingItem', () => {
+    it('applies correct props if item is active', () => {
+      expect(example.childAt(0)).toMatchSnapshot();
```
blackswanny: `childAt` is not semantically readable. Should we use `find` to look for exact element?

src/rating/examples.js
```diff
@@ -0,0 +1,92 @@
...
+  [examples.OVERRIDES]: function Story2() {
+    const CustomRatingItem = ({
+      $index,
+      $isActive,
+      $isSelected,
+      ...props
+    }: StyledRatingItemPropsT) => (
+      <div
+        {...props}
+        style={{
```
nadiia: Probably shouldn't use inline styles in our examples, but use the `styled` instead

src/rating/rating.js
```diff
@@ -0,0 +1,97 @@
...
+  updatePreview = (previewIndex?: number) => {
+    this.setState({previewIndex});
+  };
+
+  renderRatingContents = () => {
+    const {overrides = {}, value = -1} = this.props;
+    const {previewIndex} = this.state;
+
+    const [Star, starProps] = getOverrides(overrides.Star, StyledStar);
+    const [Emoticon, emoticonProps] = getOverrides(
```
nadiia: Wonder if providing a single override would be more generalized and flexible here. 
Something like `overrides.RatingItem` or `overrides.RatingPoint` or something else. And then provide our default here based in the `kind` prop: `getOverrides(overrides.RatingItem, kind === KIND.star ? StyledStar : StyledEmoticon)`

src/rating/rating.js
```diff
@@ -0,0 +1,97 @@
...
+      overrides.Emoticon,
+      StyledEmoticon,
+    );
+
+    const ratings = [];
+    const RatingItem = this.props.kind === KIND.star ? Star : Emoticon;
+    const overrideProps =
+      this.props.kind === KIND.star ? starProps : emoticonProps;
+
+    for (let x = 1; x <= 5; x++) {
```
nadiia: Should we make the `5` be overridable through props?

src/rating/__tests__/rating.test.js
austin94: Due to shallow rendering and all of the children being the same element type; I'm not sure how find would be more readable.  Do you have a specific example you had in mind?

src/rating/styled-components.js
```diff
@@ -0,0 +1,120 @@
...
+
+// @flow
+
+import {styled} from '../styles/index';
+import type {StyledRootPropsT, StyledRatingItemPropsT} from './types';
+
+export const StyledRoot = styled('ul', ({$theme}: StyledRootPropsT) => {
+  return {
+    margin: 0,
+    padding: 0,
```
nadiia: longhand prop names for `margin` and `padding`

src/rating/styled-components.js
```diff
@@ -0,0 +1,120 @@
...
+    if ($isActive) {
+      starStroke = starFill = $theme.colors.warning300;
+    }
+
+    const starSVG = `<svg width="22" height="20" viewBox="0 0 22 20" fill="none" xmlns="http://www.w3.org/2000/svg"><path d="M11 1.54816L13.1268 7.78053L13.2423 8.11905H13.6H20.4143L14.9132 11.9714L14.6055 12.1869L14.7268 12.5424L16.8321 18.7118L11.2868 14.8285L11 14.6277L10.7132 14.8285L5.16792 18.7118L7.27321 12.5424L7.39454 12.1869L7.08681 11.9714L1.58566 8.11905H8.4H8.75769L8.87321 7.78053L11 1.54816Z" fill="${encodeURIComponent(
+      starFill,
+    )}" stroke="${encodeURIComponent(starStroke)}"/></svg>`;
+
+    const styles = {
+      padding: 0,
```
nadiia: longhand prop

src/rating/styled-components.js
```diff
@@ -0,0 +1,120 @@
...
+    const happyRatingSVG = `<svg width="44" height="44" viewBox="0 0 44 44" fill="none" xmlns="http://www.w3.org/2000/svg"><circle cx="22" cy="22" r="20" fill="${encodeURIComponent(
+      emoticonFill,
+    )}"/><ellipse cx="16" cy="18" rx="2" ry="2" fill="black"/><ellipse cx="28" cy="18" rx="2" ry="2" fill="black"/><path fill-rule="evenodd" clip-rule="evenodd" d="M21.8377 29C24.8073 29 26.6585 27.2303 27.1609 26.4558C27.4615 25.9925 28.0807 25.8605 28.5441 26.161C29.0074 26.4616 29.1394 27.0808 28.8388 27.5442C28.0439 28.7697 25.614 31 21.8377 31C18.0453 31 15.8091 28.75 15.1202 27.4755C14.8576 26.9897 15.0385 26.3829 15.5244 26.1203C16.0102 25.8577 16.617 26.0386 16.8796 26.5245C17.2718 27.25 18.8842 29 21.8377 29Z" fill="black"/></svg>`;
+    const veryHappyRatingSVG = `<svg width="44" height="44" viewBox="0 0 44 44" fill="none" xmlns="http://www.w3.org/2000/svg"><circle cx="22" cy="22" r="20" fill="${encodeURIComponent(
```
nadiia: How about we move all the SVGs in a single separate file?

src/rating/styled-components.js
```diff
@@ -0,0 +1,120 @@
...
+    const ratingIcons = [
+      angryRatingSVG,
+      sadRatingSVG,
+      neutralRatingSVG,
+      happyRatingSVG,
+      veryHappyRatingSVG,
+    ];
+
+    const styles = {
+      padding: 0,
```
nadiia: longhand prop

src/rating/__tests__/styled-components.test.js
```diff
@@ -0,0 +1,63 @@
...
+describe('StyledComponents', () => {
+  describe('StyledRoot', () => {
+    it('displays correct styles by default', () => {
+      const component = shallow(<StyledRoot />);
+      expect(component.instance().getStyles()).toMatchSnapshot();
+    });
+  });
+
+  describe('StyledStar', () => {
+    it('displays correct styles by default', () => {
```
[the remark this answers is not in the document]
austin94: I've updated these tests to use the `.each` utilitiy

src/rating/__tests__/rating.test.js
blackswanny: https://airbnb.io/enzyme/docs/api/ShallowWrapper/at.html

src/rating/rating.js
blackswanny: and also need to provide new set of images

austin94: I originally thought about that, but it would complicate the component for a feature that we aren't sure exists or is needed.  If there is a Figma spec for a dynamic rating than we can probably add that functionality in at a later date.

For the Star component it's trivial, but the Emoticon example is limited to 5 items.

blackswanny: so I suggest to talk about it with @nguyensomniac or Jeff. As our customers likely to request this feature. For instance, currently some uber apps have 2 or 3 only rating icons.

src/rating/styled-components.js
austin94: I've refactored these into a shared utility file

austin94:
> My only question is around emoticons rating, it seems like it should act slightly different from the stars and not highlight any other face when a single one hovered over or selected, something like a single select instead of a scale.

This is actually how it's shown in Figma, I overlooked this originally.

The component now handles the Emoticon state as you described.

src/rating/rating.js
austin94: That would also fix some of the complexity in logic between the 2 IMHO

src/rating/rating.js
austin94: With the split to `StarRating` and `EmoticonRating` to allow dynamic sizing, the overrides now match `{ Root, Item }`

src/rating/rating.js
austin94: These components are now split. Much cleaner!

nguyensomniac: Agreed with @austin94 on this, let's wait until we know if this is a feature people need.

src/rating/emoticon-rating.js
```diff
@@ -0,0 +1,97 @@
...
+          }
+          $isSelected={x === previewIndex}
+          onClick={() => this.selectItem(x)}
+          onKeyDown={e => {
+            if (e.keyCode === SPACE_KEY_CODE || e.keyCode === ENTER_KEY_CODE) {
+              this.selectItem(x);
+            }
+          }}
+          onFocus={() => this.updatePreview(x)}
+          onMouseOver={() => this.updatePreview(x)}
```
nadiia: We may need to debounce this handler, what do you think?

src/rating/emoticon-rating.js
```diff
@@ -0,0 +1,97 @@
...
+        <Emoticon
+          key={x}
+          role="radio"
+          tabIndex={0}
+          aria-setsize={5}
+          aria-checked={x === value}
+          aria-posinset={x}
+          $index={x}
+          $isActive={
+            previewIndex !== undefined ? x === previewIndex : x === value
```
nadiia: The other option could be introducing a `$isPseudoActive` props for elements with a `key` value less than `previewIndex`. This way we could export a single `Rating` component for both `star` and `emoticons` types and only style pseudo-active `StyledStar` items but not `StyledEmoticon`.

src/rating/emoticon-rating.js
austin94: The [mouseover](https://developer.mozilla.org/en-US/docs/Web/Events/mouseover) is only triggered when the mouse enters the container. So no debounce should be necessary. 

The [mousemove](https://developer.mozilla.org/en-US/docs/Web/Events/mousemove) would be the trouble maker.

PR #544, opened 2018-11-27
docs: add new theming docs

austin94:
This PR introduces a new `Themable values` page in storybook.

screen shot 2018-11-27 at 11 37 07 am

#### License

MIT


austin94:
> can we add some description too on how one can use these values? so folks don't start copy-pasting these values, but they'll grab them from the themes?

I'll move this into it's own subsection then. Will add some introduction and theming examples.

austin94:
@gergelyke We already had some theming documentation under `src/theme/README.md`. I've moved this to the docs section, updated references from the root README, and added some additional instructions for usage with the `overrides` prop.

PR #546, opened 2018-11-27
feat(avatar): adds avatar component

chasestarr:
Updated rfc based on some changes in implementation. 

- Chose not to export a stateful component since this does not seem like a logic heavy component, but happy to change if we'd like/better follows the project conventions. 
- Added a `Root` component to facilitate an error color, and chose to export it. Unsure if this is the right decision since updates from design with a fallback *image src* may remove the need for it. Would it be preferable to not export `Root` as to not require a breaking change in the future?
- Do we have other release options when in intermediate stages like here? We spoke about adding a console statement to the nav heading component. I kind of like how React itself prepends `UNSAFE` to api which is not 100% in. And again, we could simply not allow root component to be customizable. 

simple
screen shot 2018-11-27 at 3 43 27 pm

error state
screen shot 2018-11-27 at 3 43 32 pm

overrides
screen shot 2018-11-27 at 3 43 42 pm


rfcs/avatar-component.md
```diff
@@ -40,23 +40,23 @@ export default () => (
 * `name: string` - Required
   Defines an alternative text description of the image.
 * `overrides: {}` - Optional
-  * `Avatar: ?React.ComponentType` component to use for Avatar image styling
+  * `Avatar: ?React.ComponentType` component to use for Avatar image styling.
+  * `Root: ?React.ComponentType` component to use for containing element. displays if image fails to load.
```
nadiia: the above two should be `Avatar?: React.ComponentType | {props: {}, style: {}, component: React.ComponentType}` type

rfcs/avatar-component.md
```diff
@@ -40,23 +40,23 @@ export default () => (
...
 * `size: string` - Optional. Defaults to `scale1000`
-  Defines the width of the image. Should be an option from the $theme.sizing list otherwise, will use default.
+  Defines the width/height of the image. Must be an option from the $theme.sizing list otherwise, will use default.
 * `src: string` - Required
   Image to display.
 
-## `StatefulAvatarContainer` API
-
-* All props of a `Avatar` component. Handles displaying fallback color if provided src image fails to load.
-
```
nadiia: 👍

src/avatar/styled-components.js
```diff
@@ -0,0 +1,45 @@
...
+// @flow
+
+import {styled} from '../styles';
+import type {StylePropsT} from './types';
+
+function getSize(props: StylePropsT) {
+  const {$size, $theme} = props;
+
+  const defaultSize = $theme.sizing.scale1000;
+  const size = $size || defaultSize;
```
nadiia: no need for `|| defaultSize` here

rfcs/avatar-component.md
blackswanny: we have it in our overrides.js types definition. We may use it if customer is aware of where to look for particular type

image

src/avatar/styled-components.js
chasestarr: flow complains about accessing `undefined` on the object (`$theme.sizing[undefined]`)

src/avatar/__tests__/avatar.test.js
```diff
@@ -0,0 +1,51 @@
...
+    triggerLoadError(wrapper);
+    expect(wrapper.find('img')).toHaveLength(0);
+  });
+
+  it('applies expected accessibility attributes if image fails to load', () => {
+    const name = 'user name';
+    const wrapper = mount(<Avatar name={name} src="invalid-img-src.png" />);
+    const rootNode = wrapper.childAt(0).getDOMNode();
+
+    expect(rootNode.getAttribute('aria-label')).toBeNull();
```
blackswanny: these `expects` refer to another logic, when image doesn't fail to load. I would split this test into two, so these `expects` do not confuse with the main `expects` of this test. And we can combine them under one reasonable `describe` to combine common `arrange` part

src/avatar/__tests__/styled-components.test.js
```diff
@@ -0,0 +1,32 @@
...
+// @flow
+
+import React from 'react';
+import {shallow} from 'enzyme';
+import {StyledAvatar, StyledRoot} from '../index';
+
+describe('Avatar styled components', () => {
+  const styledComponents = [
+    [StyledAvatar, 'StyledAvatar'],
+    [StyledRoot, 'StyledRoot'],
```
blackswanny: we should test Root against `$didImageFailToLoad` property as well

src/avatar/avatar.js
```diff
@@ -0,0 +1,54 @@
...
+
+import {getOverrides} from '../helpers/overrides';
+
+import {Avatar as StyledAvatar, Root as StyledRoot} from './styled-components';
+import type {PropsT, StateT} from './types';
+
+export default class Avatar extends React.Component<PropsT, StateT> {
+  static defaultProps: $Shape<PropsT> = {
+    overrides: {},
+    size: 'scale1000',
```
blackswanny: we should be able to import current Theme applied. In this case we would be able to reference exact value instead of string name. Need to think about better way to do it
cc @nadiia @gergelyke @austin94

src/avatar/examples.js
```diff
@@ -0,0 +1,120 @@
...
+export default {
+  [examples.AVATAR_EXAMPLE]: function AvatarStory() {
+    return (
+      <Row>
+        {['scale800', 'scale1000', 'scale1200', 'scale1400'].map(
+          (size, index) => (
+            <Avatar
+              name={`user name # ${index}`}
+              size={size}
+              src={`https://api.adorable.io/avatars/285/${index}@adorable.io.png`}
```
blackswanny: make sure we can use these icons under license

src/avatar/styled-components.js
```diff
@@ -0,0 +1,45 @@
...
+  return $theme.sizing[size] || defaultSize;
+}
+
+export const Avatar = styled('img', (props: StylePropsT) => {
+  const themedSize = getSize(props);
+
+  return {
+    borderRadius: '50%',
```
blackswanny: we may need to add `box-sizing: border-box;` in case user adds overrides of border and padding, so 50% will still work

src/avatar/avatar.js
nadiia: Do you mean customers import the `theme`? We shouldn't encourage doing it in order to set the size here. 
The current approach is good as an initial implementation. What we can do further to allow custom sizes is to set the default `size` prop to `scale1000` and allow any string as a value of `size` prop, for example '25px'. Then in the `styled-components.js`:
```
 function getSize(props: StylePropsT) {
  const {$size, $theme} = props;
   return $theme.sizing[$size] || $size;
}
```

blackswanny: if we were able to put `size:  currentTheme.scale.scale1000` here, that would help to not use this string constant. Strings to refer to object property are vague and can be easily compromised. It seems to be not the first example where we try to address it

blackswanny: if importing `theme` is not allowed for some reasons, we can think about how to accommodate it still. For instance, freeze theme object so it would be available only for reading. 
@nadiia  Can you elaborate, why we shoulnd't import theme?

src/avatar/examples.js
gergelyke: Adorable Avatars is released under the terms of the MIT License.

rfcs/avatar-component.md
chasestarr: @blackswanny can you add this comment to the fixit week RFC audit task? i'd rather not change conventions here when we have a specific portion of work to reconsider RFC decisions

src/avatar/styled-components.js
chasestarr: @nadiia resolving for now, but let me know if you have strong feeling about this

src/avatar/examples.js
chasestarr: something this adorable must be shared with the world :)

src/avatar/__tests__/avatar.test.js
chasestarr: good idea

src/avatar/avatar.js
chasestarr: this api follows conventions set in [Block](https://github.com/uber-web/baseui/blob/master/src/block/examples.js#L71). i can update the interface to allow arbitrary sizes like `10px`, `1rem`, etc. does that solves both of our issues? we should consider exposing a withTheme context consumer or HOC like [here](https://github.com/uber-web/baseui/blob/e3d2c136b2dcbdee8f113adbd3a8af755f2b10a0/src/styles/styled.js#L14).

src/avatar/styled-components.js
chasestarr: updated

src/avatar/__tests__/avatar.test.js
```diff
@@ -0,0 +1,58 @@
...
+  wrapper.update();
+}
+
+describe('Avatar', () => {
+  it('applies expected accessibility attributes to img element', () => {
+    const name = 'user name';
+    const wrapper = mount(<Avatar name={name} src="valid-img-src.png" />);
+
+    const image = wrapper.find('img');
+    expect(image.getDOMNode().getAttribute('alt')).toBe(name);
```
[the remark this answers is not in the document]
chasestarr: can you share the documentation for this matcher?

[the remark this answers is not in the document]
austin94: Weird, originally thought that this was being provided by `jest-enzyme`, but it looks like it's provided by `jest-core`?

Looking through their TS types...

```js
// DOM testing library extensions https://github.com/kentcdodds/dom-testing-library#custom-jest-matchers
type DomTestingLibraryType = {
  toBeInTheDOM(): void,
  toHaveTextContent(content: string): void,
  toHaveAttribute(name: string, expectedValue?: string): void,
};
```

It's included from [dom-testing-library](https://github.com/kentcdodds/dom-testing-library#custom-jest-matchers), but isn't listed in the [full matchers reference](https://jestjs.io/docs/en/expect) 🤷‍♂️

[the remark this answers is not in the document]
austin94: Lets resolve this for now. Will require some other configuration to use.

rfcs/avatar-component.md
blackswanny: added

PR #547, opened 2018-11-28
add silhouette of top bar and grids

blackswanny:
added silhouette of top bar. To add it also started to add grid system logic.

### DIFF
--- a/rfcs/top-navigation-silhouette.md
+++ b/rfcs/top-navigation-silhouette.md
@@ -1,0 +1,67 @@
+# Top Navigation Silhouette
+
+This silhouette provides proper styles for wrapper element to be used as Top Navigation. By default element has no visual styles.
+Silhouettes comply with Base UI grid layout, so they are responsive to resizing and stretching, changing sizes, fonts, etc
+
+## Usage
+
+### Basic usage
+
+```javascript
+import * as React from 'react';
+import {TopNavigation} from 'baseui/silhouette/top-navigation';
+import {Button} from 'baseui/button';
+
+export default () => (
+  <TopNavigation>
+    <Button>Menu</Button>
+    <span>Uber</span>
+  </TopNavigation>
+  );
+```
+
+### Advanced usage
+
+```javascript
+import * as React from 'react';
+import {TopNavigation} from 'baseui/silhouette/top-navigation';
+import {Button} from 'baseui/button';
+
+export default () => {
+  return
+    <TopNavigation
+      height={12}
+      overrides={{
+        Root: {
+          style: {
+            backgroundColor: 'blue',
+          }
+        }
+      }}
+    >
+    <Button>Menu</Button>
+    <span>Uber</span>
+</TopNavigation>;
+}
+```
+
+## Exports
+
+* `TopNavigation`
+* `StyledRoot`
+
+## `TopNavigation` API
+
+* `height: number` - Optional
+  Is number of rows as height of navigation in our grid system. Default is 9 rows.
+* `overrides: {Root}` - Optional
+  * `Root: ReactComponent` - Optional. DOM element representing top navigation
+
+## Dependencies
+
+This element dependes on `Block` element of this framework
+
+## Accessibility
+
+Use `nav` tag for root element.
+Use `role="navigation"`
--- a/src/silhouette/top-navigation/examples.js
+++ b/src/silhouette/top-navigation/examples.js
@@ -1,0 +1,51 @@
+/*
+Copyright (c) 2018 Uber Technologies, Inc.
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+
+// @flow
+
+import React from 'react';
+import {TopNavigation} from './';
+import {Button} from '../../button';
+import examples from './examples-list';
+
+export default {
+  [examples.DEFAULT]: function Story1() {
+    return (
+      <TopNavigation
+        overrides={{
+          Root: {
+            style: {backgroundColor: '#D1E2F9'},
+          },
+        }}
+      />
+    );
+  },
+  [examples.OVERRIDES]: function Story2() {
+    return (
+      <TopNavigation
+        height={12}
+        overrides={{
+          Root: {
+            props: {
+              as: 'h2',
+            },
+            style: {
+              display: 'flex',
+              alignItems: 'center',
+              justifyContent: 'space-between',
+              border: `1px solid #CCCCCC`,
+              marginTop: '40px',
+            },
+          },
+        }}
+      >
+        <Button style={{margin: '10px'}}>Get Started</Button>
+        <Button style={{margin: '10px'}}>About</Button>
+      </TopNavigation>
+    );
+  },
+};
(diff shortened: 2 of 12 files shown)


src/themes/creator.js
```diff
@@ -285,6 +285,21 @@ export default function createTheme(
     tooltip: {
       backgroundColor: primitives.mono900,
     },
+    grid: {
```
blackswanny: these are not yet final values for base scale. Will be changed upon design team sharing with us values for base scales, spacing, etc. Also LightTheme is not final place to put these constants. If we make it as part of themes, we will be able to customize it as part of theme, but need to repeat it in all of our themes.

rfcs/top-navigation-silhouette.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Usage
+
+### Basic usage
+
+```javascript
+import * as React from 'react';
+import {TopNavigation} from 'baseui/silhouette/top-navigation';
+
+export default () => <TopNavigation><button>Menu</button><span>Uber</span></TopNavigation>;
```
gergelyke: can you break this into multiple lines? 
if you are using the button as an example, can you use the Button component?

src/silhouette/top-navigation/examples.js
```diff
@@ -0,0 +1,51 @@
+/*
```
gergelyke: would it make sense to have only one top level category, `Silhouettes`, and all the different layout could go under that?

chasestarr:
i don't have prior context for the silhouette components, so pardon the probing question.

what is the intended developer experience for the silhouette components?
are they supposed to represent the entire page layout? or be a building block to place in the page?

schnerd:
Same question as @chasestarr. 

Also how does this work alongside `HeaderNavigation` (#525)? As a developer I'd be very confused if I saw both HeaderNavigation and TopNavigation in the storybook.

rfcs/top-navigation-silhouette.md
blackswanny: ok

src/silhouette/top-navigation/examples.js
blackswanny: can you elaborate? So to have only one folder `silhouettes` with all files gathered in there?

blackswanny:
> i don't have prior context for the silhouette components, so pardon the probing question.
> 
> what is the intended developer experience for the silhouette components?
> are they supposed to represent the entire page layout? or be a building block to place in the page?

each silhouette is just a block and doesn't design the whole page layout. Main difference between silhouettes and other components is that silhouettes are invisible and only provide layout styles. Also they fall under grid system of Base UI, whereas regular components do not.
This is my understanding. I will send email to Jeff to get some extra feedback.

> Also how does this work alongside HeaderNavigation (#525)? As a developer I'd be very confused if I saw both HeaderNavigation and TopNavigation in the storybook.

Probably we should start documentation describing silhouettes and grid concept so customer would be aware of difference and usage. So far HeaderNavigation can be placed anywhere and it's not `fixed`, cause doesn't represent layout. I assume in current implementation we can use them together as below, which may be confusing, especially because of naming
```js
<TopNavigation>
     <HeaderNavigation />
</TopNavigation>
```
cc @nadiia  @austin94 @chasestarr @schnerd @gergelyke 

image

gergelyke:
I agree with the question raised by Chase and David. Should we rather have a generic layout component, that has different types, and each type would support different placements.

So for this one, it would be something like this:

```javascript
<Layout type={Layout.TOP_BAR} placements={{
  header: HeaderNavigation,
  body: <div></div>
}}/>
```

What do you think @blackswanny @schnerd @chasestarr ?

blackswanny:
I think it's a good idea, which saves a ton of extra code. However, it would be a factory of many different layouts. So far we have 12, so this might be a big file

chasestarr:
rather than type as a prop, `silhouettes` may export multiple components. this could lead to more encapsulated implementation code and suggests semantic names for each silhouette configuration for eng/design to collaborate. 

nit on your example @gergelyke, but i don't feel strongly either way (I'm mostly trying to understand the team's preferences). it may be more idiomatic react to expose top-level render props, like in the example below, rather than nest in a `placements` obj

```js
import {HeaderFlush} from 'baseui/silhouettes';

<HeaderFlush
  header={<CustomerHeader />}
  body={<CustomBody />}
/>
```

may want to consider an implicit parent<>child relationship. what this does better than the other api is it removes the documentation/naming overhead of shoving a component into header, body, etc... props. downside is it may require guard-rails like if a dev places more than two children, throw/log an error. 

```js
<HeaderFlush>
  <CustomHeader /> // implicitly oriented at top of layout
  <CustomBody /> // implicitly oriented below previous element
</HeaderFlush>
```

gergelyke:
That sounds good too to me Chase - Oleg, would you mind starting with an RFC first, maybe as a new PR?

rfcs/top-navigation-silhouette.md
```diff
@@ -0,0 +1,67 @@
...
+This silhouette provides proper styles for wrapper element to be used as Top Navigation. By default element has no visual styles.
+Silhouettes comply with Base UI grid layout, so they are responsive to resizing and stretching, changing sizes, fonts, etc
+
+## Usage
+
+### Basic usage
+
+```javascript
+import * as React from 'react';
+import {TopNavigation} from 'baseui/silhouette/top-navigation';
```
nadiia: Can we have it imported from `baseui/silhouette` with no subdirs?

blackswanny:
> That sounds good too to me Chase 

What approach exactly? Both examples have downsides
This one explicitly tells layout parts which we need to consider, whereas every application requires many custom layout elements: header, subheader, main body, footer, left menu, side menu, right menu. Silhouettes themselves show at least 12 examples, and it's only those that might be needed. Silhouettes can be used in custom app layout here and there, but not setup the whole page as we can't comprehend those
```js
<HeaderFlush
  header={<CustomerHeader />}
  body={<CustomBody />}
/>
```

This example has the same concern as previous, but also the styles will be forced to apply even if we shuffle these children in different order, which may confuse as usually React elements reflect the same DOM tree from left to right, from top to bottom. Also our overrides approach needs to be enforced frequently to work with this example and override styles
```js
<HeaderFlush>
  <CustomHeader /> // implicitly oriented at top of layout
  <CustomBody /> // implicitly oriented below previous element
</HeaderFlush>
```

nadiia:
I think we shouldn't tight the content of a top bar to its props, but it should accept it as children and it should not care much what children it has. It should be a simple layout component not more.

gergelyke:
Can we take a step back please, and split the RFC part into a new PR, so we can discuss just that separately please?

rfcs/top-navigation-silhouette.md
blackswanny: yes. last proposal is to have import links from silhouette folder directly

blackswanny:
> Can we take a step back please, and split the RFC part into a new PR, so we can discuss just that separately please?

added RFC PR https://github.com/uber-web/baseui/pull/558

PR #549, opened 2018-11-28
fix: constrain flow $values util type

### LINKED ISSUE #539: flow $Values utility type does not match against literal
originally brought up by @brentertz in baseui [slack](https://baseui.slack.com/archives/CCGJPJA1H/p1543243130000200)

## Current Behavior

```js
const FOO = {
  bar: 'bar',
  bas: 'bas'
};

type Foo = $Values<typeof FOO>
  
// Works
const v1: Foo = 'bar';

// Errors
const v2: Foo = 1;

// Does not error, but should
const v3: Foo = 'quux';
```

## Context

Flow's `$Values` utility type by default does not compare against the string literal, but instead compares to simple string type. The referenced object should be frozen to operate as expected.

```js
const FOO = Object.freeze({
  bar: 'bar',
  bas: 'bas'
});

type Foo = $Values<typeof FOO>
  
// Works
const v1: Foo = 'bar';

// Errors
const v2: Foo = 'quux';
```

5 or so [instances](https://github.com/uber-web/baseui/search?utf8=%E2%9C%93&q=%24values%3C&type=) of this in the code base

chasestarr:
closes #539


nadiia:
Why don't we use `$Keys` in types instead?

leyanlo:
In my opinion, freezing the objects makes sense since they're effectively constants, and `Object.freeze()` enforces that idea. And you'd want `$Values` over `$Keys`, since that's what you'd be using for your types. The fact that the values are the same as the keys is just a convention.

We actually use this pattern a bunch in the freight/shipper-web app, with the addition of [keymirror](https://www.npmjs.com/package/keymirror), which simplifies things a little because we only have to worry about the left-hand side. E.g.
```js
export const STYLE_TYPE = Object.freeze(
  keymirror({
    default: null,
    toggle: null,
  })
);
```

blackswanny:
I wonder if we could use some utility method to not repeat this next time

nadiia:
We should revisit the const's types for all components.

gergelyke:
@chasestarr when you have the chance, can you create a story for that please?

PR #550, opened 2018-11-28
fix(saucelabs): add start_period

gergelyke:


nightwatch.conf.js
```diff
@@ -57,7 +57,7 @@ const sauceLabsBaseConfig = {
   username: '${SAUCE_USERNAME}',
   access_key: '${SAUCE_ACCESS_KEY}',
   globals: {
-    waitForConditionTimeout: 10000,
+    waitForConditionTimeout: 20000,
```
nadiia: How about we increase default max wait to 10sec?

PR #558, opened 2018-11-28
add top bar silhouette RFC

blackswanny:
add top bar silhouette RFC

### DIFF
--- a/rfcs/top-bar-silhouette.md
+++ b/rfcs/top-bar-silhouette.md
@@ -1,0 +1,62 @@
+# Top Bar Silhouette
+
+This silhouette provides proper styles for wrapper element to be used as Top Navigation. By default element has no visual styles.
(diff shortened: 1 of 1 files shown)


rfcs/top-bar-silhouette.md
```diff
@@ -0,0 +1,62 @@
...
+    <span>Uber</span>
+  </TopBar>
+  );
+```
+
+### Advanced usage
+
+```javascript
+import * as React from 'react';
+import {TopBar} from 'baseui/silhouette/top-bar';
```
gergelyke: as I see, this naming is misleading - we want to provide layouts, and not just topbars, but a Layout that has a topbar and a body.

What would you think about this approach?

```
<Layout type={Layout.TOP_BAR} placements={{
  header: HeaderNavigation,
  body: <div></div>
}}/>
```

PR #559, opened 2018-11-29
fix(input): change onInputChange event to be type input

jonsadka:
Referencing `e.target.value` throws flow error `Cannot get e.target.value because property value is missing in EventTarget [1].`.  Input type should be updated to `SyntheticInputEvent`

### DIFF
--- a/src/select/types.js
+++ b/src/select/types.js
@@ -86,7 +86,7 @@ export type PropsT = {
   onBlurResetsInput: boolean,
   onChange: (params: OnChangeParamsT) => void,
   onFocus: (e: SyntheticEvent<HTMLElement>) => void,
-  onInputChange: (e: SyntheticEvent<HTMLInputElement>) => void,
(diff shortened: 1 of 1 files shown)


jonsadka:
Any idea why `buildkite/baseui/saucelabs-selenium` is failing? I don't have access

chasestarr:
looks like one of the clean-up processes timed out. just re-ran to see how it goes

```

1) The button component "after all" hook:
--
  | Error: Timeout of 20000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
  | at Timeout.<anonymous> (node_modules/mocha-nightwatch/lib/runnable.js:232:19)

```

jonsadka:
Awesome, thanks yall

chasestarr:
thanks for sending the pr @jonsadka

PR #563, opened 2018-11-29
feat: introduce Breadcrumbs component

austin94:
#### Minor: New Feature

This PR introduces the new `Breadcrumbs` component. To increase flexibility this component only applies styling and chevron separators between its children.

screen shot 2018-11-29 at 12 26 15 pm

#### Tests Added + Pass

Yes

#### License

MIT

### DIFF
--- a/src/breadcrumbs/__tests__/breadcrumbs.test.js
+++ b/src/breadcrumbs/__tests__/breadcrumbs.test.js
@@ -1,0 +1,37 @@
+/*
+Copyright (c) 2018 Uber Technologies, Inc.
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+
+// @flow
+import React from 'react';
+import {shallow} from 'enzyme';
+import {StyledLink} from '../../link';
+import {Breadcrumbs} from '../';
+
+describe('Breadcrumbs', () => {
+  it('applies correct accessibility attributes to root element', () => {
+    const example = shallow(
+      <Breadcrumbs>
+        <StyledLink href="#">Parent Page</StyledLink>
+        <StyledLink href="#">Sub-Parent Page</StyledLink>
+        <span>Current Page</span>
+      </Breadcrumbs>,
+    );
+    expect(example).toHaveProp('aria-label', 'Breadcrumbs navigation');
+  });
+
+  it('displays separators in correct positions', () => {
+    expect(
+      shallow(
+        <Breadcrumbs>
+          <StyledLink href="#">Parent Page</StyledLink>
+          <StyledLink href="#">Sub-Parent Page</StyledLink>
+          <span>Current Page</span>
+        </Breadcrumbs>,
+      ),
+    ).toMatchSnapshot();
+  });
+});
--- a/src/breadcrumbs/breadcrumbs.js
+++ b/src/breadcrumbs/breadcrumbs.js
@@ -14,7 +14,7 @@ import {getOverrides} from '../helpers/overrides';
 
 function Breadcrumbs({children, overrides = {}}: BreadcrumbsPropsT) {
   const numChildren = Children.count(children);
-  const mappedChildren = [];
+  const childrenWithSeparators = [];
 
   const [Root, baseRootProps] = getOverrides(overrides.Root, StyledRoot);
   const [Icon, baseIconProps] = getOverrides(overrides.Icon, StyledIcon);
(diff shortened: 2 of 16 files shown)


nadiia:
The font size seems to be different for active and non-active elements

src/breadcrumbs/styled-components.js
```diff
@@ -0,0 +1,34 @@
...
+  return {
+    color: $theme.colors.mono900,
+    ...$theme.typography.font350,
+  };
+});
+
+export const StyledSeparator = styled('div', ({$theme}: StyledSeparatorT) => {
+  return {
+    display: 'inline-block',
+    color: $theme.colors.mono700,
```
gergelyke: can can add a breadcrumb specific color, so it would work with both the light and dark theme? something like this: https://github.com/uber-web/baseui/blob/834307495cc9a3803acc09735e693ebef52811ee/src/themes/creator.js#L173

austin94: I've added the `breadcrumbsText` and `breadcrumbsSeparatorFill` to the theme.

Here is an updated dark mode:

screen shot 2018-11-29 at 1 21 28 pm

austin94:
> The font size seems to be different for active and non-active elements

I've bumped the typography to now match the `StyledLink` size. Should be good now.

chasestarr:

Built with commit d90946baacd5a911861036e2c0c494e7ee835f9d

https://deploy-preview-563--baseui.netlify.com

src/breadcrumbs/__tests__/breadcrumbs.test.js
```diff
@@ -0,0 +1,37 @@
...
+      <Breadcrumbs>
+        <StyledLink href="#">Parent Page</StyledLink>
+        <StyledLink href="#">Sub-Parent Page</StyledLink>
+        <span>Current Page</span>
+      </Breadcrumbs>,
+    );
+    expect(example).toHaveProp('aria-label', 'Breadcrumbs navigation');
+  });
+
+  it('displays separators in correct positions', () => {
```
blackswanny: not sure that it's clear from `toMatchSnapshot` expectation

nadiia:
Just to reiterate here, we should consider looking into how Breadcrumbs component can handle the look and feel of its items. Other than that this PR looks ok.

src/breadcrumbs/breadcrumbs.js
```diff
@@ -0,0 +1,49 @@
...
+
+  const [Root, baseRootProps] = getOverrides(overrides.Root, StyledRoot);
+  const [Icon, baseIconProps] = getOverrides(overrides.Icon, StyledIcon);
+  const [Separator, baseSeparatorProps] = getOverrides(
+    overrides.Separator,
+    StyledSeparator,
+  );
+
+  Children.forEach(children, (child, index) => {
+    mappedChildren.push(child);
```
blackswanny: I would say children are not mapped , but with separators

src/breadcrumbs/examples-list.js
```diff
@@ -0,0 +1,14 @@
...
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+
+/* eslint-env node */
+/* eslint-disable flowtype/require-valid-file-annotation */
+
+module.exports = {
+  DEFAULT: 'Breadcrumbs with Link component',
```
blackswanny: can I override to make vertical aligned breadcrumbs?

src/breadcrumbs/styled-components.js
```diff
@@ -0,0 +1,34 @@
...
+    display: 'inline-block',
+    color: $theme.colors.breadcrumbsSeparatorFill,
+    marginLeft: $theme.sizing.scale300,
+    marginRight: $theme.sizing.scale300,
+  };
+});
+
+export const StyledIcon = styled(ChevronRight, () => {
+  return {
+    verticalAlign: 'text-bottom',
```
blackswanny: it only works for inline parent element. If Separator's styles or tag is changed this won't work. How about to use flexbox model to style it?

src/breadcrumbs/examples-list.js
austin94: That should be possible since all styles and the icon can be controlled via `overrides`

src/breadcrumbs/breadcrumbs.js
austin94: Renamed.

src/breadcrumbs/styled-components.js
austin94: With the `flex` model it will center the icon relative to the container; not the text. This makes the icon look slightly off-center in some instances.  

Want to go over some possible styling offline and find a better match?

PR #564, opened 2018-11-29
fix: proper fallback for size prop

phou87:
Previously, if you passed in a string for $size, Icon would set the height and width to that string, resulting in an invalid CSS property (and causing the icon to not render). This is easy to repro by supplying an icon as a Button enhancer and leaving the size as the default. (Button spreads the $size prop to the Icon eventually)

This didn't get caught before because the examples use
`enhancer={() => <Icon />}` instead of `enhancer={Icon}`, which creates an intermediate component layer between Button and Icon, which "traps" the props and prevents them from being spread all the way down.


gergelyke:
thanks for the fix @phou87 !

PR #566, opened 2018-11-29
feat: introduce Notification component

austin94:
#### Minor: New Feature

This PR introduces the `Notification` (in-line notification) component.  It includes 4 styles: `primary (default), success, warning, and error`.

screen shot 2018-11-29 at 3 17 17 pm
screen shot 2018-11-29 at 3 17 27 pm
screen shot 2018-11-29 at 3 17 36 pm
screen shot 2018-11-29 at 3 18 20 pm

#### Tests Added + Pass

Yes

#### License

MIT

### DIFF
--- a/src/notification/examples.js
+++ b/src/notification/examples.js
@@ -1,0 +1,52 @@
+/*
+Copyright (c) 2018 Uber Technologies, Inc.
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+
+// @flow
+
+import React from 'react';
+import {Notification, KIND} from './';
+import examples from './examples-list';
+
+export default {
+  [examples.DEFAULT]: function Story1() {
+    return <Notification>This is a primary notification</Notification>;
+  },
+  [examples.SUCCESS]: function Story2() {
+    return (
+      <Notification kind={KIND.success}>
+        This is a success notification
+      </Notification>
+    );
+  },
+  [examples.WARNING]: function Story3() {
+    return (
+      <Notification kind={KIND.warning}>
+        This is a warning notification
+      </Notification>
+    );
+  },
+  [examples.ERROR]: function Story4() {
+    return (
+      <Notification kind={KIND.error}>
+        This is a warning notification
+      </Notification>
+    );
+  },
+  [examples.OVERRIDES]: function Story5() {
+    return (
+      <Notification
+        overrides={{
+          Root: {
+            style: {backgroundColor: 'pink', color: 'black'},
+          },
+        }}
+      >
+        This is a custom notification
+      </Notification>
+    );
+  },
+};
--- a/src/notification/examples.js
+++ b/src/notification/examples.js
@@ -32,7 +32,7 @@ export default {
   [examples.ERROR]: function Story4() {
     return (
       <Notification kind={KIND.error}>
-        This is a warning notification
+        This is an error notification
       </Notification>
     );
   },
--- a/src/notification/notification.js
+++ b/src/notification/notification.js
@@ -1,0 +1,31 @@
+/*
+Copyright (c) 2018 Uber Technologies, Inc.
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+
+// @flow
+
+import React from 'react';
+import type {NotifcationPropsT} from './types';
+import {KIND} from './constants';
+import {StyledRoot} from './styled-components';
+import {getOverrides} from '../helpers/overrides';
+
+function Notification({children, kind, overrides = {}}: NotifcationPropsT) {
+  const [Root, baseRootProps] = getOverrides(overrides.Root, StyledRoot);
+
+  return (
+    <Root role="alert" $kind={kind} {...baseRootProps}>
+      {children}
+    </Root>
+  );
+}
+
+Notification.defaultProps = {
+  overrides: {},
+  kind: KIND.primary,
+};
+
+export default Notification;
--- a/src/notification/styled-components.js
+++ b/src/notification/styled-components.js
@@ -1,0 +1,39 @@
+/*
+Copyright (c) 2018 Uber Technologies, Inc.
+
+This source code is licensed under the MIT license found in the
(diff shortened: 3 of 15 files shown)


gergelyke:
do you think we can build this on top of the Toast component with style overrides?

chasestarr:

Built with commit 0641690ea962381667ef81ed5951d8b67fd6b167

https://deploy-preview-566--baseui.netlify.com

nadiia:
> do you think we can build this on top of the Toast component with style overrides?

That was the plan. Not build on top tho but add additional types, since the current toast component is an inline notification.

austin94:
> do you think we can build this on top of the Toast component with style overrides?

I originally went down this path, but ran into some complexity.  With the amount of style overrides necessary it was easiest to just treat them as separate components. This also allows them to be more easily themed separately without Toast items cascading into Notification.

src/notification/examples.js
```diff
@@ -0,0 +1,52 @@
...
+    return (
+      <Notification kind={KIND.error}>
+        This is an error notification
+      </Notification>
+    );
+  },
+  [examples.OVERRIDES]: function Story5() {
+    return (
+      <Notification
+        overrides={{
```
blackswanny: we should be able to override animation of Notification. Can you add the test showing it?

nadiia:
The only two style changes here based on the `kind` will be `backgroundColor` and `color`, and probably set the `closeable` default value to `false` unless this prop is set explicitly.

src/notification/notification.js
```diff
@@ -0,0 +1,31 @@
...
+import type {NotifcationPropsT} from './types';
+import {KIND} from './constants';
+import {StyledRoot} from './styled-components';
+import {getOverrides} from '../helpers/overrides';
+
+function Notification({children, kind, overrides = {}}: NotifcationPropsT) {
+  const [Root, baseRootProps] = getOverrides(overrides.Root, StyledRoot);
+
+  return (
+    <Root role="alert" $kind={kind} {...baseRootProps}>
```
blackswanny: they say that not all notifications should have `alert` role as it is considered more intense. Think for notification not error (or even warning) we shouldn't use this role

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_alert_role

blackswanny:
how to close notification? Does it have timeout or optional closable button as Toast?

src/notification/examples.js
```diff
@@ -0,0 +1,52 @@
...
+*/
+
+// @flow
+
+import React from 'react';
+import {Notification, KIND} from './';
+import examples from './examples-list';
+
+export default {
+  [examples.DEFAULT]: function Story1() {
```
blackswanny: any example of more complex content layout in notification?

austin94:
> how to close notification? Does it have timeout or optional closable button as Toast?

These are in-inline notifications, not Toasts. Normally these are not able to be closed.

src/notification/styled-components.js
```diff
@@ -0,0 +1,39 @@
...
+
+import {styled} from '../styles/index';
+import type {StyledRootPropsT} from './types';
+import {KIND} from './constants';
+
+export const StyledRoot = styled('div', ({$theme, $kind}: StyledRootPropsT) => {
+  let color = $theme.colors.notificationPrimaryText;
+  let backgroundColor = $theme.colors.notificationPrimaryBackground;
+
+  if ($kind === KIND.success) {
```
blackswanny: this seems to be a `switch...case` statement

src/notification/notification.js
austin94: Maybe I'm miss-reading the docs, but `error` and `warning` are specifically shown using the role. Could you link to an example where a warning/error shouldn't have that role?

These should be used sparingly by design/dev, but if a notification is shown it should probably use the assertive tone to be read, right?

src/notification/types.js
```diff
@@ -0,0 +1,28 @@
...
+import type {OverrideT} from '../helpers/overrides';
+
+export type OverridesT = {
+  Root?: OverrideT<*>,
+};
+
+export type NotifcationPropsT = {
+  children?: Node,
+  overrides?: OverridesT,
+  kind: $Keys<typeof KIND>,
```
blackswanny: $Values is more correct, but we need to freeze it https://github.com/uber-web/baseui/issues/539

src/notification/styled-components.js
austin94: (I'm also used to an eslint rule that doesn't allow them 😄. This is now updated to a switch.

src/notification/types.js
austin94: We need to make this update in several areas. The majority of components are using `$Keys`.

austin94: Updated and frozen.

blackswanny:
> These are in-inline notifications, not Toasts. Normally these are not able to be closed.

Ok. Can you clarify with design that it's expected? Cause I anticipate that notifications can stack one by one taking space as I see them in some Uber apps. @nguyensomniac

src/notification/notification.js
blackswanny: Error and warning's are good. But as per design we have `primary` and `success` notifications. They don't look like `assertive`

austin94: The `assertive` value is specifically a `POLITENESS_SETTING` and has not direction relation to what is actually being displayed.

It only means that the user will be immediately told of the notification, which should probably occur if a success notification is triggered.

blackswanny: ok, makes sense

nguyensomniac:
> > These are in-inline notifications, not Toasts. Normally these are not able to be closed.
> 
> Ok. Can you clarify with design that it's expected? Cause I anticipate that notifications can stack one by one taking space as I see them in some Uber apps. @nguyensomniac

Yeah these aren't meant to be closed. In theory you could restyle a toast to look like an inline notification but you really shouldn't.

austin94:
Closing in favor of #574

PR #569, opened 2018-11-30
fix: Add ellipsis to long text in Select component

### LINKED ISSUE #568: [Select] Select component does not ellipsis long values
Selecting a value with long text in the `<Select>` component does not shorten the string
## Expected Behavior

String should be shortened via some method, preferably an ellipsis.

## Current Behavior

String extends beyond the edge of the container and makes it hard to see the icons

kapture 2018-11-29 at 16 45 10

jonsadka:
#### Fixed Issues?

Fixes #568

#### Tests Added + Pass?

Yes

#### Any Dependency Changes?
No

#### Before

kapture 2018-11-29 at 16 45 10

#### After

kapture 2018-11-29 at 16 46 08

### DIFF
--- a/src/select/styled-components.js
+++ b/src/select/styled-components.js
@@ -10,6 +10,7 @@ import {TYPE} from './constants';
 import {getSvgStyles} from '../icon/styled-components';
 import {SIZE} from './constants';
 import type {SharedStylePropsT} from './types';
+import {ellipsisText} from '../styles/util';
 
 function getFont(size = SIZE.default, typography) {
   return {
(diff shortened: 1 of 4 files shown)


src/select/styled-components.js
```diff
@@ -186,6 +196,7 @@ export const StyledSingleValue = styled('div', props => {
     whiteSpace: 'nowrap',
     maxWidth: '100%',
     ...getControlPadding($size, sizing, $type),
+    ...ellipsisText(),
```
chasestarr: maybe in-line if the properties are not dynamic?

src/select/styled-components.js
```diff
@@ -35,6 +35,16 @@ function getControlPadding(size = SIZE.default, sizing, type) {
   }[size];
 }
 
+function ellipsisText() {
```
blackswanny: I would like to reuse in other components. We can put it here src/styles/util.js

austin94: It we are going to reuse we may want to drop the `maxWidth` value

schnerd:
Nice!

PR #574, opened 2018-11-30
feat(notification): add `Notification` component on top of `Toast`

nadiia:
- modify toast component to have inline notification styles and `closeable` property set to `false` by default


docs/migrations/00574.md
```diff
@@ -0,0 +1,9 @@
+## Toast
+
+#### Changed the default `closeable` value from `true` to `false`. Also changed the default notification presentation type to 'inline' that results in default styles change. To render a notification as a toast the default props needs to be overwritten.
+
+```diff
+-  <Toast>Message</Toast>
++  import {TYPE} from 'baseui/toast'
++  <Toast notificationType={TYPE.toast} closeable={true}>Message</Toast>
```
gergelyke: can we do this in a non-breaking manner, by using defaults for the original behaviour?

nadiia:
The other way I was thinking is to export separately an
‘InlineNotification’ that returns a ‘Toast’ with correctly set up defaults.

gergelyke:
Probably that would be better, as folks won't have to set any prop, just use a drop-in component

rfcs/toast-component.md
```diff
@@ -90,6 +94,7 @@ export default () => {
 These properties are passed to every Toast's presentational (styled) component that is exported:
 
 * `$kind: KindTypeT`
+* `$type: TotificationTypeT`
```
chasestarr:
```suggestion
* `$type: NotificationTypeT`
```

PR #578, opened 2018-11-30
fix(toast): add component specific theme colors

nadiia:


gergelyke:
can you add values for the dark theme too?

nadiia:
> can you add values for the dark theme too?

Will need to clarify those with the designers first, will update once determined that it uses different primitives.

PR #581, opened 2018-12-01
fix(deps): move screener-storybook to dev dependencies

gergelyke:


gergelyke:
fixes https://github.com/uber-web/baseui/issues/580

PR #583, opened 2018-12-03
fix(components): add displayName

gergelyke:
after merging this, both the storybook source viewer and the chrome react devtools will show the real name of the component, instead of names like `Styled(div)`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,6 +18,7 @@ This document helps you navigate the process.
 
 2.  Once your RFC PR is approved, start implementing the component
 
+- You can find the component template in `src/template-component`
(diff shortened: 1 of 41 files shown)


blackswanny:
for some reason this rule is not `errored` in our eslint. Should we change setup to be error, so we will catch it ahead and be able to fix it during commits

gergelyke:
@chasestarr added it to the "template" component, so I'd rather not duplicate - what do you think?

chasestarr:
oh, i'm just learning about the template component now (by the title, i guessed it provided some template ui). maybe that should be called out in the doc instead?

gergelyke:
@chasestarr that's a great idea - I've added it to the doc

PR #589, opened 2018-12-05
fix(flow) - flow fix me md imports

chasestarr:
sets up config so that flow does not complain about .md imports


chasestarr:
extracted from https://github.com/uber-web/baseui/pull/588

chasestarr:
closing this in favor of #617

PR #590, opened 2018-12-05
docs(accordion) review accordion rfc

chasestarr:


src/accordion/README.md
```diff
@@ -0,0 +1,173 @@
+# Accordion Component
+
```
gergelyke: can we add a one liner here too please?

PR #596, opened 2018-12-05
docs(checkbox) review checkbox rfc

chasestarr:


src/checkbox/README.md
```diff
@@ -0,0 +1,200 @@
+# Checkbox Component
+
```
nadiia: 1-liner: `A checkbox control that is used for selecting multiple options from a set.`

PR #598, opened 2018-12-05
docs(header-navigation) review header navigation rfc

chasestarr:


src/header-navigation/README.md
```diff
@@ -61,22 +61,28 @@ export default (props) => <HeaderNavigation
...
+  * All the children of header navigation. It accepts any components, but it is useful to use `NavigationList` and `NavigationItem` to align and follow style guides.
+
 ## `StyledNavigationList` API
 
-* `align: ALIGN.right | ALIGN.left | ALIGN.center` - Optional.
-  Alignment of elements inside of navigation list
-* `children: React$Node` - Required.
-  All the children of NavigationList.  
-  
+* `align: $Values<ALIGN> = undefined`
```
nadiia: Should it be nullable type and the default set to `null`?

nadiia: Feels like the ` ALIGN.right | ALIGN.left | ALIGN.center` desription is more user-friendly, especially when someone looking at the docs doesn't use/not familiar with flow types.

chasestarr: yeah i agree it get's rid of some readability. we were not very consistent across rfcs in the pattern, so went with flow conventions. Although, i want some syntax to describe a set of constants (a pattern that comes up often). This would make maintaining rfcs in the future

chasestarr: do you have a suggestion for an alternative syntax?

nadiia: I see, agree about maintainability and a single pattern for constants. Maybe something like `ALIGN[key]` or `ALIGN[string]`?

nadiia: or `ALIGN[Object.keys(ALIGN)]`

PR #600, opened 2018-12-05
docs(input) review input rfc

chasestarr:


src/input/README.md
```diff
@@ -0,0 +1,188 @@
...
+* All properties of the StatefulContainer except `children`
+* All properties of the Input except `$components` that are optional and used to pass overrides for default styled components set.
+
+## `StatefulContainer` API
+
+* `children: (props: Props) => React.Node`
+* `initialState: {value: string} = {value: ''}`
+  * Initial state of an uncontrolled input component.
+    * `value` - an initial input value.
+* `stateReducer: (type: $Keys<STATE_CHANGE_TYPE>', nextState: {}, currentState: {}, e: SyntheticEvent<HTMLInputElement>) => stateToSet: {}` - Optional
```
nadiia: Ideally the `$Keys<STATE_CHANGE_TYPE>` should be  changed to`$Values`. But similar comment here about using `STATE_CHANGE_TYPE.change | null` instead of  `$Keys<STATE_CHANGE_TYPE>`.

PR #601, opened 2018-12-05
docs(menu) review menu rfc

chasestarr:


src/menu/README.md
```diff
@@ -0,0 +1,194 @@
+# Menu
+
```
gergelyke: can we add a 1-liner here on what this component provides?

PR #604, opened 2018-12-05
docs(pagination) review pagination rfc

chasestarr:


src/pagination/README.md
```diff
@@ -0,0 +1,85 @@
+# Pagination Component
+
```
gergelyke: can we add a one-liner here please?

PR #605, opened 2018-12-05
docs(popover) review popover rfc

chasestarr:


src/popover/README.md
```diff
@@ -230,7 +244,7 @@ export default () => (
 )
 ```
 
-However it honestly seems like the render prop approach is a better API in this case.
+However it seems like the render prop approach is a better API in this case.
```
gergelyke: should we remove this line entirely?

PR #607, opened 2018-12-05
docs(progress-steps) review progress-steps rfc

chasestarr:


src/progress-steps/README.md
```diff
@@ -63,31 +63,39 @@ export default () => {
 
 ## `ProgressSteps` API
 
-* `current: number` - Optional, Defaults to 0
-  Defines the current active step index
-* `overrides: {Root}` - Optional
-  Overrides for presentational components. See "Presentational Components Props API" below.
-  * `[ComponentName]: ReactComponent | {props: {}, style: {}, component: ReactComponent}` - Optional
+* `current?: number = 0`
```
nadiia: Same comment about keeping `Optional | Required` in props description. We can probably only explicitly specify `Required` and omit when a prop is optional

chasestarr: I'll follow up on this change in a later pr

PR #608, opened 2018-12-05
docs(radio) review radio rfc

chasestarr:


src/radio/README.md
```diff
@@ -0,0 +1,193 @@
+# RadioGroup Component
+
```
gergelyke: can we add a one-liner on the purpose of the component?

nadiia: `A radio control that is used to select a single option from set.`

PR #621, opened 2018-12-05
Docs/audit textarea rfc

chasestarr:


src/textarea/README.md
```diff
@@ -0,0 +1,135 @@
...
+* `SIZE`
+
+## StatefulTextarea API
+
+* All properties of the input's StatefulContainer except `children`
+* All properties of the Textarea comonent
+
+## Stateful container API (imported from `baseui/input`)
+
+* `children?: (props: Props) => React.Node = null`
```
nadiia: This one should be `Required`. Also since it's reexported from input, maybe we can link to the Input's readme here.

PR #628, opened 2018-12-07
feat(file-uploader): implements file uploader

chasestarr:
WIP still, but wanted to get an initial pr up here. Based quite a lot on jerome's work. Still some unknowns concerning various ui states. Only basic render has been provided.


src/file-uploader/styled-components.js
```diff
@@ -0,0 +1,58 @@
...
+    borderWidth: props.$theme.sizing.scale0,
+    boxSizing: 'border-box',
+    paddingTop: props.$theme.sizing.scale900,
+    paddingRight: props.$theme.sizing.scale800,
+    paddingBottom: props.$theme.sizing.scale900,
+    paddingLeft: props.$theme.sizing.scale800,
+  };
+});
+
+export const FileSelectButton = styled('button', (props: StylePropsT) => {
```
gergelyke: can we use here the Button component?

src/file-uploader/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+* `$disabled: boolean`
+* `$multi: boolean`
+* `$isFocused: boolean`
+* `$error: boolean`
+* `$theme: theme`
+
+## Dependencies
+
+Does this component depend on any 3rd party packages or other internal components?
```
gergelyke: should we add `attr-accept` here?

package.json
```diff
@@ -125,6 +125,7 @@
     "webpack-cli": "^3.1.0"
   },
   "dependencies": {
+    "attr-accept": "^1.1.3",
```
gergelyke: I wonder if we should use https://www.npmjs.com/package/accepts? seems like more used / maintained

chasestarr:
it may be worth holding off on dev here until headless dropzone is released https://github.com/react-dropzone/react-dropzone/pull/717

that would streamline implementation to only include presentation

src/file-uploader/file-uploader.js
```diff
@@ -0,0 +1,167 @@
...
+      if (this.props.onChange) {
+        this.props.onChange({...nextFiles, event});
+      }
+
+      this.setState({
+        acceptedFiles: [
+          ...this.state.acceptedFiles,
+          ...nextFiles.acceptedFiles,
+        ],
+        rejectedFiles: [
```
nadiia: Why do we need to maintain the list of rejected files?

src/file-uploader/file-uploader.js
```diff
@@ -0,0 +1,167 @@
...
+            onDragEnter={this.onDragEnter}
+            onDragLeave={this.onDragLeave}
+            onDragOver={this.onDragOver}
+            $isDragActive={this.state.isDragActive}
+          >
+            <Block font="font450">Drop files here to upload</Block>
+            <Block font="font450" color="mono600">
+              or
+            </Block>
+            <FileSelectButton onClick={this.onClick}>
```
nadiia: How about making it a label to the hidden input?

src/file-uploader/file-uploader.js
```diff
@@ -0,0 +1,167 @@
...
+  render() {
+    return (
+      <Root>
+        <ContentContainer>
+          <FileDragAndDrop
+            onDrop={this.onDrop}
+            onDragEnter={this.onDragEnter}
+            onDragLeave={this.onDragLeave}
+            onDragOver={this.onDragOver}
+            $isDragActive={this.state.isDragActive}
```
nadiia: it should be passed to all elements

src/file-uploader/file-uploader.js
```diff
@@ -0,0 +1,167 @@
...
+  onWindowDragOverOrDrop = (event: SyntheticDragEvent<HTMLElement>) => {
+    event.preventDefault();
+  };
+
+  render() {
+    return (
+      <Root>
+        <ContentContainer>
+          <FileDragAndDrop
+            onDrop={this.onDrop}
```
nadiia: Would it make sense to make the whole area droppable? So the onDrag handlers will be on a container component.

package.json
chasestarr: i chose this package to be consistent with `react-dropzone` https://github.com/react-dropzone/react-dropzone/blob/master/package.json#L84 which, according to a reviewer on jerome's pr, is commonly used across the FE org

src/file-uploader/file-uploader.js
chasestarr: this component simply wraps the `react-dropzone` package, so re-using that api

src/file-uploader/file-uploader.js
chasestarr: i've changed the markup here, so not sure if the comment still applies. but i wanted to make an outer wrapper that contains all of the elements, and just the area with the dashed border accepts drops. i'd like to display this in the design meeting, so will accept feedback on how it should operate

src/file-uploader/file-uploader.js
```diff
@@ -0,0 +1,150 @@
...
+          ...styleProps
+        } = renderProps;
+
+        const prefixedStyledProps = prependStyleProps({
+          ...styleProps,
+          isDisabled: props.disabled,
+        });
+
+        return (
+          <Root>
```
nadiia: Should pass `prefixedStyledProps ` here as well

src/file-uploader/styled-components.js
```diff
@@ -0,0 +1,61 @@
...
+export const StyledFileDragAndDrop = styled('div', (props: StylePropsT) => {
+  return {
+    alignItems: 'center',
+    backgroundColor: props.$theme.colors.fileUploaderBackgroundColor,
+    borderColor: props.$isDragActive
+      ? props.$theme.colors.fileUploaderBorderColorActive
+      : props.$theme.colors.fileUploaderBorderColorDefault,
+
+    borderStyle: 'dashed',
+    borderRadius: props.$theme.borders.radius200,
```
nadiia: Not sure if we need to have the `$theme.borders.useRoundedCorners` check here. Something to consider.

src/file-uploader/styled-components.js
```diff
@@ -0,0 +1,61 @@
...
+    paddingRight: props.$theme.sizing.scale800,
+    paddingBottom: props.$theme.sizing.scale900,
+    paddingLeft: props.$theme.sizing.scale800,
+    width: '100%',
+  };
+});
+StyledFileDragAndDrop.displayName = 'StyledFileDragAndDrop';
+
+export const StyledContentMessage = styled('div', (props: StylePropsT) => ({
+  ...props.$theme.typography.font450,
```
nadiia: If we set the font on the `StyledRoot` element then the inner divs should just inherit it.

tajo:
Nice job! Is this expected for a disabled state?

disabled_outline

chasestarr:
the inline styles are applied by the `react-dropzone` package

PR #630, opened 2018-12-07
fix(block): enumerate types for css properties

chasestarr:
context: 

```
BaseUI <Block/> component `justifyContent={'end'}` doesn't put content to the right side. Change it to `justifyContent={'flex-end'}` works but give flow error. 
```

considered simply allowing just a `string`, but thought this might be helpful with ide autocomplete


src/block/README.md
```diff
@@ -48,22 +48,22 @@ All other props are spread onto the base element.
...
+* `justifySelf?: [available values](https://developer.mozilla.org/en-US/docs/Web/CSS/justify-self)`
+* `position?: 'static' | 'absolute' | 'relative' | 'fixed' | 'sticky'`
 * `width?: string`
 * `minWidth?: string`
 * `maxWidth?: string`
 * `height?: string`
 * `minHeight?: string`
 * `maxHeight?: string`
-* `overflow?: 'visible' | 'hidden' | 'scroll' | 'scrollX' | 'scrollY' | 'auto'`
+* `overflow?: 'visible' | 'hidden' | 'scroll' | 'scrollX' | 'scrollY' | 'auto' | 'inherit' | 'initial' | 'unset'`
```
nadiia: Should we make it similar description with a [link to ->](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow)?

chasestarr: block does not follow css convention for this property https://github.com/uber-web/baseui/blob/master/src/block/styled-components.js#L129-L135 so wanted to leave it separate

PR #632, opened 2018-12-07
docs(table): update table rfc based on feedback

chasestarr:
Thanks @dswho for making the first pass on this rfc!

cc @CharlieMartell @ericsoco for feedback


proposals/table-component.md
```diff
@@ -0,0 +1,221 @@
...
+* `StyledCell`
+
+## `Table` API
+
+* `columns: Array<{title: string, dataIndex: number, render?: (rowData: Array<any>) => React.Node, sortComparator?: (a: Array<any>, b: Array<any>) => number}>` - Required
+  * Describes how the column of data sorts and corresponding cells render.
+* `data: Array<Array<any>>` - Required
+  * Matrix data to be rendered within the table.
+* `estimatedRowSize?: number = 40px`
+  * Passed to `react-virtualized`. Used to estimate table height.
```
chasestarr: @CharlieMartell @ericsoco interested in your input on the react-virtualized specific api

another is `useDynamicRowHeight` on L184

proposals/table-component.md
```diff
@@ -0,0 +1,221 @@
...
+
+Component to render tabular data.
+
+## Features
+
+* Sorting
+
+## Currently Unsupported
+
+* Pagination
```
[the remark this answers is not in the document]
chasestarr: pagination is possible with the current api, see the second 'advanced' example. it would not be a tight ui integration at the moment. 

unifying the pagination story should only include new properties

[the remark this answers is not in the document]
gergelyke: cool, thanks!

I meant the whole `Currently unsupported` section - does it apply to them too?

proposals/table-component.md
```diff
@@ -0,0 +1,221 @@
...
+* `dataIndex?: number`
+  * Attribute at which to index the row data. Must provide a `sortComparator` or column may not be sorted.
+* `render: <T>(<T>, index, Array<T>) => $ReactNode`
+  * Customize the cells for this column.
+* `sortComparator?: (a: Array<any>, b: Array<any>`
+  * Defaults to [local sort](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort) at provided `dataIndex`.
+
+## Dependencies
+
+* [react-virtualized](https://bvaughn.github.io/react-virtualized/#/components/List)
```
gergelyke: should we look into https://github.com/bvaughn/react-window#frequently-asked-questions?

chasestarr: i considered for quite a while if react-window was worth it. 

we would be pulling in a much smaller sized package.. but `react-virtualized` has some nice features (called out in the rfc) that `react-window` does not support.

gergelyke: 👍

proposals/table-component.md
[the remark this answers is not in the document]
chasestarr: not that i can think of

proposals/table-component.md
```diff
@@ -0,0 +1,221 @@
...
+
+const data = [
+  ['1', 'Sarah', 'Brown', 31, 'New York No. 1 Anywhere'],
+  ['2', 'Jane', 'Smith', 32, 'San Francisco No. 1 Anywhere'],
+  ['3', 'Joe', 'Black', 33, 'Sydney No. 1 Anywhere'],
+  ['4', 'Jane', 'Red', 34, 'London No. 1 Anywhere'],
+];
+
+const columns = [
+  {
```
tajo: Missing `dataIndex`?

proposals/table-component.md
```diff
@@ -0,0 +1,221 @@
...
+  onPageChange = (nextPage, prevPage) => {
+      const {pageSize} = this.state;
+      this.fetch(nextPage, pageSize);
+  }
+
+  render = () => {
+    return (
+        <Table
+          columns={columns}
+          dataSource={this.state.data}
```
tajo: What's the difference between `dataSource` and `data`?

proposals/table-component.md
```diff
@@ -0,0 +1,221 @@
...
+
+const data = [
+  [1, 'John', 'Smith'],
+  [2, 'Lois', 'Lane'],
+];
+
+const columns = {
+  {
+    title: 'Employee ID',
+    dataIndex: 1,
```
tajo: What is the `dataIndex` used for? Could `columns` be just an array?

proposals/table-component.md
```diff
@@ -0,0 +1,221 @@
...
+
+* Pagination
+* Filtering
+* Fixed row
+* Fixed column
+* Striped rows
+* Hoverable row styling
+* Multi-column sort
+* Row headers
+* Expandable row
```
tajo: Are we planning to use `<table>`, `<tr>`, `<td>`... elements to implement our table? Expandable row is pretty hard to implement with those (I had an experience ending up with two different components - `table` and `expandable-table` and we always regretted to not build it as a single component from the start). If we want to be more future-proof, we could experiment with grid as a replacement. It would also help with responsiveness.

proposals/table-component.md
chasestarr: my thought is that it should be optional, if a `render` function is provided. what do you think?

proposals/table-component.md
chasestarr:
```suggestion
          data={this.state.data}
```

chasestarr: just a typo :)

proposals/table-component.md
chasestarr: i considered relying on position in the array, but it's likely that the data may not be ordered in the way it should be rendered. similarly, a column may be a combination of multiple pieces of data from the provided row

proposals/table-component.md
chasestarr: good ideas. thanks for this

chasestarr:
regarding loading states, do you think a simple `isLoading` prop would be sufficient for enabling that ui?

proposals/table-component.md
tajo: I think I understand it now - `dataIndex` serves as a selector for `data`. Then, this makes sense to me. 

We could also consider to make the `table` more "dumb" and let our users to slice and dice the data before they pass it in.  I feel that `dataIndex` and `render` construct doesn't save that much work unless I am missing something - the same thing can be achieved with something as 

```js
data={data.map(x => x.slice(1,4))}
```

or with re-ordering

```js
data={data.map(x => [x[3], x[1], x[2]])}
```


Also, it's one less API to learn/remember and would make our component simpler.

chasestarr: cool, i like this pattern for getting rid of `dataIndex`. does this replace render? is something like below what you have in mind?

```jsx
<Table
  data={data.map(x => [
    (<Cell>{x[1]}</Cell>),
    (<CustomCell>{x[2]}</CustomCell>),
  ])}
/>
```

tajo: That's a different thing and valid concern. `render` for column would give you option to swap the cell component for a specific column but what if I want to customize a specific row? Or one of the head cells? I can do overrides only for all head cells or all rows, right? I think we can generalize tables into these two patterns:

## Build your own table

```jsx
import {
  StyledRoot
  StyledHead
  StyledHeadCell
  StyledBody
  StyledRow
  StyledCell
} from 'baseui/table';

export default () =>
  <StyledRoot>
    <StyledHead>
      <StyledHeadCell>Column 1</StyledHeadCell>
      <StyledHeadCell>Column 2</StyledHeadCell>
    </StyledHead>
    <StyledBody>
      <StyledRow>
        <StyledCell>Foo</StyledCell>
        <StyledCell>Baz</StyledCell>
      </StyledRow>
    </StyledBody>
  </StyledRoot>;
```

This way, the user can customize/replace any node in the whole table. If there is no need for that, they can use `TableBuilder`.

## Use the table builder

```jsx
import { TableBuilder } from 'baseui/table';

export default () =>
  <TableBuilder 
    columns={['Column 1', 'Column 2']}
    data={[['Foo', 'Baz']]}
  />;
```

Internally, `TableBuilder` uses components from the previous example. It saves you a lot of imports and typing.

The idea is that `TableBuilder` should cover like 80% use-cases (the API can be a bit more permissive to achieve that). Also, you can still use our overrides to change styles for the whole class of components (for example, you could modify all `StyledHeadCells` if you wish).

In the rest 20%, we give you atomic components to build-your-own-thing.

The `render` prop for columns or this

```jsx
<TableBuilder
  data={data.map(x => [
    (<Cell>{x[1]}</Cell>),
    (<CustomCell>{x[2]}</CustomCell>),
  ])}
/>
```

seems like something in between those two worlds. It's hard to tell how much it's going to be used and makes our component/API pretty complex. I would make the first iteration as simple as possible - exporting atomic elements and minimalistic `TableBuilder` and add more features later.

**TLDR;** I am pretty much just describing what's already being proposed here. My only concern is if we want to immediately start mixing `TableBuilder` (or just `Table` in our proposal) with pass-your-own-components or should we just keep it more locked down, so you can pass only the data. There' s a lot of unsupported features we will probably want to eventually add and more complex the API is now, harder it will get.

chasestarr: thanks for putting so much thought into this response, and I agree. let's try to get a simple table implementation out and consider additional features as the needs arise. will update the rfc

proposals/table-component.md
dswho: One concern is regarding the behavior of `colspan` and `rowspan`. Can this be done easily with layout grid? I've had trouble implementing in the past using `display:table` and don't have too much experience with grid.

Also will `layout:fixed` and `layout:auto` be affected by using `grid` layout? I need the default behavior of having cells auto-size based on cell content.

One implementation of an expandable table uses native `<table>`, `<td>`, `<tr>` elements. This is in the `antd` table: https://ant.design/components/table/ which uses [rc-table](https://github.com/react-component/table) under the hood.

tajo: > One concern is regarding the behavior of colspan and rowspan. Can this be done easily with layout grid?

Yes! CSS Grid has `span` keyword that you can use in `grid-row` and `grid-column` to span over rows/columns: https://css-tricks.com/snippets/css/complete-guide-grid/#article-header-id-27

> Also will layout:fixed and layout:auto be affected by using grid layout? I need the default behavior of having cells auto-size based on cell content.

This is probably where grid shines the most. You can combine absolute values, percentage and [`fr` units](https://css-tricks.com/snippets/css/complete-guide-grid/#fr-unit) so any type of layout should be possible.

proposals/table-component.md
```diff
@@ -0,0 +1,175 @@
...
+```javascript
+import React from 'react';
+import {Table} from 'baseui/table';
+
+const data = [
+  ['1', 'Sarah', 'Brown', 31, 'New York No. 1 Anywhere'],
+  ['2', 'Jane', 'Smith', 32, 'San Francisco No. 1 Anywhere'],
+  ['3', 'Joe', 'Black', 33, 'Sydney No. 1 Anywhere'],
+  ['4', 'Jane', 'Red', 34, 'London No. 1 Anywhere'],
+].map(row => [`${row[1] row[2]}`, row[3], row[4]]); // selects data to display
```
gergelyke:
```suggestion
 ].map(row => [`${row[1]} ${row[2]}`, row[3], row[4]]); // selects data to display
```

PR #633, opened 2018-12-07
docs(versioning): add versioning policy

gergelyke:


docs/pages/introduction/versioning.md
```diff
@@ -0,0 +1,18 @@
...
+
+We recognize that you need stability from the Base UI.
+
+This document describes the practices that we follow to provide you the Base UI library. We want everyone who depends on Base UI to know when and how new features are added, and to be well-prepared when obsolete ones are removed.
+
+Development builds of Base UI include many helpful warnings. Whenever possible, we add warnings in preparation for future breaking changes. These warnings do not affect the production version of your application.
+
+We are following [SemVer](https://semver.org), with a few flavours. We do not bump major versions for the following changes:
+
+- Any component or functions that's prefixed with `Unstable_` / `unstable_` may change or got removed without notice,
```
chasestarr: Just so i'm clear on what you're thinking, i would import an unstable component like below?

```js
import {Unstable_FileUploader} from 'baseui/file-uploader';
```

docs/pages/introduction/versioning.md
```diff
@@ -0,0 +1,18 @@
...
+- Visual changes, like colors and sizes,
```
tajo: Agreed with colors. Sizes can be tricky though. If you increase a size of some component it can unexpectedly wrap the rest of content and break the whole experience - something that users would definitely report as a bug.  We should probably elaborate more on this point.

docs/pages/introduction/versioning.md
gergelyke: yes, I think so

or should we just go for `UnstableFileUploader`?

tajo: `Unstable_FileUploader` might be more obvious when scrolling through the codebase.

docs/pages/introduction/versioning.md
nadiia: Should we also mention something like any css style changes?

tajo: On the other thought, since `overrides` is so powerful concept, we will probably have plenty of breaking changes even when considering only non-visual/API changes.

gergelyke: yeah, in our case markup changes for the components would automatically trigger major bumps

nadiia: I'd vote for no major bump for style changes at this point

gergelyke: added

PR #640, opened 2018-12-10
fix(popover): fix enzyme matcher warning in tests

schnerd:
`yarn test` currently has the following warning for popover:

```
 PASS  src/popover/__tests__/popover.test.js
  ● Console

    console.warn node_modules/enzyme-matchers/lib/utils/deprecate.js:19
      Matcher toHaveTagName is deprecated and will be removed in a future release. Use the replacement, toHaveDisplayName instead.
```

This PR migrates from toHaveTagName to toHaveDisplayName to fix this issue.


schnerd:
> awesome! what did you run to update flow-typed?

```
flow-typed install --overwrite jest@23.4.2
```

Though for some reason that updated all deps, so I just committed the jest change and ignored all the other updates.

PR #641, opened 2018-12-10
docs(breakpoints): add responsive style props proposal

tajo:
**This is a proposal** - I want to start discussion. If `docs/pages` is not a right place for it, I'm happy to move it.


chasestarr:
this seems like a reasonable approach to me. anecdotally, folks i've spoken with who use `styled-system` are super into it.

gergelyke:
should we make this document into `proposals`? or does it already work this way?

chasestarr:
yeah, it should probably move into the `proposals` directory because once in master, this will render on the docs site.

tajo:
@gergelyke it's not implemented yet, so I moved it into `proposals`

PR #642, opened 2018-12-12
feat(input): implements the input mask feature

chasestarr:
```jsx
<MaskedInput placeholder="Phone number" mask="(999) 999-9999" />
```


gergelyke:
Two high-level feedback:

- is there a way to add the extra dependency as a package that's tree-shakable? it seems like this package adds ~4kb (that translates to 100ms on slower 3g networks): https://bundlephobia.com/result?p=react-input-mask@2.0.4

- probably I'd remove the date formatter example, as we are about to add the date picker (so it would be less confusing what to use for what)

what do you think?

chasestarr:
do you have an example in baseui where we've organized the code to be shakable?

from what i understand in webpack's [docs](https://webpack.js.org/guides/tree-shaking/), tree shaking only happens at the module import boundary. this feature will always require the mask code to be available if we want an encapsulated api. 

we could recommend decoupling this feature from the baseui package
https://codesandbox.io/s/znorz94yom

another alternative could be to export a `MaskedInput` component to provide as an `Input` override
https://codesandbox.io/s/pyqjw22740

gergelyke:
Yeah, I don't have a strong preference, I am ok with including it too - I just wanted to start the discussion to see what other folks think / would recommend.

src/input/base-input.js
```diff
@@ -100,34 +106,54 @@ class BaseInput<T: EventTarget> extends React.Component<
...
+
+    if (type === CUSTOM_INPUT_TYPE.textarea) {
+      return (
+        <InputContainer {...sharedProps} {...inputContainerProps}>
+          <Before {...sharedProps} {...beforeProps} />
+          <Input
+            {...sharedProps}
+            {...this.getInputProps()}
+            {...inputPropsFromOverride}
+            value={value}
```
nadiia: The `value` is included in the `this.getInputProps`. Is there a reason we want to pass `value` after the override props?

src/input/base-input.js
```diff
@@ -100,34 +106,54 @@ class BaseInput<T: EventTarget> extends React.Component<
...
-        <After {...sharedProps} {...afterProps} />
-      </InputContainer>
+      <InputMask mask={mask} maskChar={maskChar} {...this.getInputProps()}>
+        {inputProps => {
+          return (
+            <InputContainer {...sharedProps} {...inputContainerProps}>
+              <Before {...sharedProps} {...beforeProps} />
+              <Input
+                {...sharedProps}
+                {...inputProps}
```
nadiia: We probably still want to spread `this.getInputProps()` before those from render props.

nadiia:
+1 for adding 'MaskInput` as an additional component in exports.

src/input/base-input.js
chasestarr:
```suggestion
```

src/input/base-input.js
chasestarr: `inputProps` shuttles the props provided to the `InputMask` through to the `Input` component

chasestarr:
separated out into it's own component called `MaskedInput` ! Good feedback on the initial implementation. Thanks folks

PR #645, opened 2018-12-13
Change default type ramp to be consistent with Uber move theme

### LINKED ISSUE #226: [Proposal] Use system font stack for Base UI
Design is revisiting the type for the internal version of Base UI, and as part of that we'd like to take a stronger opinion on a typeface for the open-source version, i.e. using the system font stack instead of Helvetica.

Looking at what the new type looks like compared to various other typefaces (San Francisco, Helvetica) you can see, the new font for body text is proportionally similar to San Francisco, which would help make the open-source and non-open-source versions look more consistent, at least on Macs. And, as a plus, it would help make re-creating layouts much easier :)

Would love to hear some thoughts—we're still in the early stages of all of this.

nguyensomniac:
Hi all!

There's a few small PRs I want to open to help sync type styles with what designers currently use. The first step is to get both the internal and external type ramps consistent with each other.

This way we can tweak type styles without any negative side effects to the OS version of baseui and vice versa. In addition, fixes #226 so the metrics of the type we use match up better.

I suppose this is technically a breaking change...but I'd rather rip the bandaid off sooner so we can keep things in sync.

#### Fixed Issues?

#226 

#### License

MIT

### DIFF
--- a/src/themes/creator.js
+++ b/src/themes/creator.js
@@ -222,25 +222,25 @@ export default function createTheme(
         fontFamily: primitives.primaryFontFamily,
         fontSize: '14px',
         fontWeight: 'normal',
-        lineHeight: '24px',
+        lineHeight: '20px',
       },
       font350: {
         fontFamily: primitives.primaryFontFamily,
         fontSize: '14px',
         fontWeight: 'bold',
-        lineHeight: '24px',
+        lineHeight: '20px',
       },
       font400: {
         fontFamily: primitives.primaryFontFamily,
         fontSize: '16px',
         fontWeight: 'normal',
-        lineHeight: '28px',
+        lineHeight: '24px',
       },
       font450: {
         fontFamily: primitives.primaryFontFamily,
         fontSize: '16px',
         fontWeight: 'bold',
-        lineHeight: '28px',
+        lineHeight: '24px',
       },
       font500: {
         fontFamily: primitives.primaryFontFamily,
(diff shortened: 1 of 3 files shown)


src/themes/creator.js
```diff
@@ -256,27 +256,33 @@ export default function createTheme(
...
         fontFamily: primitives.primaryFontFamily,
-        fontSize: '28px',
+        fontSize: '32px',
         fontWeight: 'bold',
-        lineHeight: '40px',
+        lineHeight: '48px',
       },
       font800: {
         fontFamily: primitives.primaryFontFamily,
-        fontSize: '32px',
```
nadiia: We have a separate primitives list that are used for uber move theme creation.

chasestarr:
according to [versioning policy](https://baseui.design/?selectedKind=Introduction&selectedStory=Versioning%20policy&full=0&addons=1&stories=1&panelRight=0&addonPanel=REACT_STORYBOOK%2Freadme%2Fpanel) css change does not require a major bump

src/themes/creator.js
nguyensomniac: Yup! For type, the move and non-move primitives should be almost identical. Design isn't going to support this separate type ramp moving forward, and ideally the OS and non-OS versions should look as close to each other as possible.

nguyensomniac:
> according to [versioning policy](https://baseui.design/?selectedKind=Introduction&selectedStory=Versioning%20policy&full=0&addons=1&stories=1&panelRight=0&addonPanel=REACT_STORYBOOK%2Freadme%2Fpanel) css change does not require a major bump
> 
> also, tests are failing on a jest snapshot expecting the old font value. you'll need to run the command below to update snapshots and commit the changes
> 
> `yarn test -u`

sounds good! didn't know we had a versioning policy, thanks for the link

gergelyke:
maybe I am missing something here, but shouldn't we remove the font overrides from the move theme then? @nguyensomniac @chasestarr

nguyensomniac:
good point. will open a separate PR!

PR #648, opened 2018-12-13
fix: Various type fixes

nguyensomniac:
Fixes lots of little type/style bugs that became inconsistent with Figma.

- Uber Move theme uses 500 instead of 700 for bold. I kept the default light theme at 700 since fonts without a 500 weight fall back to regular weight.

- Buttons use a slightly bigger label font

- Toasts/inline notifs become bold

- Checkboxes use same fill/stroke when disabled

- Cards have consistent heading sizes

- Tags use font250 instead of font200 + bold


gergelyke:
thanks @nguyensomniac ! would you mind either splitting into multiple PRs, or add the affected components to the commit message?

nguyensomniac:
> thanks @nguyensomniac ! would you mind either splitting into multiple PRs, or add the affected components to the commit message?

done 👍

PR #650, opened 2018-12-13
fix: Type stack fixes

nguyensomniac:
Addresses @gergelyke's comments in #645 and changes fallback fonts to be more consistent.

Removes Open Sans from our font stack and replaces it with `system-ui`, since from my knowledge Open Sans doesn't come preinstalled on any OS and is therefore redundant.

### DIFF
--- a/src/themes/light-theme-with-move.js
+++ b/src/themes/light-theme-with-move.js
@@ -21,83 +21,32 @@ export const LightThemeMove = createTheme(
   },
   {
     typography: {
-      font200: {
-        fontFamily: primaryFontFamily,
-        fontSize: '12px',
-        fontWeight: 'normal',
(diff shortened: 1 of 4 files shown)


src/themes/light-theme-with-move.js
```diff
@@ -9,10 +9,10 @@ import createTheme from './creator.js';
 import {primitives as lightThemePrimitives} from './light-theme-primitives.js';
 
 const primaryFontFamily =
-  'UberMoveText, "Open Sans", "Helvetica Neue", Helvetica, sans-serif';
+  'UberMoveText, system-ui, "Helvetica Neue", Helvetica, Arial, sans-serif';
```
nadiia: 👍

PR #653, opened 2018-12-14
feat(tabs): add tabs implementation

anantkamath:
Add implementation of the tabs component

screen shot 2018-12-15 at 12 15 42 am


src/tabs/stateful-tabs.js
```diff
@@ -0,0 +1,67 @@
...
+      return initialState.activeKey;
+    } else {
+      return React.Children.map(
+        children,
+        (child, index) => child.key || String(index),
+      )[0];
+    }
+  };
+
+  state = {
```
nadiia: Could you move the state definition up below the `defaultProps`

src/tabs/stateful-tabs.js
```diff
@@ -0,0 +1,67 @@
...
+    // $FlowFixMe
+    activeKey: this.getInitialKey(),
+  };
+
+  render() {
+    const {initialState, stateReducer, ...restProps} = this.props;
+    return (
+      <Tabs
+        {...restProps}
+        // $FlowFixMe
```
nadiia: Would it be possible to fix this flow error?

src/tabs/styled-components.js
```diff
@@ -0,0 +1,88 @@
...
+    color: $active ? colors.black : colors.mono800,
+    cursor: $disabled ? 'not-allowed' : 'pointer',
+    height: sizing.scale1200,
+    paddingTop: sizing.scale550,
+    paddingLeft: sizing.scale300,
+    paddingRight: sizing.scale300,
+    marginLeft: sizing.scale200,
+    marginRight: sizing.scale200,
+    borderBottom:
+      $orientation === ORIENTATION.horizontal && $active
```
nadiia: We still need to make an active tab distinguished for the vertical orientation. How about we set the right border for the vertically positioned and active tab. 
I didn't find a design for a vertical case but maybe @nguyensomniac can provide more design details on it?

src/tabs/styled-components.js
```diff
@@ -0,0 +1,88 @@
...
+        ? `2px solid ${colors.primary400}`
+        : 'none',
```
nadiia: How about for non-active tabs we still set the border to `2px` but make the color `transparent`? I see that there is no height jump since we set a fixed height, but for consistency seems to be a good approach and for the case when the height of the tabs is overwritten to be not fixed.

src/tabs/styled-components.js
```diff
@@ -0,0 +1,88 @@
...
+    $active,
+    $orientation,
+    $theme: {colors, sizing, typography},
+  } = props;
+  let style = {
+    ...typography.font300,
+    boxSizing: 'border-box',
+    color: $active ? colors.black : colors.mono800,
+    cursor: $disabled ? 'not-allowed' : 'pointer',
+    height: sizing.scale1200,
```
nadiia: Do we actually want the height to be fixed? In the case of a very long title it seems not to work well.

src/tabs/tab-panel.js
```diff
@@ -0,0 +1,22 @@
...
+import type {TabPropsT} from './types.js';
+
+class TabPanel extends React.Component<TabPropsT> {
+  static defaultProps: $Shape<TabPropsT> = {
+    disabled: false,
+    onSelect: () => {},
+    onClick: () => {},
+    onKeyDown: () => {},
+    title: '',
+  };
```
nadiia: Should we have a `render` here?

src/tabs/tabs.js
```diff
@@ -0,0 +1,117 @@
...
+      if (!child) return;
+      const key = child.key || String(index);
+      const props = {
+        key,
+        id: key, // for aria-labelledby
+        active: key === activeKey,
+        disabled: disabled || child.props.disabled,
+        $orientation: orientation,
+        overrides: child.props.overrides,
+        onSelect: () => this.onChange(key),
```
nadiia: Let's change the arg passed to be an object: `this.onChange({key: key})` or `this.onChange({activeKey: key})`

src/tabs/styled-components.js
anantkamath: Yep, would love some ideas/details for highlighting active vertical tabs.
FWIW, the base ui figma only seems to show text color changes (black for active and gray for inactive tabs) and no borders for vertical tabs.

chasestarr: screen shot 2018-12-17 at 11 01 00 am

src/tabs/tab-panel.js
chasestarr: huh, yeah not quite sure how this is working

src/tabs/__tests__/tab.test.js
```diff
@@ -0,0 +1,101 @@
...
+        .fn()
+        .mockImplementation(({children}) => <span>{children}</span>),
+    };
+    // $FlowFixMe
+    const wrapper = mount(<Tab overrides={overrides}>Title</Tab>);
+    const tab = wrapper.find(overrides.Tab);
+    expect(tab).toHaveLength(1);
+  });
+
+  test('onChange is called on a click event', () => {
```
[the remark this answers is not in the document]
anantkamath: Oops, meant onSelect

src/tabs/styled-components.js
nguyensomniac: yup, that looks right!

src/tabs/README.md
```diff
@@ -0,0 +1,159 @@
...
+
+* `title: React$Node` - Required.
+  Title of the Tab to be shown in the Tab bar.
+* `children: React$Node` - Optional.
+  Content of the tab to be shown when it is selected. (Can be omitted if using `Tabs` as an externally controlled component for example)
+* `key: string` - Optional.
+  Unique key for the tab. Defaults to the child index.
+* `disabled: boolean` Optional.
+  Defaults to the `disabled` value passed from the parent`Tabs` component
+* `onSelect: () => {}` - Optional.
```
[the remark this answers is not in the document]
anantkamath: a) To keep the style of the API consistent with other baseui components, like accordion for example

b) For optional finer grained control as a principle (whatever the user's use case may be ). As it is, the API allows folks to only listen to onSelect if they want, and skip listening onClick and onKeydown

src/styles/types.js
```diff
@@ -218,6 +218,7 @@ export type SizingT = {
   scale300: string,
   scale400: string,
   scale500: string,
+  scale550?: string,
```
nguyensomniac: Do we use `scale550` anywhere? For the official set of components we should stick to the 4px baseline grid.

src/tabs/styled-components.js
```diff
@@ -0,0 +1,88 @@
...
+    paddingTop: sizing.scale550,
```
nguyensomniac: `scale600` should work well here

src/tabs/styled-components.js
anantkamath: Awesome. Marking this as resolved then

src/tabs/styled-components.js
anantkamath: Had chosen 550 because the figma appeared to show the padding as 14px. Happy to change it to scale600 :) .

src/styles/types.js
anantkamath: Used 600 like you suggested.

src/tabs/styled-components.js
anantkamath: Removed the fixed height. 
`boxSizing: 'border-box'` accomplishes the same thing right?

src/tabs/tab-panel.js
anantkamath: Do we need one? I only added this 'component' as a way to provide an API to accept props together for the tabs and the associated panels.

src/tabs/examples-list.js
```diff
@@ -0,0 +1,14 @@
...
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+/* eslint-env node */
+/* eslint-disable flowtype/require-valid-file-annotation */
+
+module.exports = {
+  TABS_EXAMPLE: 'Tabs',
+  STATEFUL_TABS_EXAMPLE: 'Stateful Tabs',
+  STYLE_PROPS_OVERRIDES: 'Stateful tabs with overrides',
```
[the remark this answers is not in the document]
anantkamath: Added

src/tabs/styled-components.js
nguyensomniac: Mind messaging me a link? I'll fix this.

src/tabs/__tests__/tab.test.js
```diff
@@ -0,0 +1,101 @@
...
+    const wrapper = mount(<Tab {...props} />);
+    renderedTab = wrapper.find(StyledTab).first();
+    expect(renderedTab.props()).toMatchSnapshot('styled tab has correct props');
+  });
+
+  test('component overrides', () => {
+    const overrides = {
+      Tab: jest
+        .fn()
+        .mockImplementation(({children}) => <span>{children}</span>),
```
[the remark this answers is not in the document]
anantkamath: Fixed

src/tabs/__tests__/styled-components.test.js
```diff
@@ -0,0 +1,38 @@
...
+
+describe('Tabs styled components', () => {
+  test.each(styledComponents)('Basic render', (Component, name) => {
+    const component = shallow(<Component />);
+    expect(component.instance().getStyles()).toMatchSnapshot(
+      `${name} has correct styles`,
+    );
+  });
+
+  test.each(styledComponents)('Basic render', (Component, name) => {
```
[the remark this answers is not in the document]
anantkamath: Added test for the $active prop.
Also now running each test twice, for each orientation.

src/tabs/__tests__/stateful-tabs.test.js
```diff
@@ -0,0 +1,65 @@
...
+LICENSE file in the root directory of this source tree.
+*/
+// @flow
+import React from 'react';
+import {mount} from 'enzyme';
+import {StatefulTabs, TabPanel, StyledTab, StyledTabContent} from '../index.js';
+import {STATE_CHANGE_TYPE} from '../constants.js';
+
+describe('StatefulTabs', () => {
+  test('basic render and interaction', () => {
```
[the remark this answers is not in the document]
anantkamath: Divided into more granular tests and added the additional tests that you suggested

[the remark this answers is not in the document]
chasestarr: 👍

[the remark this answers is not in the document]
chasestarr: thanks!

src/tabs/tab-panel.js
anantkamath: Added 
``
render() {
    return null;
  }
``

PR #655, opened 2018-12-14
fix(icons): revert to original icons

gergelyke:


src/icon/alert.js
```diff
@@ -8,26 +8,17 @@ LICENSE file in the root directory of this source tree.
 // BASEUI-GENERATED-REACT-ICON
 // DO NOT EDIT THIS FILE DIRECTLY, SEE README.md
 import * as React from 'react';
-import Icon from './icon.js';
-import type {IconPropsT} from './types.js';
+import Icon from './icon';
+import type {IconPropsT} from './types';
```
nadiia: Need to change the template to include the file extension.

src/icon/icon-exports.js
```diff
@@ -1,31 +1,24 @@
...
-export {default as Overflow} from './overflow.js';
-export {default as Plus} from './plus.js';
-export {default as Search} from './search.js';
-export {default as Spinner} from './spinner.js';
-export {default as TriangleDown} from './triangle-down.js';
-export {default as TriangleLeft} from './triangle-left.js';
-export {default as TriangleRight} from './triangle-right.js';
-export {default as TriangleUp} from './triangle-up.js';
-export {default as Upload} from './upload.js';
+export {default as Alert} from './alert';
```
nadiia: need to keep the file extension in all imports

gergelyke: ah, good catch, thanks Nadiia!

PR #657, opened 2018-12-15
feat(button-group) implements the ButtonGroup component

chasestarr:
screen shot 2018-12-17 at 9 21 29 am


src/button-group/button-group.js
```diff
@@ -0,0 +1,74 @@
...
+          first: index === 0,
+          kind: props.kind,
+          last: index === props.children.length - 1,
+          onClick: event => {
+            if (props.disabled) {
+              return;
+            }
+
+            if (child.props.onClick) {
+              child.props.onClick(event);
```
gergelyke: should we document that first the child's onClick will be called, and then the group's onClick?

src/button-group/button-group.js
```diff
@@ -0,0 +1,74 @@
...
+
+  return (
+    <Root aria-label={props.ariaLabel} {...rootProps}>
+      {React.Children.map(props.children, (child, index) => {
+        if (!React.isValidElement(child)) {
+          return null;
+        }
+
+        return React.cloneElement(child, {
+          disabled: props.disabled,
```
nadiia: Wonder if we want to allow a child-specific `disabled` prop, and then use the value from the button group only if it's set to `true`. So a single button can be disabled.

src/button-group/stateful-container.js
```diff
@@ -0,0 +1,61 @@
...
+    };
+  }
+
+  onClick = (event: SyntheticEvent<HTMLButtonElement>, index: number) => {
+    if (this.props.mode === MODE.radio) {
+      if (
+        this.state.selected.length === 0 ||
+        this.state.selected[0] !== index
+      ) {
+        this.setState({selected: [index]});
```
nadiia: We should call a `stateReducer` before setting an updated state.

src/button-group/stateful-container.js
```diff
@@ -0,0 +1,61 @@
...
+
+export default class StatefulButtonGroup extends React.Component<
+  StatefulContainerPropsT,
+  StateT,
+> {
+  constructor(props: StatefulContainerPropsT) {
+    super(props);
+
+    this.state = {
+      selected: props.selected ? [].concat(props.selected) : [],
```
nadiia: We use an `initialState: {selected}` API in other stateful components. That sometimes creates some confusion around passing `selected` for a stateless vs `initialState: {selected}` for a stateful components. But the idea of an `initialState` prop was to avoid another often confusion around 'updated the props passed but the `selected` not updated'. 
Should we be consistent with `initialState` prop for stateful components?

src/button/styled-components.js
```diff
@@ -14,6 +14,25 @@ type StylePropsT = SharedStylePropsT & {
...
+
+  if ($theme.borders.useRoundedCorners) {
+    value = $theme.borders.radius200;
+  }
+
+  return {
+    borderTopRightRadius: value,
+    borderBottomRightRadius: value,
+    borderTopLeftRadius: value,
+    borderBottomLeftRadius: value,
```
nadiia: 👍

src/button-group/README.md
```diff
@@ -0,0 +1,145 @@
...
+* `MODE`
+* `STATE_CHANGE_TYPE`
+
+## `Button` API
+
+Same as `{Button} from baseui/button`, with additional properties:
+* `first?: boolean = false`
+  * Removes right corner radii from button.
+* `last?: boolean = false`
+  * Removes left corner radii from button.
```
schnerd: As a customer it would be nice if this was done magically – would it possible to apply `:first-child, :last-child { border-radius: 0 }` to Button when used in a ButtonGroup?

tajo: I think it’s being already done magically via cloneElement. Imo much better than CSS selectors since it preserves component isolation.

chasestarr: we could possibly build the `selected` state into the core button, and remove this "extension" component entirely. i think that could better experience for devs. does the code below look reasonable, or more complicated?

```js
import React from 'react';
import {ButtonGroup} from 'baseui/button-group'
```

tajo: ^Yea, that’s a good idea!

chasestarr: 🙆 cool, i'll take some time today to rework it

chasestarr:
updated so that it reuses the core `Button` component

src/button-group/README.md
chasestarr: implemented with overrides... could not get pseudo selectors working

PR #667, opened 2018-12-18
fix(datepicker): add overrides functionality

nadiia:


src/datepicker/calendar-header.js
```diff
@@ -49,49 +52,79 @@ export default class CalendarHeader extends React.Component<HeaderPropsT> {
...
-      this.props,
+    const {date, overrides = {}} = this.props;
+    const allPrevDaysDisabled = monthDisabledBefore(date, this.props);
+    const [PrevButton, prevButtonProps] = getOverrides(
+      overrides.PrevButton,
+      ArrowLeft,
     );
-    if (allPrevDaysDisabled) {
-      return;
-    }
```
nadiia: Do we actually want avoid rendering prev and next buttons when all dates before/after are disabled?

src/datepicker/calendar-header.js
```diff
@@ -135,66 +168,93 @@ export default class CalendarHeader extends React.Component<HeaderPropsT> {
...
         maxDropdownHeight={'300px'}
-        // $FlowFixMe
-        onChange={this.handleMonthChange}
         options={monthOptions}
         searchable={false}
+        {...monthSelectProps}
+        // Adding event handlers after customers overrides in order to
+        // make sure the components functions as expected
+        // We can extract the handlers from props overrides
+        // and call it along with internal handlers by creating an inline handler
```
nadiia: Thoughts on this approach?

src/datepicker/day.js
```diff
@@ -113,23 +115,29 @@ export default class Day extends React.Component<DayPropsT, DayStateT> {
...
+      <Day
         aria-label={`day-${getDate(date)}`}
         role="option"
         tabIndex={this.props.selected ? '0' : '-1'}
         {...sharedProps}
+        {...dayProps}
+        // Adding event handlers after customers overrides in order to
+        // make sure the components functions as expected
+        // We can extract the handlers from props overrides
+        // and call it along with internal handlers by creating an inline handler
```
nadiia: Same here. What you all think about handling props overrides here?

src/datepicker/calendar-header.js
tajo: Seems reasonable? Are we doing the same thing everywhere else or what rules do we have regarding this?

src/datepicker/calendar-header.js
chasestarr: i think so, it confirms that disabled dates are not selectable rather than a rendering bug

but paging @nguyensomniac on it

src/datepicker/types.js
```diff
@@ -49,6 +66,7 @@ export type WeekPropsT = {
   onDayMouseOver: ({date: Date, event: Event}) => void,
   onDayMouseLeave: ({date: Date, event: Event}) => void,
   onSelect: ({date: Date}) => void,
+  overrides?: DatepickerOverridesT<{}>,
```
chasestarr: should `SharedStyleProps` be passed as an arg to the type?

src/datepicker/calendar-header.js
nadiia: reverted it

src/datepicker/types.js
nadiia: `SharedStyleProps` only passed to the `Day` component, so I didn't add it here

PR #668, opened 2018-12-18
fix(select): fix blur and close select when double clicked

tajo:
`onBlur` doesn't really work for us since the actual `<input />` is not visible/clickable so essentially any click (on the select component or outside) triggers the `onBlur` event. 

I think the `onBlur` should be triggered only when the select component is focused and it's transitioning into non-focused state (click outside or an option is selected).

The solution is `document.addEventListener('click',...)` and re-implemeting `onBlur` that way. Definitely not ideal but not sure what else we can do if we don't use native input/select components. One downside is that it doesn't register clicks in different iframes (but that's probably not a big concern?) and also the event parameter being passed is different (target = what user clicked on vs what's being "blurred").

With that, we also want to change the behavior of select - when it's double clicked it should close but stay focused.

**Before:**
before

**After:**
after


src/select/select.js
```diff
@@ -130,6 +135,14 @@ class Select extends React.Component<PropsT, SelectStateT> {
     this.clearValue(event);
   };
 
+  handleClickOutside = (event: Event) => {
+    const isFocused = this.state.isFocused || this.state.isPseudoFocused;
+    // $FlowFixMe
```
chasestarr: this may help? https://stackoverflow.com/questions/48433096/react-flowtype-node-contains-event-target

tajo: Hm, it seems we can't cast to `any` because there's this eslint rule `flowtype/no-weak-types`?

nadiia:
Need to add keyboard support for it as well. Currently when I tab fro the select it still maintenance focused styles.

nadiia:
The other issue is tabbing out while the dropdown is opened changed the behavior. See the screen cast below.

Was:
sel1

Became:
sel2

src/select/select.js
```diff
@@ -65,8 +65,10 @@ class Select extends React.Component<PropsT, SelectStateT> {
     if (this.props.autoFocus) {
       this.focus();
     }
+    if (__BROWSER__) {
+      document.addEventListener('click', this.handleClickOutside);
```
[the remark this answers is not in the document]
nadiia: And then we need to unsubscribe on every blur and re-subscribe on every focus.

src/select/select.js
chasestarr: cool, no worries

src/select/select.js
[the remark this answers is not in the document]
tajo: fixed

tajo:
> The other issue is tabbing out while the dropdown is opened changed the behavior.

Fixed.

PR #670, opened 2018-12-18
[Link, Card] style tweaks

nguyensomniac:
- Changes card paragraph size to match Figma
- Adds `:hover` and `:active` states to links


src/link/styled-components.js
```diff
@@ -9,11 +9,18 @@ LICENSE file in the root directory of this source tree.
 import {styled} from '../styles/index.js';
 
 export const Link = styled('a', ({$theme}) => {
-  const {colors, typography} = $theme;
+  const {colors, typography, animation} = $theme;
   return {
     color: colors.primary400,
     ...typography.font450,
     textDecoration: 'none',
+    transition: `color ${animation.timing100} ${animation.easeOutCurve}`,
```
nadiia: Please use the longhand properties instead of a shorthand.

nguyensomniac: fixed 👍

PR #673, opened 2018-12-19
fix(select): highlight the first item initially

tajo:


gergelyke:
can we add some test coverage to it?

tajo:
> can we add some test coverage to it?

Added.

PR #675, opened 2018-12-19
feat(docs): skeleton for new docs site

gergelyke:


documentation-site/pages/_app.js
```diff
@@ -0,0 +1,44 @@
...
+import App, {Container} from 'next/app';
+
+import {Provider as StyletronProvider} from 'styletron-react';
+import {LightTheme, ThemeProvider} from 'baseui';
+
+import getStyletron from '../helpers/styletron';
+import Head from '../components/meta';
+
+export default class MyApp extends App {
+  static async getInitialProps({Component, router, ctx}) {
```
tajo: Can this method be left out since unchanged?

gergelyke: oh yeah, you are right, good catch!

PR #677, opened 2018-12-19
fix(input): pass ref to input in order to attach popover and add input with a datepicker example

nadiia:


src/input/base-input.js
```diff
@@ -84,7 +85,7 @@ class BaseInput<T: EventTarget> extends React.Component<
     } = this.props;
 
     return {
-      $ref: inputRef,
+      $ref: $ref || inputRef,
```
nadiia: Thinking of deprecating `inputRef` and use `$ref` from the top level props since we don't do props spreading on input elements.

PR #690, opened 2018-12-21
fix(flow): use correct export for React Context type

kevhuang:
Running Flow in a project that uses baseui results in the following error:

```
Cannot get React.Context because property Context is missing in module react [1].

 [1]  8│ import * as React from 'react';
      9│ import {LightTheme} from '../themes/index.js';
     10│
     11│ import type {ThemeT} from './types.js';
     12│
     13│ export const ThemeContext: React.Context<ThemeT> = React.createContext(
     14│   LightTheme,
     15│ );
     16│

```

### DIFF
--- a/src/styles/theme-provider.js
+++ b/src/styles/theme-provider.js
@@ -10,7 +10,7 @@ import {LightTheme} from '../themes/index.js';
 
 import type {ThemeT} from './types.js';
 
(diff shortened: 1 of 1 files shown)


chasestarr:
do you mind looking into why flow fails in CI?

kevhuang:
I upgraded a bunch of packages, flow-bin being one of them, and it resolved this error.

PR #692, opened 2018-12-21
feat(docs-site): adds example component

chasestarr:
#### Description

adds example component for docs site

couple things that i'm unsure about:
- should the site use 'move' font like the design?
- is there a better way to read the source code from a file? putting it in the static dir and fetching it was the only method i came up with
- the source path has some complexity between dev/prod since the `/static` dir is nested within `/beta` at the moment. I added an env var for the time being. it can likely be removed when out of beta.
- i'm not excited about duplicating import paths when using the `Example` component. I at least added a webpack module alias to make path names consistent. unsure if we can introspect on the import path?

```jsx
import Input from 'examples/input/example.js';
//...
<Example title="Basic usage" path="examples/input/example.js">
  <Input />
</Example>
```

screen shot 2018-12-21 at 3 59 19 pm


documentation-site/components/example.js
```diff
@@ -0,0 +1,173 @@
...
+          {this.props.children}
+        </Block>
+
+        <Block>
+          <Source>{this.state.source}</Source>
+        </Block>
+
+        <Block paddingLeft="scale800">
+          <CodeSandboxer
+            // asdf
```
gergelyke: i guess we can remove this :)

documentation-site/components/example.js
```diff
@@ -0,0 +1,173 @@
...
+  async componentDidMount() {
+    const sourcePath = `${String(process.env.STATIC_ROOT)}${this.props.path}`;
+    const res = await fetch(sourcePath);
+    const source = await res.text();
+    this.setState({source});
+  }
+
+  handleCopy = () => {
+    this.setState({isCopied: true}, () =>
+      setTimeout(() => this.setState({isCopied: false}), 5000),
```
gergelyke: why do we need the setTimeout here?

gergelyke:
To try to answer you questions:

- `move` is for internal use only, so I think we shouldn't use it here neither
-  probably we can start with the fetch approach, and improve later if we want to? what are the cons of the current approach?

chasestarr:
current downside is duplicating the import path, but maybe it's not that bad. I worked to make it better by enabling the module alias in webpack
```
import Input from 'examples/input/example.js';
//...
<Example title="Basic usage" path="examples/input/example.js">
  <Input />
</Example>
```

documentation-site/components/example.js
```diff
@@ -0,0 +1,173 @@
...
+              'react-dom': '16.5.2',
+              'react-scripts': '2.0.3',
+              'styletron-engine-atomic': '1.0.9',
+              'styletron-react': '4.3.6',
+              'styletron-react-core': '1.3.3',
+            }}
+            providedFiles={{'index.js': {content: index}}}
+            template="create-react-app"
+          >
+            {() => <Link>open in code sandbox</Link>}
```
gergelyke:
```suggestion
             {() => <Link>Open in CodeSandbox</Link>}
```

documentation-site/components/example.js
chasestarr:
```suggestion
```

documentation-site/components/example.js
chasestarr: My thought is that a visitor would like some feedback to indicate the code was successfully copied. To do so, I replace the `copy to clipboard` text with `copied to clipboard`. Following up from that, I'd like to reset the text back to original message to communicate that it is still able to be copied

gergelyke: ahhh, makes perfect sense - missed that!

PR #695, opened 2018-12-23
fix(build): Fixes babel config to publish ES modules

### LINKED ISSUE #516: ES6 code not shipped
I have seen that the lib tries to make bundles of code transpiled in `es2015` and `es` formats.
From what I have seen is that both bundles are essentially the same.  I am just curious about it. Is that intentional or I am missing something here?

## Expected Behavior
It should have ES6 and ES2015 code published so that consuming libs can make a decision about transpiling it.

## Current Behavior
It's not happening from what I have seen here.

## Context
I am evaluating this lib for my usage.

## Your Environment
| Tech    | Version |
| ------- | ------- |
| Base UI | v1.?.?  |
| React   |    v16.4.0     |
| browser |    Google Chrome     |
| node.js |    8.11.0     |
| etc.    |         |

AjayPoshak:
Fixes #516 

#### Description
For `ES` build, now it publishes ES6 modules.

#### Scope

- [X] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


gergelyke:
before merging this, should we remove`.babelrc`?

chasestarr:
I believe it's been deleted, no? explore the source at this commit https://github.com/uber-web/baseui/tree/0354010e5d9f874e786588e0a821ff48ddc07985

gergelyke:
oh you are right, my bad

PR #698, opened 2019-01-02
feat(docs-site): Component Readme

chasestarr:
Creates the input component page. We may want to consider making the source code section of the example component collapsable? Possibly with an option to make it initially open / closed?

localhost_3000_components_input


gergelyke:
yeah, the collapsable example section would make a lot of sense - do you have any idea how we should change the card component to have it?

gergelyke:
it seems like some commits from the master appeared here too - can you clean that up before reviews?

gergelyke:
how would you feel about the following changes?

screen shot 2019-01-03 at 5 17 20 pm
screen shot 2019-01-03 at 5 17 12 pm

chasestarr:
i'm going to open a new pr to fix

PR #702, opened 2019-01-03
feat(overrides): export `mergeStyleOverrides` and add tests

leyanlo:
#### Description

Export `mergeStyleOverrides`, `StyleOverrideT`, and add tests. Can be helpful for dealing with style overrides outside of Base UI.

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change


leyanlo:
Updated the test with `$FlowFixMe`s..

PR #703, opened 2019-01-03
fix(datepicker): handles value and navigation states separately

nadiia:


src/datepicker/navigation-container.js
```diff
@@ -0,0 +1,186 @@
...
+        highlightedDate = addWeeks(
+          // adding `new Date()` as the last option to satisfy Fow
+          highlightedDate ? highlightedDate : lastHighlightedDate || new Date(),
+          1,
+        );
+        stateChangeType = STATE_CHANGE_TYPE.moveDown;
+        break;
+    }
+    this.internalSetState(stateChangeType, {highlightedDate});
+  }
```
schnerd: If I'm understanding this PR correctly, it seems like I won't get any keyboard controls unless I'm using StatefulDatepicker?

If so, I think this is very problematic. I almost always use controlled components in my app – [the react docs themselves also recommend using controlled components](https://reactjs.org/docs/uncontrolled-components.html) over uncontrolled components. I'd like to be able to use a `<Datepicker>` and have all the A11Y keyboard interaction work as intended.

Are there any downsides to just moving all this logic into `<Datepicker>`? I don't see a major use case for breaking it out into its own state container, and if there is a use case maybe it can be handled via props passed into `<Datepicker>` instead?

schnerd: > Before this change it was actually only available in an uncontrolled datepicker, this commit extracts the a11y related navigation into a separate container that can be used with a controlled datepicker.

Definitely an improvement from before.

> Providing a composed component is a good idea but with the current implementation if a calendar is used outside of a popover signing up for global events might not be a good idea.

This is true, but there may be other ways around that. For example, there could be a pseudo `focus` state, where keyboard events are only attached after the user has focused on the calendar container or one of its children.

It also seems like the API should be optimized for the majority use case. Looking at how superfine was used could be a helpful indicator here, if you search for the package name on sourcegraph, it seems like people very rarely used Calendar by itself, and almost always used one of the Datepicker variants with an input+popover. It would be unfortunate if the 90% use case for datepicker also required people to wrap it in a NavigationContainer.

PR #707, opened 2019-01-04
feat(docs-site): adds component readme

chasestarr:
screen shot 2019-01-04 at 8 46 01 am
screen shot 2019-01-04 at 8 45 46 am

based on https://github.com/uber-web/baseui/pull/698


gergelyke:
do you think it would make sense to add it to here too https://github.com/uber-web/baseui/blob/master/documentation-site/routes.js#L17 ?

PR #714, opened 2019-01-07
docs(button): move to the new docs site

gergelyke:


documentation-site/static/examples/button/with-enchancers.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import {Button} from 'baseui/button';
+
+function CloudComponent() {
```
chasestarr: can we replace this cloud component with an icon component? this code feels like noise that does not contribute to explaining the concept

documentation-site/routes.js
```diff
@@ -43,12 +43,16 @@ const routes = [
     text: 'Components',
     children: [
       {
-        text: 'Data Entry',
+        text: 'Basic Inputs',
```
nadiia: It says `Basic Inputs` but contains a button example below. We should split button example into a separate child of the 'Components' instead.

gergelyke: the idea came for Fabric UI, and how they categorized components: https://developer.microsoft.com/en-us/fabric#/components/button

what do you think, would it make sense to follow that?

nadiia: Oh I see, `Inputs` might be confusing, or at least at first. Other suggestion here can be `Basic controls`. No objections to using the current name tho.

gergelyke:
currently blocked by https://github.com/atlassian/extract-react-types/issues/41

tajo:
Looks great! I noticed there is FOUC when the page is being loaded. And it happens in both dev and prod modes. Is that fixable?

gergelyke:
@tajo oh yeah, I see - I'll look into that in a separate PR

PR #715, opened 2019-01-07
feat(menu): adds examples using react-virtualized

### LINKED ISSUE #629: [Menu] Select component with 1500 options takes 500ms to render
An example is here: https://github.com/NameFILIP/baseui-performance-test

flamegraph

chasestarr:
Adds examples for how to incorporate react-virtualized with the menu component. This enables use cases where the list is long (over 1500 items). Previously load times for the component were ~5s. This approach only renders the visible items - leading to an instantaneous action.

Right now I've set it up so that users of baseui will need to make this override themselves. Is it a common use case? I'm hesitant to add api for it since it sounds like a rare occurrence.

closes #629


src/menu/examples.js
```diff
@@ -226,4 +228,48 @@ export default {
...
+                      paddingTop: 0,
+                      paddingBottom: 0,
+                      display: 'flex',
+                      alignItems: 'center',
+                    },
+                  },
+                }}
+              />
+            )}
+            width={200}
```
nadiia: Let's not set the width here but let it extend to the control width. Also there lost box shadow on the dropdown, not sure if it's width related.

PR #716, opened 2019-01-07
docs(dnd-list): add dnd-list proposal

tajo:
**My first component proposal! So please, tear it apart!**

Since this is a first serious use-case of drag and drop (apart from the file uploader) I did some research and prototyping and put together a section about underlying implementation. 

At the beginning I really hoped that we could use some third-party library to not deal with dnd basics. However, I can't find anything that would be good for a library usage and allow the level of customization that baseui requires. So I've started to prototype my own library that implements dnd from scratch and should meet all our requirements. It doesn't seem terribly complicated and it's nice to have a full control. But...

**...is this the hill I will die on?** If there's some better option, lmk! Happy to ditch that if so.

### DIFF
--- a/proposals/dnd-list.md
+++ b/proposals/dnd-list.md
@@ -105,7 +105,7 @@ These properties are passed to presentational (styled) components that are expor
 
 Example usage:
 
-```
+```text
 tab to item
 space to begin item drag
 arrow up/down to change position
--- a/proposals/dnd-list.md
+++ b/proposals/dnd-list.md
@@ -68,6 +68,7 @@ export default () =>
 
 * `initialState: {items: Array<React.Node>}` - Required.
   Initial state of the component where `items` are list items.
+* `removable: boolean` - Optional. Should be list items removable.
 * `overrides: {Root, List, Item, DragHandle, CloseHandle}` - Optional.
   Overrides for presentational components. See "Presentational Components Props API" below
   * `[ComponentName]: ReactComponent | {props: {}, style: {}, component: ReactComponent}` - Optional
@@ -94,18 +95,19 @@ These properties are passed to presentational (styled) components that are expor
 
 ## Accessibility
 
-`tab`, `shift + tab` - to move focus between tabs.
+`tab` - to focus the List component
 
 `Space` - to drag (focus) or drop an item
 
 `Esc` - to cancel the move
 
-`Arrow Up`, `Arrow Down` - move focused item
+`Arrow Up`, `Arrow Down` - selecting / moving items
 
 Example usage:
 
 ```text
-tab to item
+tab to focust the list component
+arrow up/down to select item
 space to begin item drag
 arrow up/down to change position
 space to end item drag
--- a/proposals/dnd-list.md
+++ b/proposals/dnd-list.md
@@ -56,7 +56,7 @@ export default () =>
 
 ## `List` API
 
-* `items: Array<string>` - Required.
+* `items: Array<React.Node>` - Required.
   List items.
 * `overrides: {Root, Item, DragHandle, CloseHandle}` - Optional.
   Overrides for presentational components. See "Presentational Components Props API" below
@@ -66,18 +66,17 @@ export default () =>
 
 ## `StatefulList` API
 
-* `initialState: {items: Array<string>}` - Required.
+* `initialState: {items: Array<React.Node>}` - Required.
   Initial state of the component where `items` are list items.
 * `overrides: {Root, List, Item, DragHandle, CloseHandle}` - Optional.
   Overrides for presentational components. See "Presentational Components Props API" below
   * `[ComponentName]: ReactComponent | {props: {}, style: {}, component: ReactComponent}` - Optional
-* `onChange: ({newState: Array<string>, oldIndex: number, newIndex: number})` - Optional.
+* `onChange: ({newState: Array<React.Node>, oldIndex: number, newIndex: number})` - Optional.
   onChange handler that is called every time an item is moved or removed (newIndex = -1).
-* `stateReducer: (type: string, oldIndex: number, newIndex: number, currentState: {items: Array<string>}) => stateToSet: {items: Array<string>}` - Optional. Useful when you want to replace default array-move logic.
+* `stateReducer: (type: string, nextState: {items: Array<React.Node>}, currentState: {items: Array<React.Node>}) => stateToSet: {items: Array<React.Node>}` - Optional.
   A state change handler.
   * `type` - a state change type
-  * `oldIndex` - an original item index
-  * `newIndex` - a new item index (after being moved or removed when -1)
+  * `nextState` - a new state value to be set
   * `currentState` - current state value
   * `stateToSet` - a return value that the state will be updated with
 
--- a/proposals/dnd-list.md
+++ b/proposals/dnd-list.md
@@ -34,7 +34,7 @@ CustomItem = styled(StyledItem, props => ({
   borderBottom: props.$dragged ? `1px solid red` : 'none'
 }));
 
-export default () => 
+export default () =>
   <StatefulList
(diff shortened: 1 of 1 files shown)


proposals/dnd-list.md
```diff
@@ -0,0 +1,133 @@
+# Drag and Drop List
+
+## Usage
+
+### Basic usage
+
+```js
+import * as React from 'react';
+import {StatefulList, Item} from 'baseui/dnd-list';
```
chasestarr: what does `Item` mean here?

proposals/dnd-list.md
```diff
@@ -0,0 +1,133 @@
...
+
+## STATE_CHANGE_TYPE Constant
+
+* `change` - An item was moved or removed
+
+## Accessibility
+
+`tab`, `shift + tab` - to move focus between tabs.
+
+`Space` - to drag (focus) or drop an item
```
chasestarr: according to medium article `Esc` should cancel the move.

Somewhat unclear in this document is what steps one would take to change order with keyboard. Is it something like below?

```
tab to item
space to begin item drag
arrow up/down to change position
space to end item drag
```

proposals/dnd-list.md
```diff
@@ -0,0 +1,133 @@
...
+* `StyledItem`
+* `StyledDragHandle`
+* `StyledCloseHandle`
+* `STATE_CHANGE_TYPE`
+
+## `List` API
+
+* `items: Array<string>` - Required.
+  List items.
+* `overrides: {Root, Item, DragHandle, CloseHandle}` - Optional.
```
chasestarr: what do we assume the `CloseHandle` button does? Does there need to be some api for removing items? I think this should be not implemented initially, until there is some time for folks to use the component. Does a parent/child api better serve customization?

```jsx
<StatefulList>
  {items.map(x => <Item>{x}</Item>}
</StatefulList>

<StatefulList>
  {items.map(x => (
    <Item>
      {x}
      <Button onClick={...remove item from list}>
        remove
      </Button>
    </Item>
  }
</StatefulList>
```

proposals/dnd-list.md
```diff
@@ -0,0 +1,133 @@
...
+      List: CustomList,
+      Item: CustomItem
+    }}
+  >
+```
+
+## Exports
+
+* `StatefulList`
+* `List`
```
chasestarr: What does `List` usage look like?

proposals/dnd-list.md
tajo: > what do we assume the CloseHandle button does? Does there need to be some api for removing items?

There is an `x` in the mocks so I assumed it means "remove". Technically, it would mean moving the item from index `i` to `-1` so there would be no need for an extra API.

> Does a parent/child api better serve customization?

Good point. I felt we could always do this later if there's a good use-case. One could be having differently styled items but not sure how realistic is that.

On the other hand, this API would introduce some footguns since `List` needs to pass some props to `Items` through `React.cloneElement`, so if `Item` is a custom component it could be easily broken if those hidden props are not correctly passed down.

proposals/dnd-list.md
tajo:
```jsx
state = {
  items: ['Item 1', 'Item 2', 'Item 3']
};

render() {
  return <List
    items={this.state.items}
    onChange={({oldIndex, newIndex}) => 
      this.setState(prevState =>
        ({ items: arrayMove(prevState.items, oldIndex, newIndex)})
      )}
    }
  />
}
```

proposals/dnd-list.md
chasestarr: Cool I'm all for the encapsulated pattern. Though, still somewhat concerned about optimizing the api based on a possibly unique behavior rather than the default case. Will we need an opt-out flag at the top level?

proposals/dnd-list.md
chasestarr: 👍

proposals/dnd-list.md
tajo: > Will we need an opt-out flag at the top level?

In respect to removing items? Or can you elaborate?

chasestarr: yeah i'm thinking about the case where items should not be removed. will we need a `removable={false}` prop? Is there a way we can empower users to implement this use-case themselves and not complicate the base component?

proposals/dnd-list.md
```diff
@@ -0,0 +1,144 @@
...
+* `StyledList`
+* `StyledItem`
+* `StyledDragHandle`
+* `StyledCloseHandle`
+* `STATE_CHANGE_TYPE`
+
+## `List` API
+
+* `items: Array<string>` - Required.
```
nadiia: What is a `string` item here? Should it be a `React.Node` or a component?

proposals/dnd-list.md
```diff
@@ -0,0 +1,144 @@
...
+## `StatefulList` API
+
+* `initialState: {items: Array<string>}` - Required.
+  Initial state of the component where `items` are list items.
+* `overrides: {Root, List, Item, DragHandle, CloseHandle}` - Optional.
+  Overrides for presentational components. See "Presentational Components Props API" below
+  * `[ComponentName]: ReactComponent | {props: {}, style: {}, component: ReactComponent}` - Optional
+* `onChange: ({newState: Array<string>, oldIndex: number, newIndex: number})` - Optional.
+  onChange handler that is called every time an item is moved or removed (newIndex = -1).
+* `stateReducer: (type: string, oldIndex: number, newIndex: number, currentState: {items: Array<string>}) => stateToSet: {items: Array<string>}` - Optional. Useful when you want to replace default array-move logic.
```
nadiia: We should probably stick to the signature the rest of components have and add additional args last `stateReducer(type, nextState, currentState)`.

proposals/dnd-list.md
tajo: How about this? I can imagine we would add `disabled` as well at some point.

```jsx
// default removable = false
const items = [
  {label: 'Item 1', removable: false},
  {label: 'Item 2', removable: true},
  {label: 'Item 3', removable: false}
];

export default () =>
  <StatefulList
    initialState={{ items }}
    onChange={({newState, oldIndex, newIndex}) => {}}
  />
```

proposals/dnd-list.md
tajo: It should be the item label: `Draggable List Item One`. I guess React.Node could be a better choice because localization libs usually use some React component to translate strings. But the expectation is that the output should be a plain text to keep the design nice. Wdyt?

screen shot 2018-11-29 at 2 47 00 pm

nadiia: Let’s do React.Node, so it provides more flexibility if there is some additional markup needed.

proposals/dnd-list.md
tajo: Good point! My thought was that it's pretty hard to tell what exactly changed if you have to compare two arrays and don't have `oldIndex` / `newIndex` information. It can be even impossible if the two items are not unique (and they don't have to be).

**First, I added `onSort` prop with this signature:** 

```tsx
(currentState: Array<string>, oldIndex: number, newIndex: number) : Array<string>
```

since I think that it's a legit requirement to let users override the sorting logic. For example there can be some extra validation allowing / disallowing particular drag and drop.

**But then I saw the `stateReducer`** and thought it doesn't make sense to have this very similar hook twice. So I removed `onSort` and modified `stateReducer`. But I can add `onSort` back and assume it will be applied first before `stateReducer` is triggered?

nadiia: Since we already provide `onChange` hook, for the stateful list `onChange`+`stateReducer` should be sufficient. And for further control, a stateless list should be used. Does that sound reasonable?

tajo: It does! Updated.

proposals/dnd-list.md
tajo: Updated.

PR #718, opened 2019-01-08
fix(checkbox): change checkbox typography style from bold to regular

jonsadka:
Change typography font weight from `350` to `300` to remove the bold look.

**Before**:
screen shot 2019-01-07 at 5 11 49 pm

**After**:
screen shot 2019-01-07 at 5 11 36 pm

#### Scope
- [ ] Patch: Style Fix

### DIFF
--- a/src/checkbox/styled-components.js
+++ b/src/checkbox/styled-components.js
@@ -239,7 +239,7 @@ export const Label = styled('div', props => {
     verticalAlign: 'middle',
     ...getLabelPadding(props),
     color: getLabelColor(props),
-    ...typography.font350,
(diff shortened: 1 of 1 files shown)


nadiia:
@nguyensomniac could you confirm this change, please. It follows the current style in Figma that is currently set to Uber Move normal.

gergelyke:
closing for now, till we don't we get more information from design on when to use the bold / regular version of checkboxes

nguyensomniac:
`font350` is correct. It's 14/20 Uber Move medium, which will look bold in the OS version.

PR #720, opened 2019-01-09
docs(header-navigation): move to the new doc site

gergelyke:


src/header-navigation/examples.js
```diff
@@ -73,6 +73,7 @@ class MenuContainer extends React.Component<
   };
   state = {
     isOpen: this.props.isOpen,
+    z,
```
chasestarr: was this intentional?

chasestarr:
lgtm once lint passes

PR #727, opened 2019-01-10
feat(menu): support disabled options

salmontoday:
Adds support for disabled options, using the same treatment as Select.

This also fixes StatefulMenu's handling of clicking options. Previously, it would not properly highlight options that were clicked; it only worked through the keyboard.


salmontoday:
Whoops, wrong username

PR #728, opened 2019-01-10
feat(menu): support disabled options

phou87:
Adds support for disabled options, using the same treatment as Select.

This also fixes StatefulMenu's handling of clicking options. Previously, it would not properly highlight options that were clicked; it only worked through the keyboard.


src/menu/option-list.js
```diff
@@ -29,6 +29,7 @@ export default function OptionList({
     StyledListItem,
   );
   const sharedProps = {
+    $disabled: item.disabled,
```
nadiia: This should be mapped on a level up. Current fix only handles list type options but, for example, misses provided profile type options. Look into `getSharedProps` called from the `/menu.js`.

phou87: Gotcha, will update

PR #732, opened 2019-01-10
feat(build): git tag calls to build netlify

chasestarr:
Calls out to the netlify webhook to build the site. Trouble downside of this approach is it would mean giving up deploy previews on prs with this strategy. Waiting on a response from netlify.


nadiia:
Previews on PRs seem to be pretty helpful. Is there a way we can keep them working?

chasestarr:
waiting to hear back from netlify to see if that is possible

PR #742, opened 2019-01-11
Release v5.7.0

gergelyke:


gergelyke:
yeah, I guess we can go either way - previously we did not support the disabled state, so I'd rather go with a minor, if that's ok for everyone

PR #753, opened 2019-01-15
docs(menu): adds menu to docs site

chasestarr:
adds a new prop to the `Example` component to pass additional npm packages. This ensures that the `react-virtualized` example works correctly in the code sandbox


chasestarr:
This has an issue in extracting the prop types

PR #765, opened 2019-01-18
feat(docs): move the rating component to the new doc site

gergelyke:


gergelyke:
oh, I totally missed those - on it!

PR #766, opened 2019-01-18
feat(table): add initial implementation

Dogfalo:
#### Description

Add MVP implementation for Table component
#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change


src/table/styled-components.js
```diff
@@ -0,0 +1,68 @@
...
+    ...borders.border400,
+    borderLeft: 'none',
+    borderTop: 'none',
+    borderBottom: 'none',
+  };
+});
+export const Body = styled(Grid, (props: SharedStylePropsT) => {
+  return {};
+});
+export const Cell = styled('div', ({$theme}: SharedStylePropsT) => {
```
gergelyke: should we use table semantics here? so in the example of the cell, it should be `td`

src/table/examples.js
```diff
@@ -0,0 +1,53 @@
...
+
+for (let i = 0; i < 5; i++) {
+  data = data.concat(data);
+}
+const ControlledTable = (props: TablePropsT) => {
+  const {overrides, ...otherProps} = props;
+  return <Table {...otherProps} overrides={overrides} />;
+};
+
+export default {
```
gergelyke: can you add more examples, potentially one with sorting too?

chasestarr:
Thanks for putting up the PR Alvin! I'm going to look into some ways we can decouple `react-virtualized` from the core implementation. Have you taken a look at how segment does it [here](https://github.com/segmentio/evergreen/tree/master/src/table/src)?

src/table/styled-components.js
gergelyke: oh, I've missed that

gergelyke: in that case, should we add ARIA roles? https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Table_Role

src/table/README.md
```diff
@@ -0,0 +1,175 @@
...
+  StyledCell
+} from 'baseui/table';
+
+const data = [
+  [1, 'John', 'Smith'],
+  [2, 'Lois', 'Lane'],
+];
+
+export default () => (
+  <StyledRoot>
```
tajo: I know we have `Roots` in most of components but it seems odd to not call it `StyledTable` here? It could make the code more readable since it would be more obvious that it's a table.

src/table/examples.js
chasestarr: sorting will be added in a follow-up pr

src/table/README.md
```diff
@@ -0,0 +1,175 @@
...
+* `StyledRoot`
+* `StyledHead`
+* `StyledHeadCell`
+* `StyledBody`
+* `StyledRow`
+* `StyledCell`
+
+## `Table` API
+
+* `columns: Array<string>` - Required
```
nadiia: We could relax the type here to `React.Node`

src/table/types.js
```diff
@@ -0,0 +1,23 @@
...
+*/
+// @flow
+
+import * as React from 'react';
+
+import type {ThemeT} from '../styles/types.js';
+
+export type TablePropsT = {|
+  columns: Array<string>,
+  // eslint-disable-next-line flowtype/no-weak-types
```
nadiia: shouldn't need this comment

src/themes/creator.js
```diff
@@ -193,6 +193,10 @@ export default function createTheme(
       notificationNegativeBackground: primitives.negative50,
       notificationNegativeText: primitives.negative500,
 
+      // Table
+      tableHeadBackgroundColor: primitives.mono100,
+      tableHeadShadow: '0 1px 4px -1px hsla(0, 0%, 0%, 0.16)',
```
nadiia: Should it be `shadow400`?

nadiia:
Screencast of the table head issue:
tb1

src/table/README.md
Dogfalo: fixed in 49c9094349f1c078d78fd8f397d09871166c9cda

src/themes/creator.js
Dogfalo: yes. fixed in f9181bfe924779b8c794716b079a65720c9b8035

src/table/styled-components.js
Dogfalo: added aria roles in 7cb20ab42e612a3b2e853b8e4d7d639e0c15b03f

Dogfalo: Using `grid` role instead of table becuase

> If a table maintains a selection state, has two-dimensional navigation, or allows the user to rearrange cell order use grid or treegrid instead.

src/themes/creator.js
chasestarr: sorry, no... with `shadow400` it extended to the left/right when the shadow should only apply to bottom

Dogfalo: I didn't see that issue happening

documentation-site/pages/components/table.mdx
```diff
@@ -0,0 +1,57 @@
...
+
+export default Layout;
+
+# Table
+
+Display larger amounts of information in a scannable UI.
+
+## Examples
+
+<Example title="Basic Usage" path="examples/tag/basic.js">
```
gergelyke: i think this needs to be updated

documentation-site/pages/components/table.mdx
```diff
@@ -0,0 +1,57 @@
...
+
+<Example title="Vertical Scroll" path="examples/table/vertical-scroll.js">
+  <VerticalScroll />
+</Example>
+
+<Example title="Horizontal Scroll" path="examples/table/horizontal-scroll.js">
+  <HorizontalScroll />
+</Example>
+
+<Example
```
gergelyke: should we put a bit of explanation here when someone should use a virtualized table?

PR #768, opened 2019-01-18
feat(docs): overrides docs

gergelyke:
super early PoC - https://5c41810c6e40ec00078cc073--baseui.netlify.com/components/accordion/


chasestarr:
this is awesome! i had been thinking of something similar and thought https://github.com/jxnblk/react-x-ray might be appropriate. added it to your code:

```jsx
<XRay
  grid={false}
  color={LightTheme.colors.primary}
  backgroundColor="white"
>
  <Example overrides={overrides} />
</XRay>
```

it looks like this:
screen shot 2019-01-18 at 8 48 50 am

documentation-site/components/overrides.js
```diff
@@ -0,0 +1,92 @@
...
+import {Block} from 'baseui/block';
+import {StatefulPopover, TRIGGER_TYPE, PLACEMENT} from 'baseui/popover';
+import {Card, StyledBody} from 'baseui/card';
+
+function isStyledExport(exportName) {
+  return exportName.startsWith('Styled');
+}
+
+function getOverrideName(exportName) {
+  return exportName.replace('Styled', '');
```
nadiia: This may not be true for all components.

documentation-site/pages/components/accordion.mdx
```diff
@@ -8,10 +8,14 @@ LICENSE file in the root directory of this source tree.
 import Example from '../../components/example';
 import API from '../../components/api';
 import Layout from '../../components/layout';
+import Overrides from '../../components/overrides';
 
 import BasicAccordion from 'examples/accordion/basic.js';
 import StatefulAccordion from 'examples/accordion/stateful.js';
 
+import * as AccordionExports from 'baseui/accordion';
+import {Accordion, Panel} from 'baseui/accordion';
```
nadiia: Should do a single import and `{Accordion, Panel}` destructuring from it

nadiia:
Looks great. The tooltip for toggle icon seems to show up in the wrong place. I like the example Chase posted since it provides clearer visibility into components structure.

documentation-site/components/overrides.js
gergelyke: I've just gone through all the components, and it seems only the Block component does not export styled components, so I think we should be good with this one

PR #770, opened 2019-01-19
feat(docs): swap documentation site

gergelyke:


chasestarr:
I would do a search across the code for `beta`. looks like there are some configs missing

gergelyke:
Yeah, I'll do that, thank Chase!

PR #777, opened 2019-01-23
doc(align-layout): propose align layout (Row / Column) components based on CSS grid.

cain-wang:
This PR proposes a new set of layout components that align child components in a line (row or column).

It uses CSS grid underneath that provides a simpler yet more powerful layout syntax compared to Flex Box:

1. All layout styles can be configured on the container element without flex-grow on respective child nodes.
2. Support CSS grid gap instead of margins on child nodes.
3. Supports many CSS property shortcuts like color, font, width, height, padding, margin, etc ...

Some comparison with Block component:

These layout components focus on providing simpler styling for layout containers while Block component serves as a lower construct for other styling needs. The align-layout components can use Block component underneath.

In terms of syntactical difference:

```js
// A simple column layout with Block
<Block
  display="grid"
  gridTemplateRows="repeat(3, 1fr)"
  gridAutoFlow="column"
  alignItems="center"
  gridColumnGap="scale1000"
>
  // children
</Block>

// A simple column layout with align Column
<Column items="repeat(3, 1fr)" gap="scale1000" center>
  // children
</Column
```

While css Grid is superior than flex-box in many ways, it poses a lot of syntactical complexities to us. And we'd always need to lookup the property names every time we do layout. Since most of our layout needs are Row and Column layouts, we create these React components with simpler property names that are easier to remember.

### DIFF
--- a/src/layout/README.md
+++ b/src/layout/README.md
@@ -32,7 +32,7 @@ import {Row, Column} from '@uber/layout-align';
 * `Row`
 * `Column`
 
-## `Button` API
+## `Row` & `Column`API
 
 * `items: string`
 
--- a/src/layout/README.md
+++ b/src/layout/README.md
@@ -1,0 +1,72 @@
+# Align Layout Components
+
+## Usage
+
+```javascript
+import {Row, Column} from '@uber/layout-align';
+
+// "1fr" is similar to flex-grow: 1 on the target item
+<Row items="100px 1fr 20%">
+  <Item>100px</Item>
+  <Item>Rest</Item>
+  <Item>20%</Item>
+</Row>
+
+// automatically translate gap to row-gap or column-gap
+<Column items="100px 1fr 20%" gap="24px">
+  <Item>100px</Item>
+  <Item>Rest</Item>
+  <Item>20%</Item>
+</Row>
+
+// support common CSS properties
+<Column items="100px 1fr 20%" width="360px" height="240px" color="primary200">
+  <Item>100px</Item>
+  <Item>Rest</Item>
+  <Item>20%</Item>
+</Row>
+```
+
+## Exports
+
+* `Row`
+* `Column`
+
+## `Button` API
+
+* `items: string`
+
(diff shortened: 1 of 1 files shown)


nadiia:
It seems to have an overlap with the Block component. Would it make sense to improve/extend the `Block`?

cain-wang:
> It seems to have an overlap with the Block component. Would it make sense to improve/extend the `Block`?

Yeah, as @tajo also commented, it is a higher level of abstraction for the layout. 
We built quite extensive layout in our applications primarily using flex box, and they were quite verbose and hard to manage. We want a simpler API for layouts, especially for team members who are not expert with CSS. These components addressed 90%+ of our layout needs, by replacing most flex box styles.

We thought about extending Block. Block is very comprehensive already, and we don't want to further complicate that component. Supporting both native css grid, flex box and high level layout settings could be a confusing combination.

tajo:
> These components addressed 90%+ of our layout needs, by replacing most flex box styles.

That's pretty encouraging. I think this would be an useful addition to the baseui components.

src/layout/README.md
```diff
@@ -0,0 +1,72 @@
...
+  <Item>20%</Item>
+</Row>
+```
+
+## Exports
+
+* `Row`
+* `Column`
+
+## `Button` API
```
nadiia: `Button` should be replaced here with `Row/Column` I guess

cain-wang: Good catch! Fixed the wording.
Also added some comparison with Block component in the description above.

chasestarr:
Thank you for the proposal @cain-wang. This looks to be an elegant approach to grid layout and i'm happy that your team has found it useful. We've talked about how this can be effectively incorporated into the project and have decided to not add it to base ui. 

As-is, It's too similar to the existing `Block` component, which leads to documentation challenges in explaining when to use one or the other. Some common layout patterns will be supported by a [silhouette](https://github.com/uber-web/baseui/pull/777#issuecomment-456630076) component in the near future.

This may be most appropriate as a standalone npm package. I'm going to close this pr for now, but feel free to continue the conversation here. Again, I appreciate that you helping to improve the project.

PR #778, opened 2019-01-23
chore(babel): Adds styletron display name babel transform

chasestarr:
#### Description

Adds a babel transform to apply a `displayName` to styled components. This helps eliminate rote work of adding `displayName` manually. This _could_ be considered a breaking change if anyone relies on `displayName`. That field is essentially public api of the styled components. What level of release do we think is required for this change?

#### Scope

- [ ] Patch: Bug Fix
- [ ] Minor: New Feature
- [x] Major: Breaking Change


tajo:
It should probably be a major version since there is a good potential this will break many snapshot tests?

gergelyke:
as this change does not break the exposed API, should we just bump a patch and be explicit in the versioning policy that snapshot test updates does not count as a semver major?

chasestarr:
versioning policy says we do not bump major on `Change in undocumented APIs and internal data structures.`. a component's `displayName` is an undocumented API. basing the decision on this, the pr can ship in a patch

tajo:
Good points. Yea, let’s ship patch!

PR #779, opened 2019-01-23
feat(docs): move file uploader component

tajo:


documentation-site/routes.js
```diff
@@ -192,6 +192,15 @@ const routes = [
           },
         ],
       },
+      {
+        text: 'Unstable',
+        children: [
+          {
+            text: 'File Uploader',
```
gergelyke: should we follow the way the datepicker PR handles unstable components?

documentation-site/routes.js
```diff
@@ -72,6 +72,10 @@ const routes = [
             text: 'Textarea',
             path: '/components/textarea',
           },
+          {
+            text: 'Unstable File Uploader',
```
gergelyke: should we move this rather to the pickers?

tajo:
> Need to move the description of props to the types file.

Hm, it seems we don't have description of props yet? I should be looking into `README.md`, right? Maybe it's because the component is not stable yet @chasestarr?

nadiia:
Right, no much to describe in the API, disregard my comment above.

PR #780, opened 2019-01-23
feat(puppeteer): adding the test runner

gergelyke:


src/accordion/pup-e2e.js
```diff
@@ -0,0 +1,19 @@
...
+
+// @flow
+/*eslint-env node*/
+
+describe('Google', () => {
+  beforeAll(async () => {
+    await page.goto('https://google.com');
+  });
+
+  it('should display "google" text on page', async () => {
```
nadiia: Should we add a simple accordion test instead?

gergelyke: yeah, that makes sense

src/accordion/pup-e2e.js
```diff
@@ -0,0 +1,19 @@
...
+    await expect(page).toMatch('google');
```
KevinGrandon: `expect(page)` works here? Isn't that a puppeteer page object? Just wondering if we need to get the content from the page.

jest.e2e.config.js
```diff
@@ -0,0 +1,14 @@
...
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+
+// @flow
+/*eslint-env node*/
+
+module.exports = {
+  preset: 'jest-puppeteer',
```
KevinGrandon: Personally not a huge fan of this library since it makes more globals, but I suppose it does mean less boilerplate in each test. Up to you all.

PR #782, opened 2019-01-23
feat(datepicker): add quick select actions to date picker

chasestarr:
#### Description

Adds quick select actions to the date picker component. I did not make the options customizable for now.. but it should be straight-forward.

screen shot 2019-01-23 at 12 47 36 pm

#### Scope

- [x] Minor: New Feature

### DIFF
--- a/src/datepicker/styled-components.js
+++ b/src/datepicker/styled-components.js
@@ -40,6 +40,34 @@ export const StyledCalendarContainer = styled(
   },
 );
 
+export const StyledQuickSelectContainer = styled(
+  'div',
+  (props: SharedStylePropsT) => ({
+    maxWidth: '296px',
+    marginLeft: 'auto',
+    marginRight: 'auto',
+    marginBottom: '16px',
+  }),
+);
+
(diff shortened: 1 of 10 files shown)


src/datepicker/styled-components.js
```diff
@@ -40,6 +40,34 @@ export const StyledCalendarContainer = styled(
   },
 );
 
+export const StyledQuickSelectContainer = styled(
+  'div',
+  (props: SharedStylePropsT) => ({
+    maxWidth: '296px',
+    marginLeft: 'auto',
+    marginRight: 'auto',
+    marginBottom: '16px',
```
nadiia: Would it make sense to use theme's `scale600` instead?

chasestarr: ah right - i need to go through and use theme values. I was using px for prototyping it

nadiia:
The quick select buttons look slightly different in Figma, have grey background.

chasestarr:
updated to the tertiary buttons!

PR #788, opened 2019-01-24
feat(button-group): Align color scheme with designs

chasestarr:
#### Description

This pr applies the correct styles to the button group component. It also introduces a dev warning for deprecating the `kind` prop in future major versions. Versions other than `tertiary` do not render according to design guidelines. When we look to release a new major version, we can remove the warning and delete the selected styles on other button variants. 

I also found a bug in the implementation, so i fixed that in this pr. please see the first commit.

screen shot 2019-01-24 at 9 58 00 am

#### Scope

- [x] Minor: New Feature


src/themes/creator.js
```diff
@@ -103,18 +103,25 @@ export default function createTheme(
...
       buttonSecondaryFill: primitives.primary50,
       buttonSecondaryText: primitives.primary400,
       buttonSecondaryHover: primitives.primary100,
       buttonSecondaryActive: primitives.primary200,
+
       buttonTertiaryFill: primitives.mono200,
       buttonTertiaryText: primitives.primary400,
       buttonTertiaryHover: primitives.mono400,
       buttonTertiaryActive: primitives.mono500,
+      // button $seleected only applies to tertiary variant.
```
gergelyke:
```suggestion
      // button $selected only applies to tertiary variant.
```

PR #793, opened 2019-01-25
feat(docs): add algolia search

tajo:
search

Note that in the preview it redirects from `localhost` to `baseui.design`, that's why the search input jumps.

### DIFF
--- a/documentation-site/components/search.js
+++ b/documentation-site/components/search.js
@@ -1,0 +1,88 @@
+/*
+Copyright (c) 2018 Uber Technologies, Inc.
+
+This source code is licensed under the MIT license found in the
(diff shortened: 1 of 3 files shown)


documentation-site/components/search.js
```diff
@@ -0,0 +1,88 @@
...
+
+const SEARCH_INPUT_ID = 'algolia-doc-search';
+
+type State = {
+  enabled: boolean,
+};
+
+// can't really use baseui/input because algolia injects its
+// own markdown and breaks our component (that's fairly complex)
+const PlainInput = styled('input', ({$theme}) => ({
```
gergelyke: can we use the input component?

PR #795, opened 2019-01-25
feat(file-uploader): creates post-file drop ui

chasestarr:
#### Description

Adds UI for displaying file upload progress / errors. I chose not to create a stateful wrapper around this component because of how much file upload preferences can vary. I suggest we release this as-is and develop a wrapper if requested from developers using the component.

This component is currently marked `unstable`, and i think it should stay that way until design has a chance to review.

#### Scope

- [x] Minor: New Feature


tajo:
Looking good!

gergelyke:
I've accidentally accepted the changes in screener, but now added Alexander to the screener org, so he'll have the chance to review

PR #799, opened 2019-01-28
feat(docs-site): recognize contributors

chasestarr:
#### Description

Fetches contributors to the baseui repo and displays their avatars on the home page of the docs site

screen shot 2019-01-28 at 10 43 52 am

#### Scope

- [x] Patch: Bug Fix


gergelyke:
when visiting the netlify preview, I get the site does not render for me, and I get this error message: `this.state.contributors.filter is not a function`

chasestarr:
Looks like we are exceeding the github api rate-limits. One way I could imagine solving this would be to bake in the contributors on the site build step (in the `getInitialProps` method). Problem is, when using an `.mdx` file `getInitialProps` is never called. See this [relevant issue](https://github.com/mdx-js/mdx/issues/382).

I think I will remove mdx from this page until the use case is supported. This page here is small

gergelyke:
is that possible that github rate limiting will hit us during builds? as we have quite a lot of builds, can we use authenticated requests to have a bigger limit?

PR #800, opened 2019-01-28
feat(tag): add closeable prop to disable close button

ksheedlo:
#### Description

This change adds a boolean `closeable` prop to include or exclude the "x" button on the Tag component. Default value is true, which includes the button. Passing false will create a non-interactive tag that does not include the button.

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change


src/tag/tag.js
```diff
@@ -55,25 +57,27 @@ class Tag extends React.Component<PropsT, {}> {
       $kind: kind,
     };
     return (
-      <Root {...sharedProps} {...rootProps}>
+      <Root {...sharedProps} {...rootProps} $closeable={closeable}>
```
chasestarr: suggest moving this to `sharedProps` above

```suggestion
      <Root {...sharedProps} {...rootProps}>
```

src/tag/tag.js
```diff
@@ -55,25 +57,27 @@ class Tag extends React.Component<PropsT, {}> {
       $kind: kind,
```
chasestarr:
```suggestion
      $kind: kind,
      $closable: closable,
```

PR #807, opened 2019-01-29
fix(typography): set the default font color 

nadiia:


nadiia:
On the other hand, it may break existent usage of the `Block` if someone relies on font color inheritance while using `Block` 🤔 This would be a breaking change.

chasestarr:
do we have any breaking changes coming up? it may be worth saving this fix for then

nadiia:
Made it a typhography specific change in order to keep the `Block` component inherit the color

PR #814, opened 2019-01-30
fix(compat): use create-react-context for compatibility with React 15.x

VladimirMilenko:
#### Description

Use `create-react-context` polyfill to support React v15.x.
Styletron uses it under the hood to have support of older versions. At the same time, `baseui` doesn't use any of React 16 features, except for `createContext`.

#### Scope
- [x] Patch: Bug Fix

### DIFF
--- a/package.json
+++ b/package.json
@@ -142,6 +142,7 @@
     "webpack-cli": "^3.1.0"
   },
   "dependencies": {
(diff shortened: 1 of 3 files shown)


VladimirMilenko:
Alright, my mistake. I found a usage for createRef. Closing this PR

PR #822, opened 2019-01-30
feat(docs): add overrides documentation to all components

tajo:
It's a pretty big diff! Sorry for that but it's probably better than creating 30 tickets + PRs? This work exposed some bugs / inconsistencies in our `overrides` pattern. So I went down the rabbit hole and applied some fixes to our components. **It's nothing major but definitely should get your attention (changes in `src/` folder)**. On the other hand, you can probably just skim through `documentation-site/` folder diff since it's a lot of code and rather test it by previewing.

## Fixed bugs / improvements

- **baseui/button**: Added more examples to the Button since it's the most used component and not all cases were covered
- **baseui/card**: Overrides were not applied for the `Action` component. `Root` override was exported as `Wrapper` which breaks the override itself, renamed to `Root`
- **baseui/menu**: Top level `overrides` prop was not passed into `ListItem`
- **baseui/datepicker**: `StyledHeader` renamed to `StyledCalendarHeader` to match the override key
- **baseui/input**: Input exports `StyledInputEnhancer` however the overrides accept `StartEnhancer` and `EndEnhancer`, so I've updated the exports in order to make start/end enhancers overridable separately which imo makes more sense since those can have different looks / purposes:

```js
   // before
    export {
      /* ... */
      InputEnhancer as StyledInputEnhancer,
    } from './styled-components.js';
    
    // after 
    export {
      /* ... */
      InputEnhancer as StyledStartEnhancer,
      InputEnhancer as StyledEndEnhancer,
    } from './styled-components.js';
```

## Other notes

- `Override` component has two new props `blacklisted` and `whitelisted`.  They both expect an array of strings - override keys. For various reasons, sometimes, Styled* exports don't exactly match override keys. You can use `blacklisted` to filter those or `whitelisted` to input them all manually.
- **baseui/menu**: API might be a bit confusing in regard to `OptionProfile` and the way how the overrides works overall. We might need to split the `Menu` component into two smaller components. Or maybe we could add `type` prop to the `Menu` to decide if `OptionList` or `OptionProfile` should be rendered. Anyway, I expanded the overrides section for this specific component and tried to explain it a greater detail.
- **baseui/modal**: Not covered since it can't be displayed inline and it has many child components anyway.
- **baseui/popover**: Exports `StyledPadding` but it's not used in the component itself, maybe we should just remove it? Users can just use `Block` instead. Blacklisted in meantime.

```js
    /**
     * A drop-in component that provides the recommended padding
     * for popovers. Mostly a convenience for users so they don't
     * have to define this themselves.
     */
    export const Padding = styled('div', {
      padding: '12px',
    });
```

## Minor vs Major?

I renamed one export in `baseui/datepicker` and `baseui/input` in order to fix overrides. If we want to play it super-safe, we can do a major bump but I don't expect theses change will impact a lot of people so minor bump is probably fine too.


documentation-site/static/examples/button/shapes.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import {Button, SHAPE} from 'baseui/button';
+
+export default () => (
+  <React.Fragment>
+    <Button shape={SHAPE.default}>Default shape</Button>
+    <Button shape={SHAPE.square}>Square shape</Button>
```
gergelyke: only a single icon should be used for square and round

src/input/__tests__/styled-components.test.js
```diff
@@ -9,7 +9,7 @@ import React from 'react';
 import {shallow} from 'enzyme';
 import {
   StyledRoot,
-  StyledInputEnhancer,
+  StyledStartEnhancer as StyledInputEnhancer,
```
chasestarr: if we rename.. can we update the code below this line to `StyledStartEnhancer`? I find the unnecessary import naming hard to follow

src/input/index.js
```diff
@@ -13,7 +13,8 @@ export {default as MaskedInput} from './masked-input.js';
 // Styled elements
 export {
   Root as StyledRoot,
-  InputEnhancer as StyledInputEnhancer,
+  InputEnhancer as StyledStartEnhancer,
+  InputEnhancer as StyledEndEnhancer,
```
chasestarr: i would maintain the export until the next major

src/card/index.js
```diff
@@ -15,6 +15,6 @@ export {
   HeaderImage as StyledHeaderImage,
   Thumbnail as StyledThumbnail,
   Title as StyledTitle,
-  Root as StyledWrapper,
+  Root as StyledRoot,
```
chasestarr: can we save till next major?

nadiia:
Regarding the input's change: for future changes, it may make sense to split the styled enhancer export in two in case we actually provide different elements for those two. Wonder if we can into the new two along the older export with a warning first, and then once we do a major bump we'll remove the original export. Otherwise, it does seem to be a breaking change. Was there a specific issue with displaying overrides while having a single styled export?

tajo:
> Otherwise, it does seem to be a breaking change. 

I preserved the original `export` as @chasestarr suggested.

> Was there a specific issue with displaying overrides while having a single styled export?

We generally assume that exported `StyledComponents` match `overrides` keys - that didn't work for this case. 

We can manually adjust that; however, in this case it probably makes better sense to actually export `Start` and `End` enhancers separately since - it gives users more flexibility (they can style them differently) and it also solves the first issue.

nadiia:
That makes sense. They actually can style start and end enhancers separately as in overrides it's two separate identifiers and even tho we use a single import it still creates two separate styled elements for each enhancer.

PR #824, opened 2019-01-31
feat(block): responsive style props breakpoints

chasestarr:
#### Description

Early version of this pr. What do we think of the approach? I still need to add examples to the docs site

#### Scope

- [x] Minor: New Feature


tajo:
**I like it!** Hopefully, Styletron will correctly sort those media queries. :)

src/block/styled-components.js
```diff
@@ -8,297 +8,243 @@ LICENSE file in the root directory of this source tree.
...
+          }
+
+          const mediaQuery = mediaQueries[index];
+          if (!styles[mediaQuery]) {
+            styles[mediaQuery] = {};
+          }
+
+          styles[mediaQuery][property] = transform(v);
+        });
+      } else if (value != null) {
```
nadiia: no need for the second null check

src/block/styled-components.js
```diff
@@ -8,297 +8,243 @@ LICENSE file in the root directory of this source tree.
...
+
+      if (Array.isArray(value)) {
+        value.forEach((v, index) => {
+          // Do not create a media query for the smallest breakpoint.
+          if (index === 0) {
+            styles[property] = transform(v);
+            return;
+          }
+
+          const mediaQuery = mediaQueries[index];
```
nadiia: It seems like we never create a media query for the `small` breakpoint that makes its value unused and leaves only 2 possible default breakpoints. 
How about we consider the first value in the array as default (the smallest, as we currently do) but we don't skip the first media query in `mediaQueries`? So we'll get 3 media queries in addition to the default (considered smallest screen) value. Something like this:
```
height: '100px', 
[smallBreakpoint]: {height: '200px', }
[mediumBreakpoint]: {height: '300px', }
[largeBreakpoint]: {height: '400px', }
```

Unless I missed something while reading the code 🤔

chasestarr: yes - you're totally right. my mistake

src/typography/index.js
```diff
@@ -12,65 +12,65 @@ import type {BlockPropsT} from '../block/types.js';
 
 // Captions - aka Caption, CaptionLabel
 export const Caption1 = (props: BlockPropsT) => (
-  <Block font="font200" {...props} />
+  <Block {...props} font={props.font || 'font200'} />
```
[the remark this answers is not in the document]
chasestarr: correct - flow can't handle the possibility of providing the `string` and then possibly accepting an array in the props spread... sounds like a bug/missing feature?

PR #827, opened 2019-01-31
feat(datepicker): add a drop-in datepicker input

nadiia:


src/datepicker/datepicker.js
```diff
@@ -0,0 +1,107 @@
...
+
+  handleKeyDown = (event: KeyboardEvent) => {
+    if (!this.state.isOpen && event.keyCode === 40) {
+      this.open();
+    }
+  };
+
+  render() {
+    const {overrides = {}} = this.props;
+    const [InputC, inputProps] = getOverrides(overrides.Input, Input);
```
tajo: What the `C` stands for? :)

gergelyke:
Once merged, should we say that datepicker is not unstable anymore?

One potential issue I ran into: when the datepicker is opened in the popover, when I click outside of it the Datepicker is not closed - is that on purpose?

tajo:
Keyboard control is now global and changes the state of other instances too.

shortcuts_are_global

tajo:
When you try to open the year / month selector, the whole page jumps up.

jump

tajo:
**Nitpick:** When you click fast (double-click) on arrows, it selects the text (month or day names). `user-select: none` would fix that. 

select

tajo:
Anyway, thanks for working on this! It's really complex and overall it works very well!

src/datepicker/datepicker.js
nadiia: `Component`,  didn't want to rename the top level imports 🤷‍♀️

gergelyke: should we be explicit, and name it `InputComponent`?

tajo:
**It looks and works great!** All the bugs are fixed and the keyboard control is pretty intuitive to me now!

One one last tiny thing: The page jumps a bit when you first hit `Arrow Down` after the calendar is open.

jump

PR #836, opened 2019-02-04
test(e2e): breadcrumbs a11y

gergelyke:


gergelyke:
there is not much functionality to test here, imho - it is just visual wrapper around links. what do you think?

PR #839, opened 2019-02-04
feat(table): adds filter functionality

chasestarr:
#### Description

Creates a popover to control filtering. This is an early version of the feature, does not include tests and docs site examples at the moment. Let me know if we agree on this approach and I will continue with implementation.

#### Scope

- [x] Minor: New Feature


chasestarr:
gonna fix up this branch...

PR #840, opened 2019-02-04
feat(table): adds filter functionality

chasestarr:
#### Description

Creates a popover to control filtering. This is an early version of the feature, does not include tests and docs site examples at the moment. Let me know if we agree on this approach and I will continue with implementation.

#### Scope

- [x] Minor: New Feature


nadiia:
Looks good to me 👍

tajo:
Looks good! One issue: It's not accessible. Is that going to come later?

chasestarr:
screen recording 2019-02-12 at 10 40 am

there is a bug when if the popover re-renders from a parent component state update, the popover 'loses' it's reference.

chasestarr:
fixed it by disabling events on this popover... this probably isn't the 100% correct approach. i couldn't find any real differences between our approach and the react-popper library

chasestarr:
bump - what do we think about the `eventsEnabled: false` solution? it felt safer that doing major changes to the popover component

tajo:
> bump - what do we think about the `eventsEnabled: false` solution? it felt safer that doing major changes to the popover component

Sounds good to me. What's the use-case of keeping these events enabled?

tajo:
Is there some plan to support keyboard controls? I think you can open it now but then there is no way how to focus the content of popover? I guess this should ideally have similar/same controls as the datepicker.

chasestarr:
ah interesting, i was testing tab-navigation in storybook before. seems to not work in the docs site? i'll likely need to build a wrapper like the datepicker for the content provided to filter

chasestarr:
fyi - i'm working on making this accessible now. pulling in the package [tabbable](https://github.com/davidtheclark/tabbable) to get the list of tabbable elements. I considered doing this myself, but that package has no deps so seems reasonable to include

chasestarr:
update: i went ahead and used the package `focus-trap` because i realized i was writing essentially that code and it only depends on the `tabbable` package i was using. the popover content now sits inside of a `TrapFocus` component it only exists in the table directory right now and is not exported, but it could be more broadly used within baseui

tajo:
> update: i went ahead and used the package `focus-trap` because i realized i was writing essentially that code and it only depends on the `tabbable` package i was using. the popover content now sits inside of a `TrapFocus` component it only exists in the table directory right now and is not exported, but it could be more broadly used within baseui

Somehow I missed the update, sorry for late response! It works much better now! 

There is one slight glitch: When you check the checkbox by hitting `space` it focuses the popover's body and when you hit `space` again it starts scrolling it. I would expect that the checked checkbox preserves the focus and repeated `space` press would just uncheck it. It would make navigation much smoother, now I have to always tab back to the position where I was.

I can make a screencap if it helps.

chasestarr:
good catch - i'll work on this today

PR #848, opened 2019-02-04
test(e2e): add progress steps e2e test

gergelyke:


gergelyke:
@chasestarr I think that makes perfect sense - how do you want to go about it? should we try that for the next component we cover with e2e tests?

PR #850, opened 2019-02-05
chore(build): ship es modules

gergelyke:


rtsao:
@chasestarr Unfortunately there is not any exhaustive or canonical documentation regarding the "module" field. However, by convention and de-facto implementations in webpack and Rollup, the "module" field behaves exactly the same as the "browser" field, which can be a map, per the spec: https://github.com/defunctzombie/package-browser-field-spec

In webpack, all these resolution rules are handled by: https://github.com/webpack/enhanced-resolve. In webpack itself, the browser target merely has an additional `"browser"` value in the `resolve.mainFields` list: https://github.com/webpack/webpack/blob/88a5050fa837506ff456bcff3d8d443ec9998caa/lib/WebpackOptionsDefaulter.js#L337-L345

PR #852, opened 2019-02-05
test(button-group): adds a11y e2e tests and enables e2e test isolation

chasestarr:
#### Description

I had found it kind of strange to include e2e test related code in the `examples.js` file like [here](https://github.com/uber-web/baseui/blob/cccd99ed79e2df7eb6802e5695c4dbd02d025dda/src/checkbox/examples.js#L61). There's no context for why they are there without looking at the e2e test. Also, for all new components, they need to be registered into the e2e tests file. Wouldn't it be great if that file could just pick up any new e2e tests? _it does now!_

While i wanted to enable in-line react components like we have in enzyme tests, i found it too tough to figure out. we would need to compile the test code with the bundle shipping to the dev server. The best i came up with was to co-locate the scenarios and tests to make them as explicitly related as possible.

I have not moved any existing e2e tests over to this pattern yet. let me know what we think


chasestarr:
i'd like to include migrating existing tests in this pr. just to be sure that everything works before merging it in

PR #855, opened 2019-02-06
feat(file-uploader): adds spinner if no progress info available

chasestarr:
#### Description

Often we start a file upload and we have no idea how long it will take. The file-uploader now displays a spinner in this case

#### Scope

- [x] Minor: New Feature


src/file-uploader/file-uploader.js
```diff
@@ -114,7 +116,7 @@ function Unstable_FileUploader(props: PropsT) {
 
               {afterFileDrop && (
                 <React.Fragment>
-                  {!!props.progressAmount && (
+                  {props.progressAmount ? (
```
nadiia: For an initial `0` progress it will render spinner. Or is the expectation to have the progressAmount > 0?

chasestarr: ah right - good catch

PR #861, opened 2019-02-07
docs(comparison): add comparison page

gergelyke:


documentation-site/pages/getting-started/comparison.mdx
```diff
@@ -0,0 +1,44 @@
...
+
+## Technical overview
+
+The table below shows you a high-level comparison of the most popular
+React component libraries and Base UI. While we try to do our best
+to keep it up-to-date, it may have incorrect information. If you
+notice that, please let us know in a GitHub issue, or send a Pull Request fixing it!
+
+<Table columns={['Name', 'CSS Engine', 'Customization', 'Size (gzipped)']} data={[
+  ['Base UI', 'CSS-in-JS, using Styletron', 'Through an extensive overrides interface', <a href="https://bundlephobia.com/result?p=baseui@5.13.2">~5kB</a>],
```
tajo: It's misleading to say that Base UI has `5kB`. We don't export any components from the `index.js` so bundlephobia considers only the size of some helper functions (withTheme, styled, etc...)

tajo: you would have to import all components into a single file and bundle it

PR #868, opened 2019-02-08
feat(slider): slider rework

tajo:
screen shot 2019-02-08 at 1 11 57 pm

**This is a breaking change.** The underlying implementation was completely replaced, so it now supports all browsers, mobile and a11y. Many overrides are also different since the DOM structure is changed. Slider now implements `disabled` state based on the design.

**Also, the API got simplified**. There is no `range` anymore, but you can control the behavior by setting `min`, `max` and `step`. It should be more intuitive since that's the standardized [input range API](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range). Also, all other libraries are using the same API too (e.g. [Material UI](https://material-ui.com/lab/slider/)).

It was not my intention to introduce this change without a RFC but I somehow missed that `range` prop is not just for setting `min` and `max` values but it can be also used to force non-uniform steps. That and also, the old slider has `step` prop too (it works only if range is not defined?)! 

For now, my argument is that non-uniform steps are not a good fit for the slider anyway and a group of `radios` should be used instead. Also, there is no design for that.

 I looked backed through old PRs to get more context on the API design and it seems that first proposal was written before we even had any design? Finally, I checked our internal projects, `Slider` is used only in two places and they don't use non-uniform steps (although they use a lot of overrides).

TickBar displays only `min` and `max` values. Again, we don't have design for other bars (with more ticks?) and it's probably not necessary to have intermediate values anyway since the current value is always visible above the thumb?


src/slider/slider.js
```diff
@@ -5,200 +5,135 @@ This source code is licensed under the MIT license found in the
...
-  }
-  return tick;
-}
 
-class Slider extends React.Component<PropsT, StatelessStateT> {
-  domNode: ?Element | Text;
+// values.length should not be bigger than two
+// because our design doesn't support more than
+// two thumbs
+const limitValues = (values: number[]): number[] =>
```
gergelyke: should throw an exception if more than 2 values are supported, just to make sure folks instantly know why it does not work

src/slider/README.md
```diff
@@ -55,78 +45,65 @@ export default () => {
...
 
 ## `Slider` API
 
-* `range: Array<Point>` - Required.
-  * All points present on slider axis. First and last point represent min and max
-    value. Each `Point` can be a value itself (for primitive type) or if `Point` is
-    `Object` it should have `value` and `label`. If more than 2 elements are present
-    in array, they represent all ticks on axis and `step` property is ignored.
-* `value: Point | Array<Point>` - Required.
+* `values: Array<number>` - Required.
```
nadiia: What do you think about changing the name of `values` props from plural to single `value`? We use single `value` in select even tho it accepts an array of values for multi-select.

src/slider/slider.js
```diff
@@ -5,200 +5,141 @@ This source code is licensed under the MIT license found in the
...
-}
 
-function getTickLabel(tick: number | {label: React.Node}): React.Node {
-  if (typeof tick === 'object') {
-    return tick.label;
+// values.length should not be bigger than two
+// because our design doesn't support more than
+// two thumbs
+const limitValues = (values: number[]) => {
+  if (values.length > 2 || values.length === 0) {
```
nadiia: Possibly could do `values` props not required and set the default to `[min, min]` or `[min, max]`

src/slider/slider.js
```diff
@@ -5,200 +5,141 @@ This source code is licensed under the MIT license found in the
...
+          max={max}
+          values={values}
+          disabled={disabled}
+          onChange={values => onChange({values})}
+          renderTrack={({props, children, isDragged}) => (
+            <Track
+              onMouseDown={props.onMouseDown}
+              onTouchStart={props.onTouchStart}
+              $isDragged={isDragged}
+              style={props.style}
```
nadiia: What is this style prop? Should allow style overrides through regular overrides API

src/slider/styled-components.js
```diff
@@ -6,65 +6,56 @@ LICENSE file in the root directory of this source tree.
...
-    marginLeft: sizing.scale400,
-    borderRadius: $theme.borders.useRoundedCorners ? borders.radius100 : '0px',
-    backgroundColor: colors.mono400,
-    height: sizing.scale100,
+    paddingTop: sizing.scale1000,
+    paddingBottom: sizing.scale600,
+    paddingRight: sizing.scale600,
+    paddingLeft: sizing.scale600,
+    display: 'flex',
+    cursor: $disabled ? 'not-allowed !important' : 'inherit',
```
nadiia: Shouldn't use `!important`. Is there a reason for it here?

nadiia:
Left a few comments, otherwise looks really great and works smooth!

src/slider/README.md
tajo: Since it's now always an array, I thought plural makes more sense? But I'm fine changing it. Wdyt @chasestarr / @gergelyke?

src/slider/slider.js
tajo: It would not be much helpful here since the component is stateless, so if you omitted the `values` prop you couldn't move the slider. What's the use-case for that? Maybe the default value would make sense in the `StatefulSlider` component?

nadiia: Oh, you're totally right. It can only be applicable to the stateful slider.

src/slider/README.md
chasestarr: i prefer singular `value`. we will maintain a more similar api to the html element. also, it leaves us open to making a DX improvement of overloading the prop with a single number value in the future.

documentation-site/pages/components/slider.mdx
```diff
@@ -36,27 +35,26 @@ Each slider has a primary control handle that cues user interaction.
...
-<Example title="Slider as a Stateful Component" path="examples/slider/stateful.js">
+<Example
+  title="Slider as a Stateful Component"
+  path="examples/slider/stateful.js"
+>
   <Stateful />
 </Example>
 
-<Example title="Slider Range example" path="examples/slider/range.js">
+<Example title="Slider Range" path="examples/slider/range.js">
```
chasestarr: can you add an example for how to achieve this use-case? `For example, if the defined range is [402, 8442] with 402 step increments, but you want to show "0.4km" or "0.25mi" as the label instead of "402"` This should be simple to do through overrides, but would be nice to help migration efforts

documentation-site/pages/components/slider.mdx
```diff
@@ -36,27 +35,26 @@ Each slider has a primary control handle that cues user interaction.
...
-  <RangeLabel />
-</Example>
-
-<Example title="Slider as a stepped slider" path="examples/slider/step.js">
-  <Step />
+<Example title="Disabled Slider" path="examples/slider/disabled.js">
+  <Disabled />
 </Example>
 
-<Example title="Slider with Multiple Ticks" path="examples/slider/ticks.js">
```
chasestarr: is this use-case still possible? 
screen shot 2019-02-11 at 8 19 54 am

package.json
```diff
@@ -153,6 +153,7 @@
     "react-dropzone": "^8.0.0",
     "react-input-mask": "^2.0.4",
     "react-movable": "^1.0.1",
+    "react-range": "^1.0.2",
```
chasestarr: looked at the package and big thumbs up. great work on this lib, @tajo

documentation-site/static/examples/slider/overrides.js
```diff
@@ -2,21 +2,30 @@ import React from 'react';
...
+        onChange={({values}) => this.setState({values})}
         overrides={{
-          Axis: {
-            style: ({$theme}) => ({background: $theme.colors.mono700}),
-          },
-          AxisRange: {
-            style: ({$theme}) => ({background: $theme.colors.mono1000}),
-          },
+          Thumb: ({$values, $thumbIndex}) => (
+            <div
```
chasestarr: can you add a `tabindex` here so that keyboard navs works?

src/slider/__tests__/slider.test.js
```diff
@@ -1,248 +0,0 @@
-/*
-Copyright (c) 2018 Uber Technologies, Inc.
-
-This source code is licensed under the MIT license found in the
-LICENSE file in the root directory of this source tree.
-*/
-// @flow
-import React from 'react';
-import {mount} from 'enzyme';
-import {Slider, StyledThumb} from '../index.js';
```
chasestarr: Is the thought behind deleting the `slider.test.js` file that the stateful component tests should cover the functionality?

src/slider/stateful-slider-container.js
```diff
@@ -22,28 +22,23 @@ class StatefulSliderContainer extends React.Component<
...
   state = {...this.props.initialState};
 
-  onChange = (config: {event: *, value: Array<number>}) => {
-    const {event, ...params} = config;
-    this.internalSetState(STATE_CHANGE_TYPE.change, event, params);
+  onChange = (params: {values: Array<number>}) => {
+    this.internalSetState(STATE_CHANGE_TYPE.change, params);
     const {onChange} = this.props;
-    return onChange({event, ...params});
+    return onChange({...params});
```
chasestarr: nit - do you mind not destructuring `onChange` from props here? it feels hard to follow when the method calling it is named the same.

also, something to add to the migration documentation is that we're no longer passing `event` to the `onChange` callback

src/slider/stateful-slider-container.js
```diff
@@ -22,28 +22,23 @@ class StatefulSliderContainer extends React.Component<
...
-    type: ChangeActionT,
-    e: SyntheticInputEvent<HTMLInputElement>,
-    {value}: ParamsT,
-  ) => {
-    const nextState = {value};
+  internalSetState = (type: ChangeActionT, {values}: ParamsT) => {
+    const nextState = {values};
     const {stateReducer} = this.props;
-    const newState = stateReducer(type, nextState, this.state, e);
+    const newState = stateReducer(type, nextState, this.state);
```
chasestarr: may want to call out in the migration documentation that we are no longer passing the event

src/slider/styled-components.js
```diff
@@ -75,88 +75,55 @@ export const TickBar = styled('div', props => {
...
+    backgroundColor: $isDragged ? $theme.colors.primary : $theme.colors.mono600,
+  };
+});
+InnerThumb.displayName = 'StyledInnerThumb';
+
+export const ThumbValue = styled('div', props => {
+  const {$theme} = props;
+  return {
+    position: 'absolute',
+    top: '-24px',
```
chasestarr: is this dependent on the font size? should it be calculated so that it can handle customized theme sizing?

src/slider/__tests__/slider.test.js
tajo: The functionality of slider is already tested in `react-range`.  So it doesn't make sense to test it twice.

`baseui/slider` just adds visual components on a top of it.

documentation-site/pages/components/slider.mdx
tajo: Added this example:

screen shot 2019-02-11 at 1 35 08 pm

It also answers the multiple ticks use-case. Previously, it just displayed whatever user manually specified through the `range` prop. Since this prop doesn't exist anymore, you have to DIY through the `TickBar` override. Yep, that's a bit more work but we display the current value above the thumb by default so it doesn't seem so necessary (also no design for that).

Moreover, if we decided to solve this properly, `justify-content: space-between` doesn't cut it  since it's not responsive. There needs to be more logic to hide some elements when the viewport is too narrow. That's something that `d3` solves but I would say it's out of the scope here and we can introduce some dataviz like components separately in the future.

documentation-site/pages/components/slider.mdx
tajo: answered in a previous comment

src/slider/slider.js
tajo: default (single mid-point) added to StatefulSlider

src/slider/README.md
tajo: renamed

PR #871, opened 2019-02-11
chore(build): removes most of storybook and rebuilds source for docs site changes

chasestarr:
#### Description

to start the docs site run the script below
```
yarn documentation:watch
```

gets rid of all `examples` files. you may want to use the docs site for testing at this point, but feel free to continue using storybook (it's not being turned off!). when the component source changes, it will recompile so that the docs site and get the updates.

i've removed the storybook addons since we are not publishing it as the main docs site. i like to think of it now as the main source of visual regression testing and e2e tests. storybook also loads uber-move font by default now since that's what screener relies on. 

all `.scenario` files are automatically loaded into the storybook, no need to create a `stories.js` file. those files will also be ran through screener so there's no need to 'register' the file with the screener config.


gergelyke:
it works great 👍 

I am wondering if we can make it a bit faster though? or are we OK with a few seconds of build time?

chasestarr:
yeah i agree it's slow. right now we are transpiling the src into `dist` so that the docs site reads it as a npm module. it's using the same configs as production for dev. we could look into deviating from prod configs to improve perf

gergelyke:
probably I am the one who'll be least affected - what do you think @tajo @nadiia?

chasestarr:
to be clear, it's not any slower than it was before and it's likely faster because rebuilding is now automated.

tajo:
The watch task calls:

```js
"build:es2015": "cross-env NODE_ENV=production babel ./src --out-dir ./dist --ignore **.test.js,**/__tests__/**,**/e2e.js,template-component/**",
```

It would be great if it could `babel` only the changed file, that would speed up the rebuild dramatically.

chasestarr:
good idea - i can look into how nodemon could get us the changed file although i worry about multiple file changes very quickly. let me run a test

tajo:
I think babel —watch could replace the nodemon completely?

chasestarr:
i can't get babel --watch to react to src changes? ran the cmd below
`yarn babel ./src --out-dir ./dist --watch`

chasestarr:
ah - i was expecting some log statement to say it's rebuilding. cool, i've removed nodemon and replaced the watch task with babel. it's super fast now :) thanks @tajo !

package.json
```diff
@@ -38,9 +38,8 @@
     "prepublish": "npm run build",
     "icon:generate": "./src/icon/build-icons.js",
     "now": "now",
-    "documentation:dev": "yarn build:es2015 && next documentation-site",
-    "documentation:build": "yarn build:es2015 && next build documentation-site && next export documentation-site -o public",
-    "documentation:watch": "nodemon --watch src --exec yarn documentation:dev"
+    "documentation:dev": "yarn build:es2015 --watch & next documentation-site",
```
tajo: 👍 👍 👍

tajo:
It works beautifully now, even the HMR!

PR #873, opened 2019-02-12
react-movable v2 upgrade

tajo:
It brings much better [support for iOS Safari](https://github.com/tajo/react-movable/releases/tag/v2.0.0) and there is no need to `stopPropagation` for remove handle.

Also, underlying flow types are now correctly linked, so some updates were necessary to fix it for baseui.

### DIFF
--- a/src/dnd-list/types.js
+++ b/src/dnd-list/types.js
@@ -40,7 +40,7 @@ export type ListPropsT = {|
   /** Items (labels) to be rendered */
   items: Array<React.Node>,
   /** Handler for when drag and drop is finished and order changed or item is deleted (newIndex would be -1 in that case) */
-  onChange?: ({oldIndex: number, newIndex: number}) => mixed,
+  onChange: ({oldIndex: number, newIndex: number}) => void,
   overrides?: OverridesT,
 |};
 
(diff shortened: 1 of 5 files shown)


src/dnd-list/types.js
```diff
@@ -40,7 +40,7 @@ export type ListPropsT = {|
   /** Items (labels) to be rendered */
   items: Array<React.Node>,
   /** Handler for when drag and drop is finished and order changed or item is deleted (newIndex would be -1 in that case) */
-  onChange?: ({oldIndex: number, newIndex: number}) => mixed,
+  onChange: ({oldIndex: number, newIndex: number}) => void,
```
chasestarr: why switch from mixed?

tajo: It needs to match the `onChange` signature of `react-movable` which returns `void`.

tajo: But if you need to do some `await` shenanigans, StatefulList's `onChange` still returns `mixed`.

chasestarr: fine with me so that it can interop with the external lib

chasestarr: fyi this is technically a breaking change because it restricts the type constraint

tajo: Good point, I published new flow types so we can keep using mixed.

PR #874, opened 2019-02-12
chore(button-group): deprecate kind prop

chasestarr:
#### Description

deprecates the `kind` prop from `button-group`. design only wants a single option here

#### Scope

- [x] Major: Breaking Change


chasestarr:
@tajo how did you generate the migration file?

gergelyke:
@chasestarr it is a manual thing, you have to create that in the `docs/migrations` folder, where the name comes from the PR's id

PR #877, opened 2019-02-12
fix(input): add aria-label/labelledby

nadiia:


src/input/__tests__/__snapshots__/input.test.js.snap
```diff
@@ -3,6 +3,8 @@
 exports[`Input - basic functionality: input has correct props 1`] = `
 Object {
   "aria-invalid": false,
+  "aria-label": undefined,
```
gergelyke: do we want to provide a value for the tests?

nadiia: updated, should have default `null`

nadiia: + updated e2e setting

chasestarr:
fyi i added this same feature for `radio` here https://github.com/uber-web/baseui/blob/master/src/radio/types.js#L33-L42 camelCased may want to stay consistent?

nadiia:
> fyi i added this same feature for `radio` here https://github.com/uber-web/baseui/blob/master/src/radio/types.js#L33-L42 camelCased may want to stay consistent?

The `aria-label` is the original attribute and prop name, what do you think about sticking to it to avoid a confusion?

gergelyke:
can we support both in all the components? would that make sense?

Ideally, we'd stay consistent but at the same time we would not introduce a breaking change

chasestarr:
i understand that it's the name react uses with elements, but it honestly feels like a strange deviation on react's end. 

i'm fine with enabling both options for radio, making a note for future deprecation, and going with `aria-*` across the repo

gergelyke:
following up on this, @nadiia / @chasestarr would one of you be up for sending a PR to the radio component to make it work consistently (with the deprecation notice)?

chasestarr:
fyi i will be adding the radio change in my radio overrides pr today

PR #886, opened 2019-02-14
fix(datepicker): add couple more e2e tests

nadiia:


src/datepicker/__tests__/datepicker.e2e.js
```diff
@@ -142,4 +145,48 @@ describe('Datepicker', () => {
...
+    await page.waitFor(selectors.input);
+    await page.click(selectors.input);
+    await page.waitFor(selectors.calendar);
+    await page.waitFor('[aria-label="Sunday, March 10th 2019"]');
+    await page.click(selectors.yearSelect);
+    const dropdown = `${selectors.yearSelect} ${selectors.selectDropdown}`;
+    await page.waitFor(dropdown);
+
+    await page.$$eval('ul[role="listbox"] li', items => {
+      const arr = items.filter(item => {
```
gergelyke: nit: `items.find` may make this a bit more simple, as that would return only the single item, and not the array

PR #888, opened 2019-02-14
docs(layout): add proposal

gergelyke:
heavily inspired by ant's layout system, this is a simplified version of that


src/layout/README.md
```diff
@@ -0,0 +1,72 @@
+# Layout Component
+
+## Usage
+
+* `Layout`: The layout wrapper, can contain `Header`, `Sider`, `Content`, `Footer` or `Layout` itself,
+* `Header`: The top layout, in which any element can be nested, and must be placed in `Layout`.
+* `Sider`: The sidebar with default style and basic functions, in which any element can be nested, and must be placed in `Layout`.
```
chasestarr: can we name this `Sidebar`? i'd rather use a more common word

gergelyke: that's a good point, done

hemildesai:
Hi, I wanted to ask if there were any updates on when this feature will be available? Or if there were any examples that uses this kind of layout structure?

PR #889, opened 2019-02-14
fix(radio): decouples radio from checkbox, enables overrides

chasestarr:
#### Description

**adds a new component** `Radio` that should be used in place of `StyledRadio`
**adds deprecation warnings** to various elements of radio-related components
- `StyledRadio` component deprecation.
- `Radiomark` component deprecation.
- Limits supported overrides on `RadioGroup` to only `RadioGroupRoot`. Other overrides will need to be passed to `Radio`.
- Removes the `Radiomark` override on `Radio` component.

#### Scope

- [x] Minor: New Feature


src/radio/radio.js
```diff
@@ -5,5 +5,161 @@ This source code is licensed under the MIT license found in the
...
+    }
+  }
+
+  onMouseEnter = (e: SyntheticInputEvent<HTMLInputElement>) => {
+    this.setState({isHovered: true});
+    this.props.onMouseEnter && this.props.onMouseEnter(e);
+  };
+
+  onMouseLeave = (e: SyntheticInputEvent<HTMLInputElement>) => {
+    this.setState({isHovered: false, isActive: false});
```
nadiia: Do we need to unset the `isActive` here for the case when mouseUp happens not on the element?

src/radio/radio.js
```diff
@@ -5,5 +5,161 @@ This source code is licensed under the MIT license found in the
...
+        {...sharedProps}
+        {...rootProps}
+      >
+        {isLabelTopLeft(this.props.labelPlacement) && label}
+        <RadioMarkOuter {...sharedProps} {...radioMarkOuterProps}>
+          <RadioMarkInner {...sharedProps} {...radioMarkInnerProps} />
+        </RadioMarkOuter>
+        <Input
+          aria-invalid={this.props.isError || null}
+          aria-required={this.props.isError || null}
```
nadiia: Do we want to introduce `required` prop for it instead of setting required based on error state?

src/radio/radiogroup.js
```diff
@@ -31,70 +27,68 @@ class StatelessRadioGroup extends React.Component<PropsT, StatelessStateT> {
...
+    );
 
-    const {
-      Root = StyledRoot,
-      RadioMark = StyledRadioMark,
-      Label = StyledLabel,
-      Input = StyledInput,
-    } = overrides;
+    if (__DEV__) {
+      if (this.props.ariaLabel || this.props.ariaLabelledBy) {
```
nadiia: 👍 thanks for adding this warning!

src/radio/types.js
```diff
@@ -76,6 +93,50 @@ export type StateT = {
...
+  /** handler for focus events on trigger element. */
+  onFocus?: (e: SyntheticInputEvent<HTMLInputElement>) => void,
+  /** Handler for mouseenter events on trigger element. */
+  onMouseEnter?: (e: SyntheticInputEvent<HTMLInputElement>) => void,
+  /** Handler for mouseleave events on trigger element. */
+  onMouseLeave?: (e: SyntheticInputEvent<HTMLInputElement>) => void,
+  /** Handler for mousedown events on trigger element. */
+  onMouseDown?: (e: SyntheticInputEvent<HTMLInputElement>) => void,
+  /** Handler for mouseup events on trigger element. */
+  onMouseUp?: (e: SyntheticInputEvent<HTMLInputElement>) => void,
```
nadiia: let's return mixed type instead of void

src/radio/radio.js
chasestarr: looks like the radio mark doesn't implement active states at the moment. also based on figma, it seems like were kind of off here. i'll make a note for us to review here

PR #899, opened 2019-02-15
fix(progress-step): overflow fixes

fanglansheng:
Fixes https://jeng.uberinternal.com/browse/WPT-2512
#### Description
Added `<Select/>` in dotted.js example.
Removed overflow: hidden from `StyledContent` to support content with long dropdown.

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


fanglansheng:
@gergelyke Some help to format the PR? thanks :)

gergelyke:
thanks for the fix @fanglansheng !

it seems like there is a bit of a visual regression introduced with this change, it adds a 4px bottom padding - can we remove that?

screen shot 2019-02-15 at 2 26 51 pm

screen shot 2019-02-15 at 2 26 38 pm

fanglansheng:
Sure, remove 4px for the StyledContent right?

gergelyke:
I think so, yes

fanglansheng:
I added `height: 'calc(100% + 14px)',` to StyledContentTail to fix the gaps here after remove padding:
image

src/progress-steps/styled-components.js
```diff
@@ -130,7 +130,7 @@ export const StyledContentTail = styled(
       ':after': {
         content: '""',
         display: 'inline-block',
-        height: '100%',
+        height: 'calc(100% + 14px)',
```
chasestarr: please use a theme `sizing` value rather than the `14px` so that it can scale based on theme values. look up the `creator.js` file in the project to view available values

gergelyke:
it seems like the spacing is not quite there yet - can you help us updating these to match?

screen shot 2019-02-19 at 2 19 31 pm
screen shot 2019-02-19 at 2 19 09 pm

gergelyke:
disregard my last comment, my bad :)

PR #900, opened 2019-02-15
docs(algolia): exclude parts of the website

gergelyke:


documentation-site/components/sidebar.js
```diff
@@ -81,7 +81,7 @@ const NavigationItem = props => {
 };
 
 export default () => (
-  <List as="ul">
+  <List data-crawler={'exclude'} as="ul">
```
nadiia: Isn't it how it finds the rest of the pages? So if excluded it won't get to all the components pages?

gergelyke: as far as I know, it only affects what's added to the index, and the stopUrl is the one that stops the crawling

I'll ask them in the other PR

gergelyke:
merging it as is now, and will change if needed

PR #901, opened 2019-02-18
feat(datepicker-popover): adds support of mountNode

jcemer:
#### Description

It adds support of mountNode attribute to Datepicker and Popover. It creates feature compatibility with Modal.

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change


src/popover/popover.js
```diff
@@ -399,6 +399,16 @@ class Popover extends React.Component<PopoverPropsT, PopoverPrivateStateT> {
...
   }
 
+  getMountNode(): HTMLElement {
+    const {mountNode} = this.props;
+    if (mountNode) {
+      return mountNode;
+    }
+    // Flow thinks body could be null (cast through any)
+    // eslint-disable-next-line flowtype/no-weak-types
+    return ((document.body: any): HTMLBodyElement);
```
nadiia: Need to wrap it in the `if (__BROWSER__) {}`

schnerd:
Seems like this makes sense – probably something that may need to be supported in any component that uses portal.

Would anyone prefer a `getMountNode: () => HTMLElement` callback instead of a `mountNode: HTMLElement` prop? This is something I didn't fully explore when building `Modal` so I settled on the latter, but I'm wondering if anyone sees benefits to using the callback style:

```js
getPopoverMountNode() {
  return document.querySelector('.fake-body');
}

render() {
  return <Popover getMountNode={this.getPopoverMountNode} ...>
}
```
vs
```
getPopoverMountNode() {
  return document.querySelector('.fake-body');
}

render() {
  const mountNode = document.querySelector('.fake-body');
  return <Popover mountNode={mountNode} ...>;
}
```

If they're essentially the same, then we can roll with whats in the current PR.

gergelyke:
@jcemer would you mind updating the jest snapshots / tests?

src/popover/popover.js
jcemer: I'm not sure because I get this implementation from the Modal component and I can't see the condition there.

jcemer: It is already checked on `render`.

jcemer:
@nadiia @gergelyke I fixed the tests.

gergelyke:
thanks @jcemer ! would you mind signing our CLA ? https://cla-assistant.io/uber-web/baseui?pullRequest=901

jcemer:
@gergelyke done.

tajo:
What is the use-case of using custom mountNode for any of those components? SSR?

jcemer:
@tajo It is for the cases there is a need to attach the popover in another element than document.body. 

In our case, we are rendering inside an iframe using a portal. The popover is mounting in the parent window so there is a need to have `mountNode` to set it to the document.body of the iframe. The modal already allows us to define it.

jcemer:
@nadiia @chasestarr @gergelyke I updated the branch, let me know if there is something missing. It would be great to have it since datepicker is stable now 🎉

gergelyke:
thanks, @jcemer! as Nadiia recommended, can you please add test coverage?

jcemer:
@gergelyke @nadiia could you advise how to write it because I couldn't find a test that covers it in the modal component.

nadiia:
Let's merge this change as is.

gergelyke:
@jcemer would you mind resolving this (hopefully) last merge conflict? We'll land it after that's done

gergelyke:
and thanks for your patience! :)

jcemer:
@gergelyke done :)

chasestarr:
take a look at these docs for running the e2e tests: https://github.com/uber-web/baseui/blob/master/TESTING.md

nadiia:
Merging. Thanks again for the contribution!

PR #905, opened 2019-02-19
fix(storybook): use move font in storybook

chasestarr:
#### Description

this will likely cause many screener errors

#### Scope

- [x] Patch: Bug Fix


.storybook/config.js
```diff
@@ -5,14 +5,14 @@ import {Provider as StyletronProvider} from 'styletron-react';
...
-import {LightTheme} from '../src/themes';
+import {LightThemeMove} from '../src/themes';
 
 // Add providers for theme and styletron
 const engine = new Styletron();
 addDecorator(story => {
   return (
     <StyletronProvider value={engine}>
-      <ThemeProvider theme={LightTheme}>{story()}</ThemeProvider>
+      <ThemeProvider theme={LightThemeMove}>{story()}</ThemeProvider>
```
nadiia: Do we want to replace it or add a switch to toggle between move and system fonts?

gergelyke: we have a separate story to support theme switching for the docs site, I think

chasestarr: this is also only for the storybook, which is only here for screener now

PR #925, opened 2019-02-21
feat(datepicker): make the datepicker a stable component

nadiia:


src/datepicker/types.js
```diff
@@ -102,19 +102,19 @@ export type CalendarInternalState = {
 
 export type CalendarPropsT = {
   /** Defines if the calendar is set to be focused on an initial render. */
-  calFocusedInitially: boolean,
+  autoFocuseCalendar: boolean,
```
gergelyke: should it be `autoFocusCalendar`? or maybe simply `autoFocus`

nadiia: I'd keep the `Calendar` part since it makes a date in the calendar focused and not the input element of the datepicker. So the `autoFocus` may be confusing

nadiia:
> should we also rename the following to be consistent?
> 
> * `isHighlighted` -> `highlighted`
> * `isFocused` -> `focused `

`isFocused` is an internal state's prop name, so I'd probably keep it. Don't have a strong opinion on it, can rename if that looks better.
Will rename the `isHighlighted` prop.

src/datepicker/types.js
gergelyke: oh, okay, that makes perfect sense - thanks for the clarification!

gergelyke:
Yeah, I don't have a strong opinion either - it just feels strange to have a single prop starting with `is`, while all the rest are not

PR #928, opened 2019-02-21
feat(tag): static

gergelyke:


chasestarr:
can we simplify the api so that if `closeable === false` turn off the hover effect?

gergelyke:
that makes perfect sense

PR #929, opened 2019-02-21
fix(tag): closable prop disables hover

gergelyke:


gergelyke:
@chasestarr @nadiia I've re-opened a new PR; instead of a feature I've just opted for a patch - what do you think?

PR #931, opened 2019-02-22
fix(tabs): do not render non-active tabs content

nadiia:


src/tabs/tabs.js
```diff
@@ -83,7 +83,7 @@ export default class Tabs extends React.Component<TabsPropsT> {
           {...tabContentProps}
           {...props}
         >
-          {child.props.children}
+          {sharedProps.$active ? child.props.children : null}
```
chasestarr: nit: my mental understanding on `sharedProps` in the code base is that it only affects the visual style. when changing the `sharedProps`, i would be surprised to have the rendered markup change. is it reasonable to initialize an `isActive` var, and use that here and on L74?

nadiia: Definitely, let me change this quick fix

PR #941, opened 2019-02-22
feat(table): adds heading cell example

chasestarr:
#### Description

screen shot 2019-02-22 at 1 39 10 pm


chasestarr:
going to close this pr so that design only has to review 1 screener diff

see: https://github.com/uber-web/baseui/pull/948

PR #944, opened 2019-02-23
feat(table): adds delta cell example

chasestarr:
#### Description

depends on https://github.com/uber-web/baseui/pull/941

screen shot 2019-02-22 at 4 08 04 pm


chasestarr:
going to close this pr so that design only needs to review 1 screener diff

see https://github.com/uber-web/baseui/pull/948

PR #951, opened 2019-02-25
feat(i18n): add i18n support

gergelyke:
this is a PoC implementation for the i18n support - please let me know what you think, and if everyone is onboard, I'll extend it for the rest of the codebase in this PR


src/datepicker/datepicker.js
```diff
@@ -138,58 +139,72 @@ export default class Datepicker extends React.Component<
       overrides.Popover,
       Popover,
     );
+
+    if (__DEV__ && this.props['aria-label']) {
+      // eslint-disable-next-line no-console
+      console.warn(
+        `Providing i18n data through properties will be removed in the next major version. Please use the LocalProvider.`,
```
tajo: Do we really want to remove the props? They are straightforward to use and can be handy for some tweaks and overrides even when the `LocalProvider` is used. Also, it's not a big overhead for us and seems like an unnecessary breaking change that requires a lot of grunt work to fix.

// The only thing I would change is to group them into a single `this.props.translate` object so it's consistent across all components and doesn't pollute the top-level props but that ship already sailed.

src/locale/types.js
```diff
@@ -0,0 +1,15 @@
...
+LICENSE file in the root directory of this source tree.
+*/
+// @flow
+
+export type DatepickerLocaleT = {
+  ariaLabel: string,
+};
+
+export type LocaleT = {|
+  datepicker: DatepickerLocaleT,
```
tajo: Should we be overriding the English locale and keep all `LocaleT` props optional?

tajo: I mean user can do that but it should be a `deepMerge` and seems nice to do it by default.

gergelyke: that makes perfect sense 👍

src/datepicker/datepicker.js
gergelyke: That can work too - the idea behind it was to make sure we only have a single API interface to maintain (and make it straightforward from a consumer point of view too), but I guess we can go either way.

What do you think @nadiia @tajo @schnerd ?

gergelyke: Let's keep the props as they are - I am going to update this PR

src/datepicker/datepicker.js
```diff
@@ -170,7 +162,7 @@ export default class Datepicker extends React.Component<
               <InputComponent
                 aria-disabled={this.props.disabled}
                 aria-label={
-                  locale.datepicker.ariaLabel || this.props['aria-label']
+                  this.props['aria-label'] || locale.datepicker.ariaLabel
```
tajo: Nice. I wanted to point this one out. 👍

src/locale/en_US.js
```diff
@@ -0,0 +1,19 @@
+/*
+Copyright (c) 2018 Uber Technologies, Inc.
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+// @flow
+
+const en_US = {
+  datepicker: {
```
chasestarr: could we co-locate the english labels with the component code? with the theme colors, i find it hard to jump between the different contexts. it's also tough to 'know' where these labels are located when coming to the project for the first time

src/locale/types.js
```diff
@@ -0,0 +1,18 @@
+/*
+Copyright (c) 2018 Uber Technologies, Inc.
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+// @flow
+
+export type DatepickerLocaleT = {|
+  ariaLabel: string,
```
chasestarr: same as my other comment, could this be co-located with the component code?

src/locale/en_US.js
gergelyke: yeah, that makes sense

src/datepicker/datepicker.js
nadiia: Agree on keeping the aria attr in props 👍 For those who don't use i18n would be slightly too much work to do in order to change those if removed, specifically for an input case, for example.

src/button-group/button-group.js
```diff
@@ -49,12 +51,18 @@ function getBorderRadii(index, length) {
...
 
-export default function ButtonGroup(props: PropsT) {
+type LocaleT = {|locale?: ButtonGroupLocaleT|};
+export function ButtonGroupRoot(props: {|...PropsT, ...LocaleT|}) {
   const {overrides = {}} = props;
   const [Root, rootProps] = getOverrides(overrides.Root, StyledRoot);
 
   return (
-    <Root aria-label={props.ariaLabel} {...rootProps}>
+    <Root
```
chasestarr: could context consumer be placed here rather than in the wrapper component? not sure why locale needs to be provided through props

src/modal/modal.js
```diff
@@ -275,47 +276,51 @@ class Modal extends React.Component<ModalPropsT, ModalStateT> {
...
+                  // aria-modal replaces the need to apply aria-hidden="true" to all other page
+                  // content underneath the modal.
+                  // https://www.w3.org/TR/wai-aria-practices-1.1/examples/dialog-modal/dialog.html
+                  'true'
+                }
+                role={role}
+                $ref={this.getRef('Dialog')}
                 {...sharedProps}
-                {...getOverrideProps(CloseOverride)}
+                {...getOverrideProps(DialogOverride)}
```
chasestarr: 😮 👍

src/button-group/button-group.js
gergelyke: yeah, sadly enyzme does not fully support contexts, so sadly the test framework affected our production code :( 

is this correct @tajo ?

tajo: Yea, this is a workaround for Enzyme. [It can't](https://github.com/airbnb/enzyme/issues/1908#issuecomment-439747826) shallow render the `renderProp`.

chasestarr: could we simply `mount` and avoid the indirection? i've found that code bases grow in puzzling ways when ad hoc refactoring to enable test patterns

gergelyke: I agree, 100%.

We've tried that too, it won't work - I think Vojtech will want to bring up the issues with enzyme, and start a chat on what we can do about it

for now, would you be ok with it as is, and fix it once we figure out what to do with enzyme?

chasestarr: for sure - i would personally prefer to remove the test than add the additional indirection, but up to you

src/button-group/button-group.js
```diff
@@ -89,8 +99,15 @@ export default function ButtonGroup(props: PropsT) {
   );
 }
```
chasestarr:
```suggestion

// The wrapper component below was created to continue to support enzyme tests for the ButtonGroup
// component. Enzyme at the moment does not support React context @ 16.3. To get around the limitation
// in enzyme, we create a wrapper around the core ButtonGroup and pass context as a prop. In our tests,
// only ButtonGroupRoot will be tested.
// https://github.com/airbnb/enzyme/issues/1908#issuecomment-439747826
```

PR #963, opened 2019-02-26
docs: markdown element improvements and React bump

tajo:
I ported some changes from the new styletron docs into base ui docs:

- bump to the latest React (so we can use/test hooks in our docs)
- inline code is styled similarly to GitHub
- blockquote is styled too, so feel free to use it to highlight / quote some information
- enable cross-linking, it uses `prefetched next/link` when internal, `target="_blank"` when external
- markdown links use now `baseui/link`
- heading anchors are styled similarly to GitHub, they appear only if you hover over the heading
- more robust function when creating the anchor slug, it can handle a variety of special characters and even React Components (so if you put a link or inlined code into the heading it recursively extracts the text)

## Before
screen shot 2019-02-26 at 3 32 06 pm

## After
screen shot 2019-02-26 at 3 30 58 pm
screen shot 2019-02-26 at 3 31 07 pm


gergelyke:
Thanks for improving these! One issue I found: seems like the Example's headings are gone

screen shot 2019-02-26 at 3 47 31 pm

tajo:
> Thanks for improving these! One issue I found: seems like the Example's headings are gone

Ooops. Looking into that.

Also unit tests are failing, need to update Enzyme and some tests.

PR #966, opened 2019-02-27
fix(table): gets rid of react-virtualized warnings

chasestarr:


tajo:
Why `virtual-horizontal-scroll.js` sets `fixedWidth: true`. This example should be about `left<->right` scrolling? Should we add there hundreds of columns to demonstrate that?

PR #968, opened 2019-02-28
feat(layout): adds layout component

chasestarr:
#### Description

based on the proposal [here](https://github.com/uber-web/baseui/blob/9803d1c53f412da071a7b38e5f9f6140e4347d9c/src/layout/README.md) and references AntD's layout component. chose to implement as straightforward as possible without sidebar collapse like ant or responsiveness concerns. those could definitely be improvements for the future but i think it may be premature to bake in those patterns.

screen shot 2019-02-28 at 12 44 35 pm

#### Scope

- [x] Minor: New Feature


gergelyke:
should we dogfood this, and start using it for the docs site?

chasestarr:
sure - i replace the current layout with this

src/layout/index.js
```diff
@@ -0,0 +1,17 @@
...
+LICENSE file in the root directory of this source tree.
+*/
+// @flow
+export {default as Content} from './content.js';
+export {default as Header} from './header.js';
+export {default as Layout} from './layout.js';
+export {default as Sidebar} from './sidebar.js';
+
+// Styled elements
+// export {} from './styled-components.js';
```
gergelyke: should we remove it?

gergelyke: or is it a todo to export them?

chasestarr: ah good catch - i meant to export them

tajo:
Looking good. 👍 

Should we add an example when `sidebarCount > 1`? It seems there's some code handling it as a special case?

**Ad responsiveness**: Frankly, I would like to see it as a part of this mvp. Reasons:

- it might be pretty hard to do it as a afterthought (generally it's easier to do the mobile first) and we don't want to risk breaking changes or big visual regressions (and layout is very visible and impactful)
- it limits the utility a lot, solving responsiveness and making the sidebar collapsing is not trivial and would definitely motivate developers to use this component more, without it, you can create these layouts just with a couple of divs and few lines of CSS
- as @gergelyke said, it would be great to dogfood it and we want to make baseui.design responsive as well

Alternatively, we could extend the RFC and describe what the mobile story is. There are some concerns to address (for example, if and how are we going to let users to define breakpoints?). This way we would know there will be no surprises.

chasestarr:
👍 i can add responsiveness next week.

chasestarr:
closing.. not quite sure if design is ready to go through with this implementation

PR #972, opened 2019-03-01
docs(title): add title to all pages

gergelyke:


gergelyke:
semanticely speaking, there should be a single h1 on any single pages, right?

chasestarr:
there definitely _should_ be, but not enforced. just something to keep in mind

PR #975, opened 2019-03-01
fix(select): render the dropdown in a popover

chasestarr:
#### Description

Changes a solid amount of code to enable this. Greatest hits outlined below:

- **adds innerRef prop to dropdown component** - to ensure that the dropdown element renders with the same width as the input element, I added a ref here so it can be inspected from the parent `Select` component. 
- **removes most styles from dropdown-container** - leaving these alone added extra padding inside of the popover.
- **popover onClickOutside callback now passes the event** - did this to maintain a consistent api on select.
- **inlines the renderMenu code** - since this was not re-used, avoided passing args through a function.
- **removes one test** - because the dropdown is now rendering in a portal, enzyme can't access the element. I may be able to add it back once https://github.com/uber-web/baseui/pull/901 lands.
- **removes aria-activedescendant from menu** - it was being [used incorrectly](https://developers.google.com/web/fundamentals/accessibility/semantics-aria/aria-labels-and-relationships#aria-activedescendant) and threw errors in e2e tests. making a note to properly implement after this pr.
- **updates scrollItemIntoView util in menu** - this was causing 'jumping' because it would scroll the outer-most window as well as the select dropdown scroll. I noticed some pref degradation with selects over 200 items long. Will add another pr to possibly debounce the call.
- **fixes popover mounting prematurely** - when transitioning `isMounted` from `false` to `true`, it would call the popover unmount code unintentionally. this lead to a bug where page loads followed by fast clicks on the select component would choose an option in the dropdown menu.
- **makes some e2e tests more reliable** - select and datepicker tests were failing maybe 1 out of 5-6 runs. adds some more waits to ensure proper execution.

#### Scope

- [x] Patch: Bug Fix


src/select/select.js
```diff
@@ -578,7 +556,9 @@ class Select extends React.Component<PropsT, SelectStateT> {
           <MultiValue
             value={value}
             key={`value-${i}-${value[this.props.valueKey]}`}
-            removeValue={() => {
+            removeValue={event => {
+              // stop propagation so that clicks on tag action button does not open dropdown
+              event.stopPropagation();
```
tajo: It's better to avoid `stopPropagation` whenever possible since it can lead to [strange bugs](https://css-tricks.com/dangers-stopping-event-propagation/).

The better solution would be to implement this logic on the other side. So the dropdown would check if the event comes from `role=button` element  and if so, it would ignore it (our `Tag` component is now missing that role so it should be added first). 

It can be turned into more general function:

```js
export function checkIfInteractive(target: Element, rootElement: Element) {
  const DISABLED_ELEMENTS = [
    'input',
    'textarea',
    'select',
    'option',
    'button',
    'a'
  ];
  while (target !== rootElement) {
    if (DISABLED_ELEMENTS.includes(target.tagName.toLowerCase())) {
      return true;
    }
    const role = target.getAttribute('role');
    if (role === 'button' || role === 'link') {
      return true;
    }
    if (target.tagName) target = target.parentElement!;
  }
  return false;
}
```

Where the `rootElement` is the dropdown and `target` is the tag. Then if `checkIfInteractive` returns `true`, we can ignore it.

tajo:
It looks and works good!

nadiia:
The focused state doesn't go away when select loses focus: 
screen shot 2019-03-05 at 4 27 53 pm

nadiia:
The dropdown doesn't stick to the input when the height changes: 
dropdown

src/select/select.js
chasestarr: updated - thanks

nadiia:
Opening the dropdown makes the page jump, got two examples on datepicker (checked it since e2e tests fail on it and it reused select for the month and year select). The second one is actually reproduceable on the master too, probably not related. But the first one is not on master but I'm not 100% sure is related to your change:

jump

jump2

src/select/select.js
```diff
@@ -18,44 +35,59 @@ import {
...
   wrapper: ?HTMLElement;
   input: ?HTMLInputElement;
   dragging: boolean;
+  // focusAfterClear is a flag to indicate that the dropdowm menu should open after a selected
+  // option has been cleared.
   focusAfterClear: boolean;
+  // openAfterFocus is a flag to indicate that the dropdown menu should open when the component is
+  // focused. Developers have the option to disable initial clicks opening the dropdown menu. If not
+  // disabled, clicks will set this flag true. Upon focusing, look to this to see if the menu should
+  // be opened, or only focus.
```
nadiia: 👍

src/select/select.js
```diff
@@ -98,139 +132,101 @@ class Select extends React.Component<PropsT, SelectStateT> {
...
-    }
-    // $FlowFixMe
-    if (event.target.tagName === 'INPUT') {
+    // Case comes up when text has been typed into the input field. If no text provided,
+    // the 'input' element will have essentially 0 width therefore will not be clickable.
+    // When click outside does not reset input, text provided will stay rendered after clicks away
+    // from the select component. Upon subsequent clicks on the provided text, open the dropdown
+    // menu, in addition to text edit operations.
+    if (event.target === this.input) {
+      // CHASE: not sure why this condition is here. I cannot replicate a situation where clicks
```
nadiia: I think that was to handle double click on input and toggle the dropdown. @tajo is that correct?

nadiia: Just checked the double click on this version and it seems to be broken.

chasestarr:
isolated the datepicker bugs to `scrollItemIntoView` function https://github.com/uber-web/baseui/blob/d40039869c5198095c2a752d65c393ec8201fddb/src/menu/utils.js#L36-L68

chasestarr:
added more fixes here. updated the top comment but duplicating the changes below.

- **adds innerRef prop to dropdown component** - to ensure that the dropdown element renders with the same width as the input element, I added a ref here so it can be inspected from the parent `Select` component. 
- **removes aria-activedescendant from menu** - it was being used incorrectly and threw errors in e2e tests. making a note to properly implement after this pr.
- **updates scrollItemIntoView util in menu** - this was causing 'jumping' because it would scroll the outer-most window as well as the select dropdown scroll. I noticed some pref degradation with selects over 200 items long. Will add another pr to possibly debounce the call.
- **fixes popover mounting prematurely** - when transitioning `isMounted` from `false` to `true`, it would call the popover unmount code unintentionally. this lead to a bug where page loads followed by fast clicks on the select component would choose an option in the dropdown menu.
- **makes some e2e tests more reliable** - select and datepicker tests were failing maybe 1 out of 5-6 runs. adds some more waits to ensure proper execution.

gergelyke:
this is great! if I am correct, this closes https://github.com/uber-web/baseui/issues/813 too?

chasestarr:
Yes - it should fix that

src/datepicker/__tests__/datepicker.e2e.js
```diff
@@ -165,7 +164,7 @@ describe('Datepicker', () => {
       option.click();
       return option;
     });
-    await page.waitFor(dropdown, {hidden: true});
```
nadiia: We could keep it to make sure the dropdown closes after a year is selected

src/datepicker/__tests__/datepicker.e2e.js
```diff
@@ -155,8 +155,7 @@ describe('Datepicker', () => {
     await page.waitFor(selectors.calendar);
     await page.waitFor(selectors.day);
     await page.click(selectors.yearSelect);
-    const dropdown = `${selectors.yearSelect} ${selectors.selectDropdown}`;
-    await page.waitFor(dropdown);
+    await page.waitFor(selectors.selectDropdown);
```
nadiia: I'd keep it more specific since we have a few dropdowns here

src/popover/popover.js
```diff
@@ -66,21 +67,37 @@ class Popover extends React.Component<PopoverPropsT, PopoverPrivateStateT> {
...
   ) {
+    // Handles the case where popover content changes size and creates a gap between the anchor and
+    // the popover. Popper.js only schedules updates on resize and scroll events. In the case of
+    // the Select component, when options were filtered in the dropdown menu it creates a gap
+    // between it and the input element.
+    if (this.popperRef.current) {
+      const {height} = this.popperRef.current.getBoundingClientRect();
+      if (this.popperHeight !== height) {
+        this.popperHeight = height;
+        this.popper && this.popper.scheduleUpdate();
```
nadiia: 👍

src/select/select.js
```diff
@@ -69,168 +107,126 @@ class Select extends React.Component<PropsT, SelectStateT> {
...
     if (this.dragging) return;
-    // Clear the value
     this.clearValue(event);
   };
 
-  handleClickOutside = (event: Event) => {
-    if (this.isValueJustSelected) {
-      this.isValueJustSelected = false;
+  handleClick = (event: MouseEvent | TouchEvent) => {
+    if (this.props.disabled || !isClick(event) || !isLeftClick(event)) {
```
nadiia: It should probably still be `if (this.props.disabled || (!isClick(event) && !isLeftClick(event))) {`

chasestarr:
```suggestion
    if (this.props.disabled || (!isClick(event) && !isLeftClick(event))) {
```

PR #977, opened 2019-03-03
fix(popover): render without error after ssr

### LINKED ISSUE #976: Popovers violate invariant rendering after ssr
I pulled in the latest alpha of [fusion-react](https://github.com/fusionjs/fusion-react/releases/tag/v2.0.0-alpha.2), and was testing out baseui components which are server side rendered, but the popover component was throwing invariant violations, `Invariant Violation: Portals are not currently supported by the server renderer. Render them conditionally so that they only appear on the client render.`:

image

## Current Behavior
Popovers initially ssr rendered with the following code will error

```
<Popover isOpen={true} content={...}>{...}</Popover>
```

## Expected Behavior
Component renders without error

## Current Workaround
```
function ExComp() {
  // hooks way of `componentDidMount`
  const [initialized, setInitialized] = useState(false);
  useEffect(() => {
    setInitialized(true);
  });

  return <Popover isOpen={initialized}content={...}>{...}</Popover>
}
```
which still throws warnings...
image

## Your Environment

```
❯ npm ls react-dom
└── react-dom@16.8.3
❯ npm ls baseui
└── baseui@5.14.0
```

To get the same env, run a [fusion.js](https://fusionjs.com/) app. Add baseui of course, and then either build yourself the latest alpha of [fusion-react](https://github.com/fusionjs/fusion-react/releases/tag/v2.0.0-alpha.2)  or just alias the following forked & scoped package which includes it...
```
yarn add fusion-apollo@npm:@enjoylife/fusion-apollo # to get the alpha dist of fusion-react i'm usin
```

- [x ] I have searched the [issues](https://github.com/uber-web/baseui/issues) of this repository and believe that this is not a duplicate.

enjoylife:
We have to check for mount as the call to createPortal will error with `Portals are not currently supported by the server renderer...` if rehydrating after ssr. See, https://github.com/facebook/react/issues/13097#issuecomment-405658104,  for rationale of the given changes.

Fixes #976

#### Scope

- [x ] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


chasestarr:
Thanks for the contribution. Some thoughts, interested in what you think. 

I suppose the `popover` component currently assumes that it will initially render closed. What's the use-case to having the popover open by default? It can sometimes be useful to limit certain patterns so that we maintain a consistent UX.

enjoylife:
@chasestarr well at the most abstract it would be any data driven popover with a user's data being truthy. At a more concrete level a tutorial driven with tooltips is one use case I've had to put together. `<TutorialTip open={isNewUser} />`

PR #978, opened 2019-03-04
chore(project): remove style snapshot tests

gergelyke:
Removing all style related snapshot tests, while keeping the ones we use for testing props passed in.

Going from 432 snapshot tests to 69.


documentation-site/routes.js
```diff
@@ -50,7 +50,7 @@ const routes = [
         path: '/theming/custom-themes',
       },
       {
-        text: 'Theming values',
+        text: 'w',
```
chasestarr:
```suggestion
        text: 'Theming values',
```

gergelyke: ups :D thanks Chase!

PR #979, opened 2019-03-04
fix(docs): point getting started docs to styletron

gergelyke:


gergelyke:
there was a false positive in screener, merging as is

PR #983, opened 2019-03-05
feat(tag): add solid, outlined tags, click handlers

nadiia:


documentation-site/pages/components/tag.mdx
```diff
@@ -35,20 +37,28 @@ Compact UI to display a collection of consice information.
   <Basic />
 </Example>
 
-<Example title="Tag with all available kinds" path="examples/tag/kinds.js">
+<Example title="Tags of all available kinds (closable/non-clickable)" path="examples/tag/kinds.js">
```
gergelyke: Should we be more descriptive on what are these examples? Maybe something like

`Tags of all available closable kinds, with non-clickable body`?

documentation-site/static/examples/tag/clickable-non-closeable.js
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import {Tag, KIND, VARIANT} from 'baseui/tag';
+
+const kinds = ['neutral', 'primary', 'positive', 'warning', 'negative'];
```
gergelyke: should we use `Object.keys(KIND)` instead, so we don't have to update this example is KINDs are added?

nadiia: For sure, I just didn't want to add a `custom` example here

src/tag/types.js
```diff
@@ -28,19 +29,23 @@ export type PropsT = {
...
   color?: string,
   /** Handler for events on Action button element. `children` provides which Tag was clicked. */
-  onActionClick: (
-    e: SyntheticEvent<HTMLInputElement>,
-    children?: React$Node,
-  ) => void,
+  onActionClick: (e: Event, children?: React$Node) => void,
+  /** Handler for events on Action button element. `children` provides which Tag was clicked. */
+  onActionKeyDown: (e: Event, children?: React$Node) => void,
+  /** Passing an onVlivk handler that also makes the tag clickable. */
```
chasestarr:
```suggestion
  /** Passing an onClick handler that also makes the tag clickable. */
```

chasestarr: what use-case does `onActionKeyDown` solve? Seems to duplicate most of the functionality of `onActionClick`. Similarly with `onKeyDown`? I would personally hold off until someone needs it.

nadiia: Since we pass the keyDown handler to the action element (to handle keyboard navigation) we don't want to have it overwritten by users overrides, therefore, pulling it up to the top level

PR #985, opened 2019-03-05
fix(menu): add marginBottom: 0 for list items

chrisirhc:
#### Description

Some CSS libraries/UI kits add a bottom margin to list items by default (I know it's a bad practice).
Set this up so the menu items don't look wonky when using it with those libraries, to aid users who are migrating off other UI kits.

#### Scope

- [X] Patch: Bug Fix


gergelyke:
thanks @chrisirhc !

PR #992, opened 2019-03-06
fix(breadcrumbs): key prop

gergelyke:
closes  https://github.com/uber-web/baseui/issues/969

fixes https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318


gergelyke:
merging as is, screener was a false positive

PR #999, opened 2019-03-08
docs: Make baseweb.design responsive

tajo:
**It's not perfect but there is now much better support for mobile devices**. The sidebar is collapsible (on mobile) and there are various tweaks and changes in the header and sidebar. Also I removed many borders to keep the overall design lighter.

Things to tackle next: 
- Make the sidebar top levels collapsible (with a little chevron?). 
- Algolia search is being difficult to style.

I had to do a lot of overrides. It might make sense to not use `baseui/HeaderNavigation` and implement it from scratch in a cleaner way. The sidebar is also overly complicated since it uses recursive approach, might be overkill.

## Desktop view 
screen shot 2019-03-07 at 5 47 47 pm

## Mobile view
screen shot 2019-03-07 at 5 43 07 pm

## Open Search
screen shot 2019-03-07 at 5 43 35 pm

## Open Sidebar
screen shot 2019-03-07 at 5 43 16 pm


documentation-site/components/sidebar.js
```diff
@@ -50,13 +43,31 @@ const NavigationLink = props => {
...
 const NavigationItem = props => {
-  const {route, index, level = 1} = props;
+  const {route, index, level = 1, path} = props;
   return (
     <Block
       font={levelToFont[level]}
-      paddingLeft={levelToPadding[level]}
       paddingBottom={level !== 3 ? 'scale100' : 0}
-      paddingTop={index === 0 && level === 3 ? 'scale100' : 0}
+      marginTop={index === 0 && level === 3 ? 'scale300' : 'scale300'}
```
leyanlo: Looks like this is always scale300

documentation-site/components/sidebar.js
```diff
@@ -50,13 +43,31 @@ const NavigationLink = props => {
...
+      overrides={{
+        Block: {
+          style: ({$theme}) => ({
+            paddingLeft: !route.children ? $theme.sizing.scale600 : '0px',
+            borderLeft:
+              (path && path === route.path) || (!path && route.path === '/')
```
leyanlo: Do we need the first `path` in `path && path === route.path`?

documentation-site/components/sidebar.js
```diff
@@ -50,13 +43,31 @@ const NavigationLink = props => {
...
+                ? `3px solid ${$theme.colors.primary300}`
+                : route.children
+                  ? 'none'
+                  : '3px solid transparent',
+            textTransform: route.children ? 'uppercase' : 'none',
+            ...{
```
leyanlo: I think this can be unwrapped and simplified to
```
...(route.children && level === 2 ? $theme.typography.font350 : {}),
```

documentation-site/components/sidebar.js
```diff
@@ -50,13 +43,31 @@ const NavigationLink = props => {
...
     <Block
       font={levelToFont[level]}
-      paddingLeft={levelToPadding[level]}
       paddingBottom={level !== 3 ? 'scale100' : 0}
-      paddingTop={index === 0 && level === 3 ? 'scale100' : 0}
+      marginTop={index === 0 && level === 3 ? 'scale300' : 'scale300'}
+      overrides={{
+        Block: {
+          style: ({$theme}) => ({
+            paddingLeft: !route.children ? $theme.sizing.scale600 : '0px',
```
leyanlo: Could just use a number `0`

documentation-site/components/header-navigation.js
```diff
@@ -0,0 +1,156 @@
...
+import Search from './search';
+import {version} from '../../package.json';
+
+export const HEADER_BREAKPOINT = '@media screen and (min-width: 640px)';
+
+const Hamburger = styled('div', ({$theme}) => ({
+  display: 'block',
+  userSelect: 'none',
+  paddingLeft: $theme.sizing.scale600,
+  paddingTop: '2px',
```
leyanlo:
```
paddingTop: $theme.sizing.scale0,
```

tajo:
Note: Um, the sidebar sometimes disappear when navigating (will fix).

documentation-site/components/header-navigation.js
chasestarr: I personally like using github's suggest feature for single line changes https://help.github.com/en/articles/incorporating-feedback-in-your-pull-request#applying-a-suggested-change

```suggestion
  paddingTop: $theme.sizing.scale0,
```

documentation-site/components/header-navigation.js
```diff
@@ -0,0 +1,156 @@
...
+} from 'baseui/header-navigation';
+import {styled} from 'baseui';
+import Menu from 'baseui/icon/menu';
+import Logo from '../images/base-web.svg';
+import GithubLogo from './github-logo';
+import SlackLogo from './slack-logo';
+import Search from './search';
+import {version} from '../../package.json';
+
+export const HEADER_BREAKPOINT = '@media screen and (min-width: 640px)';
```
chasestarr: Feels kind of bad to specify this outside of the theme - Does base ui not provide enough here or was this just the most straightforward approach?

documentation-site/components/layout.js
```diff
@@ -7,147 +7,58 @@ LICENSE file in the root directory of this source tree.
...
-          </Link>
-        </Block>
-      </NavigationList>
-    </HeaderNavigation>
+const SidebarWrapper = styled('div', ({$theme, $isOpen}) => ({
+  display: $isOpen ? 'block' : 'none',
+  paddingTop: $theme.sizing.scale800,
+  marginLeft: $theme.sizing.scale1000,
+  marginRight: $theme.sizing.scale800,
+  '@media screen and (min-width: 820px)': {
```
chasestarr: same theming question as above, should this be integrated with the global breakpoints?

documentation-site/components/header-navigation.js
```diff
@@ -0,0 +1,156 @@
...
+                  color: 'inherit',
+                  textDecoration: 'none',
+                },
+              },
+            }}
+            target="_blank"
+            as="a"
+            href="https://github.com/uber-web/baseui/releases"
+          >
+            <Tag closeable={false} color="#333" kind="custom">
```
gergelyke: the Tag itself does not have `cursor:pointer`, should we add it?

nadiia: A noop function passed to the Tag should work.

documentation-site/components/header-navigation.js
tajo: The closest breakpoint was `medium: 600px` and that's too narrow for my use-case. I could override the theme and add more breakpoints but I don't see a problem to add arbitrary number of breakpoints as long as they all use mobile-first min-width approach. 

One downside is the array-like API that we have in `Block` - it makes sense to have a small set of breakpoints there but if those don't cut it for you, it should be safe to use custom?

tajo: Or if we want to enforce a limited set of breakpoints we would probably need like 6-8 of them. Three breakpoints are probably not enough to fit all UIs.

chasestarr: Yeah i understand. I wouldn't want to be burdened by the need to specify all values in the block array. This is fine with me, could be something we can point to for how to approach responsive

gergelyke:
one super minor thing - the slack and github icons seem to be vertically un-aligned with the search box
Screen Shot 2019-03-08 at 12 37 54 PM

documentation-site/components/header-navigation.js
leyanlo: Ohh didn't realize that icon that did that. Amazing

documentation-site/pages/index.js
```diff
@@ -28,26 +28,25 @@ type Contributor = {
 const cardOverrides = {
   Root: {
     style: ({$theme}) => ({
-      marginRight: $theme.sizing.scale500,
-      marginBottom: $theme.sizing.scale500,
-      width: '340px',
+      marginLeft: '16px',
```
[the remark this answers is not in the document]
leyanlo:
```suggestion
      marginLeft: $theme.sizing.scale600,
```

documentation-site/components/sidebar.js
leyanlo:
```suggestion
      marginTop="scale300"
```

documentation-site/components/sidebar.js
leyanlo:
```suggestion
              (path === route.path) || (!path && route.path === '/')
```

tajo: This is actually needed because `route.path` can be undefined

PR #1001, opened 2019-03-11
docs(typography): link to theming values

gergelyke:


chasestarr:
I believe https://github.com/uber-web/baseui/pull/975 should fix the flakey e2e tests

PR #1012, opened 2019-03-12
docs: add notes about longhand vs shorthand props

tajo:
There was some confusion about shorthand vs longhand CSS properties so I've added some description into the Understanding Overrides page and overrides inspector.

### DIFF
--- a/documentation-site/components/overrides.js
+++ b/documentation-site/components/overrides.js
@@ -84,6 +84,14 @@ class Overrides extends React.Component {
             },
           })}
         </Block>
+        <Block as="p" font="font400" marginTop="scale900">
+          <b>Note:</b> You should always use longhand CSS properties. Mixing
+          shorthands and longhands will lead into{' '}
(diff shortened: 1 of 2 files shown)


documentation-site/components/overrides.js
```diff
@@ -84,6 +84,14 @@ class Overrides extends React.Component {
             },
           })}
         </Block>
+        <Block as="p" font="font400" marginTop="scale900">
```
gergelyke: should we use the warning inline notifications instead?

tajo: You mean the `<Blockquote>` or the baseui notification component? I don't think it should be that super visible since it's not even 100% related to this specific location - it's more bold and detailed in the Overrides page.

gergelyke: yeah, I meant the base notification component - but yeah, your point makes perfect sense, as always, I was just being the devil advocate 😈

PR #1015, opened 2019-03-13
feat(docs): add dark theme switcher

gergelyke:
This PR just enables the development of the Dark Theme - the look and feel will be off for the time being!

If you'd like to try it out, just visit any url using the `theme=dark` query string - once you want to change back to the light theme, use `theme=light`

Screen Shot 2019-03-13 at 1 47 03 PM


tajo:
I wonder why the Header stayed white since `baseui/header-navigation` gets darkened

gergelyke:
Oh yeah, it was not the HeaderNavigation, but the div behind it :) just fixed it

PR #1016, opened 2019-03-13
feat(menu): enables keyboard navigation between nested menus, provides proper focus feedback

chasestarr:
#### Description

Screen Shot 2019-03-13 at 3 05 44 PM

- **Ensures targeted focus coloring** - Only one menu may be 'active' at any time. If the menu is active, highlighted menu item is blue. If inactive, highlighted menu item is black.
- **Fixes doubled item highlight** - Bug is described in [this issue](https://github.com/uber-web/baseui/issues/993). Mouse hover now triggers a state update, and this change removes highlight from css `:hover` pseudo-selector. Keyboard navigation between items will overwrite a mouse highlight and vice-versa. Always prefers the most recent action.
- **Fixes keyboard navigation with multiple menus** - Only initializes keyboard navigation on menu focus. Previously if multiple menus were on the page, they would all change the highlighted item.
- **Enables keyboard navigation between nested menus** - Adds a new `NestedMenu` component to wrap nested menus to track refs for each. Left arrow navigates to parent menu, right arrow navigates to child menu. This works through context.
- **Adds more contextual child menu example** - it's now a 'file' menu, rather than a series of numbers and also shows how only a couple items create a child menu.

#### Scope

- [x] Minor: New Feature
- [x] Manually tested `Select` and `Datepicker` components for stability.
- [x] unit/e2e tests
- [x] update the disabled item colors

### DIFF
--- a/src/menu/menu.js
+++ b/src/menu/menu.js
@@ -59,8 +59,8 @@ export default function Menu(props: StatelessMenuPropsT) {
             $ref={ref}
             $isFocused={isFocused}
             $isHighlighted={isHighlighted}
-            {...optionProps}
             {...rest}
+            {...optionProps}
           />
         );
       })}
--- a/src/menu/menu.js
+++ b/src/menu/menu.js
@@ -30,11 +30,11 @@ export default function Menu(props: StatelessMenuPropsT) {
     <List
       role="listbox"
       $ref={rootRef}
-      onMouseEnter={props.focusMenu}
-      onMouseOver={props.focusMenu}
-      onMouseLeave={props.unfocusMenu}
-      onFocus={props.focusMenu}
-      onBlur={props.unfocusMenu}
+      onMouseEnter={focusMenu}
+      onMouseOver={focusMenu}
+      onMouseLeave={unfocusMenu}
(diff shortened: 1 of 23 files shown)


nadiia:
Two minor issues from the example in docs:
- The tab order seems to be odd when tabbing from the submenu.
- The submenu stays open when tabbed outside of the menu.

src/menu/menu.js
```diff
@@ -9,32 +9,58 @@ import * as React from 'react';
...
-            tabIndex={index === 0 ? 0 : -1} // Allows tab focus into first element
-            {...getSharedProps(requiredProps)}
+            resetMenu={resetMenu}
+            role="option"
+            $disabled={disabled}
+            $ref={ref}
+            $isFocused={isFocused}
+            $isHighlighted={isHighlighted}
             {...optionProps}
+            {...rest}
```
nadiia: `{...optionProps}` should go last

chasestarr:
updated so that tabbing is trapped when focused on child menus. restricts keyboard navigation to arrows

tajo:
Looks and works great!

PR #1019, opened 2019-03-14
feat(sidenav): initial implementation of the side navigation component

nadiia:


src/side-navigation/nav-item.js
```diff
@@ -0,0 +1,120 @@
...
+        <ItemContainer
+          key={`${index}-level${level}-${item.title}`}
+          {...sharedProps}
+          {...itemContainerProps}
+        >
+          {typeof renderItem === 'function' ? (
+            renderItem(item)
+          ) : (
+            <>
+              {item.path ? (
```
gergelyke: should we decouple the routing logic from this component, and just expose a change handler? maybe something like this https://github.com/trendmicro-frontend/react-sidenav#react-router-v4-with-react-v16?

nadiia: There is no much routing logic, we may remove the default link usage in the markup tho, what do you think?

gergelyke: in the current implementation, if one does not provide a path, the elements won't be rendered as clickable elements, right?

yeah, probably that would help

what do you think about that approach I linked? maybe just adding an `id` to the route elements, and let consumers handle navigation in the change handlers?

src/side-navigation/types.js
```diff
@@ -0,0 +1,78 @@
...
+  isActive: ?(item: *, activePath: string) => mixed,
+  /** Overrides for the internal elements and components */
+  overrides: SideNavComponentsT,
+  /** Optional render function that is called instead default item rendering */
+  renderItem: ?(item: *) => React.Node,
+};
+
+export type SideNavPropsT = NavPropsT & {
+  /** List of navigation items */
+  items: Array<*>,
```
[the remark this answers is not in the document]
nadiia: I didn't want to restrict it to only those three props in case they want to have some extended data in it. But can definitely change it to have a more specific type.

src/side-navigation/types.js
```diff
@@ -0,0 +1,78 @@
...
+  Link?: OverrideT<*>,
+  NavItem?: OverrideT<*>,
+  SubNavContainer?: OverrideT<*>,
+};
+
+export type NavPropsT = {
+  /** Defines the current active path. Used for the defaut calculation of the $active prop */
+  activePath: string,
+  /** isActive is called on the nav item. If return true the item will be rendered as an active one */
+  isActive: ?(item: *, activePath: string) => mixed,
```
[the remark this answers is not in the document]
nadiia: Sure, any ideas on naming?

src/side-navigation/nav-item.js
```diff
@@ -0,0 +1,120 @@
...
+      if (typeof onSelect === 'function') {
+        onSelect({item});
+      }
+    }
+    this.props.onKeyDown && this.props.onKeyDown(e);
+  };
+
+  render() {
+    const {
+      activePath,
```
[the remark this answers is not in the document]
nadiia: I see your point, but due to the nesting items rendered within nav-item it needs it to do the check for nested items.

src/side-navigation/types.js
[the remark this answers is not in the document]
chasestarr: `activePredicate`?

src/side-navigation/nav-item.js
nadiia: I think the clickable part we should still handle internally. The `path` prop currently serves as an identifier, but maybe the naming is not perfect. Don't want to use `id` here as it's a valid attr. 
I'll change the link to be just a regular clickable element and let the customers handle it through the `onChange` only.

src/side-navigation/types.js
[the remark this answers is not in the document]
nadiia: sounds great! will change it

src/side-navigation/nav-item.js
gergelyke: oh yeah, that makes sense - maybe instead of the path, we should add a boolean prop like `navigatable` or something like this?

nadiia: How about `linkId` or `itemId` name?

gergelyke: either sounds great to me

src/side-navigation/styled-components.js
```diff
@@ -0,0 +1,75 @@
...
+          colors.primary
+        }`
+      : 'transparent',
+    boxSizing: 'border-box',
+    borderLeftWidth: '4px',
+    borderLeftStyle: 'solid',
+    borderLeftColor: $active ? colors.primary : 'transparent',
+    color: $active ? colors.primary : null,
+    cursor: $selectable ? 'pointer' : 'default',
+    // outline: 'none',
```
chasestarr:
```suggestion
    outline: 'none',
```

not sure if you wanted the outline or not

tajo:
The navigation relies 100% on `onClick`, right? It has some downsides:

- crawlers can see those links
- users can't see where the link goes to by hovering over it

For SPA, `onClick={() => history.push()}` is fine but what if I want to also add some external links?

```
 <a href="https://www.example.com" target="_blank">Example</a>
```

It would be great to replicate (allow) the functionality that Next.js has:

- all links are rendered as `<a href="...">`, not divs or buttons
- JS intercepts the default navigation action if it's internal link and does history.push instead of a full reload

nadiia:
@tajo I added an example of sidenav items with a regular <a> tag usage. Is that what you were thinking  of?

documentation-site/static/examples/side-navigation/basic.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import {Navigation} from 'baseui/side-navigation';
+import nav from './data.js';
```
gergelyke: should we include this in this file instead of referencing this? the reader of the documentation site would have no way to open this from the site

also, currently all examples are self-contained on the doc site, so I'd add it here to be consistent too

tajo: Agreed. It's nice to have examples self-contained even if it means some copy&pasting.

tajo:
> @tajo I added an example of sidenav items with a regular `<a>` tag usage. Is that what you were thinking of?

Yea, the `renderItem` API is  nice! Tiny thing: If you select one of those `#` links and refresh, it doesn't set the correct `activeItemId` but that should be fixable just by updating the example code. 

Is there some reason why not to ship this as a default `NavItem`?

```jsx
const renderItem = function(item, itemProps) {
  const {onClick, onKeyDown, tabIndex, role, ...rest} = itemProps;
  return (
    <StyledLink href={item.itemId} onClick={onClick} onKeyDown={onKeyDown}>
      <StyledNavItem {...rest}>
        {item.title}
        {item.itemId ? ` (${item.itemId})` : ''}
      </StyledNavItem>
    </StyledLink>
  );
};
```

I almost can't see a use-case where I wouldn't want to use `<a>` by default since it's a navigation component?

nadiia:
> > @tajo I added an example of sidenav items with a regular `<a>` tag usage. Is that what you were thinking of?
> 
> Yea, the `renderItem` API is nice! Tiny thing: If you select one of those `#` links and refresh, it doesn't set the correct `activeItemId` but that should be fixable just by updating the example code.
> 
> Is there some reason why not to ship this as a default `NavItem`?
> 
> ```js-jsx
> const renderItem = function(item, itemProps) {
>   const {onClick, onKeyDown, tabIndex, role, ...rest} = itemProps;
>   return (
>     <StyledLink href={item.itemId} onClick={onClick} onKeyDown={onKeyDown}>
>       <StyledNavItem {...rest}>
>         {item.title}
>         {item.itemId ? ` (${item.itemId})` : ''}
>       </StyledNavItem>
>     </StyledLink>
>   );
> };
> ```
> 
> I almost can't see a use-case where I wouldn't want to use `<a>` by default since it's a navigation component?

See the comment from Gergely above with an [example](https://github.com/trendmicro-frontend/react-sidenav#react-router-v4-with-react-v16). 
Shipping the NavItem that renders as a link probably would have more out of the box use cases. But in most cases, the link component would still need to be replaced with some router link component.
@gergelyke @tajo do you have a strong opinion on one or the other approach?

gergelyke:
I don't - whatever the two of you think is the best 👍

tajo:
So the react-router integration could look like:

```jsx
withRouter(<Navigation
  items={nav}
  activeItemId={activePath}
  onChange={({e, item}) => {
    e.preventDefault();
    this.props.history.push(`/${item.path}`);
    this.setState({ activePath: item.id })
  }}
/>)

Next.js has `Router.push`.

But yea, if you wanted to use the `Link` component that react-router / next.js are exporting, then you would always need to implement the renderItem.

One benefit of the default `<a>` is that external links would just work out of the box. If you want to mix both external and internal links, then you have to implement `onClick` logic that decides when to call `e.preventDefault`.

PR #1032, opened 2019-03-15
feat(datepicker): adds support to error

jcemer:
#### Description

It adds support of error attribute to Datepicker passing it down to the Input component.

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change


src/datepicker/datepicker.js
```diff
@@ -184,6 +185,7 @@ export default class Datepicker extends React.Component<
                 aria-label={
                   this.props['aria-label'] || locale.datepicker.ariaLabel
                 }
+                error={this.props.error}
```
chasestarr: you should be able to pass `error` to the `Input` override on datepicker. Enabled by [L137](https://github.com/uber-web/baseui/pull/1032/files#diff-6240fc17e068eaeef7095937a1d63eaeL137) and [L195](https://github.com/uber-web/baseui/pull/1032/files#diff-6240fc17e068eaeef7095937a1d63eaeL195)

chasestarr: We may want to consider enabling this pattern for https://github.com/uber-web/baseui/pull/1031 - what do you think?

jcemer: I'm not sure because `error` looks similar to `required` and `placeholder` that are props of the datepicker.

chasestarr: ok - fine with me for now. may be something we want to take on in the future

PR #1034, opened 2019-03-15
feat: add data-baseweb attribute

tajo:
This adds a new `data-baseweb="component"` attribute to all component roots (there are some exceptions). The idea is that we can later use this attribute to easily detect/scrape/test/screenshot our components across projects.

### DIFF
--- a/src/input/base-input.js
+++ b/src/input/base-input.js
@@ -132,7 +132,11 @@ class BaseInput<T: EventTarget> extends React.Component<
     const [Before, beforeProps] = getOverrides(BeforeOverride, NullComponent);
     const [After, afterProps] = getOverrides(AfterOverride, NullComponent);
     return (
-      <InputContainer {...sharedProps} {...inputContainerProps}>
+      <InputContainer
+        data-baseweb={this.props['data-baseweb'] || 'base-input'}
+        {...sharedProps}
+        {...inputContainerProps}
+      >
         <Before {...sharedProps} {...beforeProps} />
         <Input {...sharedProps} {...this.getInputProps()} {...inputProps}>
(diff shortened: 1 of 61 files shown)


src/input/base-input.js
```diff
@@ -132,7 +132,11 @@ class BaseInput<T: EventTarget> extends React.Component<
     const [Before, beforeProps] = getOverrides(BeforeOverride, NullComponent);
     const [After, afterProps] = getOverrides(AfterOverride, NullComponent);
     return (
-      <InputContainer {...sharedProps} {...inputContainerProps}>
+      <InputContainer
+        data-baseweb={this.props['data-baseweb'] || 'base-input'}
```
nadiia: How about we don't add it to the API?

src/popover/popover.js
```diff
@@ -473,6 +473,7 @@ class Popover extends React.Component<PopoverPropsT, PopoverPrivateStateT> {
       <Body
         key="popover-body"
         $ref={this.popperRef}
+        data-baseweb={this.props['data-baseweb'] || 'popover'}
```
nadiia: same here

src/toast/toast.js
```diff
@@ -161,6 +161,7 @@ class Toast extends React.Component<ToastPropsT, ToastPrivateStateT> {
           <Body
             tabIndex={0}
             role="alert"
+            data-baseweb={this.props['data-baseweb'] || 'toast'}
```
nadiia: Also here, why do we want to get this value from props?

nadiia: Oh I see, these all are added for the reused internally components, correct? So we don't expose it still...

tajo: Yep. For example, `notification -> toast`, `textarea->input` ...

PR #1038, opened 2019-03-18
fix(spinner): size and color props

gergelyke:


src/spinner/spinner.js
```diff
@@ -56,6 +56,7 @@ class Spinner extends React.Component<SpinnerPropsT> {
           clipRule="evenodd"
           d="M16.9497 7.05015C14.2161 4.31648 9.78392 4.31648 7.05025 7.05015C6.65973 7.44067 6.02656 7.44067 5.63604 7.05015C5.24551 6.65962 5.24551 6.02646 5.63604 5.63593C9.15076 2.12121 14.8492 2.12121 18.364 5.63593C18.7545 6.02646 18.7545 6.65962 18.364 7.05015C17.9734 7.44067 17.3403 7.44067 16.9497 7.05015Z"
           {...activePathProps}
+          $color={this.props.color}
```
nadiia: let's pass it before the prop overrides

src/icon/styled-components.js
```diff
@@ -14,8 +14,6 @@ export function getSvgStyles({$theme, $size, $color}: StyledComponentParamsT) {
       $size = $theme.sizing[$size];
     } else if (typeof $size === 'number') {
       $size = `${$size}px`;
-    } else {
-      $size = $theme.sizing.scale600;
```
nadiia: The custom size string still won't work I think.

src/spinner/spinner.js
nadiia: And pass the `$color` prop to `Track` element also?

src/icon/styled-components.js
gergelyke: discussed this offline

PR #1040, opened 2019-03-19
fix(types): change handler returns mixed instead of any

gergelyke:


gergelyke:
closes https://github.com/uber-web/baseui/issues/626

gergelyke:
depends on https://github.com/tajo/react-range/pull/4

PR #1041, opened 2019-03-19
docs(icon-provider): add proposal

gergelyke:


tajo:
The concept is good! I assume, to implement this we will introduce some general `<ContextedIcon />` component that accesses the context and fallbacks to the default set of icons and we will use it in all of our components?

One thing that might be worth to document is that the icon component should have some standardized API so it can be swapped 1:1, props like `size`, `color` and `title`.  Are those accepted by `react-icons/fa`?

gergelyke:
Yeah, something like that, I haven't thought about the implementation for now, just the high-level API. Do you think it's enough if we link to the icon component, and we say that the icon component passed in must implement that API?

tajo:
> Yeah, something like that, I haven't thought about the implementation for now, just the high-level API. Do you think it's enough if we link to the icon component, and we say that the icon component passed in must implement that API?

I think it should be enough to just document those 3 props and encourage developers to use them. They don't have to but then their icons can end up with wrong sizes / colors.

chasestarr:
What was the thought given to new `icon-provider` vs embedding more information into the theme? Conceptually, how do we see this different from theming an application (especially since the provider is being imported from the top level like `theme-provider`)? Do we expect anyone to use one and not the other? What benefit is gained from doing so?

gergelyke:
Closing this in favor of extending the theme

PR #1048, opened 2019-03-20
feat(theme): add icons to themes

gergelyke:
The purpose of this PR is to enable the global override of icons across components.

Please note, that the icon files are generated, and the changes are coming from `src/icon/icon-template.txt`


src/icon/icon-exports.js
```diff
@@ -1,30 +1,24 @@
-/*
-Copyright (c) 2018 Uber Technologies, Inc.
-
-This source code is licensed under the MIT license found in the
-LICENSE file in the root directory of this source tree.
-*/
```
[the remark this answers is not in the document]
gergelyke: oh yeah, it seems like we've manually edited this file - let me update the generator script

documentation-site/routes.js
```diff
@@ -58,9 +58,13 @@ const routes = [
         path: '/theming/theming-values',
       },
       {
-        text: 'Overrides',
+        text: 'Overriding components',
```
tajo: This makes the whole sidebar wider. Also, overrides can override `styles` / `props` only?

tajo:
Why don't we just add those icons to the default theme (`creator.js`) instead of having conditions in every `icon/*` file? I think it would be overall more transparent, for example, developers might `console.log(theme)` and they would see that icons is something that can be themed.

documentation-site/pages/theming/overriding-icons.mdx
```diff
@@ -0,0 +1,35 @@
...
+
+With Base Web, you can use your custom icon set. To do that, you have to provide your icons in the theme.
+
+You can use any React components as Icon replacements, as long as they accept the following properties:
+
+* `size`, to set their width and height,
+* `color`, to set their colors,
+* `title`, to set the title of the icon for accessibility purposes.
+
+You can find more about the [Icon API here](/components/icon/#icon-api).
```
chasestarr:
```suggestion
You can find more about the [Icon API here](/components/icon/#icon-api) and also refer to this [list of available icons](https://baseweb.design/components/icon/#available-icons).
```

PR #1050, opened 2019-03-20
fix(tabs): enables more nested override patterns

chasestarr:
#### Description

map children was previously not cloning the element, therefore losing some nested props. adds a warning to migrate from `TabPanel` component to `Tab`. 

#### Scope

- [x] Patch: Bug Fix


src/tabs/tabs.js
```diff
@@ -34,18 +33,30 @@ export default class Tabs extends React.Component<TabsPropsT> {
...
         active: key === activeKey,
         disabled: disabled || child.props.disabled,
         $orientation: orientation,
-        overrides: child.props.overrides,
         onSelect: () => this.onChange({activeKey: key}),
-      };
-      return <Tab {...props}>{child.props.title}</Tab>;
+      });
+
+      // const key = child.key || String(index);
```
gergelyke: should we remove this?

src/tabs/tabs.js
```diff
@@ -34,18 +33,30 @@ export default class Tabs extends React.Component<TabsPropsT> {
     // eslint-disable-next-line flowtype/no-weak-types
     const tabs = React.Children.map(children, (child: any, index) => {
       if (!child) return;
+
       const key = child.key || String(index);
-      const props = {
-        key,
+      return React.cloneElement(child, {
+        key: key,
```
gergelyke: can we keep this simply `key,`?

src/tabs/tabs.js
chasestarr: fixed thank you

src/tabs/tabs.js
chasestarr:
```suggestion
        key,
```

PR #1051, opened 2019-03-20
chore: run documentation link checker in workers queue

KevinGrandon:
#### Description

The workers queue has better autoscaling and generally more agents. This should result in faster builds.

#### Scope

Refactor only, no release necessary.


KevinGrandon:
Going to look at this at a later date actually once CI is healthy.

PR #1052, opened 2019-03-20
fix(styles): import react for types

kahwee:
#### Description

This only affects flow type. Need to import `* as React` in order for flow to find `React.Node`

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


gergelyke:
thanks @kahwee, I've totally missed that!

kahwee:
Np! I noticed I couldn't pass flow in one of my projects that uses baseui. :)

Thanks.

kahwee:
Will any Flow experts be kind enough to chime in? 

I changed `React.Node` to `typeof React.Component` to try to address the Flow type complaints.

It does seem to work.

src/styles/types.js
```diff
@@ -5,6 +5,7 @@ This source code is licensed under the MIT license found in the
 LICENSE file in the root directory of this source tree.
 */
 // @flow
+import * as React from 'react';
```
AlexMSmithCA: Good catch!  As per the [Flow React documentation](https://flow.org/en/docs/react/types/), React types must be imported in files where they are used.

`import * as React from 'react'` is shorthand for including access to all types.

Without this, the `React.Node` usages below would resolve to `any` which potentially results in a lot of uncovered code.

kahwee: Thanks! What do you think of the change from `Overflow?: React.Node,` to `Overflow?: typeof React.Component,` I can't seem to find any guides that suggest this is a good way. It just happened to work for me.

src/styles/types.js
```diff
@@ -317,27 +318,27 @@ export type ThemeT = {|
...
-  Overflow?: React.Node,
-  Plus?: React.Node,
-  Search?: React.Node,
-  Spinner?: React.Node,
-  TriangleDown?: React.Node,
-  TriangleLeft?: React.Node,
-  TriangleRight?: React.Node,
-  TriangleUp?: React.Node,
-  Upload?: React.Node,
+  Alert?: typeof React.Component,
```
AlexMSmithCA: Are there concrete implementations for each of these?

If we can, we should avoid `typeof React.Component' here and instead use the concrete types for each of these if they exist.  This is particularly useful if we want to constrain the prop types.

src/styles/types.js
AlexMSmithCA: Ah, added a comment below before I saw this.  I don't have a lot of context on this code, but from a quick glance it appears that there are corresponding concrete implementations for of these icons.

For example, `ArrowDown`:
https://github.com/uber-web/baseui/blob/73fc65624f2cbf067bd731c4e5af9faf1eeacc4f/src/icon/arrow-down.js#L15-L37

My assumption would be that the props passed into `ArrowDown` should always match those passed into `theme.icons.ArrowDown`.  If that's the case, we could use the same `IconPropsT` type.  Something like:

```
import type {IconPropsT} from '<some_path_to_type>';
...
export type IconT = {
   ...
   ArrowDown?: React.ComponentType<IconPropsT>,
   ...
};
```

chasestarr:
Thanks Alex! Tested this change locally and found `React.ComponentType<IconPropsT>` to work well. Where `IconPropsT` is imported as `import type {IconPropsT} from '../icon/types.js';`

rtsao:
Ah, didn't realize there already was a PR

PR #1059, opened 2019-03-21
fix(components): remove z-index usage

nadiia:
#### Description

Remove z-index usage in the modal, pagination, popover and table components

#### Scope

- [ ] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


gergelyke:
should we add a screener.io scenario with an open select dropdown in a modal? or something similar, to have a visual test case for it?

tajo:
Can we also remove it from `create.js` and `styles/types.js`

nadiia:
> Can we also remove it from `create.js` and `styles/types.js`

It will be a breaking change. We can remove it on the next major bump.

PR #1060, opened 2019-03-21
docs: add proposal for component heading

tajo:
[rendered](https://github.com/uber-web/baseui/blob/2a1a98ec10802f7e2014cf906dcb956c04fc08c2/proposals/heading-component.md)


proposals/heading-component.md
```diff
@@ -0,0 +1,73 @@
...
+
+There is proposal for an `<h>` element and [document outline algorithm](http://adrianroselli.com/2016/08/there-is-no-document-outline-algorithm.html) to solve this problem once for all but there is no browser support yet. Fortunately, we [can use React Context](https://medium.com/@Heydon/managing-heading-levels-in-design-systems-18be9a746fa3) to implement our own document outline algorithm.
+
+## Usage
+
+```jsx
+import * as React from 'react';
+import {Heading} from 'baseui/heading';
+export default () => (
+  <Heading heading="Base Web">
```
chasestarr: I find the api for the `Heading` component confusing. It's hard to imagine what the resulting markup looks like without viewing an example like what's provided below. While not knowing the markup is common when working with react components, I'm concerned when it has a very similar mapping to html `<Hn>` elements.

I would prefer to see a similar API to http://adrianroselli.com/2016/08/there-is-no-document-outline-algorithm.html where there are two components. One for `Headings` and another for `Sections`. This way, each component is semantically consistent with its related html element

tajo: Yea, I've actually almost submitted this proposal with `Section` and `Heading` components since that's what [this article](https://medium.com/@Heydon/managing-heading-levels-in-design-systems-18be9a746fa3) also proposes. 

However, I was not convinced it's not even more confusing because the `Section` component doesn't render the `<section>` element (and it probably should not). It **only manages the context value/leveling** and then it becomes somewhat unnecessary because you can do the same thing with a single component.

But I agree that it might be also surprising to see `<Heading heading="foo">Related text</Heading>`.

Also should you be able to do both:

```jsx
<Heading>
<Section>
  <Heading>
</Section>
```

and

```jsx
<Section>
  <Heading>
  <Section>
    <Heading>
  </Section>
</Section>
```

If so, should the ouput be the same?

chasestarr: Good point about not rendering a `section` element. Maybe an alternative name is more appropriate? `HeadingLevel`? 

My thought is that `Section` would be a context provider. Could `Heading` throw if no context is found? That way, only the second option would be technically possible.

schnerd:
Glancing at the API initially I sort of assumed that `<Section>` would render an [html section element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section), people reading code that uses this component might assume something similar?

Additionally, if I actually want to use hierarchical/semantic markup then I might end up with code like the following:

```js
<Heading>1</Heading>
<Section>
  <section>
    <Heading>1a</Heading>
  </section>
</Section>
```

While it's a mouthful, maybe something like `<WithNextHeadingLevel>` would be avoid the "section" ambiguity and more explicitly describes what's going on?

tajo:
@schnerd Yea, I am not sold on it either. Imo, it doesn't make sense to render the `<section>` element since it doesn't bring any value to users today. Maybe once there is some official outline algorithm supported by screen readers but then we could just remove this whole component.

**How about just renaming `<Section>` to `<HeadingLevel>`?** Additional `WithNext` doesn't really make it more descriptive and `with` evokes HoCs. Maybe `<HeadingLevelProvider>` would also make sense but `<HeadingLevel>` might be a good compromise?

tajo:
Cool, I'm going forward with `<HeadingLevel />`

PR #1063, opened 2019-03-21
feat(control-components): audit dark theme colors

chasestarr:
#### Description

Aligns these components closer to dark theme colors in figma

#### Scope

- [x] Minor: New Feature


documentation-site/static/examples/radio/disabled.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import {Radio, RadioGroup} from 'baseui/radio';
+
+export default function Disabled() {
+  return (
+    <RadioGroup name="radio group" value="1">
+      <Radio disabled value="1">
```
nadiia: adding `disabled` prop to the `RadioGroup` should make all the radios disabled

src/checkbox/styled-components.js
```diff
@@ -226,12 +226,20 @@ export const Toggle = styled('div', props => ({
...
+      return props.$theme.colors.sliderTrackFillDisabled;
+    }
+
+    if (props.$isActive && props.$checked) {
+      return props.$theme.colors.sliderTrackFillSelectedActive;
+    }
+
+    if (props.$isHovered && props.$checked) {
+      return props.$theme.colors.sliderTrackFillSelectedHover;
+    }
```
nadiia: Should there be another condition for just `props.$checked` case (when not active or hovered)?

src/slider/styled-components.js
```diff
@@ -38,18 +38,15 @@ Track.displayName = 'StyledTrack';
...
   return {
     borderRadius: $theme.borders.useRoundedCorners ? borders.radius100 : '0px',
     background: getTrackBackground({
       values: $value,
       colors:
         $value.length === 1
-          ? [$disabled ? colors.mono600 : colors.primary, colors.mono400]
-          : [
-              colors.mono400,
-              $disabled ? colors.mono600 : colors.primary,
```
nadiia: Don't we want to keep color change in a disabled state?

src/checkbox/styled-components.js
chasestarr: updated the colors on this element. checked/default have the same color, so no need to create another if statement

src/slider/styled-components.js
chasestarr: based on figma designs, disabled is indicated by drag handle color

documentation-site/static/examples/radio/disabled.js
chasestarr: thanks!

chasestarr:
merged, but can make any updates in a follow-up pr

PR #1067, opened 2019-03-22
[WIP] Fix Flow types

rtsao:


chasestarr:
Thanks Ryan. Is `ElementType` more appropriate than `ComponentType<T>`? Take a look at the discussion here: https://github.com/uber-web/baseui/pull/1052#issuecomment-475453101

When looking at the flow docs for [elementtype](https://flow.org/en/docs/react/types/#toc-react-elementtype), it's unclear to me what it means by `jsx intrinsics (strings)`. Does that mean it would accept something like `<svg>...</svg>`?

rtsao:
Ah, didn't realize there already was a PR

PR #1072, opened 2019-03-22
fix(toast): tabindex on close

gergelyke:
closes https://github.com/uber-web/baseui/issues/1066


chasestarr:
tabbing to close button, then pressing 'spacebar' or 'enter' does not trigger close. I imagine it being focus-able is not all that's intended in the issue? https://deploy-preview-1072--baseui.netlify.com/components/toast/

gergelyke:
ah, good catch, thanks Chase - let me look into this further

PR #1073, opened 2019-03-22
fix(toast): use flex

gergelyke:
closes https://github.com/uber-web/baseui/issues/1065

### DIFF
--- a/src/toast/toast.js
+++ b/src/toast/toast.js
@@ -169,6 +169,9 @@ class Toast extends React.Component<ToastPropsT, ToastPrivateStateT> {
             onMouseEnter={this.onMouseEnter}
             onMouseLeave={this.onMouseLeave}
           >
+            {typeof children === 'function'
+              ? children({dismiss: this.dismiss})
+              : children}
             {closeable ? (
               <DeleteAltIcon
                 role="button"
@@ -185,9 +188,6 @@ class Toast extends React.Component<ToastPropsT, ToastPrivateStateT> {
                 overrides={closeIconOverrides}
               />
             ) : null}
-            {typeof children === 'function'
-              ? children({dismiss: this.dismiss})
-              : children}
           </Body>
         )}
       </LocaleContext.Consumer>
--- a/src/toast/__tests__/toast.scenario.js
+++ b/src/toast/__tests__/toast.scenario.js
@@ -19,7 +19,7 @@ export const component = () => (
     <Toast>
       {({dismiss}) => {
         return (
-          <React.Fragment>
+          <div>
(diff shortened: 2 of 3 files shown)


chasestarr:
we've generally avoided grid so far, unless opted-in by developers since it doesn't have full support from older browsers. could this be refactored to use flex?

gergelyke:
sure, I'll do that

gergelyke:
@chasestarr updated, what do you think?

src/toast/toast.js
```diff
@@ -185,9 +186,13 @@ class Toast extends React.Component<ToastPropsT, ToastPrivateStateT> {
                 overrides={closeIconOverrides}
               />
             ) : null}
-            {typeof children === 'function'
-              ? children({dismiss: this.dismiss})
-              : children}
+            {typeof children === 'function' ? (
+              children({dismiss: this.dismiss})
+            ) : (
+              <Block display="flex" flex="1">
```
nadiia: Could we avoid using this additional wrapper? It's not in overrides so created an element that can't be customized.

chasestarr: yeah, probably just add 'flex' to `Body` with `justifyItems: 'space-between'` not totally sure what's going on with the markup ordering in this component. I imagine re-order close icon and children would avoid the `order` css property

chasestarr:
check out this sandbox with a way to position without adding a new element https://codesandbox.io/s/382vlqyyqq you'll need to change the ordering of the close icon and children in the component file so that it renders in the correct order without css

nadiia:
> check out this sandbox with a way to position without adding a new element https://codesandbox.io/s/382vlqyyqq you'll need to change the ordering of the close icon and children in the component file so that it renders in the correct order without css

Right, rendering it after the content makes sense. Let's move it @gergelyke. The wrapper element may still be useful if there are more than one children.

gergelyke:
@chasestarr @nadiia updated, what do you think?

src/toast/toast.js
```diff
@@ -169,6 +176,13 @@ class Toast extends React.Component<ToastPropsT, ToastPrivateStateT> {
             onMouseEnter={this.onMouseEnter}
             onMouseLeave={this.onMouseLeave}
           >
+            {typeof children === 'function' ? (
+              children({dismiss: this.dismiss})
+            ) : (
+              <ToastContent {...sharedProps} {...toastContentProps}>
```
nadiia: Does it make sense to wrap the children in both cases? `function` | React.Children

chasestarr:
my issue is with the new `ToastContent` override. can't we get the proper outcome without it?

PR #1076, opened 2019-03-22
feat(radio): adds description

chasestarr:
#### Description

Screen Shot 2019-03-22 at 3 26 17 PM

#### Scope

- [x] Minor: New Feature


src/radio/types.js
```diff
@@ -100,6 +101,8 @@ export type RadioPropsT = {
   checked?: boolean,
   /** Label of radio. */
   children?: React$Node,
+  /** Add more detail about a this radio element. */
```
gergelyke:
```suggestion
  /** Add more detail about a radio element. */
```

schnerd:
@chasestarr should I be able to click on the description to select the radio?

PR #1077, opened 2019-03-22
See PRs  #1147 - #1151 feat(layer): layers proposal

nadiia:


src/layer/rfc.md
```diff
@@ -0,0 +1,68 @@
...
+import * as React from 'react';
+import {Layer} from 'baseui/layers';
+
+class Modal extends React.Component {
+  ...
+  render() {
+    if (!this.props.isOpen) {
+      return null;
+    }
+    return <Layer>{this.renderModal()}</ Layer>;
```
gergelyke: What API would the `Layer` component have? I wonder if it would make sense to have at least an `index` prop, so if for any reason someone has to manually set the z-index (like because of a third-party app)?

nadiia: I was thinking index for the specific placement in the layers order, but not for z-index. I think if there is a need to set up z-index the overrides should be used.

gergelyke: that makes sense

gergelyke:
am I correct to assume, that if folks in their application import this and start using it, if they create layers for themselves, our components will still respect those?

tajo:
So what would be the implementation of `<Layer>`?

```jsx
const Layer = ({ children }) => ReactDOM.createPortal(children, mountNode)
```

where the `mountNode` comes from the `<LayerManager>`?

> In the current RFC we want to solve the first two issues (z-index and visibility management) on a component library level.

> There are many more other layers related issues that include focus management, hover management, keyboard navigation, click events and hotkeys usage.

I think it would be nice to tackle all those other things as well so we have some idea what the final API 
will be even if we introduce them in multiple iterations. 

There are interesting aspects like some layers should disable the body scrolling (modals), some should trap the focus etc. So I wonder how could we make all those things configurable (more components or more props?).

It might be easier to prevent future breaking changes this way.

src/layer/rfc.md
```diff
@@ -0,0 +1,68 @@
+# Layers in BaseWeb
```
gergelyke: can we moe this to the documentation site?

tajo:
Could we make `LayerManager` as a part of the `ThemeProvider` so everyone gets it automatically? A lot of our components will need it to function, right?

nadiia:
> Could we make `LayerManager` as a part of the `ThemeProvider` so everyone gets it automatically? A lot of our components will need it to function, right?
> 
> We could also rename `ThemeProvider` to `BaseProvider` but maybe not worth it.

I like the idea of combining the two. It also functions without the LayersManager in the way as before attaching the layers to the body elm.

tajo:
>  It also functions without the LayersManager in the way as before attaching the layers to the body elm.

But once we want to also tackle the focus management etc, it will require to have a central authority (the context provider) in place, right?

chasestarr:
right now, theme provider can safely be nested. will that be maintained if combined?

tajo:
> right now, theme provider can safely be nested. will that be maintained if combined?

We can detect that LayerManager is already present and not instantiate it again?

nadiia:
> > right now, theme provider can safely be nested. will that be maintained if combined?
> 
> We can detect that LayerManager is already present and not instantiate it again?

That's a good point, for the layers manager it's important to have it as the outermost wrapper, and the `ThemeProvider` may only be added as a component or part of app wrapper for instance.

chasestarr:
ah good idea. seems like a nice approach

nadiia:
PRs #1147 - #1150 are merged

PR #1078, opened 2019-03-22
feat(input): adds autocomplete prop

chasestarr:
#### Scope

- [x] Minor: New Feature


src/input/types.js
```diff
@@ -72,6 +72,8 @@ export type BaseInputPropsT<T> = {
   'aria-describedby': ?string,
   /** Defines styles for inputs that are grouped with other controls. */
   adjoined: AdjoinedT,
+  /** Determines if browser should provide value suggestions. */
+  autocomplete: boolean,
```
gergelyke: shouldn't it be a string value? https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete

chasestarr: thank you

src/input/base-input.js
```diff
@@ -29,6 +29,7 @@ class BaseInput<T: EventTarget> extends React.Component<
     'aria-labelledby': null,
     'aria-describedby': null,
     adjoined: ADJOINED.none,
+    autocomplete: 'on',
```
tajo: Should we be consistent with `autoFocus` and use `autoComplete` ?

chasestarr: sure - I can update that

PR #1080, opened 2019-03-25
fix(table): add roles to styled components

gergelyke:


src/table/styled-components.js
```diff
@@ -19,12 +21,16 @@ export const StyledTable = styled('div', ({$theme}: SharedStylePropsT) => {
   };
 });
 
+export const StyledTable = (props: mixed) => (
```
chasestarr: I think `*` would be more appropriate than `mixed` https://stackoverflow.com/questions/44106189/what-is-the-use-of-asterisk-type-in-flow-and-what-is-the-equivalent-of-that

```suggestion
export const StyledTable = (props: *) => (
```

gergelyke: sounds good - let me update this across to PR

PR #1082, opened 2019-03-25
fix(icon): applies isolated styles to globally overridden icons

chasestarr:
#### Description

Some components allow for style overrides of icons, and apply their own styles to the icons. For example, the close icon on 'toast' has the cursor be a pointer on hover. Refactored the icons to call the main `Icon` component if globally overridden. 

#### Scope

- [x] Patch: Bug Fix


gergelyke:
would you mind adding back the routing for the icon overrides docs page in this PR?

PR #1086, opened 2019-03-26
feat(select): creatable

tajo:
The **creatable select** enables users to create new options along with choosing existing options. This can be also combined with the multi option.

creatable


src/select/select.js
```diff
@@ -402,16 +402,24 @@ class Select extends React.Component<PropsT, SelectStateT> {
...
+        $selected: boolean,
+        $disabled: boolean,
+        $isHighlighted: boolean,
+      },
     },
-  }): React.Node => {
+  ): React.Node => {
+    if (option.isCreatable) {
+      return (
+        locale.select.create + ' ' + '"' + option[this.props.labelKey] + '"'
```
gergelyke: would a template literal be more readable here?

tajo: Yea but it somehow kills code highlighting in my VS Code 😞

chasestarr:
the code LGTM overall - do you mind adding a multi-pick example? That's the example shown in `react-select` here: https://jedwatson.github.io/react-select/

tajo:
> https://jedwatson.github.io/react-select/

tajo:
> the code LGTM overall - do you mind adding a multi-pick example? That's the example shown in `react-select` here: https://jedwatson.github.io/react-select/

Added

src/select/select.js
```diff
@@ -736,20 +744,39 @@ class Select extends React.Component<PropsT, SelectStateT> {
...
+        opt =>
+          opt[this.props.labelKey].toLowerCase() !==
+          filterValue.toLowerCase().trim(),
+      )
+    ) {
+      // $FlowFixMe
+      options.push({
+        id: filterValue,
+        [this.props.labelKey]: filterValue,
+        [this.props.valueKey]: filterValue,
```
nadiia: `valueKey` is the `id` by default

nadiia:
Similar issue when typed in same as already selected value 
creatable2

src/select/select.js
tajo: Hm, if `valueKey = id` then it just gets overridden and everything is fine?

Or we just don't need `id` and should delete the

```js
id: filterValue
```

row?

nadiia: Yes, it still works as expected, just duplicated keys 🤷‍♀️

PR #1087, opened 2019-03-27
fix(select): don't suggest creation of an option that was already picked

tajo:
Fixes problems mentioned in:

https://github.com/uber-web/baseui/pull/1086#pullrequestreview-219212436
https://github.com/uber-web/baseui/pull/1086#issuecomment-476900871

It now considers already created/selected values when filtering the list of options, so it doesn't suggest to create a new option if it already was created/selected.


tajo:
> should we add some test coverage for this?

Added

PR #1088, opened 2019-03-27
fix(link): re-enable styletron overrides

gergelyke:


src/link/index.js
```diff
@@ -11,3 +11,4 @@ import {Link} from './styled-components.js';
 
 //$FlowFixMe
 export const StyledLink = props => <Link data-baseweb="link" {...props} />;
+StyledLink.__STYLETRON__ = Link.__STYLETRON__;
```
tajo: Thanks, I forgot about this!

PR #1098, opened 2019-03-28
fix(styles): use longhand everywhere

gergelyke:


kahwee:
Wouldn’t this break all the code for users of baseui when they specify shorthand to override?

gergelyke:
that's a good question - what's your take on this @nadiia @chasestarr @tajo ?

chasestarr:
It will require CSS changes for developers. We've talked about this previously and have included this kind of change in the [versioning policy](https://baseweb.design/getting-started/versioning-policy/). Based on that doc, it would not justify a major bump

nadiia:
> Wouldn’t this break some of the code for users of baseui when they specify shorthand to override baseui’s previous shorthand? This is a pretty big change for me as a user.

It can be a breaking change. But the css is not a part of public API and the change like replace `display: flex` with `display: block` can be a breaking also. 
We could potentially apply this change at the next major bump, but wouldn't do a major bump for every css change.

tajo:
We now [officially say](https://baseweb.design/theming/understanding-overrides/#introducing-overrides) that you should always use long-hand properties. 

This is a pretty low-risk change. If you are unlucky, you might end-up with some wrong padding/margin/border radius and styletron will be throwing warnings so it should be easy to fix.

kahwee:
I don’t imagine this is difficult to fix as well. Do need to encourage engineers to eye ball their site over the affected components though. This is unlike the usual baseui updates where I can update blindly.

PR #1100, opened 2019-03-29
Release v6.12.3

gergelyke:


chasestarr:
I think the css change should be in a minor, to better indicate there may be code needing an update

gergelyke:
my thinking was that it is a bugfix, as we do not add anything to the public api, right? also, I think neither minor or patch should indicate potential code changes

gergelyke:
Probably we can also look at this as a feature that enables more precise overrides? I am ok with a minor too - what do you think?

PR #1104, opened 2019-04-01
fix(avatar): make src prop optional for Avatar

ashish-kale:
closes https://github.com/uber-web/baseui/issues/1085

#### Description

- Made src prop optional for Avatar
- Render Initials when => no src set or invalid src

#### Scope

- [x] Patch: Bug Fix


src/avatar/__tests__/avatar.test.js
```diff
@@ -79,4 +79,12 @@ describe('Avatar', () => {
     triggerLoadError(wrapper);
     expect(wrapper.find(StyledInitials).text()).toBe('U');
   });
+
+  it('only renders initial from name if src is not provided', () => {
+    const name = 'user';
+    const wrapper = mount(<Avatar name={name} />);
+
+    triggerLoadError(wrapper);
```
chasestarr: triggering an img src load error shouldn't be necessary. can it be removed and still pass ci? (may still need to call `wrapper.update()`)

```suggestion
```

ashish-kale: done.

PR #1117, opened 2019-04-03
feat(navigation-elements): dark theme

gergelyke:


src/themes/dark-theme-colors.js
```diff
@@ -152,6 +152,17 @@ export default {
...
+    // Pagination
+    paginationTriangleDown: primitives.mono100,
+
+    // Header navigation
+    headerNavigationFill: primitives.mono700,
+
+    // Tab
+    tabBarFill: primitives.mono1000,
+    tabColor: primitives.mono300,
+    tabColorActive: primitives.mono100,
```
chasestarr: you can use `colorPrimary` instead of `tabColorActive`

gergelyke: @chasestarr should be ready for another pass

PR #1118, opened 2019-04-04
feat(menu): dark theme

gergelyke:


src/popover/styled-components.js
```diff
@@ -83,7 +83,7 @@ export function getInnerStyles({$theme}: SharedStylePropsT) {
     ? $theme.borders.radius300
     : '0px';
   return {
-    backgroundColor: $theme.colors.backgroundAlt,
+    backgroundColor: $theme.colors.popoverFill,
```
chasestarr: i'm seeing `mono700` and `mono100` as menu backgrounds in figma, which should maintain the `backgroundAlt` value

gergelyke: thanks Chase, updated

PR #1121, opened 2019-04-04
fix(dev-env): windows support

gergelyke:


package.json
```diff
@@ -41,7 +41,7 @@
     "predocumentation:dev": "node ./scripts/post-generator.js",
     "predocumentation:build": "node ./scripts/post-generator.js",
     "predocumentation:serve": "node ./scripts/post-generator.js",
-    "documentation:dev": "yarn build:es2015 --watch & next documentation-site",
+    "documentation:dev": "shx yarn build:es2015 --watch & next documentation-site",
```
chasestarr: should the other scripts use `shx`? like the `documentation:build` below?

gergelyke: yeah, that makes sense - added there too

gergelyke:
closing in favor of https://github.com/uber-web/baseui/pull/1134

PR #1122, opened 2019-04-04
feat(spinner): dark theme

gergelyke:


src/styles/types.js
```diff
@@ -238,6 +238,10 @@ export type ColorsT = {
   toastPositiveBackground: string,
   toastWarningBackground: string,
   toastNegativeBackground: string,
+
+  // Spinner
+  spinnerTrackFill: string,
+  spinnerTrackOpacity: number,
```
gergelyke: I am a bit hesitant about this - this would the first time we'd introduce opacity into the theme. What do you all think about this?

chasestarr: I think there is some opacity happening with tags with dark theme. may want to reference that

chasestarr: looks like both light and dark have an opacity value of `0.16` in figma. why introduce the theme value if both are the same? light color value should change to `mono900` if using opacity

gergelyke: good point - I haven't looked at the light theme's figma, will change things around

thanks Chase!

gergelyke: updated

PR #1125, opened 2019-04-04
feat(dark-theme) list dark colors

chasestarr:
#### Scope

- [x] Minor: New Feature


gergelyke:
I am not sure if we want to fix this as part of this PR, but I think the vertical line should be positioned below the circles?

Screen Shot 2019-04-04 at 3 30 53 PM

chasestarr:
yeah i noticed that bug, investigated it for a half-hour but thought it might be best to fix dark theme for now

PR #1130, opened 2019-04-05
feat(progress-bar): dark theme

gergelyke:


gergelyke:
merging as all is passed, but the heroku app times out

PR #1132, opened 2019-04-05
Tag dark theme

chasestarr:
#### Description

Screen Shot 2019-04-05 at 12 26 37 PM

#### Scope

- [x] Minor: New Feature


src/themes/dark-theme-colors.js
```diff
@@ -107,6 +107,8 @@ export default {
     tagPositiveBackground: primitives.positive500,
     tagWarningBackground: primitives.warning500,
     tagNegativeBackground: primitives.negative500,
+    tagRGBGradient: '0, 0, 0',
+    tagRGBGradientSecondary: '255, 255, 255',
```
gergelyke: should we use the `hexToRgb` function here too, to keep this file consistent, and to only use hex values here? https://github.com/uber-web/baseui/pull/1130/files#diff-b9a386a5a8a07482d2baf6988558bc63R29

src/tag/styled-components.js
```diff
@@ -24,7 +24,7 @@ function getAlpha(props: SharedPropsT, hovered: boolean = false) {
   if (hovered) {
     return '0.8';
   }
-  return '0.92';
+  return '0.6';
```
nadiia: The `light` tag seems to have a too dark background in the light theme with the current change. Maybe move this value to the theme as well?

chasestarr:
i realized i'm way off with this diff. tried to make a light change, but seems like it will require more to dial it in well

PR #1137, opened 2019-04-06
feat(file uploader): button override

### LINKED ISSUE #1045: [File-uploader] Allow to override button component
## Feature description
That would be great to have an opportunity to override button component. I have a request from UI-designer to implement file-uploader as only a button without drag-and-drop area. I easily changed styles for drag-and-drop but button is fixed and also overrides base button.

IgorWilbert:
Fixes #1045 

#### Description

Work in progress.

#### Scope

- [ ] Patch: Bug Fix
- [X] Minor: New Feature
- [ ] Major: Breaking Change


IgorWilbert:
@gergelyke I tested it on the documentation website and this code seems to be enough to override the button component. However, I have some doubts:

- The button component can be overridden in many ways, with different states, shapes, sizes etc. Should we support total override? If so, is there anything else to be done to enable it?
- From the description of the issue, it seems the button is supposed to be a self-contained file uploader (e.g. a file uploader without drag-and-drop area). Is there anything missing for this?
- I essentially copied and pasted the code from the buttons overridden present in the date picker. Could you please give some suggestions to improve the code and write a new example on the documentation for the file uploader?

Thank you!

src/file-uploader/file-uploader.js
```diff
@@ -52,6 +53,7 @@ function FileUploader(props: PropsT) {
     overrides.HiddenInput,
     StyledHiddenInput,
   );
+  const [Button, buttonProps] = getOverrides(overrides.Button, StyledButton);
```
chasestarr: I'd like to maintain using the baseui button component here so that it maintains a consistent interface for devs looking to customize it. Check out this example of 'overriding': https://github.com/uber-web/baseui/blob/master/src/datepicker/datepicker.js#L141

IgorWilbert: I believe this is what you meant. I tested it and it seems to work. Please tell me if there is something else you want me to do. Thanks!

src/file-uploader/file-uploader.js
```diff
@@ -111,13 +112,9 @@ function FileUploader(props: PropsT) {
                         disabled={props.disabled}
                         kind={KIND.minimal}
                         onClick={open}
-                        overrides={{
```
chasestarr: This is the right approach!
I believe `<Button>` should be replaced with `<ButtonComponent>` now. Also, the `Button` override needs to be specified in the types file here: https://github.com/uber-web/baseui/blob/master/src/file-uploader/types.js#L22 (I would give it a more descriptive title though)

IgorWilbert: Done! I chose to specify it in the types file as `ButtonComponent`, mostly for consistency with the other overrides. Let me know if you want me to change this or something else. Thanks!

src/file-uploader/file-uploader.js
```diff
@@ -167,33 +169,25 @@ function FileUploader(props: PropsT) {
                         </ContentMessage>
                       )}
                       {props.errorMessage ? (
-                        <Button
+                        <ButtonComponent
                           kind={KIND.minimal}
                           onClick={() => {
                             props.onRetry && props.onRetry();
                           }}
-                          overrides={{
```
chasestarr: can you maintain this prop? same thing on L189. Other than this LGTM. thanks!

IgorWilbert: Sorry, I missed these two, it should be fine now. Thanks!

IgorWilbert:
Hey @chasestarr ! Is there anything else I need to do to make the remaining tests pass? Thanks!

chasestarr:
Thank you @IgorWilbert !

PR #1138, opened 2019-04-07
feat(Menu/Select): add onHighlightedItemChanged handler 

### LINKED ISSUE #994: [Menu/Select] add onHighlightedItemChanged handler
## Feature description
Add a way to listen to changes in the highlighted item in a menu or select. It is called when the highlighted item changes and it can be called with null if no option is highlighted.

## Use case
Display additional information based on the highlighted item in the menu in a separate view.

IgorWilbert:
Fixes #994 

#### Description

Work in progress.

#### Scope

- [ ] Patch: Bug Fix
- [X] Minor: New Feature
- [ ] Major: Breaking Change


IgorWilbert:
Hi @chasestarr ! This is my first attempt to solve it.

- It prints the array index of the highlighted item whenever it changes, for both arrow keys and mouse hover events;
- Right now it always prints the index, regardless of whether the item is `disabled` or not;
- In the initial state no item is highlighted, so it does not print anything;
- I tried other things such as making `getBackgroundColor` function on `styled-components.js` call a helper fuction or using callbacks inside `this.internalSetState` on `stateful-container,js`, but none worked quite right;

Let me know your thoughts about it, thanks!

chasestarr:
Hi @IgorWilbert, thanks! Check out some features of base that may be helpful demonstrated here: https://codesandbox.io/s/w6o67nqr2l

In particular, `stateful` components have a `stateReducer` prop that is called on each internal state change. The use-case described in #994 is possible when only using the `Menu` component, but we have no way to pass props through the `Select` component to the dropdown menu. This may be a good reference for how to 'override' internal components: https://github.com/uber-web/baseui/blob/master/src/datepicker/datepicker.js#L141

chasestarr:
The interface I'd like to see is something like the example below. This does not require importing anything from the `StatefulMenu` component, but rather providing additional props to the internal menu. 

```jsx
<Select
  overrides={{StatefulMenu: {props: {stateReducer: reducerFn}}}
/>
```

No need to add a `StyledStatefulMenu`, but rather create a override where the default the call looks like 
```js
const [OverriddenStatefulMenu, statefulMenuProps] = getOverrides(overrides.StatefulMenu,StatefulMenu);

// ...

<OverriddenStatefulMenu
  // ...already existing props
  {...statefulMenuProps}
/>
```

Haven't tested if the code provided here is correct, but should be a decent starting point

IgorWilbert:
Hi @chasestarr ! I guess the code I pushed is working, but I'm not sure how to test it. In the following codepen:
https://codesandbox.io/s/238xqp5mmp
The `basic.js` gives a Select in which `changeType select` is printed on click in an item, but it does not respond to hover or arrows. The one in `controlledExample.js` on the other hand, gives the error `reducerFn` is not defined. I think I am missing something quite obvious. Any idea of what it might be? Thanks!

chasestarr:
codesandbox can only reference published versions of the baseui package. so your changes are not reflected there. I would edit a `scenario` file and `yarn storybook` to double check.

IgorWilbert:
Hi @chasestarr ! I know, I only used codesandbox to show you the code I was writting without committing it. Now I managed to create an example in a scenario file that responds to both mouse hover and arrow keys. I also created an example in the documentation site that seems to work as it should. Let me know if you want me to change something else. Thanks!

chasestarr:
looks like linting fails. can you run `yarn test` locally?

IgorWilbert:
Hi @chasestarr ! I ran `yarn test` and this was the result:
error
Unfortunately, I didn't have time to investigate and try to fix it, I'll do so as soon as I get home. It would be helpful if you had time to give me some hints. Thanks!

chasestarr:
most of those errors are from [prettier](https://prettier.io/). follow the instructions on that site to align with the style guide.

also ensure that the uber header comment (found at the top of other files) is included as well as a `// @flow` comment.

You may need to use `eslint-disable` to suppress the `no/console` error

IgorWilbert:
Hi @chasestarr ! I did what you suggested and now all tests passed. Thank you!

src/select/dropdown.js
```diff
@@ -150,6 +154,7 @@ export default class SelectDropdown extends React.Component<DropdownPropsT> {
               Option: overrides.DropdownOption || {},
             },
           )}
+          {...statefulMenuProps}
```
nadiia: We need to omit `overrides` from the `statefulMenuProps ` since we merge it above.

PR #1140, opened 2019-04-09
feat(tag): dark theme

chasestarr:
#### Description

Enables tag dark theme color scheme. Previously: https://github.com/uber-web/baseui/pull/1132

Early on considered taking a light touch approach to applying the correct colors in dark mode, but found that the color modulations between light/dark themes to be inconsistent. Because of that, simply applying correct gradients would not work nicely. Also realized that although figma displays color values as overlays, they actually map one to one with standardized color values. Check out the figma doc I put together to map everything precisely:

tag-color-mapping

Since all of the color values are just modulations from the `400` color ramp unit, I explored programmatically lightening and darkening for each tag state but found that it led to inconsistent results. Notice that most of the colors look fine, but the `positive` text color looks rather 'neon'. Because of these inconsistencies I chose to exactly define each color value. While this approach can have rather large line counts, I believe it is better to provide a perfect component.

inconsistent-approaches

I still could not avoid from color modulations entirely because this component supports custom colors. To perform the shifts, I pulled in the `polished` package. It doesn't work for all provided colors because it relies on a middle value between light/dark, but developers should be able to test out what works for them.

Screen Shot 2019-04-09 at 2 38 04 PM
Screen Shot 2019-04-09 at 2 38 23 PM

---

Screen Shot 2019-04-09 at 2 04 08 PM
Screen Shot 2019-04-09 at 2 03 51 PM

#### Scope

- [x] Minor: New Feature


tajo:
Nice, that's a lot of manual work! The disabled states might not have enough contrast but that might be acceptable for that use-case?

chasestarr:
> Nice, that's a lot of manual work! The disabled states might not have enough contrast but that might be acceptable for that use-case?

I thought the same thing, but assume that this was reviewed by design folks. Given the current implementation it should be simple to update

nadiia:
Minor thing: 
- for closable/non-clickable outlined tags would make sense to have the hovered action's background color to match the border color:
Screen Shot 2019-04-09 at 4 31 25 PM
It's a lot of tag specific color props. Wonder if designers can simplify it slightly.
Great job!

PR #1146, opened 2019-04-12
fix(menu): accept rootRef through props

nadiia:


chasestarr:
the changes generally lgtm once tests pass

PR #1148, opened 2019-04-12
feat(layer): initial implementation of the TetherBehavior  (Part 2)

nadiia:


src/layer/index.js
```diff
@@ -7,3 +7,5 @@ LICENSE file in the root directory of this source tree.
 // @flow
 export {default as LayersManager} from './layers-manager.js';
 export {default as Layer} from './layer.js';
+export {default as TetherBehavior} from './tether.js';
+export {PLACEMENT as TETHER_PLACEMENT} from './constants.js';
```
chasestarr: can the type be renamed to `TETHER_PLACEMENT` to avoid the unnecessary indirection?

src/layer/tether.js
```diff
@@ -0,0 +1,107 @@
...
+import React from 'react';
+import Popper from 'popper.js';
+import {toPopperPlacement, parsePopperOffset} from './utils.js';
+import {PLACEMENT} from './constants.js';
+import type {TetherPropsT, TetherStateT, PopperDataObjectT} from './types.js';
+
+class Tether extends React.Component<TetherPropsT, TetherStateT> {
+  static defaultProps = {
+    anchorRef: null,
+    ignoreBoundary: false,
```
chasestarr: I often regret building this feature into the popover. It establishes a wrapper interface into the popper.js library for no substantial gain. we don't perform any manipulation on this value before provided to popper js. i think a more appropriate interface is to expose a `popperOptions` prop that get's passed to popper. This would enable devs to customize however they need to.

```js
this.popper = new Popper(this.props.anchorRef, this.props.popperRef, {
    // ... other parameters
    preventOverflow: {enabled: !this.props.ignoreBoundary},
    ...this.props.popoverOptions,
  },
});
```

Not sure if absolutely necessary for this pr, but seems like a good chance to introduce more productive patterns.

chasestarr: relevant gh issue: https://github.com/uber-web/baseui/issues/1152

PR #1149, opened 2019-04-12
fix(modal): render modal with Layer (Part 3)

nadiia:


src/modal/modal.js
```diff
@@ -334,11 +334,14 @@ class Modal extends React.Component<ModalPropsT, ModalStateT> {
...
     }
-    const mountNode = this.getMountNode();
-    if (!mountNode) {
-      return null;
-    }
-    return ReactDOM.createPortal(this.renderModal(), mountNode);
+    return (
+      <Layer
+        mountNode={this.props.mountNode}
+        onMount={this.props.isOpen ? this.autoFocus : undefined}
```
chasestarr: I can't find the class method `autoFocus`. was this change tested?

nadiia: It's tested with an e2e test. The method is on the line 199.

chasestarr: ah sorry

PR #1150, opened 2019-04-12
fix(popover): render popover with Layer and TetherBehavior  (Part 4)

nadiia:


src/popover/popover.js
```diff
@@ -8,9 +8,8 @@ LICENSE file in the root directory of this source tree.
 /* global document */
 /* eslint-disable react/no-find-dom-node */
 import * as React from 'react';
-import ReactDOM from 'react-dom';
 
-import Popper from 'popper.js';
+// import Popper from 'popper.js';
```
chasestarr:
```suggestion
```

src/popover/popover.js
```diff
@@ -512,8 +448,24 @@ class Popover extends React.Component<PopoverPropsT, PopoverPrivateStateT> {
...
+          <Layer
+            mountNode={this.props.mountNode}
+            onMount={() => this.setState({isLayerMounted: true})}
+            onUnmount={() => this.setState({isLayerMounted: false})}
+          >
+            <TetherBehavior
+              anchorRef={this.anchorRef.current}
+              arrowRef={this.arrowRef.current}
+              popperRef={this.popperRef.current}
+              ignoreBoundary={this.props.ignoreBoundary}
```
chasestarr: following up on my comment from the other pr. a consistent `popover` interface can be maintained as well as enabling the change I suggested earlier. 

```jsx
<TetherBehavior
  popperOptions={{
    preventOverflow: {enabled: !this.props.ignoreBoundary},
    ...this.props.popperOptions,
  }}
/>
```

src/popover/types.js
```diff
@@ -9,14 +9,19 @@ LICENSE file in the root directory of this source tree.
 import * as React from 'react';
 import type {ThemeT} from '../styles/types.js';
 import type {OverrideT} from '../helpers/overrides.js';
+import type {
+  PopperDataObjectT as TetherPopperDataObjectT,
```
chasestarr: do these need to be renamed on import?

```js
export {PopperDataObjectT} from '../layer/types.js';
```

src/popover/popover.js
nadiia: Addressed it in the #1150. We can remove this prop for the popover in the next major version. I've added a comment on it here.

PR #1151, opened 2019-04-12
feat(pagination): replace pagination dropdown with the Select component

nadiia:


chasestarr:
what advantage do we get from not using popover here? i'm concerned about the position data leaking out of the tether abstraction

chasestarr:
this can likely merge into the v7 branch if you prefer

PR #1156, opened 2019-04-15
feat(timezone-picker): Timezone Picker

chasestarr:
#### Description

Screen Shot 2019-04-15 at 12 13 49 PM

some interesting reading related to timezones/date-time programming:
https://zachholman.com/talk/utc-is-enough-for-everyone-right
https://medium.com/@toastui/handling-time-zone-in-javascript-547e67aa842d

#### Scope

- [x] Minor: New Feature


tajo:
Looking great! Should we provide an option to translate labels through some optional function?

chasestarr:
> Looking great! Should we provide an option to translate labels through some optional function?

Yes - I will edit this pr today

src/datepicker/types.js
```diff
@@ -270,6 +270,11 @@ export type TimezonePickerPropsT = {
    * exact specificity, provide a date. Otherwise it will default to the relevant timezone at render.
    */
   date?: Date,
+  /**
+   * Customize the option's label. Useful for translations and optionally mapping from
+   * 'America/Los_Angeles' to 'Pacific Time'.
+   */
+  mapLabels?: OptionT => string,
```
tajo: Any reason why to pass the whole `OptionT` and not just the label? Also, the select's label is type `React.Node`?

chasestarr: good point about `React.Node` will update. 

in this case, the `id` is the IANA title (`America/Los_Angeles`), and the label contains some additional information like the offset, and short code (`PST`/`PDT`). I thought that extra context may be useful

chasestarr:
```suggestion
  mapLabels?: OptionT => React.Node,
```

chasestarr:
Just realized that I need to handle the `date` prop update case. Will update

chasestarr:
can't reproduce the puppeteer fail locally, gonna look into this today

PR #1162, opened 2019-04-16
fix(link): get color values from related theme tokens

ervasive:
#### Description

Make the Link component use the related theme tokens for its text color values.

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/src/themes/creator.js
+++ b/src/themes/creator.js
@@ -157,7 +157,7 @@ export default function createTheme(
       linkText: primitives.primary400,
       linkVisited: primitives.primary500,
       linkHover: primitives.primary600,
-      linkActive: primitives.primary400,
+      linkActive: primitives.primary600,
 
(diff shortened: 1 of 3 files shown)


tajo:
Thanks! Just waiting to confirm with designers. It seems they might want to use a different color.

src/themes/creator.js
```diff
@@ -157,6 +157,7 @@ export default function createTheme(
       linkText: primitives.primary400,
       linkVisited: primitives.primary500,
       linkHover: primitives.primary600,
+      linkActive: primitives.primary400,
```
nguyensomniac: Can you change this to `primary600`?

ervasive: Should I also update this in dark theme?

PR #1163, opened 2019-04-17
docs(datetime-picker-components): adds proposal

chasestarr:
#### Description

Adds proposal for new `TimePicker`, `TimezonePicker` components and updates to `DatePicker` component.

relevant pr: https://github.com/uber-web/baseui/pull/1156

[rendered rfc](https://github.com/uber-web/baseui/blob/123ecdca0e56d6135483116ea29f0ae874f17178/proposals/datetime-picker-tools.md)


chasestarr:
We could introduce an abstraction layer on top of the collection of 'pickers'. Thinking about something like below

```js
function Pickers() {
  return (
    <DateTimePickersManager>
      <DatePicker />
      <TimePicker />
      <DatePicker />
      <TimePicker />
    </DateTimePickersManager>
  );
}
```

If a picker component is within the manager context, register it and handle the sibling communication amongst each other. It would reduce [this sort of complexity](https://github.com/uber-web/baseui/pull/1163/files#diff-439d17e2e9ed1a337c96c903593399fdR110) where the date picker must update the state of its sibling time picker. May be a premature abstraction, but keep in mind that I made the proposal with this type of extension considered.

proposals/datetime-picker-tools.md
```diff
@@ -0,0 +1,147 @@
...
+
+Date/time selections need more precision than a single date or range of dates (what is currently implemented). This proposal outlines two new components and changes to one existing component. `TimePicker`, `TimezonePicker`, and `DatePicker`.
+
+## APIs and Components
+
+`TimePicker` - used to select a time, decoupled from a date. E.g. `11:50AM`.
+
+API:
+
+* `precisionInMinutes`: `number` Range from 1 - 60. 15 default.
```
tajo: `minuteStep` might work better? Other libraries tend to use it and it seems more obvious? 

Also, did we consider including seconds? I think sooner or later somebody will request that. Then, it might make sense to keep `step` in seconds.

proposals/datetime-picker-tools.md
```diff
@@ -0,0 +1,147 @@
...
+
+## Motivation
+
+Date/time selections need more precision than a single date or range of dates (what is currently implemented). This proposal outlines two new components and changes to one existing component. `TimePicker`, `TimezonePicker`, and `DatePicker`.
+
+## APIs and Components
+
+`TimePicker` - used to select a time, decoupled from a date. E.g. `11:50AM`.
+
+API:
```
tajo: I think the most of world uses 24h time format. It would be nice to have a mode for that? Or even having some mapping function to create your own format (but 12h vs 24h might be good enough).

chasestarr: added a 12/24 format

proposals/datetime-picker-tools.md
chasestarr: changed to `step` in seconds. also changed `callback` argument and `value` prop to simply be a number

tajo:
> We could introduce an abstraction layer on top of the collection of 'pickers'. 

Yea, this could be a nice addition!

PR #1168, opened 2019-04-18
feat(time-picker): implements component

chasestarr:
#### Description

Screen Shot 2019-04-18 at 4 53 03 PM
Screen Shot 2019-04-18 at 4 53 10 PM

See [proposal](https://github.com/uber-web/baseui/blob/master/proposals/datetime-picker-tools.md) for more information

#### Scope

- [x] Minor: New Feature


chasestarr:
going to convert the value and onchange handler to work with `Date` rather than `number`

gergelyke:
code looks great! when testing it in the browser, I've ran into an issue an a question:

* neither with `enter`, nor with a mouse click I can select values
* should we strip whitespaces when using the keyboard? so either `10:00`, `10 : 00` or `10: 00` would yield the same results

chasestarr:
> * neither with `enter`, nor with a mouse click I can select values
thank you - fixed an issue with default time values

> * should we strip whitespaces when using the keyboard? so either `10:00`, `10 : 00` or `10: 00` would yield the same results
I removed spaces from the labels

PR #1178, opened 2019-04-22
Calendar dropdown quick select

chasestarr:
#### Description

Screen Shot 2019-04-22 at 3 32 01 PM

- refactors quick select to dropdown menu
- adds new `quickSelectOptions` prop
- removes `QuickSelectLabel` and `QuickSelectButton` overrides
- adds `QuickSelect` override
- moves 'day headers' into the blue calendar header
- combines year/month selectors into one dropdown menu
- removes the `MonthSelect` and `YearSelect` overrides
- adds the `MonthYearSelect` overrides

todos:
- render day items with circular highlights
- time dropdowns (will add once `TimeSelector` is merged)
- double calendar view

#### Scope

- [ ] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


src/datepicker/calendar-header.js
```diff
@@ -198,91 +212,94 @@ export default class CalendarHeader extends React.Component<HeaderPropsT> {
...
-      yearSelectProps && yearSelectProps.overrides,
+      monthYearSelectProps && monthYearSelectProps.overrides,
     );
-    const maxYear = maxDate ? getYear(maxDate) : 2100;
-    const minYear = minDate ? getYear(minDate) : 1900;
-    const yearOptions = [];
+
+    const MONTHS = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11];
+    const maxYear = maxDate ? getYear(maxDate) : 2030;
+    const minYear = minDate ? getYear(minDate) : 1980;
```
gergelyke: should we keep the original values? 1980 can be a quite limiting default, if it's used for birthdates for example

chasestarr: i had considered birthdates, but thought that a calendar view is generally a poor UX for choosing them. Most sites simply have a masked input. Honestly, I tightened constraints here because it was slow to operate with all of the original values. Is a virtualized list worth pulling in here? Maybe we could roll our own virtualization to not pay as much for the bundle increase...

gergelyke: that makes sense, thanks for explaining Chase 👍 

yeah, probably it's not worth doing that - let's use the updated min and max values

tajo:
Screen Shot 2019-04-23 at 11 12 22 AM

Limiting the range is fine but if I go to `Jan 1980` and use the left arrow, the month dropdown gets blank. It would be nice to show `Dec 1979`. Or disable the left arrow. The same applies for the future dates.

chasestarr:
> Limiting the range is fine but if I go to `Jan 1980` and use the left arrow, the month dropdown gets blank. It would be nice to show `Dec 1979`. Or disable the left arrow. The same applies for the future dates.

thank you, disabled the buttons and added tests

chasestarr:
Do we consider overrides deprecations breaking?

gergelyke:
as we have them documented on the website, probably we should - what do you think?

chasestarr:
I tend to lean towards making the most 'safe' semver change. Let's release this in a major

nadiia:
The "Datepicker with quick select actions" example's initial select doesn't work correctly: while it's supposed to be a range picker it doesn't let one select a range until a quick select is used, after it the range selection works as supposed to.
range

chasestarr:
> The "Datepicker with quick select actions" example's initial select doesn't work correctly: while it's supposed to be a range picker it doesn't let one select a range until a quick select is used, after it the range selection works as supposed to.
looks like an [existing issue](https://baseweb.design/components/datepicker/#datepicker-with-quick-select-actions) but i can look into it

> Edit: It also doesn't let one to tab through the datepicker past the selected value.
Can you provide more description on the expected/actual behavior?

src/datepicker/calendar.js
```diff
@@ -285,13 +269,50 @@ export default class Calendar extends React.Component<
...
   };
 
   onDayMouseLeave = (data: {event: Event, date: Date}) => {
     const {date} = data;
     this.setHighlightedDate(date);
-    this.props.onDayMouseLeave(data);
+    this.props.onDayMouseLeave && this.props.onDayMouseLeave(data);
+  };
+
+  handleDateChange = (data: {date: ?Date | Array<Date>}) => {
```
nadiia: What do you think about having the date value always as an array (range or single datepickjer). Since we're releasing a major update we could add this change too. It will align it with the Selects API approach, simplify the internals, and remove the confusion regarding passing an initial empty array value for the range picker.

src/datepicker/calendar.js
```diff
@@ -285,13 +269,50 @@ export default class Calendar extends React.Component<
...
+      onChange({date});
+    }
+
+    // const [firstValue, ...restDates] = [].concat(this.props.value);
+    // const nextDate = applyTime(firstValue, time);
+    // if (this.props.range) {
+    //   onChange({date: [nextDate, ...restDates]});
+    // } else {
+    //   onChange({date: nextDate});
+    // }
```
nadiia: remove commented out code

src/datepicker/datepicker.js
```diff
@@ -64,7 +54,7 @@ export default class Datepicker extends React.Component<
       isPseudoFocused,
       ...(calendarFocused === null ? {} : {calendarFocused}),
     });
-    this.props.onChange(data);
+    this.props.onChange && this.props.onChange(data);
```
nadiia: Why not keep the default prop so we don't need to make one liner checks every time we call the handler?

src/datepicker/datepicker.js
```diff
@@ -79,7 +69,7 @@ export default class Datepicker extends React.Component<
 
   formatDisplayValue(date: ?Date | Array<Date>) {
     const formatDisplayValue = this.props.formatDisplayValue || this.formatDate;
-    return formatDisplayValue(date, this.props.formatString);
+    return formatDisplayValue(date, this.props.formatString || 'YYYY/MM/dd');
```
nadiia: Is there a reason we moved it to here instead of default props?

src/datepicker/datepicker.js
```diff
@@ -197,7 +187,7 @@ export default class Datepicker extends React.Component<
                 onFocus={this.open}
                 onBlur={this.handleInputBlur}
                 onKeyDown={this.handleKeyDown}
-                placeholder={this.props.placeholder}
+                placeholder={this.props.placeholder || 'YYYY/MM/DD'}
```
nadiia: same here

src/datepicker/stateful-container.js
```diff
@@ -23,14 +23,15 @@ class StatefulContainer extends React.Component<
   ContainerStateT,
 > {
   static defaultProps = {
-    initialState: {value: null},
+    initialState: {value: null, time: null},
```
nadiia: Since we don't include time here in the `value` would it make sense to rename `value` to `date`?

src/datepicker/datepicker.js
chasestarr: the `onChange` prop is optional in types now, so we need to check at each call-site

src/datepicker/datepicker.js
chasestarr: `formatString` is optional in types. I could have destructured with a default value, or in-lined like this

src/datepicker/stateful-container.js
chasestarr: removed `time`

src/datepicker/calendar-header.js
```diff
@@ -34,39 +42,72 @@ const navBtnStyle = ({$theme}) => ({
...
+    if (this.props.onYearChange) {
+      // $FlowFixMe
+      this.props.onYearChange({date: setYear(this.props.date, value[0].id)});
+    }
   };
 
   increaseMonth = () => {
-    this.props.onMonthChange({date: addMonths(this.props.date, 1)});
+    if (this.props.onMonthChange) {
+      // $FlowFixMe
```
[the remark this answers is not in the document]
chasestarr: tbh I could not get the flow error to go away.. even after the if-statement

src/datepicker/calendar-header.js
```diff
@@ -34,39 +42,72 @@ const navBtnStyle = ({$theme}) => ({
...
+      this.props.onMonthChange({date: addMonths(this.props.date, 1)});
+    }
   };
 
   decreaseMonth = () => {
-    this.props.onMonthChange({date: subMonths(this.props.date, 1)});
+    if (this.props.onMonthChange) {
+      // $FlowFixMe
```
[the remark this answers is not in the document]
chasestarr: addressed in other comment

nadiia:
> > The "Datepicker with quick select actions" example's initial select doesn't work correctly: while it's supposed to be a range picker it doesn't let one select a range until a quick select is used, after it the range selection works as supposed to.
> > looks like an [existing issue](https://baseweb.design/components/datepicker/#datepicker-with-quick-select-actions) but i can look into it
> 
> > Edit: It also doesn't let one to tab through the datepicker past the selected value.
>  Can you provide more description on the expected/actual behavior?

Since the quick select only applicable to the range picker the expected behavior would be it allows one to select a date range one the component rendered.

src/datepicker/calendar.js
chasestarr: i'd rather enable a better developer experience and contain the complexity for ourselves

src/datepicker/styled-components.js
```diff
@@ -89,28 +76,56 @@ export const StyledCalendarHeader = styled(
...
+
+export const StyledMonthYearSelectIconContainer = styled('span', props => {
+  return {
+    alignItems: 'center',
+    display: 'flex',
+    marginLeft: props.$theme.sizing.scale500,
+  };
+});
+
+function getArrowBtnStyle({$theme, $disabled}: SharedStylePropsT) {
```
IgorWilbert: I believe the same way we use the acronym `Btn` for `Button`, we could use for other terms that occur very often. For example, `OverriddenStatefulPopover` could be written as `OvdStfPopover`, `StyledMonthYearSelectIconContainer` as `SldMonthYearSltIcCtr`. Or maybe assign special and unique characters to represent words like `styled` `overridden` and `stateful`. This is something to be done in the far future, but I guess it could reduce some variable names considerably.

src/datepicker/__tests__/datepicker.scenario.js
```diff
@@ -17,19 +17,12 @@ export const component = () => (
     aria-label="Select a date"
     highlightedDate={new Date('March 10, 2019')}
     overrides={{
-      MonthSelect: {
+      MonthYearSelect: {
```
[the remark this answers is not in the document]
chasestarr: My preference is to be explicit rather than terse. This override target the month and year select, not a full date

src/datepicker/__tests__/datepicker.e2e.js
```diff
@@ -178,19 +177,71 @@ describe('Datepicker', () => {
...
     });
 
-    await page.waitFor(selectors.monthDropdown, {hidden: true});
+    await page.waitFor(selectors.monthYearDropdown, {hidden: true});
     await page.waitFor(selectors.calendar);
     await page.waitFor(selectors.day6);
   });
+
+  it('disables previous month button if minimum month is selected', async () => {
+    await mount(page, 'datepicker');
```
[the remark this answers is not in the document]
chasestarr: I think the great part about async await is that we can forget about promise complexity. We can understand and read the code as if these actions are not asynchronous

PR #1184, opened 2019-04-24
fix(tooltip): use predefined theme tokens

ervasive:
#### Description

Make the `Tooltip` component use predefined theme tokens for its background and text colors.

There are a few minor issues with how the tooltip currently sets its colors:
1. Tooltip's color token(s) are defined [outside](https://github.com/uber-web/baseui/blob/master/src/themes/creator.js#L576-L578)  the ["normal" theme colors](https://github.com/uber-web/baseui/blob/master/src/themes/creator.js#L26-L408) (which makes theming a bit unintuitive and inconsistent).
2. Because of the previous point, themes don't actually match their [type](https://github.com/uber-web/baseui/blob/master/src/styles/types.js#L481-L491), but flow can't catch this mismatch because of [this](https://github.com/uber-web/baseui/blob/master/src/themes/creator.js#L582).
3. The tooltip's styling is a bit limited as we can set its background color independently but not its text color.

P.S. Regarding the flow not catching themes type mismatch. I see that there is a [todo note](https://github.com/uber-web/baseui/blob/master/src/themes/creator.js#L581) to fix the type validation, but I think something like this could be added as a quick temporary solution:

```diff
- // $FlowFixMe
- return deepMerge(theme, overrides);

+ if (!overrides) {
+    return theme;
+  } else {
+    // $FlowFixMe
+    return deepMerge(theme, overrides);
+  }
```

Thanks

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


src/themes/creator.js
```diff
@@ -573,9 +577,6 @@ export default function createTheme(
     zIndex: {
       modal: 2000,
     },
-    tooltip: {
-      backgroundColor: primitives.mono900,
```
gergelyke: do you think we can keep this here? I know this is inconsistent with the rest of the color values, but this would be a breaking change - I'd recommend adding a note here to change this in the next major

ervasive: Should I add it to the `ThemeT` type then? Because right now this field is incompatible with the specified type?

gergelyke: oh, that's a good catch - yes, we should add it to the ThemeT type. also, can you make sure to add a comment too, that this is something we are removing in the next major?

ervasive: Done

chasestarr:
Thank You

ervasive:
Happy to contribute

PR #1185, opened 2019-04-24
fix(select): tag override

gergelyke:
There was an issue in the MultiValue overrides, as style overrides were not applied to the Tag component.

I am not sure if this is the right way to handle this - please let me know how I should update this PR, if needed.

### DIFF
--- a/src/select/multi-value.js
+++ b/src/select/multi-value.js
@@ -22,6 +22,7 @@ export default function MultiValue(props: any) {
             marginBottom: '0',
             marginLeft: sizing.scale0,
             marginTop: '0',
+            ...multiValueProps.$style,
           }),
         },
       }}


chasestarr:
It's working correctly, see here: https://codesandbox.io/s/73x565572q

it's passing props through to the `Tag` component - so overrides must be provided through props

nadiia:
We can be more specific in the documentation of overrides saying that a specific identifier is for a component and not a styled element overrides. And then maybe in a single place document how to deal with overrides for an underlying component vs styled element's overrides.
So we shouldn't spread styles to only one selected element of an underlying component but use the approach from Chase's comment.

gergelyke:
I see, thanks Chase, Nadiia!

PR #1187, opened 2019-04-24
fix(select): quick re-selection, mobile bugs

tajo:
Solves https://github.com/uber-web/baseui/issues/1179  simply by disabling the fade out animation. It's barely noticeable anyway.

On iOS/Safari: If input's `font-size` is less than `16px`, Safari zooms in and makes the select component almost unusable / very awkward to use. So I bumped the font-size from `14px` to `16px` and notified designers about this weird feature since they are also redoing all inputs. For plain inputs, you can at least easily override the fontSize but select is also doing some measurements and input auto-sizing.

On all mobile devices, the creatable feature didn't work because the touch on menu was triggering `closeMenu` even before the item was selected.


src/popover/constants.js
```diff
@@ -42,7 +42,7 @@ export const POPOVER_MARGIN = 8;
 
 export const ARROW_SIZE = 6;
 
-export const ANIMATE_OUT_TIME = 500;
```
nadiia: Should it be overwritten just in the select's popover usage case instead of turning it off for the popover globally? This would affect the tooltip also.

tajo: We could. Personally I am not a big fan of fade-in/fade-out animations since it often makes application looking sluggish but up to you. Also, should it be configurable then?

gergelyke: I think making it configurable makes perfect sense - should we add this to the theme?

tajo: I added it as new Popover prop for now. I think the theme makes sense once we would like to use the same value across multiple components.

PR #1188, opened 2019-04-25
fix(block): does not apply undefined value in transform

chasestarr:
#### Scope

- [x] Patch: Bug Fix

### DIFF
--- a/src/block/styled-components.js
+++ b/src/block/styled-components.js
@@ -34,7 +34,11 @@ function build(breakpoints: BreakpointsT) {
         value.forEach((v, index) => {
           // Do not create a media query for the smallest breakpoint.
           if (index === 0) {
-            styles[property] = transform(v);
+            // styletron will throw when value is undefined. if so, do not apply rule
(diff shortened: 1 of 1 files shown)


src/block/styled-components.js
```diff
@@ -34,7 +34,11 @@ function build(breakpoints: BreakpointsT) {
         value.forEach((v, index) => {
           // Do not create a media query for the smallest breakpoint.
           if (index === 0) {
-            styles[property] = transform(v);
+            // styletron will throw when value is undefined. if so, do not apply rule
+            const transformedValue = transform(v);
+            if (transformedValue !== undefined) {
+              styles[property] = transform(v);
```
nadiia: There are a couple more usage of `transform` below. How about adding this check in the transform function itself?

nadiia: Oh, I see, it's different for every props.

chasestarr: thank you 🙏

PR #1195, opened 2019-04-26
chore: remove undefined attributes

leyanlo:
#### Description

Remove undefined attributes. This is not required, but given the choice between setting a value to `undefined` and not setting the value at all, it’s cleaner to avoid setting the value. It ends up functionally the same, and snapshots of the objects are slightly cleaner. Also, doing this allows you to check for the key using `foo.hasOwnProperty('bar')`, which is slightly safer than checking `typeof foo.bar === 'undefined'`, since it guards against prototypical inheritance.

#### Scope

- [ ] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


chasestarr:
can you share a bit about why this change is required?

leyanlo:
@chasestarr Updated the description. Let me know if that makes sense!

documentation-site/components/markdown-elements.js
```diff
@@ -110,7 +110,7 @@ export const DocLink = ({children, href}: {children: string, href: string}) => {
     (parts[0] === '' && parts[1] !== '') || !href.includes('http');
   return (
     <Link href={href} prefetch={internal}>
-      <StyledLink href={href} target={internal ? undefined : '_blank'}>
+      <StyledLink href={href} {...(internal ? {target: '_blank'} : {})}>
```
Hongarc: Old code same:
```js
if (internal) {
	return {target: undefined}
} else {
	return {target: '_blank'}
}
```

But your code:
```js
if (internal) {
	return {target: '_blank'}
} else {
	return {};
}
```

leyanlo: Good catch! I updated the code to be correct now. Thanks!

src/layer/types.js
```diff
@@ -15,7 +15,7 @@ export type LayersManagerPropsT = {
 };
 
 export type LayersContextT = {
-  host: typeof undefined | ?HTMLElement,
```
nadiia: We need to keep the `typeof undefined | ?HTMLElement` type here, since in the case when `LayersManager` is not added in the app the host value will be `undefined` and we'll render the content in the `document.body` instead.

leyanlo: Isn’t `typeof undefined` a subset of `?HTMLElement` though?
https://flow.org/en/docs/types/maybe/

`?HTMLElement` is equivalent to `HTMLElement & typeof undefined & null` already.

PR #1199, opened 2019-04-29
fix(button-group): add button overrides fix for button group style

illogikal:
#### Description

Fixes issue with button-group button overrides where the border radius styles wouldn't get passed in if you set the style of the BaseButton. Code below to reproduce.

```javascript
import React from 'react';
import {ButtonGroup} from 'baseui/button-group';
import Upload from 'baseui/icon/upload.js';

export default () => (
  <React.Fragment>
    <ButtonGroup>
      <Button
        overrides={{
          BaseButton: {
            style: props => ({color: '#000'}),
          },
        }}
      >
        Label
      </Button>
      <Button
        overrides={{
          BaseButton: {
            style: props => ({color: '#000'}),
          },
        }}
      >
        Label
      </Button>
      <Button
        overrides={{
          BaseButton: {
            style: props => ({color: '#000'}),
          },
        }}
      >
        Label
      </Button>
    </ButtonGroup>
  </React.Fragment>
);
```

#### Scope

- [x] Patch: Bug Fix

### DIFF
--- a/documentation-site/pages/getting-started/comparison.mdx
+++ b/documentation-site/pages/getting-started/comparison.mdx
@@ -26,7 +26,7 @@ notice that, please let us know in a GitHub issue, or send a Pull Request fixing
 
 <Table columns={['Name', 'CSS Engine', 'Customization']} data={[
   ['Base Web', 'CSS-in-JS, using Styletron', 'Through an extensive overrides interface'],
-  ['Material UI', 'CSS-in-JS, using JSS', 'Through CSS overrides'],
+  ['Material-UI', 'CSS-in-JS, using JSS', 'Through CSS overrides'],
   ['Ant Design', 'Less', 'Through Less variables'],
   ['React Bootstrap', 'SCSS, inherited from Bootstrap', 'Through Bootstrap themes'],
   ['Semantic UI', 'Less', 'Through Less variables'],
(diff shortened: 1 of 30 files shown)


gergelyke:
Thanks for the contribution @illogikal ! One small comment - can we drop the new dependency? We are trying to keep the size as small as possible

illogikal:
We could but object-path is by far the best object deep key accessor in my opinion and it could probably be sprinkled around the codebase to reduce some LOC, but I haven't looked at everything. If you insist am happy to comply, but for use cases like where I used it, it's really convenient :)

gergelyke:
how about the following - let's merge this PR without the extra dependency, and after that we can take a look at the codebase if we should pull it in ?

illogikal:
Fixed

gergelyke:
it seems like there is an issue with type checking - can you run `npm test` locally and fix the issue?

chasestarr:
we may want to wait for @sandgraham's changes to the button styles. It should remove the extra override logic when mapping over `ButtonGroup` children

chasestarr:
thank you for the contribution @illogikal. I'm going to close this pr since the button redesign has been merged. we plan to make a release in the next couple days. Feel free to re-open if you are experiencing this problem once v7 is released

PR #1200, opened 2019-04-29
fix(docs): some a11y fixes

gergelyke:


gergelyke:
closes https://github.com/uber-web/baseui/issues/1166

except for the search input - that's controlled by a third party js from algolia, so I cannot change that

PR #1203, opened 2019-04-30
feat: add BaseProvider

nadiia:
#### Description

Adds a single wrapper that adds Layers and Theme providers.

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -13,10 +13,7 @@ addDecorator((story, context) => {
   return (
     <StyletronProvider value={engine}>
       <BaseProvider
-        themeProvider={{
-          theme:
-            context.kind === 'baseui-dark' ? DarkThemeMove : LightThemeMove,
-        }}
+        theme={context.kind === 'baseui-dark' ? DarkThemeMove : LightThemeMove}
       >
         {story()}
       </BaseProvider>
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -4,8 +4,7 @@ import {configure, addDecorator} from '@storybook/react';
 import {Provider as StyletronProvider} from 'styletron-react';
 import {Client as Styletron} from 'styletron-engine-atomic';
 
-import {LayersManager} from '../src/layer';
-import {ThemeProvider} from '../src/styles';
+import {BaseProvider} from '../src/index';
 import {LightThemeMove, DarkThemeMove} from '../src/themes';
 
 // Add providers for theme and styletron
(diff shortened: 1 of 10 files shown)


nadiia:
Would making the providers optional be beneficent?

.storybook/config.js
```diff
@@ -4,24 +4,22 @@ import {configure, addDecorator} from '@storybook/react';
...
-        <ThemeProvider
-          theme={
-            context.kind === 'baseui-dark' ? DarkThemeMove : LightThemeMove
-          }
-        >
-          {story()}
-        </ThemeProvider>
-      </LayersManager>
+      <BaseProvider
+        themeProvider={{
```
tajo: why not just 

```jsx
<BaseProvider theme={...} />
```

there are no other props that ThemeProvider accepts anyway?

tajo:
> Would making the providers optional be beneficent?

Like having `BaseProvider` without a theme?

.storybook/config.js
nadiia: In case we decide to extend the ThemeProvider 🤷‍♀ Or if there are more providers to be added that have more than one prop, so it's consistent.

nadiia:
> > Would making the providers optional be beneficent?
> 
> Like having `BaseProvider` without a theme?

Yeah, or without the LayersProvider. For example, we can add those based on props passed. In this case BaseProvider can be nested.

.storybook/config.js
chasestarr: I think the api to this should be designed so that consumers do not need to know that there are multiple 'providers' bundled into one. if there needs to be a new value passed through, we can add it to the top level props

chasestarr:
will nested layer managers negatively affect layout? what i'm trying to understand is - if I arbitrarily nest `BaseProviders` for a custom theme, but that have unexpected side-effects on layering?

.storybook/config.js
gergelyke: I agree, just using the top level `theme` prop seems like a cleaner API

gergelyke:
To play the devil's advocate - why do we need the BaseProvider in the first place? I'd be worried, that we'll introduce confusion on when to use `BaseProvider` vs  the `ThemeProvider` or the `LayerManager`

tajo:
> To play the devil's advocate - why do we need the BaseProvider in the first place? I'd be worried, that we'll introduce confusion on when to use `BaseProvider` vs the `ThemeProvider` or the `LayerManager`

I think the goal is to stop using the `ThemeProvider` and having a single Provider for both. We could also add `LocaleProvider`.

nadiia:
> To play the devil's advocate - why do we need the BaseProvider in the first place? I'd be worried, that we'll introduce confusion on when to use `BaseProvider` vs the `ThemeProvider` or the `LayerManager`

I would agree with it. But for the context, the idea was to reduce the number of wrappers customers need to wrap their app with and in case we intro a new provider we'd be able to add it to the BaseProvider without the need to codemod apps code.

PR #1210, opened 2019-05-01
feat(theme): theme docs respect OS theme preference

MarkReeder:
#### Description

In browsers that expose `prefers-color-scheme` this will match the docs theme to the OS theme. Currently supported browsers are Safari and Firefox:
https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme

base-web-respect-os-theme

The theme is set on initial page load and whenever the OS theme is changed. The theme toggle button continues to functional as it did previously.

#### Scope

- [ ] Patch: Bug Fix
- [✅] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/documentation-site/pages/_app.js
+++ b/documentation-site/pages/_app.js
@@ -52,7 +52,6 @@ export default class MyApp extends App {
     super(props);
     this.state = {
       theme: LightTheme,
-      matchMediaListeners: [],
     };
     this.mediaQueryListener = this.mediaQueryListener.bind(this);
   }
--- a/documentation-site/pages/_app.js
+++ b/documentation-site/pages/_app.js
@@ -34,6 +34,9 @@ const themes = {
   DarkThemeMove,
 };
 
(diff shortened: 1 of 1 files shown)


tajo:
I love the feature!

> The theme toggle button continues to functional as it did previously. 

If I use it to switch and refresh the page, the theme doesn't persist. I wonder if it should take priority over the OS settings.

MarkReeder:
@tajo your comment showed up when I hit the squash and merge button or I would have responded prior to merging.

I considered that, but it complicated the code and would have been potentially problematic if there was no way to clear the preference. The root of the question is what is the intended function of the theme toggle button? If it represents an override and did take precedence, clicking the toggle button once would both toggle the theme and lock you into a state where changing the OS theme would no longer change the docs site theme.

This does raise the issue that I have potentially broken theme persistence on all browsers, not just ones that support `prefers-color-scheme` I'll double check and submit a fix if that's the case.

MarkReeder:
@tajo fix in #1219

PR #1227, opened 2019-05-02
fix(progress-steps): fix the overlaping of the circle and vert line

nadiia:


chasestarr:
Screen Shot 2019-05-02 at 2 36 11 PM

Looks like it is overlapping in this case

PR #1229, opened 2019-05-03
Remove engines fields, too restrictive and not being used

tajo:
Node is used only for our build tools and we don't really keep track with Node releases anyway. 

Also, **right now you can't deploy next.js app with baseui as a serverless lamba** since their environment currently supports only Node 8.10 and we are enforcing 8.11, so the build fails.

For `yarn` and `npm`, we don't have practical reasons to be restrictive.

### DIFF
--- a/package.json
+++ b/package.json
@@ -158,11 +158,6 @@
     "react-dom": ">= 16.4.0 < 17",
     "styletron-react": "^4.4.6"
   },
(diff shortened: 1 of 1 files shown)


chasestarr:
I wonder if this is why the now deploy i set up silently fail 🤔

PR #1230, opened 2019-05-03
chore(project): fossa scan

xinding33:
#### Description

I'm from FOSSA and I'm working with the Uber OSPO to automate license scanning. In addition to the changes in this PR, we may also need to add an API key as an environmental variable in the build environment. I don't work with Buildkite frequently so I apologize if I'm making the incorrect changes to pipeline.yml. I'm essentially trying to install the FOSSA open source dependency discovery component to the build system and run it after the code has been built. Open to suggestions on how best to add the FOSSA tool to your existing CI/CD pipeline.

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -76,3 +76,10 @@ steps:
           - e2e-server-healthy
     agents:
       queue: builders
+    # Wait until all the previous tests have completed
+    - wait
+    # Run FOSSA to scan for OSS license compliance
+  - name: ':fossa-oss-scan:'
+    commands:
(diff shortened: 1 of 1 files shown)


.buildkite/pipeline.yml
```diff
@@ -76,3 +76,10 @@ steps:
           - e2e-server-healthy
     agents:
       queue: builders
+    # Wait until all the previous tests have completed
+    - wait
```
gergelyke:
```suggestion
    - # Wait until all the previous tests have completed
    - wait
    - # Run FOSSA to scan for OSS license compliance
```

gergelyke: you can remove the wait - everything runs in parallel

PR #1231, opened 2019-05-03
feat: accessibility addition suggestions for pagination

Martskin:
Accessibility addition suggestions for pagination

#### Description

Adds aria attributes to pagination controls for a11y. 

- `aria-haspopup="true"`
- `aria-expanded="false"`
- `aria-label="next page. current page 2 of 20"`
- `role="menu"`

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/src/pagination/pagination.js
+++ b/src/pagination/pagination.js
@@ -24,6 +24,7 @@ import ChevronLeft from '../icon/chevron-left.js';
 import ChevronRight from '../icon/chevron-right.js';
 import {getOverrides} from '../helpers/overrides.js';
 import type {PaginationPropsT, PaginationStateT} from './types.js';
+import type {LocaleT} from '../locale/types.js';
 import type {OnItemSelectFnT} from '../menu/types.js';
 
 type MenuItemT = {
@@ -101,7 +102,7 @@ export default class Pagination extends React.PureComponent<
     this.setState({isMenuOpen});
   };
 
-  constructAriaWayfinderLabel = (locale: object, prefix: string) => {
+  constructAriaWayfinderLabel = (locale: LocaleT, prefix: string) => {
     const {currentPage, numPages, labels} = this.props;
     return (
       prefix +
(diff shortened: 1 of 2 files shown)


src/pagination/pagination.js
```diff
@@ -175,33 +211,43 @@ export default class Pagination extends React.PureComponent<
...
-                      component: StyledDropdownMenu,
-                      // Access $style manually because it has gone through transformation
-                      // from the override helper function already
-                      // $FlowFixMe
-                      style: dropdownMenuProps.$style,
-                    },
-                  }}
-                  {...dropdownMenuProps}
-                />
+                <div role="menu">
```
chasestarr: The [dropdown menu](https://github.com/uber-web/baseui/blob/master/src/menu/menu.js#L31) should have a `listbox` role. Is `menu` more appropriate? If so, we try to allow consumers of baseui to customize all 'sub-components'. can the role be passed to `List` override?

```jsx
<DropdownMenu
  overrides={{
    List: {
      props: {role: 'menu'},
    },
  }}
/>
```

Martskin: Thanks @chasestarr !
Ahh, I see. Good suggestion.. I changed the role to an override and I added an override for the options as well to be `menuitem`.

src/pagination/pagination.js
```diff
@@ -97,6 +101,23 @@ export default class Pagination extends React.PureComponent<
     this.setState({isMenuOpen});
   };
 
+  constructAriaWayfinderLabel = (locale: object, prefix: string) => {
```
Martskin: @chasestarr 
I know I'm not doing this type checking correctly. Any suggestions?

chasestarr: `object` can be replaced with `LocaleT` from [here](https://github.com/uber-web/baseui/blob/26a8b9a38ef5621d5055fd185ca3a88b5a8804ca/src/locale/types.js)

Martskin:
My pleasure! Thanks for merging! @chasestarr

Issue #1109, opened 2019-04-03
Inconsistent use of is-prefix in boolean props


conradreuter:
Some of the components are inconsistent in the use of the `is`-prefix for boolean props.

Button
- isLoading
- isSelected
- disabled

Checkbox
- isIndeterminate
- checked
- disabled

Select
- isLoading
- clearable
- disabled
- error
- ... more

Tag
- isHovered
- isFocused
- closable
- disabled

Modal
- isOpen
- closable

Tooltip, Popover
- isOpen
- ignoreBoundary
- showArrow

I propose to drop the prefix for consistency.

conradreuter:
Not completely related to this ticket, but I will still squeeze it in: Would it be a big deal to rename `disabled` while still capturing the intent? I was thinking of `off`. but I don't like it that much.

PR #1234, opened 2019-05-04
chore(language) - Fix inconsistent boolean props

IgorWilbert:
For #1109

#### Description

Changes isLoading to loading in: select, button, table, pagination
Changes isSelected to selected in: button, button-group, datepicker, dnd-list, pagination, rating
Changes isIndeterminate to determinate in: checkbox
Changes isHovered to hovered in: checkbox, datepicker, radio,tag

#### Scope

- [X] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


IgorWilbert:
Hello! Instead of making consistency checks for each component, I changed all occurrences across the whole project, because I thought it would be favourable for future consistency. E. g. if we decide to make a new component with both subcomponents `checkbox` and `radio, and `checkbox` uses `isHovered` while `radio` uses `hovered`, we would still have the same problem.

gergelyke:
hi Igor! to me it seems like a huge breaking change, with minor benefits - I'd either leave things as is, or make sure we support both prop names

what do you all think @chasestarr @nadiia @tajo @sandgraham ?

chasestarr:
I'd personally rather not make this change right now since there will be many other breaking changes in the v7 release. we generally want to make the upgrade process as painless as possible. maybe this change can be saved for a later date?

gergelyke:
I think that makes perfect sense

gergelyke:
closing this for now, let's revisit later

PR #1249, opened 2019-05-07
docs(providers): use baseprovider in the docs

gergelyke:


documentation-site/pages/theming/custom-themes.mdx
```diff
@@ -35,14 +35,14 @@ All themeable values are available within components that provide the `overrides
 
 To learn more on how overrides work, check out [Better Reusable React Components with the Overrides Pattern](https://medium.com/@dschnr/better-reusable-react-components-with-the-overrides-pattern-9eca2339f646) article.
 
-## The `ThemeProvider`
+## The `BaseProvider`
 
-With the `ThemeProvider`, you can apply a theme to your application globally or locally. By default, Base Web ships with a light and a dark theme.
+With the `BaseProvider`, you can apply a theme to your application globally or locally. By default, Base Web ships with a light and a dark theme.
```
nadiia: How about changing it in a way that says `use BaseProvider to apply a theme globally and ThemeProvider to apply it locally`?

documentation-site/pages/theming/custom-themes.mdx
```diff
@@ -57,19 +57,19 @@ export default function App() {
 ```
 
 If you'd like to use a different theme in one part of your application, \
-you can locally wrap that part with the ThemeProvider to override the theme locally:
```
nadiia: In this part of the docs the `ThemeProvider` usage would be the correct recommendation.

documentation-site/pages/theming/custom-themes.mdx
gergelyke: is there a difference between the two? if not, I think we should tell folks to use the same thing everywhere to keep things simple

documentation-site/pages/theming/custom-themes.mdx
gergelyke: would it work with the BaseProvider too? if so, I think we should let folks know only one way to do it, so it won't be confusing

documentation-site/pages/theming/custom-themes.mdx
nadiia: The difference is that BaseProvider adds more Providers than a single Theme one. We don't want another LayersProvider to be added in a separate part of the app, unless it's intended.

gergelyke: oh, that makes sense - let me update this PR

gergelyke:
thanks for the comments @nadiia  - just updated the PR, please let me know what you think

Issue #150, opened 2018-08-24
Demo integrating form components with ReduxForm


arkadyp:
It would be super helpful to have a few canonical examples of how to integrate the form components with ReduxForm. 

Additionally, it may be worth considering adding some ReduxForm HOC to wrap the components:
```js
// Wrapped component
export function ReduxFormTextInput(props) {
  const {
    input: {onChange, value},
    placeholder,
    disabled,
    ...otherProps
  } = props;
  return (
    <TextInput
      onChange={onChange}
      value={value}
      placeholder={placeholder}
      disabled={disabled}
      {...otherProps}
    />
  );
}

// Example use:
<Field
  component={ReduxFormTextInput}
  placeholder="placeholder"
  name="fieldName"
/>
```
If BaseUI doesn't expose these wrappers, I think that each project that wants to integrate with ReduxForm will need to reimplement this boilerplate.

I wanted to kick off a discussion of where these examples/wrappers belong. Maybe it makes more sense to create a separate `baseui-reduxform` repo to prevent bleeding concerns?

kahwee:
I use `redux-form` a lot too and it will be nice to have a `baseui-redux-form` repo that exposes the render methods.

Warbring3r:
Warning: If onChange from redux-form as a HoC is naively passed through to a StatefulSelect component, the first time the StatefulSelect element is changed, the value will not be updated. If the redux-form onChange event is not passed through, the StatefulSelect works as expected. The current fix is to not pass the onChange event through from redux-form as a HoC to Base-UI, but I don't think that's ideal.

Warbring3r:
Here is how we got BaseUI and redux-form working nicely together.

First, the redux-form wrapper:
`
         <Field
              name="trailerType"
              label={'Trailer Type*'}
              component={ReduxFormSelectField}
              placeholder={'Select one'}
              required={true}
              options={trailerOptions}
              maxDropdownHeight={'200px'}
              validate={[validateTrailerType]}
              labelKey="label"
              valueKey="id"
              onChange={this.trailerChange}
              searchable={false}
              selectOrSearch={'select'}
              $style={{ width: inputWidth }}
              overrides={{
                DropdownContainer: {
                  style: {
                    width: '280px',
                  },
                },
              }}
            />
`

Then, the HoC around BaseUI:

`
// @flow
import React from 'react';
import { Select, TYPE, type OnChangeParamsT } from 'baseui/select';
...
[...]

PR #1252, opened 2019-05-08
fix(select): show placeholder when value is empty string

leyanlo:
#### Description

Properly show the Select placeholder when value is an empty string. This happens when integrating the Select component with ReduxForm, following the pattern described in #150.

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/src/select/select.js
+++ b/src/select/select.js
@@ -442,7 +442,7 @@ class Select extends React.Component<PropsT, SelectStateT> {
    */
   getValueArray(value: ValueT): Array<OptionT> {
     if (!Array.isArray(value)) {
-      if (value === null || value === undefined) return [];
+      if (value === '' || value === null || value === undefined) return [];
       value = [value];
     }
     return value.map(value => expandValue(value, this.props));


src/select/select.js
```diff
@@ -442,7 +442,7 @@ class Select extends React.Component<PropsT, SelectStateT> {
    */
   getValueArray(value: ValueT): Array<OptionT> {
     if (!Array.isArray(value)) {
-      if (value === null || value === undefined) return [];
+      if (value === '' || value === null || value === undefined) return [];
```
chasestarr: `value` is typed as an object, not a string. does flow not display an error?

leyanlo: Ugh, my bad it’s because of ReduxForm. They type `props.input.value` as `any`, which I was just passing along to Select:
> The value of this form field. It will be a boolean for checkboxes, and a string for all other input types. If there is no value in the Redux state for this field, **it will default to ''**. This is to ensure that the input is controlled. If you require that the value be of another type (e.g. Date or Number), you must provide initialValues to your form with the desired type of this field.
https://redux-form.com/8.2.0/docs/api/field.md/#-code-input-value-any-code-

I guess I will update my code to protect against this and only set the value when it’s not an empty string. Should I comment on #150 with this approach?

chasestarr: i think commenting there would be helpful. we generally don't want to enable interface options that contradict the component's "contract". if consuming code deviates, "your mileage may vary"

leyanlo:
Added a comment to #150 about this issue. It may be worth compiling some of these techniques on integrating with common form libraries, e.g. redux-form, formik, final-form, in our documentation..

PR #1260, opened 2019-05-08
feat(select): update search icon color based on state

leyanlo:
#### Description

Update search icon color based on state to match StyledControlContainer’s border-color: changes to negative400 on error and primary400 on focus.

Screen Shot 2019-05-08 at 4 38 06 PM

Screen Shot 2019-05-08 at 4 36 24 PM

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/src/select/styled-components.js
+++ b/src/select/styled-components.js
@@ -330,12 +330,18 @@ export const getLoadingIconStyles = (props: SharedStylePropsT) => {
 };
 
 export const StyledSearchIcon = styled('div', (props: SharedStylePropsT) => {
-  const {$disabled, $theme} = props;
+  const {$disabled, $error, $isFocused, $isPseudoFocused, $theme} = props;
   const {colors, sizing} = $theme;
   return {
     // $FlowFixMe
     ...getSvgStyles(props),
-    color: $disabled ? colors.inputTextDisabled : colors.foregroundAlt,
+    color: $disabled
+      ? colors.inputTextDisabled
+      : $error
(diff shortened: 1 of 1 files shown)


nadiia:
Looked through the Figma and it looks like the current spec is inconsistent. There are more style updates going to be introduced to all inputs in the v7. Maybe we should hold off on this change or sync with the designers on it.

gergelyke:
on second thought, should we merge this, to make sure we a fix for this in v6?

gergelyke:
hi @leyanlo  - we've touched based on this with design, and they prefer to keep it grey, as in the near future, we'll change base web to use less blue

leyanlo:
@nadiia @gergelyke Is there a way to do this with overrides today? I’m digging through the code, and it doesn’t seem possible to me unless I’m missing something. In [`renderSearch()`](https://github.com/uber-web/baseui/blob/master/src/select/select.js#L733), `SearchIcon` does not receive `sharedProps`, which has the `$isFocused` prop I need in order to do styling. Only `StyledSearchIcon` receives these props, but this component is not overridable.

Should I make another pull request exposing this `StyledSearchIcon` as a `SearchIconContainer` as an override?

gergelyke:
> Should I make another pull request exposing this `StyledSearchIcon` as `SearchIconContainer` as an override?

I think that makes perfect sense - @nadiia what do you think?

PR #1261, opened 2019-05-09
docs(typography): group components together and expand Caption2 example

leyanlo:
#### Description

Group components together and expand Caption2 example.

#### Scope

- [ ] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


nadiia:
The idea of the example was also to show the usage of available typography options and how to combine those. That's what an example from designers looks like =)

leyanlo:
Ohh I see it now! Oops 🤦‍♂️

PR #1262, opened 2019-05-09
feat(inputs): update to latest styles

sandgraham:
#### Description

Updates `Input`, `Select`, and `Menu` styles to match latest Base designs:

- Update rounded corners to square corners
- Update to thicker border width
- Update "active" (focus) state colors (active is now `black`)
- Update "negative" (error) state colors (lighter `red`)
- Update "disabled" state colors (slightly different `grey`s)
- Add positive state
- Add large sizes
- Enhancers now take on state of input (generally take on border color)

#### Existing inputs:

Screen Shot 2019-05-08 at 5 21 45 PM
Screen Shot 2019-05-08 at 5 22 31 PM

#### Existing select & menu:
Screen Shot 2019-05-10 at 10 20 08 AM

#### New inputs:

Screen Shot 2019-05-10 at 10 21 47 AM

#### Updated selects & menu:

Screen Shot 2019-05-10 at 10 18 52 AM
Screen Shot 2019-05-10 at 10 19 06 AM

#### Updated sizing for inputs (new large size):

Screen Shot 2019-05-10 at 10 23 09 AM

#### Scope

- [ ] Patch: Bug Fix
- [ ] Minor: New Feature
- [x] Major: Breaking Change


gergelyke:
this looks great! should we include the select component in this change too? if the input and the select are used in the same form, they look odd next to each other

sandgraham:
@gergelyke Yeah they seem like an odd pair. Let me update that too- need to get some design input since I don't think those have been mocked up yet.

src/input/styled-components.js
```diff
@@ -62,96 +30,232 @@ export const Root = styled('div', props => {
...
+      paddingRight: sizing.scale400,
+      paddingLeft: sizing.scale400,
+    },
+    [SIZE.default]: {
+      paddingRight: sizing.scale500,
+      paddingLeft: sizing.scale500,
+    },
+    [SIZE.large]: {
+      paddingRight: sizing.scale600,
+      paddingLeft: sizing.scale600,
```
sandgraham: @nadiia is something like this okay?

nadiia: Yes, perfect!

gergelyke:
it seems like this PR changes the side-navigation too (the elements are a lot higher) - is that intentional?

sandgraham:
I'm not sure why that has changed.
Here is where the background color is controlled:
```js
// src/side-navigation/styled-components.js
// L40
backgroundColor: $active
      ? `linear-gradient(0deg, rgba(255, 255, 255, 0.92), rgba(255, 255, 255, 0.92)), ${
          colors.primary
        }`
      : 'transparent',
```
I haven't changed `primary` or the `$active` prop so far as I know... 🤔
Will have to dig into this a bit.

sandgraham:
Looks like that was introduced here: https://github.com/uber-web/baseui/commit/c5fe7472e996c38ba77bd2f459b027f2e190ce0f#diff-97acb7bcb62a2cc242b1860993b8425fL44

`linear-gradient` can only be used where `<image>` can be used (in CSS). So the property needs to be `background` or `background-image`. In this case we are using 2 background layers (one to lighten our primary blue) and so we either use short hand`background` to apply both or split into `background-color` and `background-image` for the gradient if we require long hand forms.

gergelyke:
it's strange, as on the current master, it looks good 🤔

sandgraham:
oh i rebased onto `v7` - i included a quick fix in this branch

src/input/types.js
```diff
@@ -38,6 +38,8 @@ export type SharedPropsT = {
   $disabled: boolean,
   /** Renders UI in 'error' state */
   $error: boolean,
+  /** Renders UI in 'positive' state */
+  $positive: boolean,
```
tajo: what happens if I use both `$positive` and `$negative` at the same time?

tajo: it seems that `$error` wins, a warning could be nice?

sandgraham: Good point, I'll add the warning. Might be worth using a `STATE` enum or something if these get more complicated.

PR #1264, opened 2019-05-09
fix(docs): update base-web-white.svg to be consistent with base-web.svg

leyanlo:
#### Description

Update base-web-white.svg to be consistent with base-web.svg. Toggling dark mode does not cause the logo to jump around anymore, but only change color.

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


gergelyke:
Thank you @leyanlo !

PR #1266, opened 2019-05-09
fix(tether): print out a warning message if an anchor ref is not set

nadiia:


src/layer/tether.js
```diff
@@ -43,7 +43,14 @@ class Tether extends React.Component<TetherPropsT, TetherStateT> {
         this.popper && this.popper.scheduleUpdate();
       }
       if (this.state.isMounted !== prevState.isMounted) {
-        this.initializePopper();
+        if (!this.props.anchorRef) {
+          // eslint-disable-next-line no-console
+          console.warn(
+            `[baseui][TetherBehavior] ref has not been passed to the Popper's anchor element`,
```
chasestarr: this lgtm. would it be helpful to provide a link to docs + example describing the problem?

gergelyke: should change the wording a bit, so it offers some advice on how to fix it?

chasestarr: i also wonder if this should be limited to `__DEV__`

nadiia: Good point! Yes, just `__DEV__`.

PR #1267, opened 2019-05-09
docs: add debug mode

tajo:
**This PR needs to be merged after https://github.com/uber-web/baseui/pull/1241** since it relies on `styletron-react@5` and simplified `baseui/styled`. Not doing the bump in this PR to prevent the conflicts.

It adds debugging mode when deving. Arguably not that useful in our unique case, since almost all docs components come from baseui and are pre-compiled, so the source maps point to `node_modules` but still. 

debug


tajo:
> should we also add a few lines on debugging base web to the docs?

I will udpate styletron.org docs since it's styletron related

PR #1271, opened 2019-05-10
fix(icons): only import icons used by the component

gergelyke:


gergelyke:
fixes https://github.com/uber-web/baseui/pull/1271

src/accordion/panel.js
```diff
@@ -8,10 +8,8 @@ LICENSE file in the root directory of this source tree.
 import * as React from 'react';
 import {LocaleContext} from '../locale/index.js';
 import {getOverrides, mergeOverrides} from '../helpers/overrides.js';
-import {
-  Plus as PlusIcon,
-  CheckIndeterminate as CheckIndeterminateIcon,
-} from '../icon/index.js';
+import {PlusIcon} from '../icon/plus.js';
+import {CheckIndeterminateIcon} from '../icon/check-indeterminate.js';
```
nadiia: the two above have a default export, so no destructuring needed here

gergelyke: 👍

PR #1275, opened 2019-05-12
fix(docs): update react router sidenav example

aevarisak:
#### Description

Fixes wrong destructuring naming in the React Router side navigation example. This should be correct unless the example should rather be in the `onChange={({event: e, item}) => {` form.

#### Scope

- [ ] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


aevarisak:
I am btw unable to set the label on this as is said in your contributing rules

nadiia:
Thank you for the fix! @aevarisak

PR #1277, opened 2019-05-13
docs(legacy): deploy legacy docs sites

gergelyke:


scripts/build-docs.js
```diff
@@ -0,0 +1,168 @@
...
+
+Once it's finished, go to the Netlify team we have, and update the manual deployments, if needed.
+
+If you need to add a new major legacy version, just add it to 'majors' configuration object,
+create a new deployment in Netlify, and finally add a CNAME in Cloudflare.
+`);
+
+const majors = {
+  v1: {
+    buildCommaid: 'yarn build-storybook',
```
sandgraham: nit, `buildCommaid` -> `buildCommand`?

scripts/build-docs.js
```diff
@@ -0,0 +1,168 @@
...
+}
+
+function getLatestMajors(releases) {
+  return {
+    v1: releases.find(versionFilter('v1')),
+    v2: releases.find(versionFilter('v2')),
+    v3: releases.find(versionFilter('v3')),
+    v4: releases.find(versionFilter('v4')),
+    v5: releases.find(versionFilter('v5')),
+  };
```
sandgraham: could this use the config object? or should this stay separate?

scripts/build-docs.js
gergelyke: ah, yeah, thanks!

scripts/build-docs.js
gergelyke: that's a good catch, let me update

PR #1283, opened 2019-05-15
fix(toast): add tabIndex to <Toast/> body so it can be focused

jcready:
#### Description

Adds a `tabIndex` attribute to the `<Toast/>` body so it can be focused and allow the the `onFocus`/`onBlur` event handlers to work correctly.

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/src/toast/toast.js
+++ b/src/toast/toast.js
@@ -164,7 +164,7 @@ class Toast extends React.Component<ToastPropsT, ToastPrivateStateT> {
             {...sharedProps}
             {...bodyProps}
             // the properties below have to go after overrides
-            tabIndex={0}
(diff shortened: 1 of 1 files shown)


src/toast/toast.js
```diff
@@ -164,6 +164,7 @@ class Toast extends React.Component<ToastPropsT, ToastPrivateStateT> {
             {...sharedProps}
             {...bodyProps}
             // the properties below have to go after overrides
+            tabIndex={-1}
```
nadiia: It has to go before the overrides and shared props, not after.

PR #1286, opened 2019-05-15
feat(block): add backgroundColor prop to Block

### LINKED ISSUE #859: [block] Adding more props, e.g. backgroundColor
These are fairly common use cases for a couple of projects I'm working on.

Can we have the following props available for Block?

* `backgroundColor?: string`
  * Accepts all themeable color properties (`primary200`, etc.).
* `lineHeight?`: [available values](https://developer.mozilla.org/en-US/docs/Web/CSS/line-height)
* `fontWeight?`: [available values](https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight)
* `marginX?: string`
  * Accepts all themeable sizing properties (`scale100`, etc.).
* `marginY?: string`
  * Accepts all themeable sizing properties (`scale100`, etc.).
* `paddingX?: string`
  * Accepts all themeable sizing properties (`scale100`, etc.).
* `paddingY?: string`
  * Accepts all themeable sizing properties (`scale100`, etc.).

Where `X` is the `Left` and `Right` of `margin` / `padding`

Where `Y` is the `Top` and `Bottom` of `margin` / `padding`

leyanlo:
Fixes #859

#### Description

Add backgroundColor prop to Block, following the pattern of the color prop

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/src/block/__tests__/block.test.js
+++ b/src/block/__tests__/block.test.js
@@ -31,6 +31,20 @@ describe('Block', () => {
     });
   });
 
+  it('renders themed backgroundColor if provided', () => {
+    expect(
+      retrieveStyles(<Block backgroundColor="primary200">test</Block>),
+    ).toMatchObject({backgroundColor: '$theme.colors.primary200'});
+  });
+
+  it('renders provided backgroundColor if no matching themeable value is found', () => {
+    expect(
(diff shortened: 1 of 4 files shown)


gergelyke:
Thank you!

PR #1291, opened 2019-05-16
fix(menu): add aria-activedescendant

nadiia:
`aria-activedescendant` has to be an `id` of the active menu element. In the current PR I added a simple default id values for the menu items. 
I propose we don't add id values by default but put the functionality in place to handle it if there are ids provided for the items through `getRequiredItemProps` handler. It's due to the `id` attribute value has to be unique on a page, with SSR using `Math.random` or UUIDs will result in invalid checksum warnings.


nadiia:
Added an `Accessibility` section to the menu docs with a note regarding the `aria-activedescendant` attr.

PR #1294, opened 2019-05-17
fix(nav-item): add missing item prop and update doc site sidebar

sandgraham:
#### Description

Initial problem: The documentation site sidebar looks wrong.

Screen Shot 2019-05-17 at 10 34 38 AM

Turns out the sidebar was still using a now deleted `renderItem` property on the `Navigation` component to customize each item in the nav.

This PR fixes the issue by updating the sidebar to use the `NavItem` override on `Navigation`.

There is also a tangential fix to pass through a missing `item` property to the `NavItem`. This was already in the `NavItemPropsT` type but was not being passed through.

Screen Shot 2019-05-17 at 10 38 15 AM

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/src/side-navigation/nav-item.js
+++ b/src/side-navigation/nav-item.js
@@ -49,7 +49,7 @@ export default class NavItem extends React.Component<NavItemPropsT> {
             }
           : {})}
       >
-        <NavItem {...sharedProps} {...itemProps}>
+        <NavItem item={item} {...sharedProps} {...itemProps}>
           {item.title}
         </NavItem>
       </NavLink>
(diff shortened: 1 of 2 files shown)


src/side-navigation/nav-item.js
```diff
@@ -49,7 +49,7 @@ export default class NavItem extends React.Component<NavItemPropsT> {
             }
           : {})}
       >
-        <NavItem {...sharedProps} {...itemProps}>
+        <NavItem item={item} {...sharedProps} {...itemProps}>
```
nadiia: `NavItem` is a styled element by default so passing the `item` prop here will pass it onto the `StyledNavItem` as an invalid attribute that will also result in a warning from React.

PR #1296, opened 2019-05-17
feat: theme editor POC

jh3y:
#### Description
- integrates an inline theme editor for the base web documentation site
- user can edit a theme and see update applied to site globally
- user can copy to clipboard, import and export theme

#### Note
- This is a rough POC
- Code is far from perfect and any suggestions are welcome
- UX hasn't been truly considered yet so any ideas are appreciated
- Styling would be good bounce ideas on too

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/documentation-site/components/theme-editor.js
+++ b/documentation-site/components/theme-editor.js
@@ -1,0 +1,288 @@
+/*
+Copyright (c) 2018-2019 Uber Technologies, Inc.
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+
+/* eslint-disable flowtype/require-valid-file-annotation */
+/* eslint-env browser */
+
+import React, {Fragment, useEffect, useRef, useState} from 'react';
+import {
+  styled,
+  createTheme,
+  DarkTheme,
+  DarkThemeMove,
+  LightTheme,
+  LightThemeMove,
+} from 'baseui';
+import {FormControl} from 'baseui/form-control';
+import {StatefulInput as Input} from 'baseui/input';
+import {Button, KIND} from 'baseui/button';
+import {Block} from 'baseui/block';
+import {H1} from 'baseui/typography';
+import {Accordion, StatefulPanel as Panel} from 'baseui/accordion';
+
+const Container = styled('div', ({$theme, $open}) => ({
+  backgroundColor: $theme.colors.background,
+  filter: 'drop-shadow(0 0 10px #eee)',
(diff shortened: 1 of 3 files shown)


jh3y:
Yeah I have purposely left the UX a little rough with the hope that it would spark some discussion around how it could look/work.

You would prefer to have the `Copy` action removed then? I wasn't sure, I did kinda like the idea of hitting the button and then pasting it into an IDE.

Also, there's no icon on the tab and the styling is a little lacking. I imagine `Update` would be the main CTA. I could make that full width potentially.

chasestarr:
Copy paste gives me the text (with slashes) below, to me it did not feel like actionable data but may be better if the text rendering was improved. I've been talking to teams that are interested in taking advantage of this tool, and they hope that they can work with designers to get the details dialed in, and then be sent the custom theme. I'm expecting the primary audience to not be developers, but designers.

`\"breakpoints\":{\"small\":320,\"medium\":600,\"large\":1280},\...`


some more random thoughts below:

- Could the theme be updated automatically? It would be great to have this mirror a tool like figma or sketch as much as possible.
- The reset button should likely have a confirmation step before wiping out all of the information.
- It would be nice to have this tool be context aware. For example, if the site visitor was on the button page, it would surface the button variables more prominently.
- A large form is likely not the best way to display and edit this information. It would be ideal to see color previews and maybe color pickers? Could the data be organized so that there are less 'folders'? 


I made a quick mockup of a layout with some icons, it may be too complicated, but it's ideal to use icons because once users try the buttons for the first time the UI gets out of the way from creative work. The icons in the mockup are totally random btw

Screen Shot 2019-06-18 at 10 49 15 AM

chasestarr:
Here's another direction: https://element.eleme.io/#/en-US/theme/preview where the theme editor is an encapsulated 'page'

jh3y:
I've removed the "copy" functionality and updated.

There are some good ideas and as much collab/discussion as possible would be ideal. I would prefer icons. However, I haven't used icons as I assumed these may be provided by a designer at a later stage.

I made it a big form as that was the simplest thing. I didn't want to invest a lot of time into a solution that may be discarded.

The thing that needs to be factored in with a design is performance. You can do updating automatically. Perf takes a hit though unless you do something like mutate the theme object or debounce the input potentially. The theme object is large. Should the whole thing be editable? How?

gergelyke:
Closing this in favor of the yard theme editor

PR #1297, opened 2019-05-17
fix(select): add missing StyledIconsContainer export and fix override…

nadiia:


tajo:
Those overrides are not getting highlighted:

- DropdownOption
- Popover
- StatefulMenu

So maybe they don't work?

chasestarr:
^^ those are components with their own overrides right?

nadiia:
> ^^ those are components with their own overrides right?

Yes, these are the components. They don't get highlighted but since that's the part of the documentation that fleshes out the list of the overrides we should include it in here.

PR #1303, opened 2019-05-21
fix(datepicker): formatting

gergelyke:
Closes https://github.com/uber-web/baseui/issues/1301


chasestarr:
how are the flex-grid changes related?

leyanlo:
> how are the flex-grid changes related?

Ah looks like I forgot to update the name when I updated the files. Thanks for fixing these, though maybe would make sense in a separate PR..

PR #1305, opened 2019-05-22
fix(masked-input): avoid passing overrides to MaskOverride

leyanlo:
Avoid passing overrides to MaskOverride. These overrides do nothing here and show up in the DOM and cause React warnings.

Note that the snapshot `MaskedInput - basic functionality: Masked input has correct props 1` now matches `Input - basic functionality: input has correct props 1`.

Also:
* Still allows for `{Input: {style}}` overrides
* Export MaskedInputPropsT type
* Remove weak Flow type

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


gergelyke:
Thank you @leyanlo 🤗

PR #1308, opened 2019-05-22
chore: consistently use longhand border-radius properties

leyanlo:
Consistently use longhand border-radius properties so overrides avoid Styletron warnings

#### Scope

- [ ] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


leyanlo:
> please remove enumeration from the `documentation-site` directory, and this will lgtm

@chasestarr Would it be ok to keep the changes in the `documentation-site` directory? I understand that none of these are ever exported or reused so would never have Styletron warnings, but in my experience, I find it impossible to predict when some code will be copied and used improperly elsewhere. I would recommend keeping these consistent throughout the whole codebase including documentation if possible.

Let me know if this sounds ok; if there are any concerns I am missing, let me know and I’ll make the updates. Thanks!

chasestarr:
👍 sgtm

kirbydcool:
@chasestarr @leyanlo fyi this is broken in master. Longhand properties should be `borderTopLeftRadius` rather than `borderRadiusTopLeft`

PR #1310, opened 2019-05-22
fix(formcontrol): handling $error in caption

gergelyke:


documentation-site/static/examples/form-control/validation.js
```diff
@@ -0,0 +1,28 @@
...
+          onChange={event => {
+            if (Number(event.target.value) < 11) {
+              return setError(true);
+            }
+            setError(false);
+          }}
+          size={SIZE.compact}
+        />
+      </FormControl>
+    </React.Fragment>
```
chasestarr: I'm not convinced that shuttling error through `FormControl` to child is valuable since, it's just as simple to pass directly to the `Input`. Is there more to this that I'm missing?

```jsx
return (
  <FormControl label="Input label" caption="Enter a number bigger than 10">
    <Input
      onChange={event => {setValue(event.target.value)}}
      value={value}
      size={SIZE.compact}
      error={value && Number(value) < 11}
    />
  </FormControl>
);
```

gergelyke: yeah, you are right - I was just playing with the concept

gergelyke: @chasestarr updated the diff - as per the latest designs, both the caption and the input turns red in the error state

chasestarr: image for context
Screen Shot 2019-05-22 at 3 33 10 PM

should it follow a `kind` enum pattern like the input? a boolean can't handle 3 different states

gergelyke:
closing it for now, we'll pick it up once v7 is merged to the master

PR #1311, opened 2019-05-22
feat(heading): add component Heading

tajo:
This add a new component Heading (based on the [proposal](https://github.com/uber-web/baseui/blob/master/proposals/heading-component.md))

It also removes the documentation of `baseui/typography/heading` (silent deprecation for now).


documentation-site/static/examples/heading/basic.js
```diff
@@ -0,0 +1,24 @@
+import * as React from 'react';
+import {Heading, HeadingLevel} from 'baseui/heading';
+
+export default () => (
+  <HeadingLevel>
+    <Heading>Base Web [L1]</Heading>
```
gergelyke: should we extend this example with a paragraph of Lorem ipsum text? To make it super clear how it works

gergelyke:
if we are deprecating `baseui/typography/heading`, should we add warning messages that users should start using the new component?

tajo:
> if we are deprecating `baseui/typography/heading`, should we add warning messages that users should start using the new component?

I thought we could use even a slower approach for this and start with a silent deprecation aka removing the documentation, then a warning in v8 and removal in v9. It's a tiny component but it can be annoying to make the switch. Also, we already have a lot of changes coming with v7 so I don't want to add more pain points into it.

documentation-site/static/examples/heading/basic.js
tajo: Added.

tajo:
> lgtm - should we also add a screener test?

Added.

PR #1320, opened 2019-05-24
fix: updating Select/StyledValueContainer to fix text color in dark mode

DianaSuvorova:
#### Description
In dark mode color of the input blends with the background , so one can't see what is the value

iQ1hC9fWjT

#### Scope

- [X] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


DianaSuvorova:
uhm,  not sure why probot/pr-label  and screener are not happy. Any suggetions?

gergelyke:
Hi @DianaSuvorova - thanks a lot for the fix!

It seems like screener is unhappy because this commit changed the visual of the datepicker:

Screen Shot 2019-05-24 at 1 45 50 PM

can you take a look?

PR #1322, opened 2019-05-25
feat(timepicker): add prop to allow manual entry of times

dguenther:
#### Description

Our project wants to use a timepicker that provides steps in 15 minute increments, but allows for manual entry of times that may not exist as steps in the list. I added a `creatable` prop to the timepicker that adds corresponding options to the list once the user has entered a time that doesn't already exist.

If baseui isn't interested in merging this it's no problem, but I figured I'd share it upstream to get some feedback. The `stringToOptions` function feels particularly brittle to me, especially the regex, so I'm curious if there are any suggestions for alternate ways to parse the filter string into a time.

As a side effect of adding this, passing in a Date to `value` which doesn't correspond to one of the steps in the list now displays a time rather than a blank field -- this seemed like desirable behavior to me, but I could adjust if not. Also fixed a rounding issue in `secondsToHourMinute` by using a `Date` object to do the conversion, otherwise times would be off by a minute in some cases.

#### Demo

May-25-2019 00-56-15

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change


src/datepicker/timepicker.js
```diff
@@ -54,15 +52,73 @@ function secondsToLabel(seconds, format = '12') {
   return `${zeroPrefix(hours)}:${zeroPrefix(minutes)}`;
 }
 
+/**
+ * Converts a time string, e.g. 10:00, to one or more possible TimePicker
+ * options representing that time.
+ */
+function stringToOptions(str, format): Array<OptionT> {
+  // leading zero is optional, AM/PM is optional
+  const twelveHourRegex = /^(1[0-2]|0?[1-9]):([0-5][0-9]) ?([AaPp][Mm]?)?$/;
```
chasestarr: I'm not quite sure what's going on in the examples below, but it would be great if we can refine this regex to handle some of the strange cases below. It's not that bad, and would be happy to merge as-is if we can't figure it out.

Screen Shot 2019-05-29 at 9 17 42 AM
Screen Shot 2019-05-29 at 9 17 31 AM
Screen Shot 2019-05-29 at 9 17 10 AM

src/datepicker/types.js
```diff
@@ -257,25 +257,27 @@ export type StatefulDatepickerPropsT<T> = $Diff<
...
+  format: '12' | '24',
   /** Callback for when time selection changes. */
   onChange?: Date => mixed,
   overrides?: {
     Select?: OverrideT<*>,
   },
+  /** Set to true to allow times that aren't displayed in the options list to be entered manually. */
+  creatable: boolean,
   /** Amount of seconds between each option time. Defaults to 900 (15 minutes). */
-  step?: number,
```
chasestarr: Please maintain the optional prop here and in `format`, and similarly make `creatable` optional. I understand that it is defined in default props, but when publishing an api to consumers, it should reflect what they should care about.

src/datepicker/timepicker.js
```diff
@@ -54,15 +52,73 @@ function secondsToLabel(seconds, format = '12') {
...
+        hoursMinutes = [{hours: twentyFourHours, minutes}];
+      }
+      break;
+    }
+    case '24': {
+      hoursMinutes = [{hours, minutes}];
+      break;
+    }
+    default:
+      (format: empty);
```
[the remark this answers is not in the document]
dguenther: Asserting the expression's value as `empty` causes Flow to ensure the switch is exhaustive, i.e. there are no values that can fall through and hit the `default` case. I switched this to the format you suggested

src/datepicker/timepicker.js
dguenther: Hmm, I noticed this as well... These search behaviors seem to replicate on the existing timepicker as well, I just tested it on the doc site. Definitely something I'd like to figure out how to fix, but I don't think it relates to changes in this PR.

src/datepicker/types.js
dguenther: 👍 I changed these back to the original format, let me know if there are any additional changes I should make here

src/datepicker/timepicker.js
chasestarr: sounds good to me. we can look into this later

PR #1326, opened 2019-05-28
docs(styletron-codemod): adds instructions to run

chasestarr:
adds instructions to run the styled function `// $FlowFixMe` codemod

### DIFF
--- a/docs/migrations/01207.md
+++ b/docs/migrations/01207.md
@@ -70,8 +70,25 @@ const MyStyledComponent =
   });
 ```
 
-All usage of `$ref` has been removed. You can now move from `$ref` to `ref` and you will see a warning
-in the console describing the change.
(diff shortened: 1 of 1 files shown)


docs/migrations/01207.md
```diff
@@ -70,8 +70,25 @@ const MyStyledComponent =
...
+instructions below to run the codemod locally.
+
+```
+npm install -g jscodeshift
+jscodeshift -t node_modules/baseui/codemods/styled-flowfixme.js <transform-path>
+```
+
+Use the -d option for a dry-run and use -p to print the output for comparison.
+
+## $ref => ref
```
gergelyke: I think you have to use H3 here, as `## Styletron update to v5` is an H2

chasestarr:
```suggestion
### $ref => ref
```

PR #1331, opened 2019-05-29
feat(block): add textOverflow and whiteSpace props to Block

leyanlo:
#### Description

Add textOverflow and whiteSpace props to Block. These seem like they would be commonly used in the typography components, which are all based on Block.

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change


gergelyke:
@leyanlo can you help with the merge conflict resolution, please?

leyanlo:
Updated.

I’m wondering if it would make sense to order the properties somewhat logically, rather than chronologically. Maybe something like the order they use here:
https://github.com/constverum/stylelint-config-rational-order

Also would make it easier to see what we have and don’t have..

gergelyke:
I think that would make sense - can you open a PR with that?

PR #1334, opened 2019-05-29
chore: bump styletron deps

tajo:
Bumps styletron deps in order to support the new Chrome extension.


tajo:
this version of Styletron is busted

PR #1335, opened 2019-05-29
feat(phone-input): add a new phone input component

sandgraham:
#### Description

This PR adds a new Phone Input component to Base Web. It looks something like this:

Screen Shot 2019-06-04 at 9 58 31 AM

Screen Shot 2019-06-04 at 9 58 13 AM

As part of this PR I have also added `react-virtualized` as a peer dependency. It was required for a virtual list in the country selection dropdown.

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change


gergelyke:
it seems like this PR includes changes already on v7 branch- can you remove them?

sandgraham:
ok- rebased onto v7

documentation-site/static/examples/phone-input/controlled.js
```diff
@@ -0,0 +1,24 @@
...
+  return (
+    <PhoneInput
+      inputValue={input}
+      countryValue={country}
+      onInputChange={event => {
+        setInput(event.target.value);
+      }}
+      onCountryChange={event => {
+        setInput(
+          input.replace('+' + country.dialCode, '+' + event.option.dialCode),
```
gergelyke: may worth using template literals to be consistent with line 6, and the rest of the PR?

src/phone-input/country-select-dropdown.js
```diff
@@ -0,0 +1,106 @@
...
+import {getOverrides} from '../helpers/overrides.js';
+
+import type {CountrySelectDropdownPropsT} from './types.js';
+
+export default function CountrySelectDropdown(
+  props: CountrySelectDropdownPropsT,
+) {
+  const {
+    children,
+    dropdownHeight = '400px',
```
gergelyke: nit: may worth to move this into a constants file? like here https://github.com/uber-web/baseui/blob/master/src/button/constants.js

src/phone-input/phone-input.js
```diff
@@ -0,0 +1,61 @@
...
+
+import type {PropsT} from './types.js';
+
+export default function PhoneInput(props: PropsT) {
+  const {
+    inputValue,
+    onInputChange,
+    size = SIZE.default,
+    overrides = {
+      Input: {},
```
gergelyke: I think usually we only define an empty object here https://github.com/uber-web/baseui/blob/master/src/select/select.js#L817

src/phone-input/types.js
```diff
@@ -0,0 +1,134 @@
...
+  size?: SizeT,
+  /** Defines a maximum dropdown height. The edge of the viewport will shrink the dropdown accordingly. */
+  dropdownHeight?: string,
+  /** Defines a maximum dropdown width. The edge of the viewport will shrink the dropdown accordingly. */
+  dropdownWidth?: string,
+  overrides?: OverridesT,
+};
+
+export type StatefulPhoneInputContainerPropsT = {
+  children: PropsT => React$Node,
```
chasestarr: Please use `React.Node` syntax

src/phone-input/types.js
```diff
@@ -0,0 +1,134 @@
...
+  initialState: StateT,
+  stateReducer: StateReducerT,
+  onChange: (state: StateT) => mixed,
```
chasestarr: is `onChange` redundant of `stateReducer`?

src/phone-input/stateful-phone-input-container.js
```diff
@@ -0,0 +1,119 @@
...
+  StatefulPhoneInputContainerPropsT,
+  StateT,
+  StateReducerT,
+  StateChangeT,
+  StateChangePayloadT,
+  InputChangeEventT,
+  CountryChangeEventT,
+} from './types.js';
+
+const defaultStateReducer: StateReducerT = (state, type, payload) => {
```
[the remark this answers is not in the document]
chasestarr: I also worry about allowing consumers to replace this function. It's doing somewhat complicated actions. Can it be refactored to 
- first generate the next state using this logic
- pass the next state through the reducer (giving consumers a chance to change the next state)
- set state with the result of reducer

doing so gives consumers a way to customize changes in a way that does not require reimplemented the entire state transition

src/phone-input/styled-components.js
```diff
@@ -0,0 +1,60 @@
...
+import {StyledDropdownListItem} from '../select/index.js';
+
+export const StyledCountrySelectDropdownContainer = styled(
+  StyledList,
+  ({$height = '400px'}) => {
+    return {height: $height};
+  },
+);
+
+export const StyledCountrySelectDropdownListItem = styled(
```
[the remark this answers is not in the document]
chasestarr: Screen Shot 2019-05-30 at 9 06 48 AM

src/phone-input/flag.js
```diff
@@ -0,0 +1,22 @@
...
+import {SIZE} from './constants.js';
+import * as flagComponents from './flags/index.js';
+
+export default function Flag(props: {size: string, iso2: string}) {
+  const FlagComponent = flagComponents['Flag' + props.iso2.toUpperCase()];
+  const width = {
+    [SIZE.compact]: '22px',
+    [SIZE.default]: '30px',
+    [SIZE.large]: '37px',
+  }[props.size || SIZE.default];
```
chasestarr: small nit, but would appreciate being more clear rather than immediately accessing the property after initialization. suggest moving the accessor to L21

chasestarr: what were your thoughts behind in-lining string values rather than reading from theme? I worry about the flag icons being out of sync from alternate scales

src/phone-input/country-select-dropdown.js
```diff
@@ -0,0 +1,106 @@
...
+  const [NameColumn, nameColumnProps] = getOverrides(
+    overrides.CountrySelectDropdownNameColumn,
+    DefaultNameColumn,
+  );
+  const [Dialcode, dialcodeProps] = getOverrides(
+    overrides.CountrySelectDropdownDialcodeColumn,
+    DefaultDialcodeColumn,
+  );
+  return (
+    <Container $height={dropdownHeight} {...containerProps}>
```
chasestarr: Follow this [example](https://github.com/uber-web/baseui/pull/1207/files#diff-d2689559620ff39ce2a75901e58a7526R13) for how to get ref forwarding working. Right now, I'm unable to navigate the country menu with arrow keys

src/phone-input/country-select-dropdown.js
```diff
@@ -0,0 +1,106 @@
...
+                  <ListItem
+                    key={key}
+                    style={style}
+                    {...rest}
+                    {...listItemProps}
+                  >
+                    <FlagColumn {...flagColumnProps}>
+                      <Flag
+                        iso2={children[index].props.item.id}
+                        size={SIZE.compact}
```
chasestarr: is directly defining `SIZE.compact` intentional? should this change in coordination with the input size?

src/phone-input/country-select.js
```diff
@@ -0,0 +1,118 @@
...
+          };
+        }
+      },
+    },
+    ValueContainer: {
+      style: {
+        width: {
+          [SIZE.compact]: '34px',
+          [SIZE.default]: '42px',
+          [SIZE.large]: '50px',
```
chasestarr: similar comment to the `flags.js` file. why define the values here rather than in theme, and may be more readable if we don't immediately access the property on map

src/phone-input/country-select-dropdown.js
sandgraham: ended up pulling these out into constants because they were used in a few places

src/phone-input/types.js
```diff
@@ -0,0 +1,134 @@
...
+  onCountryChange: (event: CountryChangeEventT) => mixed,
+  /** A function that can be used to map iso codes to localized country names */
+  mapIsoToLabel?: mapIsoToLabelT,
+  /** Defines the size of the text input. */
+  size?: SizeT,
+  /** Defines a maximum dropdown height. The edge of the viewport will shrink the dropdown accordingly. */
+  dropdownHeight?: string,
+  /** Defines a maximum dropdown width. The edge of the viewport will shrink the dropdown accordingly. */
+  dropdownWidth?: string,
+  overrides?: OverridesT,
```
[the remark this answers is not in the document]
sandgraham: removed the over dry types

src/phone-input/types.js
sandgraham: i originally intended it to be a simplified way to listen to state changes, but i think i will update the reducer based on your other points below so this will probably be even more redundant at that point and i will remove it

src/phone-input/styled-components.js
```diff
@@ -0,0 +1,60 @@
...
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+// @flow
+
+import {styled} from '../styles/index.js';
+import {StyledList} from '../menu/index.js';
+import {StyledDropdownListItem} from '../select/index.js';
+
+export const StyledCountrySelectDropdownContainer = styled(
```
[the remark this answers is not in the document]
sandgraham: hm, these do pass. i don't have any types on these functions and flow seems fine with that 🤔

src/phone-input/stateful-phone-input-container.js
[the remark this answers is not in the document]
sandgraham: gotcha, this makes sense - i didn't really grok what the reducer pattern was doing until now

src/phone-input/flag.js
sandgraham: 1. gotcha - i thought i saw the immediate accessor pattern in a few places throughout codebase so i used it here, but happy to move out as needed

2. i think i forgot to move those out - makes sense to have in theme for sure

src/phone-input/country-select-dropdown.js
sandgraham: ah i think it is because i am using a `Select` and not a `StatefulSelect` - maybe i can switch those out.

src/phone-input/country-select-dropdown.js
sandgraham: does the normal select menu change in size based on top level `size` prop?

personally, i don't *think* the size of the input should change the size of the dropdown but that is purely my own intuition so should double check with design.

src/phone-input/flag.js
chasestarr: 🤷‍♂ not a big deal

src/phone-input/country-select-dropdown.js
chasestarr: that makes sense - given that the size never changes, should the various flag sizes be maintained?

chasestarr:
I was thinking about the peer dep on react-virtualized this morning. I believe it should be included in `dependencies` since it will be tree-shaken out if this component is not being used. Having a peer dep requires all consumers (even if they do not use the phone-input) to install the package

src/phone-input/country-select-dropdown.js
sandgraham: ah those are to match the input size itself:

phone input sizes

sandgraham: a dev could use overrides to make things larger but maybe i should add a prop for sizing the dropdown holistically

chasestarr: it lgtm for now, maybe we can allow further customization when needed

nadiia: They needed for the selected state I assume that would be different for different size inputs.

src/phone-input/country-select-dropdown.js
nadiia: Could we use Select's `maxDropdownHeight` props instead of passing the dropdown hight on the Container? The `maxDropdownHeight` is passed on the Menu's `List` element.

src/phone-input/country-select.js
```diff
@@ -0,0 +1,116 @@
...
+    DefaultSelect,
+  );
+  const selectOverrides = mergeOverrides(baseOverrides, overrides);
+  return (
+    <Select
+      size={size}
+      value={[countryValue]}
+      onChange={event => {
+        if (inputRef && inputRef.current) {
+          inputRef.current.focus();
```
nadiia: Could you add a comment on why we need to do `focus` on `change`?

src/phone-input/phone-input.js
```diff
@@ -0,0 +1,56 @@
...
+    inputValue,
+    onInputChange,
+    size = SIZE.default,
+    overrides = {},
+  } = props;
+  const inputRef = useRef(null);
+  const baseOverrides = {
+    Input: {
+      style: {
+        paddingLeft: '4px',
```
nadiia: value should probably come from the `$theme`

src/phone-input/stateful-phone-input-container.js
```diff
@@ -0,0 +1,113 @@
...
+
+const defaultStateReducer: StateReducerT = (type, nextState) => nextState;
+
+export default class StatefulPhoneInputContainer extends React.Component<
+  StatefulPhoneInputContainerPropsT,
+  StateT,
+> {
+  static defaultProps = {
+    initialState: {
+      inputValue: `+${getCountry('US').dialCode} `,
```
nadiia: Would be nice if we can make it dependant on the `countryValue` instead of requiring both to be passed in sync.

src/phone-input/stateful-phone-input-container.js
```diff
@@ -0,0 +1,113 @@
...
+      countryValue: getCountry('US'),
+    },
+    overrides: {
+      Input: {},
+      CountrySelect: {},
+      CountrySelectDropdown: {},
+      CountrySelectDropdownListItem: {},
+      CountrySelectDropdownFlagColumn: {},
+      CountrySelectDropdownNameColumn: {},
+      CountrySelectDropdownDialcodeColumn: {},
```
nadiia: Do we need to provide a default for every override? Could we just provide a default empty object for the `overrides`?

src/phone-input/stateful-phone-input-container.js
```diff
@@ -0,0 +1,113 @@
...
+    onInputChange: () => {},
+    onCountryChange: () => {},
+    stateReducer: defaultStateReducer,
+  };
+
+  state = {...this.props.initialState};
+
+  internalSetState = (type: StateChangeT, payload: StateChangePayloadT) => {
+    const nextState: StateT = this.internalStateReducer(type, payload);
+    this.setState(this.props.stateReducer(type, nextState));
```
nadiia: It has to pass a current state as a 3rd arg to the `stateReducer`.

src/phone-input/stateful-phone-input-container.js
```diff
@@ -0,0 +1,113 @@
...
+  };
+
+  internalStateReducer = (type: StateChangeT, payload: StateChangePayloadT) => {
+    switch (type) {
+      case STATE_CHANGE_TYPE.inputValueChange:
+        if (typeof payload === 'string') {
+          return {...this.state, inputValue: payload};
```
nadiia: We should be able to pass he only the value that is changed, so no need to spread the current state.

src/phone-input/stateful-phone-input-container.js
```diff
@@ -0,0 +1,113 @@
...
+        } else {
+          return this.state;
+        }
+      case STATE_CHANGE_TYPE.countryValueChange: {
+        if (typeof payload !== 'string') {
+          // Replace (if possible) the current country dialcode
+          let newInputValue = this.state.inputValue.replace(
+            `+${this.state.countryValue.dialCode}`,
```
nadiia: We should do a regex that replaces it only at the beginning of the string.

src/phone-input/stateful-phone-input-container.js
```diff
@@ -0,0 +1,113 @@
...
+            `+${payload.dialCode}`,
+          );
+          // If the replacement did nothing, just use the new dialcode
+          newInputValue =
+            this.state.inputValue === newInputValue
+              ? `+${payload.dialCode} `
```
nadiia: Will it result in dropping the phone number entered? Also, this says that the input value won't consistently contain a country code, which is not clear if intended.

src/phone-input/stateful-phone-input-container.js
```diff
@@ -0,0 +1,113 @@
...
+              : newInputValue;
+          return {
+            inputValue: newInputValue,
+            countryValue: payload,
+          };
+        } else {
+          return this.state;
```
nadiia: Maybe an empty obj if no changes are made? Same on the line 85.

src/phone-input/stateful-phone-input-container.js
```diff
@@ -0,0 +1,113 @@
...
+  onInputChange = (event: SyntheticInputEvent<HTMLInputElement>) => {
+    this.internalSetState(
+      STATE_CHANGE_TYPE.inputValueChange,
+      event.target.value,
+    );
+    this.props.onInputChange(event);
+  };
+
+  onCountryChange = (event: CountryChangeEventT) => {
+    if (event.option) {
```
nadiia: When will it not satisfy the condition?

src/phone-input/types.js
```diff
@@ -0,0 +1,138 @@
...
+  'aria-label'?: string,
+  /** Sets aria-labelledby attribute. */
+  'aria-labelledby'?: string,
+  /** Sets aria-describedby attribute. */
+  'aria-describedby'?: string,
+  initialState?: StateT,
+  stateReducer?: StateReducerT,
+  onInputChange?: (event: SyntheticInputEvent<HTMLInputElement>) => mixed,
+  onCountryChange?: (event: CountryChangeEventT) => mixed,
+  mapIsoToLabel?: mapIsoToLabelT,
```
nadiia: Could you add descriptions to all the props above? Same for the `StatefulPhoneInputContainerPropsT `.

nadiia:
A few comments on the component functionality:
- Input type should be set to `number`.
- Probably makes sense to make the country code in input non-clearable, or change the selected country based on the entered country code.
- When the country code is removed in the input and some phone number entered instead, then when trying to change the country selection it clears the input (clears already entered phone number).
- Input in a focused state has a black border, for the current realease, it should still be blue.

Overall looks great. Good job!

leyanlo:
> * Input type should be set to `number`.

Please don’t make the input type `number`. `number` will give you an input box with an up/down arrow, which is not how telephone numbers should work.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number

Also, on mobile, this will give you a full qwerty keyboard. You probably want input type set to `tel`:
http://mobileinputtypes.com/

src/phone-input/country-select-dropdown.js
sandgraham: 😕it seems like the container needs to have an explicit height set or nothing shows up.

src/phone-input/stateful-phone-input-container.js
sandgraham: i can update this so that if we cannot replace a recognizable dial code in the input string, we prepend it to the current input value instead.

so if the input value is `4882399` and you select `US` from dropdown it will prepend the country code: `+1 4882399`. (right now it would change the string to `+1 `)

Regarding the second point, i think this touches on the awkwardness around my decision to have the country code inserted into the input string, rather than being a separate un-editable part of the UI. It allows the user to delete the country code from the input.

originally, i had the state and presentation separate. i changed it because i wanted to let the user opt out of entering a country code, say if for some reason our dropdown did not cover their needs.

but now i think maybe i should have the state and presentation stay separate. in the case where a developer did not want to require the country code, i can add a prop `requireCountryCode` which when set to `false`, adds an `Other` option to the dropdown.

even then, why use this component if no country code is required? you would be better off just using your own text field and adding a mask if you did not require any country code logic.

src/phone-input/stateful-phone-input-container.js
sandgraham: ah this is for flow, the option property is optional in the original event type so even though this should always exist i need to check for existence so flow does not complain.

PR #1338, opened 2019-05-29
fix: fix input sizes, colors and rounded borders in dropdowns and popovers

nadiia:


src/themes/creator.js
```diff
@@ -497,14 +545,16 @@ export default function createTheme(
...
       scale500: '12px',
-      scale550: '14px',
-      scale600: '16px',
-      scale700: '20px',
-      scale800: '24px',
-      scale900: '32px',
-      scale1000: '40px',
-      scale1200: '48px',
-      scale1400: '56px',
+      scale600: '14px',
```
gergelyke: if we go down on this path, and change values for these props, we should probably write a migration guide, as folks depend on these values a lot for overrides / Block props

src/select/styled-components.js
```diff
@@ -309,7 +309,6 @@ export const StyledInput = styled<SharedStylePropsArgT>('input', props => {
   return {
     ...getFont($size, typography),
     boxSizing: 'content-box',
-    fontSize: '16px', // prevents iOS to zoom in when focused
```
chasestarr: should this be maintained so that ios does not zoom?

src/themes/creator.js
chasestarr: it may be straightforward to write a codemod for this.

- look for strings in files that import baseui block. 
- if strings match, replace with new scale title. (may have false-positives)
- search for baseui components with overrides prop. replace scales from the props function (this is likely tricky, but we could have high confidence with unit tests)

src/select/styled-components.js
gergelyke: oh, that's a really good catch, we should keep that

nadiia: The default input font size is 16px and for the large and compact ones it doesn't look good with the hardcoded 16px font.

PR #1339, opened 2019-05-29
fix(compatiblity): ie11 starts-with

gergelyke:
* adds a polyfill for the doc site
* removes usage from component code

### DIFF
--- a/src/datepicker/timepicker.js
+++ b/src/datepicker/timepicker.js
@@ -89,9 +89,8 @@ function stringToOptions(
           {hours: twelveHours + 12, minutes},
         ];
       } else {
-        const twentyFourHours = meridiem.toLowerCase().startsWith('a')
-          ? twelveHours
-          : twelveHours + 12;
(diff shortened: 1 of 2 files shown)


src/datepicker/timepicker.js
```diff
@@ -89,9 +89,8 @@ function stringToOptions(
           {hours: twelveHours + 12, minutes},
         ];
       } else {
-        const twentyFourHours = meridiem.toLowerCase().startsWith('a')
-          ? twelveHours
-          : twelveHours + 12;
+        const twentyFourHours =
+          meridiem.toLowerCase()[0] === 'a' ? twelveHours : twelveHours + 12;
```
leyanlo: Wouldn’t it be slightly faster to do the [0] before the toLowerCase()?

gergelyke: maybe - however, I believe v8 will optimize for the best performance, so I'd leave things as is

PR #1340, opened 2019-05-30
feat(documentation): adds new showcase example in README and docs

### LINKED ISSUE #1272: New showcase example on CodeSandbox
## Feature description

I think it would be interesting to have an editable example in README, like the ones from [Blueprint](https://codesandbox.io/s/nko3k41y60) and [Material](https://codesandbox.io/s/4j7m47vlm4) (however the Material one may be too simple). I tried to run the [example](https://github.com/tajo/fusion-baseui) by @tajo on codesandbox, but had no luck so far. Apparently, the dependencies, as they are today, are too complex for codesandbox, as can be seen [here](https://codesandbox.io/s/github/IgorWilbert/fusion-baseui). Does anyone else thinks that it would be interesting to have a slightly bigger example on README as some sort of showcase? If so, should we aim at making the example by @tajo run on codesandbox or incorporating it as a separate page inside the documentation? Or perhaps creating something new but more simple, just for codesandbox?  Thanks!

IgorWilbert:
Fixes #1272 

#### Description

Adds [this showcase example](https://codesandbox.io/s/patient-sky-nn8t7) in README and documentation

#### Scope

- [ ] Patch: Bug Fix
- [X] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,10 @@ export default function Hello() {
 }
 ```
 
+## Example
+
+An example of an application using Base Web can be found [here](https://github.com/tajo/fusion-baseui). You can also check how it works on [CodeSandbox](https://codesandbox.io/s/patient-sky-nn8t7).
+
 ## Docs
 
 To read the documentation, please visit [baseweb.design](https://baseweb.design).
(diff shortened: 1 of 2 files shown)


gergelyke:
thank you @IgorWilbert !

PR #1344, opened 2019-05-31
fix(select): left/right arrows should navigate input

chasestarr:
#### Description

Left/right arrow keys were being captured by the menu component. See integration test to reproduction steps.

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/src/menu/stateful-container.js
+++ b/src/menu/stateful-container.js
@@ -121,17 +121,18 @@ export default class MenuStatefulContainer extends React.Component<
 
   // Handler for arrow keys
   handleArrowKey = (event: KeyboardEvent) => {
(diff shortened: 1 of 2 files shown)


gergelyke:
Should this change be applied to https://5cf150903b0d0f00086aa9a5--baseui.netlify.com/components/datepicker/#timepicker too?

chasestarr:
hmm, it works as expected if you were to type something into there, then keyboard nav. the initial text rendered in the box is not applied to the `input` element at all 😉, and therefore un-editable. I can look into this change, but is a subtly different requirement

PR #1346, opened 2019-05-31
fix(menu): improves scroll item performance

chasestarr:
#### Description

Previously if the menu component had many items, it would re-render each item on highlight changes (keyboard/mouse events). This pr adds props diffing to prevent unnecessary re-renders on menu items. If a user holds the ArrowDown key, scrolling is much smoother, rather than being unresponsive. See before/after performance screenshots below. Notice the many small leaf renders in the before image. 'keydown' events previously took ~100ms each vs ~20ms now.

before
Screen Shot 2019-05-31 at 1 19 49 PM

after
Screen Shot 2019-05-31 at 1 18 38 PM

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


tajo:
Not completely related (happens in the master too) but just found out that for virtualized lists, arrows don't trigger scrolling so you can't really use keyboard only. Should we fix that?

src/menu/option-list.js
```diff
@@ -58,3 +58,13 @@ OptionList.defaultProps = {
...
+
+function compare(prevProps, nextProps) {
+  const highlighted = prevProps.$isHighlighted === nextProps.$isHighlighted;
+  const size = prevProps.size === nextProps.size;
+  return highlighted && size;
+}
+
+// requires updated flow parsing in the v7 branch
+// $FlowFixMe
+export default React.memo(OptionList, compare);
```
nadiia: We don't need the second param here since it does a shallow comparison by default.

chasestarr: equality checks on functions are unreliable - for example on event handlers :/

chasestarr:
this has bugs related to filtering list items, will look more into it. ty e2e tests :)

chasestarr:
I spent a decent amount of time writing a test that can be use to catch regressions to the work in this pr but could not figure out a robust method of measuring react render times during e2e tests

src/menu/stateful-container.js
```diff
@@ -195,21 +217,10 @@ export default class MenuStatefulContainer extends React.Component<
...
-      onMouseEnter: () => {
-        this.internalSetState(STATE_CHANGE_TYPES.mouseEnter, {
-          highlightedIndex: index,
-          activedescendantId,
-        });
-      },
+      // binds so that in-line functions can be avoided. this is to ensure
+      // referential equality when option-list compares props in memoized compoent
+      onClick: this.handleItemClick.bind(this, activedescendantId, index, item),
+      onMouseEnter: this.handleMouseEnter.bind(this, activedescendantId, index),
```
nadiia: I think this will still create a new bound function every time the `getRequiredItemProps` is called.

chasestarr: it maintains referential equality so that we can compare in memo function

PR #1350, opened 2019-05-31
docs: deemphasize version selector

tajo:
It's probably not used very often but it stands out a lot. Changed the KIND from `primary` to `minimal` and removed the chevron since that's [not a pattern](https://baseweb.design/components/popover/#popover-opens-on-click) in our design system.

## Before

Screen Shot 2019-05-31 at 3 10 10 PM
Screen Shot 2019-05-31 at 3 28 40 PM

## After

Screen Shot 2019-05-31 at 3 21 27 PM
Screen Shot 2019-05-31 at 3 23 08 PM


tajo:
> Looks good 👍 may be not obvious that it's a clickable select

Yea, but it still looks like a link / clickable target so I think that's fine. We actually don't have an official dropdown component yet.

PR #1361, opened 2019-06-04
fix(v7): addresses designer feedback

chasestarr:


src/button/constants.js
```diff
@@ -20,7 +20,9 @@ export const SHAPE = {
 };
 
 export const SIZE = {
+  medium: 'default',
```
nadiia: We shouldn't introduce a new value that just refrences the `default`. Also, it's ok to have different naming from the designers' ones. Considering they change it pretty often, introduce new or deprecate older ones, we won't be able to follow it without breaking changes.

src/button/constants.js
```diff
@@ -20,7 +20,9 @@ export const SHAPE = {
...
   default: 'default',
   compact: 'compact',
+  small: 'compact',
```
nadiia: Same here. Saying that I'm inclined toward not introducing this change and stick to the current naming.

src/button/constants.js
gergelyke: I think ideally we should use the same naming to reduce friction

chasestarr: it doesn't need to be included right now. so i'm going to remove so that this can land without deliberation

chasestarr:
```suggestion
```

src/button/constants.js
chasestarr:
```suggestion
```

PR #1364, opened 2019-06-04
feat(form-control): positive option

chasestarr:
Screen Shot 2019-06-04 at 11 12 08 AM

#### Scope

- [x] Minor: New Feature


src/form-control/__tests__/form-control.test.js
```diff
@@ -77,37 +82,85 @@ describe('FormControl - Label and Caption for controls', () => {
...
+      <FormControl error={error}>
+        <Input />
+      </FormControl>,
+    );
+    const errorRendered = rendered.find(Caption).first();
+    expect(error).toHaveBeenCalledWith(sharedProps);
+    expect(errorRendered).toHaveText('Error test');
+  });
+
+  test('Accepts a function for error', () => {
```
nadiia: nit: `a function for positive`

src/form-control/form-control.js
```diff
@@ -43,18 +49,36 @@ export default class FormControl extends React.Component<FormControlPropsT> {
...
     } = this.props;
 
     const onlyChildProps = React.Children.only(children).props;
-    const sharedProps = getSharedProps(onlyChildProps, STYLETRON_PROP_MAPPER);
-    sharedProps.$error = this.props.error || sharedProps.$error;
+
+    const sharedProps = {
+      $disabled: !!disabled,
+      $error: !!error,
+      $positive: !!positive,
```
nadiia: Seems like `required` and `size` are dropped here. We should pass it as part of shared props unless we intend to deprecate it.

src/form-control/__tests__/form-control.test.js
chasestarr:
```suggestion
  test('Accepts a function for positive', () => {
```

src/form-control/form-control.js
chasestarr: those two fields served no purpose before since it was [not accepted in props](https://github.com/uber-web/baseui/pull/1364/files/830598fc4099478830dfda8bfc877c0bcb933247#diff-ec657471cac72f23fd3c44aef25bb001R12) and had [no effect on styles](https://github.com/uber-web/baseui/blob/830598fc4099478830dfda8bfc877c0bcb933247/src/form-control/styled-components.js) I chose to remove so that the code was more clear

nadiia: I see, we don't provide `required` and `size` props on the form control component currently. Out of scope here, but may need to check with the design on requirements for it for different input sizes.

PR #1367, opened 2019-06-04
feat(phone-input): add new phone input component

sandgraham:
#### Description

This PR adds a new Phone Input component to Base Web. It looks something like this:

Screen Shot 2019-06-04 at 9 58 31 AM

Screen Shot 2019-06-04 at 9 58 13 AM

As part of this PR I have also added `react-virtualized` as a peer dependency. It was required for a virtual list in the country selection dropdown.

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change


sandgraham:
I had to open a new PR after probot killed the old one. I think i have addressed all the feedback from the original PR though.

Main things to look for now: I changed the state/interface a bit. I no longer update the input text when you select a country. The dial code is read only now, which simplifies setting the initial state and some of the inner logic fo the component.

src/phone-input/__tests__/phone-input.e2e.js
```diff
@@ -0,0 +1,96 @@
...
+    await page.click(selectors.phoneInputSelect);
+    // verify dropdown is open
+    await page.waitFor(selectors.phoneInputSelectDropdown);
+    // // verify US option is visible
+    await page.waitForSelector(countryListItemForIso(unitedStates.iso), {
+      visible: true,
+    });
+  });
+
+  it('allows a user to select a country from the dropdown, \
```
chasestarr: not blocking since it can be added as a follow-up, but it would be nice to have integration tests that ensure this component can be navigated using keyboard controls

nadiia:
Older PR is here #1335 .

src/phone-input/__tests__/phone-input.e2e.js
```diff
@@ -0,0 +1,96 @@
...
+const unitedKingdom = {iso: 'GB', dialCode: '+44'};
+
+describe('PhoneInput', () => {
+  beforeEach(async () => {
+    await mount(page, 'phone-input');
+    await page.waitFor(selectors.phoneInput);
+  });
+
+  it('passes basic a11y tests', async () => {
+    const accessibilityReport = await analyzeAccessibility(page);
```
nadiia: Need to wait for the component rendered, not page, to make sure accessibility check runs on the component.

src/phone-input/__tests__/phone-input.scenario.js
```diff
@@ -0,0 +1,20 @@
...
+import React from 'react';
+
+import {StatefulPhoneInput} from '../index.js';
+import {Block} from '../../block/index.js';
+
+export const name = 'phone-input';
+
+export const component = () => (
+  <Block>
+    <StatefulPhoneInput aria-label="stateful phone input example" />
```
nadiia: Is there a default `aria-label` provided?

src/phone-input/stateful-phone-input-container.js
```diff
@@ -0,0 +1,67 @@
...
+      text: '',
+      country: COUNTRIES.US,
+    },
+    onTextChange: () => {},
+    onCountryChange: () => {},
+    stateReducer: defaultStateReducer,
+    overrides: {},
+  };
+
+  state = {...this.props.initialState};
```
nadiia: If we do `state = {text: '', country: COUNTRIES.US, ...this.props.initialState}` then we won't require the both `text` and `country` always to be passed together.

src/phone-input/__tests__/phone-input.e2e.js
sandgraham: oh i moved that into the `beforeAll`

src/phone-input/__tests__/phone-input.scenario.js
sandgraham: yeah, it is `Please choose a country dial code and enter a phone number.`

src/phone-input/stateful-phone-input-container.js
sandgraham: ah good point, i will update

tajo:
It would be nice to set `autocomplete="tel"` by default (it's `on` now).

https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete

Ideally, it could also respect `tel-country-code` to select correct country.

sandgraham:
oh cool i will add that - didn't know about these

PR #1368, opened 2019-06-04
feat(input): adds input before/after examples

chasestarr:
Adds an example of an input with a 'clear' icon and positive/negative after icons.

Screen Shot 2019-06-04 at 2 26 56 PM

#### Scope

- [x] Minor: New Feature


documentation-site/static/examples/input/before-after.js
```diff
@@ -1,7 +1,43 @@
...
+    };
+  });
+
+  return (
+    <Input
+      onChange={event => setValue(event.target.value)}
+      value={value}
+      placeholder="Input with a clear button"
+      overrides={{
+        After: () => {
```
nadiia: Would be nice to move out and not inline the `After` component.

documentation-site/static/examples/input/before-after.js
```diff
@@ -27,5 +63,44 @@ export default () => (
...
     />
+    <Block as="br" />
+
+    <InputWithClear />
+    <Block as="br" />
+
+    <StatefulInput
+      error
+      overrides={{
+        After: () => (
```
nadiia: Same here.

tajo:
So we want to keep the "clear" functionality out of the input's core API?

chasestarr:
> So we want to keep the "clear" functionality out of the input's core API?

It may be something we want to incorporate in, but it feels straightforward to do like in this example

PR #1369, opened 2019-06-04
feat(payment-card): add component Payment Card

tajo:
payment-card

Originally, I thought it could be a sub-export of `baseui/input` but then I saw that `phone-input` is also a standalone component so following the same convention. It's also good for code-splitting (no need to rely on unreliable tree-shaking).

Also waiting for the final sign-off from @nguyensomniac that we can open-source the card icons but should be fine.


src/payment-card/payment-card.js
```diff
@@ -0,0 +1,124 @@
...
+  const number = input.replace(/[^0-9]/gi, '');
+  const validatedValue = valid.number(number);
+  if (validatedValue.card && Array.isArray(validatedValue.card.lengths)) {
+    return number.slice(
+      0,
+      validatedValue.card.lengths[validatedValue.card.lengths.length - 1],
+    );
+  }
+  // CC number NEVER can have more than 19 digits
+  return number.slice(0, 19);
```
gergelyke: can we limit the length based on the credit card type?

tajo: Yep, that's what happens right above! This is a fallback when the type is not detected (yet).

gergelyke: oh, I've overlooked it, great!

documentation-site/static/examples/payment-card/controlled.js
```diff
@@ -0,0 +1,37 @@
...
+          onChange={event => this.setState({value: event.target.value})}
+          placeholder="Enter payment card number"
+          value={this.state.value}
+          size={SIZE.default}
+        />
+        <Block as="br" />
+        <PaymentCard
+          onChange={event => this.setState({value: event.target.value})}
+          placeholder="Enter payment card number"
+          value={this.state.value}
```
nadiia: All 3 inputs share the value. Ok to showcase the sizing but may be confusing in the context of functionality.

src/payment-card/__tests__/payment-card.test.js
```diff
@@ -0,0 +1,29 @@
...
+  expect(sanitizeNumber(' 4111/ ')).toBe('4111');
+  expect(sanitizeNumber('kkk4-1-1-1abc')).toBe('4111');
+
+  expect(sanitizeNumber('5555555555554444 1234')).toBe('5555555555554444');
+  expect(sanitizeNumber('abc4111 1111 1111 1111 333 444')).toBe(
+    '4111111111111111333',
+  );
+});
+
+//asdasd 5555 5555 5555 4444 1111
```
nadiia: nit: remove commented out example

src/payment-card/types.js
```diff
@@ -0,0 +1,150 @@
...
+  ENHANCER_POSITION,
+} from '../input/constants.js';
+
+export type AdjoinedT = $Keys<typeof ADJOINED>;
+
+export type SizeT = $Keys<typeof SIZE>;
+
+export type StateTypeT = $Keys<typeof STATE_CHANGE_TYPE>;
+
+export type InternalStateT = {
```
nadiia: This seems to be not used.

src/payment-card/types.js
```diff
@@ -0,0 +1,150 @@
...
+import * as React from 'react';
+import type {OverrideT} from '../helpers/overrides.js';
+import {
+  STATE_CHANGE_TYPE,
+  ADJOINED,
+  SIZE,
+  ENHANCER_POSITION,
+} from '../input/constants.js';
+
+export type AdjoinedT = $Keys<typeof ADJOINED>;
```
nadiia: This seems to be not used.

tajo:
> Looks good! A question about type exports: do we need all of those that are same as in Input to be recreates and exported in PaymentCard component?

I cleaned up the `types` file. It now reuses all input types and only extends it with the `IconWrapper` override. That's actually the only different between `input` and `payment-card` APIs.

src/payment-card/index.js
```diff
@@ -0,0 +1,14 @@
...
+LICENSE file in the root directory of this source tree.
+*/
+// @flow
+export {default as StatefulPaymentCard} from './stateful-payment-card.js';
+export {default as StatefulContainer} from '../input/stateful-container.js';
+export {default as PaymentCard} from './payment-card.js';
+export {default as valid} from 'card-validator';
+export {SIZE} from '../input/index.js';
+// Styled elements
+export {IconWrapper as StyledIconWrapper} from './styled-components.js';
```
nadiia: Types export is missing.

src/payment-card/types.js
```diff
@@ -0,0 +1,29 @@
...
+
+import type {
+  InputComponentsT,
+  InputPropsT,
+  StInputPropsDiffT,
+} from '../input/types.js';
+import type {OverrideT} from '../helpers/overrides.js';
+
+export type PaymentCardComponentsT = {
+  ...InputComponentsT,
```
chasestarr: fyi spreads from inexact types will mark the 'spread-ed' fields as optional [try-flow](https://flow.org/try/#0C4TwDgpgBAogHgQwMbCgXigbwD5QGYD2BAXFAM7ABOAlgHYDmU2AvgNwBQ7okUAkrREQp0WfEVIUaDKG07doAZTCUICACbxkqDJigA6A5pQAaKACMElUrQCuAWzMRKMjvKhKV6-oK0jdBvW8hYFMLKyhbBycXTiQCWgooHxRSD1UNYL8w0gBGAAZZOITUOmTgVOV0oN8dbKh8tiA)

src/payment-card/payment-card.js
```diff
@@ -0,0 +1,124 @@
...
+        CardTypeToComponent[type || 'generic'] || GenericIcon,
+        {
+          size: iconSize[size],
+        },
+      )}
+    </IconWrapper>
+  );
+
+  return (
+    <Input
```
nudge: some extra attributes may be useful here:
`inputMode: 'numeric'`
`autoComplete: 'cc-number'`

`inputMode` hints to the browser to reveal a numeric keypad - useful for mobile clients
`autoComplete` hints to the browser this field supports card autofill

src/payment-card/payment-card.js
```diff
@@ -0,0 +1,124 @@
...
+      size={size}
+      data-baseweb="payment-card"
+      overrides={{
+        ...overrides,
+        Before: BeforeComponent,
+      }}
+      onChange={e => {
+        e.target.value = sanitizeNumber(e.target.value);
```
nudge: This will probably cause the cursor to jump to the end of the text input if the user tries to edit characters in the middle of the field.

src/payment-card/types.js
tajo: Interesting. Should be fine in this case, since all those props were already optional.

src/payment-card/payment-card.js
tajo: `autoComplete: 'cc-number'` was there already but I've added the `inputMode` as well

src/payment-card/payment-card.js
tajo: fixed, that was a fun one :)

chasestarr:
Could we get some e2e tests to validate the edge-cases we talked about the other day? Unit tests are great, but may not be representative of real usage

ex:
```
type 411122
validate expected number representation
press arrow-left twice
type 33
validate expected number representation 
```

tajo:
> Could we get some e2e tests to validate the edge-cases we talked about the other day? Unit tests are great, but may not be representative of real usage

Yep, added a suite of e2e tests!

PR #1377, opened 2019-06-05
fix(calendar): remove dayjs

gergelyke:


chasestarr:
could you write an integration test to validate the edge case reported? https://github.com/uber-web/baseui/issues/1301

gergelyke:
that's a good call - I can give it a shot

PR #1381, opened 2019-06-06
feat(a11y-checker): Utility to inspect components for violations at runtime

chasestarr:
#### Description

Screen Shot 2019-06-06 at 9 14 50 AM

Inspired by recent presentations by facebook developers. [[1](https://twitter.com/aweary/status/1124849839567659008), [2](https://twitter.com/jwyung/status/1124871810640531456)] It looks unclear whether or not it will be open-sourced, but does feel like a huge win for users of baseui. This is a very early prototype, simply validates that we can get this information from axe-core and display it to the page. No tests have been completed to measure performance impact (it may not scale well currently). 

See the image below for what fb's implementation looks like. Couple unique things to point out.
- The popover mentions a `BaseAccessibleElement` component that's likely fb specific. This indicates that the tool is not simply pulling data from axe, but have further customizations with context.
- The red overlay more appealing than a simple box around the element. This could mean two things; either their shared components have a state where they are 'in-violation' or they are doing some fancy imperative dom manipulation to get computed styles from the element. Probably the latter, but we could think bigger and build the feature set from baseui core

fb-a11y-checker

#### Implementation
- `CheckAlly` component is added as a wrapper around all children we would like to validate. This wraps the elements in a span and runs `axe-core` on that tree.
- Running the validation check in `requestIdleCallback` so that this action will operate when the browser is not performing more critical activity. This may not be necessary for an mvp, but was something I wanted to ensure was possible.
- `axe-core` returns an array of violations with selectors for the relevant nodes. For each node, we render a violation popover. It was a bit tricky to get this working with only a selector, but I realized that we can imperatively get the element and pass that to tether rather than going through react refs. Not sure if there is a better way to do this - open to suggestions!
- 'add input component' button at the top is to validate that this implementation works when dynamically adding elements with violations to the page (it does!).


gergelyke:
looks great! how this would be used in consumer's applications?

documentation-site/pages/ally.js
```diff
@@ -0,0 +1,203 @@
...
+        node.setAttribute(
+          'style',
+          `border: solid 1px ${theme.colors.negative300};`,
+        );
+
+        node.addEventListener('mouseenter', handleMouseEnter);
+        node.addEventListener('mouseleave', handleMouseLeave);
+      }
+
+      () => {
```
nadiia: Did you mean to return this function?

chasestarr:
> looks great! how this would be used in consumer's applications?

Wrap whatever dom tree you want to validate in the component. This could be installed in the `BaseProvider` component so that all apps have it, but that _may_ be too heavy-handed

gergelyke:
> > looks great! how this would be used in consumer's applications?
> 
> Wrap whatever dom tree you want to validate in the component. This could be installed in the `BaseProvider` component so that all apps have it, but that _may_ be too heavy-handed

So the next step here is to add a documentation page on it, on how to use it?

documentation-site/pages/ally.js
chasestarr: ty 🙏 
```suggestion
      return () => {
```

chasestarr:
we could release this as an unstable component and do some tests on real applications. it looks kind of bad right now, and likely has performance issues

gergelyke:
> we could release this as an unstable component and do some tests on real applications. it looks kind of bad right now, and likely has performance issues

that sounds great

chasestarr:
bikeshed 🚨🚨🚨alert

what should this be titled?

gergelyke:
> bikeshed 🚨🚨🚨alert
> 
> what should this be titled?

Should we simply call it `A11y`?

chasestarr:
updated this so that baseui exports a `Unstable_A11y` component from `baseui/a11y`

chasestarr:
added a docs page

PR #1385, opened 2019-06-09
docs: adds react-live to documentation website

### LINKED ISSUE #1273: Code "compilation" inside documentation
## Feature description
I was checking the [Semantic React UI library](https://react.semantic-ui.com/elements/button/#types-button) and I found it amazing how you can edit code and compile it immediately to see the changes in the components. Upon closer inspection, I found that apparently it is all custom-made to work like this, as can be seen [here](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/docs/src/components/ComponentDoc/ComponentExample/ComponentExample.js) and [here](https://github.com/Semantic-Org/Semantic-UI-React/blob/7da1dd029471f653bac513661d27ac5092ef523d/docs/src/examples/elements/Button/Types/index.js). Opening codesandbox every time one wants to edit code is quite annoying in my opinion, so do you think it would be interesting to implement something like this on Base in the near future? Thanks!

IgorWilbert:
Fixes #1273.

#### Description

This is a work in progress, to implement `react-live` in a similar way to [Styletron](https://www.styletron.org/)

#### Scope

- [ ] Patch: Bug Fix
- [X] Minor: New Feature
- [ ] Major: Breaking Change


IgorWilbert:
@tajo @gergelyke I would like to know your thoughts about this. I installed `react-live` and made a small test, it seems to work fine. I tried but I wasn't able to test it with a Base component yet because it requires the definition of a scope, which is a bit complex.

tajo:
Thanks @IgorWilbert!

Yea, this is a good approach! We use `react-live` at [styletron.org](https://www.styletron.org) as well. 

> I tried but I wasn't able to test it with a Base component yet because it requires the definition of a scope, which is a bit complex. 

Yep, this will require some manual work! You need to define some global scope (functions as `styled`) and then component scopes. We can't just put everything into a single scope since there are some naming conflicts for sure (like multiple components export `KIND` variable etc).

Also, the styling/syntax highlighting should look identical with current examples (including the dark mode) but that should be fairly straightforward.

So, do you feel like going through component examples and defining their scopes? :)

IgorWilbert:
Thanks @tajo ! Yes, I would like to work on the scopes. It just might take some time for me to do it right, hope that's ok for you :)

tajo:
Sure! Also check [this file](https://github.com/styletron/styletron.org/blob/master/components/code.js#L28-L31). You need to filter out imports / exports since react-live doesn't understand them. However, we want to keep displaying them to users so they can copy&paste the whole example.

IgorWilbert:
**Update:** sorry about the garbage code.

documentation-site/pages/components/avatar.js
```diff
@@ -0,0 +1,96 @@
...
+            $size={size}
+            {...avatarProps}
+          />
+        )}
+      </Root>
+    );
+  }
+}`;
+
+// const code = `
```
IgorWilbert: 1) I tried to pass a simple example as `code`. Not surprisingly, it does not compile because `Avatar is not defined`.

documentation-site/pages/components/code.js
```diff
@@ -0,0 +1,65 @@
...
+//   withStyle,
+//   withTransform,
+//   createStyled,
+//   useStyletron
+// } from "styletron-react";
+// import { driver, getInitialStyle } from "styletron-standard";
+// import { DESKTOP_BREAKPOINT } from "../const";
+
+// import {Avatar} from '../../../src/avatar/avatar';
+import Avatar from 'baseui/avatar';
```
IgorWilbert: 2) Since `Avatar` is undefined, I tried to pass it to `scope`. The import above returned `undefined`. `scope` is working when I pass something simple such as `headerProps` object for the example from the previous commit.

documentation-site/pages/components/code.js
```diff
@@ -0,0 +1,65 @@
...
+//   styled,
+//   withStyle,
+//   withTransform,
+//   createStyled,
+//   useStyletron
+// } from "styletron-react";
+// import { driver, getInitialStyle } from "styletron-standard";
+// import { DESKTOP_BREAKPOINT } from "../const";
+
+// import {Avatar} from '../../../src/avatar/avatar';
```
IgorWilbert: 3) This import caused an `Unexpected token **i**mport` error. The path above really goes to `src/avatar/avatar.js`. To solve this, I tried using `FILTERED` for the `scope`, in `transformScope`. It just threw a different `Unexpected token` error.

documentation-site/pages/components/avatar.js
```diff
@@ -0,0 +1,96 @@
...
+//           <li>Instagram</li>
+//           <li>WhatsApp</li>
+//           <li>Oculus</li>
+//         </ul>
+//       </div>
+//     );
+//   }
+// }`;
+
+const code = `export default class Avatar extends React.Component<PropsT, StateT> {
```
IgorWilbert: 4) Then I saw examples from `react-live` in [here](https://react-live.kitten.sh/). It seems that you can render components in `code`, as long as you pass the entire class inside `code` as well. This is not what we want, but I wanted to test it. It gives  `Unexpected token React.Component**<**PropsT, StateT>` as error.

documentation-site/pages/components/code.js
tajo: It should be 

```js
import {Avatar} from 'baseui/avatar';
```

documentation-site/pages/components/code.js
tajo: yea, you can't pass imports / exports to `react-live`, it has some [limitations](https://github.com/FormidableLabs/react-live#what-code-can-i-use)

documentation-site/pages/components/avatar.js
tajo: Right, you just need to pass it through the `scope` prop. Also, not sure if it can handle flow?

documentation-site/pages/components/avatar.js
tajo: Flow might not work, just pure JS.

documentation-site/pages/components/code.js
IgorWilbert: You are right, thanks. I have tried `import {Avatar} from 'baseui/avatar';`as well, has the same results.

IgorWilbert:
@tajo it turns out I just made some syntax mistakes, no need to worry about Flow for now. At the moment it looks like the image below:
sample
I would like to know how should I do this. Essentially, do you want me to make a `Code` component that takes care of filtering out the `import/export` (essentially a replacement for `Example` ), turn the code of the examples into strings (to be passed as `code`), import the components and the code examples into `mdx` files and pass both as parameters to `Code` component?

IgorWilbert:
**Update:**

- Used escape character to be able to use the original code of `sizes` example;
Update
- Isolated the code to create `react-live` components in a new `code` component;

tajo:
> * In which folder this new `code` component should be placed? Should it replace `documentation-site/components/code.js` ?

It could extend this component with `isEditable` prop. If it's `true` then render the `react-live`. It will be still useful to keep the non-editable code snippets around. 

> * Should the `scope` be defined for each example (like I just did) or it should be defined inside the code of `baseui` components?

That would work but maybe you can save a lot of boilerplate if we introduce **some hierarchy so the final scope is composed**:

1. **Top level**: `styled`, often used across examples when demonstrating overrides
2. **Component level**: A map of all imports for single component. 
3. **Specific example level**: It will combine the first two and that should be enough for most of cases. However: 
    - It can also add some one-offs. 
    - Add additional component scopes: This [example](https://github.com/uber-web/baseui/blob/master/documentation-site/static/examples/card/image-cta.js) of card also imports `Button`, so you need to add `Button` imports into the scope of this Card example.
    - "Manually" resolve possible (and rare) naming collisions. If two components are using an export with the same name (like the `KIND` constant) and both are used in the same example, we have to rename them. In other words, import as `import {KIND as ButtonKIND} from 'button'` would not match the predefine scope of Button.

> * What should be the style for this new editable `code` ? I have in mind something similar to [styletron](https://www.styletron.org/): A button to hide/show source code and an _editable_ label in the upper right corner.

I think we can preserve exactly same styles and UX, just adding the _editable_ label.

documentation-site/static/examples/avatar/sizes1.js
```diff
@@ -0,0 +1,26 @@
+import {Avatar} from 'baseui/avatar';
+import Code from '../../../pages/components/code';
+
+const scope = {Avatar};
+
+const code = `import * as React from 'react';
```
tajo: Ideally, we want to 100% reuse the existing examples from `documentation-site/static/examples/` and keeping them as .js files. You can load them through webpack like this:

```js
const code = require('!!raw-loader!./example.js');
```

IgorWilbert:
Hello @tajo ! Sorry for taking so long to work on this, I had some family issues, but now everything is fine. 

### Updates

- I successfully modified the `Code` component to receive `isEditable`, `code` and `scope` props that are passed to the `Example` component. However, I had some issues while fixing the UI. In a variation of `Code` component that uses `react-live`, I'd need instances of `LivePreview`, `LiveEditor` and `LiveError` inside a `LiveProvider` tag. The problem with this is that the rendered preview would be hidden inside the dropdown, but I wanted only the editable code to be there. After some thought, I realized it might be cleaner to implement it as a variation of `Example` component instead of `Code` so that I can "separate" `LivePreview` from `LiveEditor` and `LiveError` just like I needed (result below). Let me know if that works for you.

result

- Right now when I click on the `Hide Source` button the changes made to the `react-live` code are reset. Let me know if this behavior is unwanted so that I can fix it.
- If you don't want any immediate changes, I will start working on parsing the example files the way you mentioned and defining the `scope` for the examples.

tajo:
**Thanks for working on this!**

We are about to reorganize the examples a bit, add Flow and TypeScript options and also worki on a new "component playground" interface that will make the component live-editable (by code and API knobs).

With that in mind, it might not be necessary to have every single example live-editable as well. Could you hold off with this PR until we make these changes? Then we'll see how it fits together.

tajo:
An early prototype sneak-peak

playground

IgorWilbert:
No problem @tajo ! This "component playground" interface looks amazing! I just would like to know if you think adding `react-live` will be necessary at all after these changes. `react-live` might give you more freedom to experiment with code, but considering how people usually work with `base` on their projects, this might not be necessary (or even desired). The "component playground" would be easier to understand. Thanks!

gergelyke:
Closing this in favor of yard

tajo:
> react-live might give you more freedom to experiment with code

The playground uses react-live, so you can edit the code as well.

PR #1389, opened 2019-06-11
fix: refactor useRoundedCorners to be more discrete

sandgraham:
#### Description

This PR adds a few theme variables for customizing the border radii of components.

```
buttonBorderRadius
inputBorderRadius
popoverBorderRadius
surfaceBorderRadius
```

This frees some projects who still need rounded corners to more easily upgrade to `v7`. They can now tweak a few theme variables rather than having to set up a bunch of overrides.

##### `useRoundedCorners`

This PR does not remove `useRoundedCorners`! Despite the name being a bit misleading post-`v7`, some people might have it set to `false` and removing it outright would cause breaking style changes. We can remove it in a future major release and offer projects more discrete theme variables (in moderation) or overrides.

##### New variables

Each of the new variables affects multiple components. For instance, `inputBorderRadius` affects `Input`, `Select`, `Textarea`, `PhoneInput, and `PaymentCard`. I tried to group these into meaningful collections so as to reduce the amount of new variables and respect what components would likely change in unison. They may not be the right groupings and some iteration may be required.

##### `Table`

The `Table` component still has hard-coded rounded corners so as to avoid breaking style changes. In the future this could have a square version based on `useRoundedCorners` or derive from `cardBorderRadius` or another variable.

##### Migrating

People already on `v7` should not notice any change in styling. The values used in this PR are equivalent to the current default border radii.

People who want to migrate to `v7` from a previous version, while maintaining the rounded corner style, can now safely do so by setting their project's theme variables like so:

```js
const theme = {
  borders: {
      /** Checkbox, Datepicker (Range), Progress Bar, Slider, Tag */
      useRoundedCorners: true,
      /** Button, Button Group */
      buttonBorderRadius: '4px'
      /** Input, Select, Textarea */
      inputBorderRadius: '4px'
      /** Popover, Menu, Tooltip */
      popoverBorderRadius: '8px'
      /** Card, Datepicker, Modal, Toast, Notification */
      surfaceBorderRadius: '4px',
  },
};
```

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


src/datepicker/styled-components.js
```diff
@@ -184,50 +204,46 @@ function calculateBorderRadius(props): ?BorderRadiusT {
...
+          ? getBorderRadius(radius, 0)
+          : getBorderRadius(0, radius);
       } else {
-        if ($hasRangeSelected) {
-          return $startDate
-            ? getBorderRadius(borders.radius200, 0)
-            : getBorderRadius(0, borders.radius200);
+        if ($hasRangeHighlighted) {
+          return $hasRangeOnRight
+            ? getBorderRadius(radius, 0)
```
nadiia: This should not be the same as `datepickerInteractionBorderRadius`. In case we want to provide easy customization of the selected date corners it should be decoupled from the main look border radius. I'd rather not provide a theme override for the selected date corners radius considering we'll be updating it furthermore accordingly to the latest design.

src/styles/types.js
```diff
@@ -483,11 +483,25 @@ export type BordersT = {
...
+  datepickerBorderRadius: string,
+  datepickerInteractionBorderRadius: string,
+  inputBorderRadius: string,
+  menuBorderRadius: string,
+  toastBorderRadius: string,
+  progressBarBorderRadius: string,
+  sliderBorderRadius: string,
+  sliderThumbBorderRadius: string,
+  tableBorderRadius: string,
+  tagBorderRadius: string,
```
nadiia: What about popover corners?

sandgraham: Where are do the popover border radius styles get applied? They weren't in the popover or toast directories 🤔

sandgraham: Ah - the popover does not have border radius styles by default right? I added a `toastBorderRadius` for the toast though.

src/datepicker/styled-components.js
sandgraham: We can hardcode these (current), make another variable for them (`datepickerRangeBorderRadius`), or share the `datepickerInteractionBorderRadius` (which I think is equivalent to the current corner styling). Which one do you like?

nadiia: What do you think about leaving this part as is for now using `borders.radius200`?

src/styles/types.js
nadiia: It has a `boxShadow` that shapes the corners. So we need to add border radius for the rounded corners case.

sandgraham: I don't understand - how can the boxShadow shape the corners?

src/datepicker/styled-components.js
sandgraham: I prefer a new variable - It seems like we will need one eventually (assuming we can't couple this with the `datepickerInteractionBorderRadius`) and I'm not sure why we should wait. Will the overall `datepicker` design change so greatly the variable no longer makes sense?

Once it is time to square things off we can just change the default theme value and then people can opt out of those changes by tweaking their theme.

chasestarr: we should be judicious when exposing theme knobs to consumers. what is the core issue this pr is meant to fix? i see it as making it simple to set 'surfaces' and 'inputs' to have rounded corners (button, card, input, menu, toast). the rest does not really matter, and in the worst situation others can be easily added. to me, radii theme values other than the ones i enumerated above should be removed

nadiia: Look at the WIP datepicker in Figma. We didn't let the selected date to be modified by setting the `useRoundedCorners` and it's not something directly depends on rounded corners usage in UI.

src/styles/types.js
nadiia: Here is an example to help understand the issue: https://codesandbox.io/embed/popover-opens-on-click-4m55s

nadiia: An example above is just to showcase the discussed issue. A real world example would be `Menu` in a `Popover` when rounded corners set to be used.

src/datepicker/styled-components.js
sandgraham: How does this look then?
```
buttonBorderRadius,
inputBorderRadius,
cardBorderRadius,
menuBorderRadius,
popoverBorderRadius  // Do we need both popover and toast?
```
This means some corners will be hard coded- the developer will need to use `override` to customize.

chasestarr: this lgtm - agreed on the popover only

sandgraham:
@chasestarr @nadiia I updated the PR (the description as well) so that there are 4 new variables. Each one corresponds to a group of components that *likely* should change together. I also am no longer removing the `useRoundedCorners` variable- it will still apply to some things so as to avoid breaking styles. The description has more detailed info though so lemme know if this looks okay. 🙏

src/datepicker/styled-components.js
```diff
@@ -15,16 +15,15 @@ export const StyledRoot = styled<SharedStylePropsT>('div', props => {
...
     backgroundColor: colors.datepickerBackground,
     textAlign: 'center',
-    borderTopLeftRadius: borderRadius,
-    borderTopRightRadius: borderRadius,
-    borderBottomRightRadius: borderRadius,
-    borderBottomLeftRadius: borderRadius,
+    borderTopLeftRadius: borders.cardBorderRadius,
+    borderTopRightRadius: borders.cardBorderRadius,
+    borderBottomRightRadius: borders.cardBorderRadius,
+    borderBottomLeftRadius: borders.cardBorderRadius,
```
nadiia: Looking at this change I wonder what's the value of keeping the `useRoundedCorners` if we removed the usage of it?

src/menu/styled-components.js
```diff
@@ -27,6 +27,10 @@ export const StyledList = styled<StyledPropsT>('ul', ({$theme}) => {
     paddingBottom: $theme.sizing.scale300,
     paddingLeft: 0,
     paddingRight: 0,
+    borderTopLeftRadius: $theme.borders.popoverBorderRadius,
+    borderTopRightRadius: $theme.borders.popoverBorderRadius,
+    borderBottomRightRadius: $theme.borders.popoverBorderRadius,
+    borderBottomLeftRadius: $theme.borders.popoverBorderRadius,
```
nadiia: Feels strange to tie the menu to popover props since it's only a single use case of the menu component.

src/datepicker/styled-components.js
sandgraham: Ah, it does still get used in some places. The checkbox, datepicker range, progress bar, slider, and tag. I can't remove it outright without creating a breaking style change (either for people who have it true or those who have it false).

sandgraham: This PR does not fix the confusion around `useRoundedCorners` anymore. I think we will need to wait until a future major release to remove it outright.

src/menu/styled-components.js
sandgraham: Yeah, I did so because pre `v7` they both had border radius `8px` and it kind of made sense to me that menu was an instance of a popover. Are there cases where it is not a popover? If so I may need to make a new variable for it.

sandgraham:
@nadiia I agree it is a bit ambiguous how the variables affect the system. I added some comments to `creator.js` that maps them to various components but that is not a great solution.

Probably 75% of our components used `radius200` (`4px`) as the default in the past. In `v7` we basically changed the default to `0px`.

> Popovers used the larger `radius300` (`8px`) and in a couple edge cases (such as the Slider & Checkbox) we used `radius100` (`2px`). The `Tag` was the only component to use `radius400` (`12px`).

How about changing `cardBorderRadius` to `defaultBorderRadius`. Then it makes more sense for components like `datepicker` or `modal`.

src/themes/creator.js
```diff
@@ -572,7 +572,16 @@ export default function createTheme(
...
+      /** Checkbox, Datepicker (Range), Progress Bar, Slider, Tag */
       useRoundedCorners: true,
+      /** Button, ButtonGroup */
+      buttonBorderRadius: '0px',
+      /** Input, Select, Textarea */
+      inputBorderRadius: '0px',
+      /** Popover, Menu, Tooltip */
+      popoverBorderRadius: '0px',
+      /** Card, Datepicker, Modal, Toast, Notification */
+      defaultBorderRadius: '0px',
```
chasestarr: `surfaceBorderRadius`? Why does popover need to have a different value than card?

chasestarr: or, directly use a radius value rather than abstract through a semantic name

sandgraham: Popover, Menu, and Tooltip all used to have 8px corners, so for people coming from <=v6 they would need those grouped.

chasestarr: that makes sense, 'default' seems to convey the wrong concept - it's not as if this will change every usage of border radius

sandgraham: It should convey the basic shape/silhouette that most UI elements will have.

`baseBorderRadius`
`basicBorderRadius`
`commonBorderRadius`
`containerBorderRadius`

sandgraham: I didn't notice the `surfaceBorderRadius` suggestion - I think we can go with that. Not sure there is a perfect semantic grouping for these left over components.

Honestly we probably only need 2 variables: `surfaceBorderRadius` and `popoverBorderRadius` 🤔... but for now this will be fine - can revisit this in an upcoming major when we finish the design migration.

PR #1392, opened 2019-06-12
fix(dnd-list): prevent default to avoid double removals

MarkReeder:
#### Description

Fixes scenarios where `onChange` is improperly called multiple times.

#### Scope

- [X] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/src/dnd-list/list.js
+++ b/src/dnd-list/list.js
@@ -94,13 +94,14 @@ class StatelessList extends React.Component<ListPropsT> {
                 {removable && (
                   <CloseHandle
                     {...sharedProps}
-                    onClick={() =>
+                    onClick={evt => {
+                      evt.preventDefault();
                       onChange &&
-                      onChange({
-                        oldIndex: typeof index !== 'undefined' ? index : 0,
-                        newIndex: -1,
-                      })
-                    }
+                        onChange({
+                          oldIndex: typeof index !== 'undefined' ? index : 0,
+                          newIndex: -1,
+                        });
+                    }}
                     {...closeHandleProps}
                   >
                     <Delete size={24} color="#CCC" />


tajo:
When does the double removal happen?

MarkReeder:
> When does the double removal happen?

I attempted to create a reduced case to show this happening, but it wasn't straightforward to reproduce outside of my code. Here's a quick gif of it happening:

dnd-double-removal

Basically, I click once on the removal of the last item and the first item also gets removed.

tajo:
Do you use the latest `baseui`? 
Do you apply any overrides? 

I would be great identify the root cause before using `e.preventDefault` as a bandaid.

MarkReeder:
I updated my project in place to use the latest `baseui`, same issue. The only override I have on the `List` is:
```
overrides={{
  Root: {
    style: {
      overflow: 'auto',
    },
  },
}}
```

but removing that doesn't fix the issue.

MarkReeder:
Did a bit more digging, logged the `evt.nativeEvent` on the `onClick` and event is firing on both the close icon svg `path` element and the `button` element.

Here are the results of `console.log({evt}, evt.nativeEvent, evt.target, evt.target.tagName, evt.target === evt.currentTarget)`:
```
{evt: Class} MouseEvent {isTrusted: true, screenX: 2114, screenY: 903, clientX: 834, clientY: 769, …} <path fill-rule=​"evenodd" clip-rule=​"evenodd" d=​"M7.29289 7.29289C7.68342 6.90237 8.31658 6.90237 8.70711 7.29289L12 10.5858L15.2929 7.29289C15.6834 6.90237 16.3166 6.90237 16.7071 7.29289C17.0976 7.68342 17.0976 8.31658 16.7071 8.70711L13.4142 12L16.7071 15.2929C17.0976 15.6834 17.0976 16.3166 16.7071 16.7071C16.3166 17.0976 15.6834 17.0976 15.2929 16.7071L12 13.4142L8.70711 16.7071C8.31658 17.0976 7.68342 17.0976 7.29289 16.7071C6.90237 16.3166 6.90237 15.6834 7.29289 15.2929L10.5858 12L7.29289 8.70711C6.90237 8.31658 6.90237 7.68342 7.29289 7.29289Z">​</path>​ "path" false
{evt: Class} MouseEvent {isTrusted: true, screenX: 2114, screenY: 903, clientX: 834, clientY: 769, …} <button class=​"__debug-31 ax cj cl ap ck fd b8 de c4 cp ah aw">​…​</button>​ "BUTTON" true
```

This is happening in the latest version of React, `16.8.6`.

To add some info, I attempted a workaround where I compared `evt.target` and `evt.currentTarget` and only ran the `onChange` if they matched. This caused the `onChange` to never fire. My current understanding of the situation is that something in the re-render caused when `onChange` is fired causes the `onClick` to be triggered on both the children of the element where the event hander exists, as well as the element itself. This only happens in some situations, that I have not been able to completely narrow down, but my current assumption is that it has to do with number of elements being rendered or some other performance characteristic of the app.

MarkReeder:
Alternatively... What are the cases where `index` is legitimately `undefined`?
This default value is associated with the issue I'm seeing: `oldIndex: typeof index !== 'undefined' ? index : 0` because the 2nd time through, `index` is undefined and is defaulted to `0` and the first element is deleted.

Would you prefer this?
```
typeof index !== 'undefined' && onChange &&
  onChange({
    oldIndex: index,
    newIndex: -1,
  });
```

tajo:
> Did a bit more digging, logged the evt.nativeEvent on the onClick and event is firing on both the close icon svg path element and the button element.

That's just event bubbling, should be fine? 

> To add some info, I attempted a workaround where I compared evt.target and evt.currentTarget and only ran the onChange if they matched. This caused the onChange to never fire. 

Afaik, that makes sense too, since `target` is an element you've clicked on and `currentTarget` is an element that listener was attached to. In case of React, there is a single global listener attached to the React root?

> What are the cases where index is legitimately undefined?

I can't recall why I've added that `undefined` check there.

MarkReeder:
> > Did a bit more digging, logged the evt.nativeEvent on the onClick and event is firing on both the close icon svg path element and the button element.
> 
> That's just event bubbling, should be fine?
> 

Except that's causing the `onChange` to fire twice, the first time correctly, the second time with an `undefined` `index`. The `undefined` `index` is what's causing the first item to be deleted because of the defaulting of the `oldIndex` to `0` when `index` is `undefined`. This is why I proposed an alternative of just not firing `onChange` when `index` is `undefined.

> > To add some info, I attempted a workaround where I compared evt.target and evt.currentTarget and only ran the onChange if they matched. This caused the onChange to never fire.
> 
> Afaik, that makes sense too, since `target` is an element you've clicked on and `currentTarget` is an element that listener was attached to. In case of React, there is a single global listener attached to the React root?
> 

Agreed- it makes sense, and helps narrow down the issue to the re-render of the `List` that causes the second `onChange` to fire.

> > What are the cases where index is legitimately undefined?
> 
> I can't recall why I've added that `undefined` check there.

If index were just passed through unchanged, I could handle not deleting anything in my code's `onChange`, or we could just not call `onChange` at all...

nadiia:
The fix with the `index` value check seems to be reasonable. @tajo what do you think?

MarkReeder:
Just updated my code locally to confirm that `index` is `undefined` and I must have made an incorrect assumption previously because the second time it's fired, the value of `index` is `0`. Unfortunately, that approach won't actually work.

tajo:
> The fix with the `index` value check seems to be reasonable. @tajo what do you think?

I need to dig deeper what's up with the `undefined` check first. I've added it as a fix for some bug,  piggybacked it to some other PR, didn't write a good commit message and now I don't remember what it was. 😅

tajo:
So it's caused by `<label>` that wraps the whole `dnd-list`. It seems that clicking on a label fires an additional `event` that gets captured by the inner `X` `button`. Also, react-movable defaults to `index = 0` if `e.target` is not a valid list item. 

Although, not completely clear on the details why browsers do this, `e.preventDefault()` actually seems like a reasonable fix in this case so merging this PR.

PR #1400, opened 2019-06-14
feat(aspect-ratio-box): add aspect-ratio-box utility component

leyanlo:
#### Description

Add AspectRatioBox component, which allows you to create a container with a given aspect ratio. Useful for wrapping images with known aspect ratios from causing reflows as images load, or building UIs with using squares of unknown width.

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change


leyanlo:
On second thought, maybe this would be better suited to follow the Card pattern, where the StyledBody is just used outright, rather than passed in as an override. Let me try this again.

leyanlo:
Nevermind, it seems that the Card component documentation is misleading! All the Card examples follow a pattern with styled components as children, but which end up with a redundant StyledBody in the DOM!

Implemented AspectRatioBox using Blocks instead in #1414, to take advantage of Block’s styling props and simplify usage. Maybe will send another PR at some point updating the Card component documentation to reduce redundancy..

PR #1403, opened 2019-06-14
fix(link): links inherit font size of the parent

### LINKED ISSUE #1383: Links should inherit font size
Links should inherit the text size and line height of their container.

## Current Behavior

Links are always a certain font size. This makes them unusable in contexts like table cells where the font size is smaller than default page content.

## Expected Behavior

Links should inherit the text size and line height of their container.

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v7.0.0  |
| React   | 16.9.0-alpha.0 |
| browser | All |

qqingvarqq:
Fixes #1383 

#### Description
Links CSS Properties `font-size, line-height `, inherit values from the parent container.
#### Scope

- [x] Patch: Bug Fix


gergelyke:
Thanks for the PR @qqingvarqq ! One small thing that the VRT suite brought up is the change of font size in the card component for the link, compared to the master. Can you take a look at that please?

qqingvarqq:
Of course
I will take a look on this.

qqingvarqq:
So, a few things.
Regression occurs, because links inherit `font-size` and `line-height` properties, and in `card-image-link` story, the link is placed inside Action component which has a different font size.
The same I noticed in the `link` story and in the documentation page (link page).
now it's fixed.
Probably this PR should be marked as breaking changes because previously links always have static size, and now font size inherits from the parent.
And should I add story for links where links have different font sizes depending on context?

chasestarr:
I believe inheriting is going to be strictly better than having a set font-size. Any consumers that explicitly defined font-sizes will be maintained. Based on our versioning policy, css changes do not require a major bump. Do you agree @gergelyke? I'm fine merging this change as-is

gergelyke:
> I believe inheriting is going to be strictly better than having a set font-size. Any consumers that explicitly defined font-sizes will be maintained. Based on our versioning policy, css changes do not require a major bump. Do you agree @gergelyke? I'm fine merging this change as-is

makes sense 👍

Issue #1236, opened 2019-05-06
[Accessibility] ARIA Improvement for Modal component


paschalidi:
Hello 👋 🌸 everybody! Thanks for the amazing work you are going here and thanks for giving us such a nice project to look up to! 🏆 

As currently the modal functions, there might be some improvements to be made on the accessibility side. 
[I take w3 accessibility practises as guide](https://www.w3.org/TR/wai-aria-practices/#dialog_modal). Following the link we can see that in the section "Keyboard Interaction" are some points that can still be covered.

## Current Behavior

1. When a dialog opens, focus is not on an actionable element inside the dialog. 
2. Pressing `Tab` button when focus is on the last tabbable element inside the dialog, moves focus outside of the dialog.

## Expected Behavior
1. If focus is on the last tabbable element inside the dialog, moves focus to the first tabbable element inside the dialog. 
2. Pressing `Tab` button when a dialog opens, focus moves to an element inside the dialog. See notes below regarding initial focus placement.

Thanks for the amazing job people! Let me know if this makes sense!

gergelyke:
thanks for the kind words, and for opening this issue @paschalidi !

I think it definitely makes sense - would you mind chiming in @schnerd ?

schnerd:
@paschalidi Thanks for reporting this! We're definitely striving to follow w3 accessibility practices and it's possible that we missed out on a couple things.

> When a dialog opens, focus is not on an actionable element inside the dialog.

As of right now we shift focus to the dialog container element itself when it opens (https://github.com/uber-web/baseui/blob/master/src/modal/modal.js#L199). It's difficult for us to automatically shift focus to the "first tabbable element" because that may be an `input` or some other element rendered into the `Modal` body by users of Base UI. Due to this complexity we decided to just focus the dialog container itself and then let the user implement focus on their own if they need to.

That said, to may be possible to use document.querySelector to find the first focusable element in the modal [[1](https://gomakethings.com/how-to-get-the-first-and-last-focusable-elements-in-the-dom/)] and trigger a focus event on it. Any other alternative approaches you can recommend?

One other thing I noticed when testing this is that our current DOM structure makes the Close (X) button the first tabbable item in the dialog. That should maybe be fixed.

> Pressing Tab button when focus is on the last tabbable element inside the dialog, moves focus outside of the dialog.

This is a good catch – we probably need some logic that listens for focus/blur events while a modal is open and then loops focus back around to the first focusable element (related to strategy above).

[...]

PR #1406, opened 2019-06-14
feat(modal): a11y enhancements (focus order & trapping)

sandgraham:
#### Description

This PR adds a few ARIA a11y enhancements to the Modal component as suggested in issue #1236. 

**Focus trapping**: This means as a user tabs through the Modal, focus will never leave the Modal.

**Focus first interactive element**: When `autofocus` is enabled, the Modal will focus the first focusable element in the Modal. If no interactive elements are available or `autofocus={false}`, the Modal container will be focused.

tabbable

**New `FocusOnce` utility**: A small wrapper component that allows a developer to make an element focus-able for just one focus event. This is useful for the following scenario:
> If content is large enough that focusing the first interactive element could cause the beginning of content to scroll out of view, it is advisable to add tabindex=-1 to a static element at the top of the dialog, such as the dialog title or first paragraph, and initially focus that element.

*[From WAI-ARIA Authoring Practices 1.1](https://www.w3.org/TR/wai-aria-practices/#dialog_modal)*

**Modal close button**: I moved the Modal's close button to be last in the tab order, otherwise it would always be selected first. I also added a focus state to the button so that there would be some visual indication of focus.

Screen Shot 2019-06-14 at 1 26 32 PM

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change

#### Todo

- [x] Add documentation for `FocusOnce`
- [x] Update documentation for `Modal` a11y concerns
- [x] Add gifs to PR description


gergelyke:
works really nice! 👏

src/table/filter.js
```diff
@@ -75,31 +75,33 @@ export default function Filter(props: FilterProps) {
...
-              onClick={() => {
-                onReset();
-              }}
-            >
-              Reset
-            </Button>
-          </Footer>
-        </TrapFocus>
+        <FocusTrap>
+          <div>
```
nadiia: What is this `div` added? Any element added has to be not overridable.

sandgraham: `FocusTrap` requires a single node be passed and cannot be a Fragment so I wrapped this in a div. The older implementation (`TrapFocus`) would wrapper whatever was inside it in an extra div so I figured it was equivalent here.

sandgraham:
@gergelyke I will need screener approval - the modal now focuses the first interactive element so there are a few visual diffs there.

gergelyke:
> @gergelyke I will need screener approval when you can- the modal now focuses the first interactive element so there are a few visual diffs there.

done

sandgraham:
@chasestarr I don't feel strongly about `Modal` owning `FocusOnce`. I thought about pulling it out, but didn't have another use case in mind.

---

So the a11y recommendation here is to focus the first element in the modal if nothing below it is focusable or if the first interactive element will cause a scroll/jump on focus.

I suppose we could build that in but it adds some ugliness to the Modal container logic for what seems like a minor edge case. I figured giving folks an easy way to address these cases with `FocusOnce` was a nice compromise. That said, I also understand having Modal provide all this by default is a nice enforcement of the a11y recommendations.

I am leaning towards keeping it separate for now but not 10/10 on this.

PR #1409, opened 2019-06-14
fix(table): proper STYLETRON context propagation for the table element

gergelyke:


src/table/index.js
```diff
@@ -18,6 +18,7 @@ import {StyledTable as Table} from './styled-components.js';
 export const StyledTable = props => (
   <Table data-baseweb="table-custom" {...props} />
 );
+StyledTable.__STYLETRON__ = Table.__STYLETRON__;
```
nadiia: Lets mobe the data-baseweb attr to the Table in `styled-components.js` file, so we won't need this additional wrapper.

PR #1414, opened 2019-06-18
feat(aspect-ratio-box): add aspect-ratio-box utility component

leyanlo:
#### Description

Add AspectRatioBox component, which allows you to create a container with a given aspect ratio. Useful for wrapping images with known aspect ratios from causing reflows as images load, or building UIs with using squares of unknown width.

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change


koulmomo:
We should add a link to: https://css-tricks.com/aspect-ratio-boxes/#article-header-id-6 or inline the logic for **why** this works as a code comment.

leyanlo:
> We should add a link to: [css-tricks.com/aspect-ratio-boxes/#article-header-id-6](https://css-tricks.com/aspect-ratio-boxes/#article-header-id-6) or inline the logic for **why** this works as a code comment.

Good idea! Updated documentation with link to css-tricks article for why this works.

PR #1420, opened 2019-06-19
feat(popover): accepting popperOptions

### LINKED ISSUE #1152: Allow passing custom Popper.js modifiers to Popover Component
## Feature Request
It would be awesome if Popover exposed a way for users to pass custom Popper.js modifiers for some positioning edge-cases.

## Use Case
In my specific usecase, my Popover is anchored to an element inside a `overflow: scroll` sidebar. The [default](https://popper.js.org/popper-documentation.html#modifiers..preventOverflow.boundariesElement) modifier `preventOverflow.boundariesElement` is set to `scrollParent`, which is the sidebar. But this causes the Popover to be "trapped" within the sidebar and undesirably renders it on top of sidebar content. Specifying the `preventOverflow.boundariesElement` to the `window` solves this issue and correctly renders the Popover outside the sidebar.

## Alternatives considered
I have in the past positioned my popover manually outside of the sidebar by passing Popover an Override to the Body element style. I manually found the `top` `left` of my anchor and passed that to the Popover Body through a CSS `transform`. This approach however throws out the positioning advantages Popper.js provides and makes anchor scrolling position updates painful.

Setting `ignoreBoundary` to `true` also does not position the popover outside of the sidebar boundaries.

jh3y:
#### Description
- Resolves #1152 by supporting `popperOptions` on `Popover` component
- adds `popperOptions?: any` to types. Similar to how `Tether` is doing so.
- Extends current default options passed in.

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/src/popover/types.js
+++ b/src/popover/types.js
@@ -80,7 +80,8 @@ export type BasePopoverPropsT = {
   mountNode?: HTMLElement,
   /** How long should be fade out animation in ms, default 0ms */
   animateOutTime?: number,
-  /** Popper options override */
+  /** Popper options override
+   * https://popper.js.org/popper-documentation.html#Popper.Defaults */
   // eslint-disable-next-line flowtype/no-weak-types
   popperOptions?: any,
 };
--- a/src/popover/types.js
+++ b/src/popover/types.js
@@ -80,6 +80,9 @@ export type BasePopoverPropsT = {
   mountNode?: HTMLElement,
   /** How long should be fade out animation in ms, default 0ms */
   animateOutTime?: number,
+  /** Popper options override */
+  // eslint-disable-next-line flowtype/no-weak-types
+  popperOptions?: any,
 };
 
 // Props for stateless render logic
--- a/src/popover/popover.js
+++ b/src/popover/popover.js
@@ -432,7 +432,11 @@ class Popover extends React.Component<PopoverPropsT, PopoverPrivateStateT> {
 
   render() {
     const rendered = [this.renderAnchor()];
-
+    const defaultPopperOptions = {
+      modifiers: {
+        preventOverflow: {enabled: !this.props.ignoreBoundary},
+      },
+    };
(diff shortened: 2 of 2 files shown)


src/popover/types.js
```diff
@@ -80,6 +80,9 @@ export type BasePopoverPropsT = {
   mountNode?: HTMLElement,
   /** How long should be fade out animation in ms, default 0ms */
   animateOutTime?: number,
+  /** Popper options override */
```
gergelyke: can we add a link to their docs, so everyone will know what options are available?

src/popover/types.js
```diff
@@ -80,6 +80,10 @@ export type BasePopoverPropsT = {
   mountNode?: HTMLElement,
   /** How long should be fade out animation in ms, default 0ms */
   animateOutTime?: number,
+  /** Popper options override
+   * https://popper.js.org/popper-documentation.html#Popper.Defaults */
+  // eslint-disable-next-line flowtype/no-weak-types
+  popperOptions?: any,
```
chasestarr: popper js provides flow types [here](https://github.com/FezVrasta/popper.js/blob/master/packages/popper/index.js.flow#L104) can that be used?

jh3y: How does this work with `flow`. Can I do something like `import typeof {Options} from 'popper.js'`? Or will I need to copy the entire module over into the `Layer` types file?

chasestarr: the syntax is probably `import type {Options} from 'popper.js';`

jh3y: I wasn't 100% sure it was working. There is one file with the flow decs under `/node_modules/popper.js/dist/esm/poppper.js.flow`(intentional 3 ps) that has a module declaration of `Popper`. `Options` is within that module. I thought it might be a straight import but my IntelliSense didn't pick up on anything.

PR #1421, opened 2019-06-19
feat: resolve #1282 exposing events in handlers and supporting onClick

### LINKED ISSUE #1282: onClick prop on StatefulPopover
When using StatefulPopover on a clickable element, it propagates the click event and calls the click handler for the outer element.  To able to stop propagation on click on the anchor of a StatefulPopover, it would be nice to have an onClick prop that would pass the event from the anchor, so stopPropagation can be called on it. 

## Feature description
An onClick prop on the StatefulPopover that passes the event created by the anchor. 

The stateless Popover seems to have this functionality.

jh3y:
#### Description
- Resolves #1282 by adding `onClick` support
- Does pose the question of exposing other event handler events

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/src/popover/stateful-container.js
+++ b/src/popover/stateful-container.js
@@ -49,7 +49,10 @@ class StatefulContainer extends React.Component<
     this.close();
   };
 
-  onClick = () => {
+  onClick = (e: Event) => {
+    if (this.props.onClick) {
+      this.props.onClick(e);
+    }
     if (this.state.isOpen) {
       this.close();
     } else {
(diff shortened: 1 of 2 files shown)


chasestarr:
yeah - exposing all of the event handlers an interesting consideration. It may be appropriate to leave the interface as-is, where if you need access to the lower level abstraction it requires a bit more work. What do you think @jh3y ?

jh3y:
I'd be happy to leave as-is 👍

chasestarr:
> I'd be happy to leave as-is 👍

sgtm - thanks for looking into this

ksjogo:
Are the TypeScript types missing this onClick handler?

PR #1422, opened 2019-06-19
feat: resolve #1129 by always rendering tab content

### LINKED ISSUE #1129: [Tabs] Consider making SEO friendly
The Tab component does not render Tab panel content until the user interacts with the tab navigation. This prevents content from being crawled properly because search engines don't interact with components.

Would be great to discuss the possibility of rendering all content and use CSS show/hide instead of rendering it on the fly.

jh3y:
#### Description
- Always render tab content resolving #1129 
- Tab content was already being shown/hidden via CSS

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change


src/tabs/tabs.js
```diff
@@ -84,7 +84,7 @@ export default class Tabs extends React.Component<TabsPropsT> {
           {...tabContentProps}
           {...props}
         >
-          {isActive ? child.props.children : null}
+          {child.props.children}
```
chasestarr: It would be nice for performance to optionally not render children. Or rather the other way around where it defaults to the current functionality, and will apply this seo benefit based on a boolean prop.

jh3y: Where there is a `boolean` attribute passed at the `Tabs` level? Or the `Tab` level potentially?
I could add something like `renderAll` as a prop on `Tabs` for now. What do you think @chasestarr?

chasestarr: I would rather make it global on the container `Tabs` component. Could you see a use-case for doing so individually?

chasestarr:
an e2e test would be welcome on this pr to protect from a regression here

src/tabs/tabs.js
jh3y: Depends on the amount of tabs or scenario. You could potentially have some tab where it's critical that its content is rendered but other tabs not so much. It's an edge case that may be worth considering in the future 👍

PR #1423, opened 2019-06-19
feat: resolve #1328 allowing custom keys in options

### LINKED ISSUE #1328: Select Component onChange Flow type error
Select Component onChange Flow type Definitions 

## Current Behavior
if options have any other fields beside `id` and `name`, flow would error.

## Expected Behavior

onChange should return the same flow type  as the options provided as props to the component

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v6.20  |

- [ X] I have searched the [issues](https://github.com/uber-web/baseui/issues) of this repository and believe that this is not a duplicate.

---------------------------------

To illustrate the issue with one of the test  examples[ select-search-multi](https://github.com/uber-web/baseui/blob/0734e91495b331fc0817a4d2c6b87eac159f4b3e/src/select/__tests__/select-search-multi.scenario.js)

If I have added an `onChange` to this component  where I try to access `option.color`  

```js
  <StatefulSelect
    options={[
      {id: 'AliceBlue', color: '#F0F8FF'},
      {id: 'AntiqueWhite', color: '#FAEBD7'},
      {id: 'Aqua', color: '#00FFFF'},
      {id: 'Aquamarine', color: '#7FFFD4'},
      {id: 'Azure', color: '#F0FFFF'},
      {id: 'Beige', color: '#F5F5DC'},
      {id: 'DarkBlue', color: '#00008B'},
      {id: 'DarkCyan', color: '#008B8B'},
    ]}
    overrides={{ValueContainer: {props: {'data-id': 'selected'}}}}
    closeOnSelect={false}
    // added
    onChange={({option}) => {
      console.log(option.color);  // flow error
    }}
    // end of  added
    labelKey="id"
    multi
    valueKey="color"
    type={TYPE.search}
  />
);

```

```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/select/__tests__/select-search-multi.scenario.js:31:26

Cannot get option.color because:
 • property color is missing in null or undefined [1].
 • property color is missing in object type [2].

        src/select/__tests__/select-search-multi.scenario.js
        28│     closeOnSelect={false}
        29│     onChange={({option}) => {
        30│       // eslint-disable-next-line no-console
        31│       console.log(option.color);
        32│     }}
        33│     labelKey="id"
        34│     multi

        src/select/types.js
 [1][2] 32│   option: ?OptionT,

```

flow is unhappy since option type  is defined as below and  does not expect user  to provide any custom fields such as `color` in this scenario.

```js
export type OptionT = $ReadOnly<{
  id?: string | number,
  label?: React.Node,
  disabled?: boolean,
  clearableValue?: boolean,
  isCreatable?: boolean,
}>;
```

jh3y:
#### Description
- Resolves #1328 by allowing custom options in flow type declaration

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/src/select/__tests__/select.test.js
+++ b/src/select/__tests__/select.test.js
@@ -79,7 +79,15 @@ describe('Select component', function() {
           {id: 'AntiqueWhite', color: '#FAEBD7'},
         ]}
         closeOnSelect={false}
-        onChange={({option}) => option.color}
+        onChange={({option}) => {
+          /* eslint-disable no-console */
+          // $FlowFixMe
+          console.info(option.color);
+          if (option.color) {
+            console.info(option.color);
+          }
+          /* eslint-enable no-console */
+        }}
         labelKey="id"
         multi
         valueKey="color"
--- a/src/select/__tests__/select.test.js
+++ b/src/select/__tests__/select.test.js
@@ -71,4 +71,19 @@ describe('Select component', function() {
       value: [],
     });
   });
+  test('select flow allows custom keys in options objects', function() {
+    wrapper = mount(
+      <Select
+        options={[
+          {id: 'AliceBlue', color: '#F0F8FF'},
+          {id: 'AntiqueWhite', color: '#FAEBD7'},
+        ]}
+        closeOnSelect={false}
+        onChange={({option}) => option.color}
+        labelKey="id"
+        multi
+        valueKey="color"
+      />,
+    );
+  });
 });
--- a/src/select/__tests__/select.test.js
+++ b/src/select/__tests__/select.test.js
@@ -83,7 +83,7 @@ describe('Select component', function() {
           /* eslint-disable no-console */
           // $FlowFixMe
           console.info(option.color);
-          if (option.color) {
+          if (option !== null && option !== undefined && option.color) {
             console.info(option.color);
           }
           /* eslint-enable no-console */
--- a/src/select/types.js
+++ b/src/select/types.js
@@ -22,6 +22,8 @@ export type OptionT = $ReadOnly<{
   disabled?: boolean,
   clearableValue?: boolean,
   isCreatable?: boolean,
+  // eslint-disable-next-line flowtype/no-weak-types
+  [key: string]: any,
 }>;
 
 export type ValueT = $ReadOnlyArray<OptionT>;
--- a/src/select/types.js
+++ b/src/select/types.js
@@ -22,7 +22,8 @@ export type OptionT = $ReadOnly<{
   disabled?: boolean,
   clearableValue?: boolean,
   isCreatable?: boolean,
-  [key: string]: mixed,
+  // eslint-disable-next-line flowtype/no-weak-types
+  [string]: any,
 }>;
 
 export type ValueT = $ReadOnlyArray<OptionT>;
--- a/src/select/types.js
+++ b/src/select/types.js
@@ -22,8 +22,7 @@ export type OptionT = $ReadOnly<{
   disabled?: boolean,
   clearableValue?: boolean,
   isCreatable?: boolean,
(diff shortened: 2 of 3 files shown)


src/select/types.js
```diff
@@ -22,6 +22,8 @@ export type OptionT = $ReadOnly<{
   disabled?: boolean,
   clearableValue?: boolean,
   isCreatable?: boolean,
+  // eslint-disable-next-line flowtype/no-weak-types
+  [key: string]: any,
```
chasestarr: does replacing `any` with `mixed` essentially put it back to before this change? i'm a bit hesitant to throw out all of the type safety on this option.

jh3y: I think `mixed` will work. My understanding is that `any` allows for dynamic typing of values which I don't believe would be necessary here 👍

chasestarr:
do you mind adding a test like [here](https://github.com/uber-web/baseui/compare/styletron-wrappers?expand=1#diff-8f98dc816df5f5416abb399ee3489f07)? that pr is not merged yet, but check out the relevant change to the [`test` command](https://github.com/uber-web/baseui/compare/styletron-wrappers?expand=1#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R19)

doing so will ensure that this change operates as we expect

jh3y:
I tried to add a type checking test here. Is there something I'm missing with the test syntax?

src/select/__tests__/select.test.js
```diff
@@ -71,4 +71,19 @@ describe('Select component', function() {
...
   });
+  test('select flow allows custom keys in options objects', function() {
+    wrapper = mount(
+      <Select
+        options={[
+          {id: 'AliceBlue', color: '#F0F8FF'},
+          {id: 'AntiqueWhite', color: '#FAEBD7'},
+        ]}
+        closeOnSelect={false}
+        onChange={({option}) => option.color}
```
chasestarr: You'll want to add a `$FlowFixMe` above lines that we expect to fail the `onChange` call back should look something like below.

```js
onChange={({option}) => {
  // accessing items that may not exist on option continues to fail
  // $FlowFixMe
  console.log(option.color);

  if (option.color) {
    // it works when we assert its existence
    console.log(option.color);
  }
}}
```

jh3y: I tried
```
test('select flow allows custom keys in options objects', function() {
    wrapper = mount(
      <Select
        options={[
          {id: 'AliceBlue', color: '#F0F8FF'},
          {id: 'AntiqueWhite', color: '#FAEBD7'},
        ]}
        closeOnSelect={false}
        onChange={({option}) => {
          /* eslint-disable no-console */
          // $FlowFixMe
          console.info(option.color);
          if (option.color) {
            console.info(option.color);
          }
          /* eslint-enable no-console */
        }}
        labelKey="id"
        multi
        valueKey="color"
      />,
    );
  });
```
But this still throws a `flow` error locally because `color` is not defined.

chasestarr: which line gives that error? it's probably because we don't know what the type of `color` is so cannot be safely provided to `console.info`

chasestarr: or needs to be more exhaustively checked

`if (option.color !== null && option.color !== undefined)`

jh3y: Fixed with `if (option !== null && option !== undefined && option.color)` 👍

PR #1427, opened 2019-06-19
docs(tag): Tag docs update

TravisL12:
#### Description

Updates the `<Tag>` documentation to clearly define the prop type values as well as have a more verbose `Show Source` examples.

#### Prop type definitions
Two new sections are added that define the `kind` and `variant` prop types:
image

#### More Verbose Code Examples
The existing example code are nested map functions which make it difficult to understand what options are available for the `variant` and `kind` prop types.

An example of this is:
image

This PR clears this up by showing more verbose code examples:
image

The `variant` value is still being looped over though the displayed examples represent the values of these props (`solid`, `light`, `outlined`):
image

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change


gergelyke:
Thanks for the pr @TravisL12 !

documentation-site/static/examples/tag/variants.js
```diff
@@ -0,0 +1,18 @@
+import * as React from 'react';
+import {Tag} from 'baseui/tag';
+
+export default () => (
+  <>
+    <React.Fragment>
```
gergelyke: do we need the ` <React.Fragment>` wrappers here?

chasestarr:
```suggestion
```

tajo:
Tbh, I would take it even a step further and expand it completely but this is good too.

TravisL12:
Thanks for the input @tajo and @gergelyke. Sorry I got tied up and didn't have a chance to address your comments.

@tajo in regards to taking this a bit further, I agree but I also have noticed that the problem that I was addressing in this PR seems to be occurring throughout the documentation. So I'm privy to expanding the scope of updating docs with a bigger vision.

documentation-site/static/examples/tag/variants.js
TravisL12: These are only needed to maintain a row of `Tag` items being displayed. A simple `div` would have also sufficed.

PR #1428, opened 2019-06-19
feat(layer): add z-index support for the layer component

nadiia:
#### Description

`LayersManager` now accepts `zIndex` property that is used by all the layers in its context to have it set as a value of `z-index` style on each layer.
`LayersManager` now warns in the DEV mode if `LayersManager`s are nested and the nested managers reuse the host created by the outer most layers manager.

Option to consider:
We could create stacking contexts on the app and layers containers. That will completely illiminate overlap between those to contexts. It potentially may cause some unexpected results with third-party libraries usage if any injects elements outside of the app wrapper container, for example in the document body element.

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change


documentation-site/pages/components/layer.mdx
```diff
@@ -33,7 +55,33 @@ css property usage as layers completely rely on the stacking context.
   <LayerBasic />
 </Example>
 
-# Tether behavior
+### Thing you need to know when using z-index in an application
```
gergelyke:
```suggestion
### Things you need to know when using z-index in an application
```

src/layer/types.js
```diff
@@ -12,10 +12,13 @@ import {TETHER_PLACEMENT} from './constants.js';
 /** LayersManager */
 export type LayersManagerPropsT = {
   children: React.Node,
+  // z-index to be set for all the layers.
+  zIndex?: ?number,
```
chasestarr: if a `zIndex` property is provided, it needs to be non-null correct? same for the similar types below

```suggestion
  zIndex?: number,
```

documentation-site/pages/components/layer.mdx
```diff
@@ -23,6 +24,27 @@ and visibility handling, what layer goes on top. There are many other layer
 related issues that include focus management, hover management, keyboard navigation,
 click events and hotkeys usage.
 
+## LayersManager
+
+In order to start leveraging the layers approach add a `LayersManager` to your application. 
+It need to be added at the root of the application. If you use [`BaseProvider`](https://baseweb.design/components/base-provider/)
```
tajo:
```suggestion
It needs to be added at the root of the application. If you use [`BaseProvider`](https://baseweb.design/components/base-provider/) 
```

documentation-site/pages/components/layer.mdx
```diff
@@ -23,6 +24,27 @@ and visibility handling, what layer goes on top. There are many other layer
...
 click events and hotkeys usage.
 
+## LayersManager
+
+In order to start leveraging the layers approach add a `LayersManager` to your application. 
+It need to be added at the root of the application. If you use [`BaseProvider`](https://baseweb.design/components/base-provider/) 
+it will add the `LayersManager` for you. Just make sure you add the `BaseProvider` at the root 
+of your application. 
+
+`LayersManager` adds two sibling containers at the root of your application. Your main
```
tajo: How about

> `LayersManager` creates two adjacent divs and appends them to the `document.body`. The main React root will be mounted into the first div. The second div is used to host rendered layers.

terms like "sibling containers" or "main application" might not be clear.

nadiia: Edited, slightly changed since it doesn't append the divs to the body but to the container the app is supposed to be added to.

PR #1432, opened 2019-06-20
feat(input): add clearable property

sandgraham:
#### Description

This PR adds a `clearable` property to the `Input` component.

```
clearable: boolean = false
```

When `clearable={true}`, the input will function similarly to an input with `type="search"`. If the input has a value, a small icon will appear on the right side. When pressed, this icon will clear the input's current value. Pressing the `Escape` key will also clear the input's current value.

Screen Shot 2019-06-20 at 3 53 29 PM

Other inputs such as `PhoneInput`, `PaymentCard`, and `Textarea` will respect this property as well. `Select` already has `clearable` built in and those familiar with it should find the input version to function very similarly. Note, `clearable` defaults to `true` in the `Select` component, but not for inputs.

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change


gergelyke:
can you also update the doc page (remove the example with overrides, and add this)?

sandgraham:
@gergelyke I also added a new screener to verify the clear input icon shows up with the prop. Could you approve that when you get a chance?

src/input/types.js
```diff
@@ -14,6 +14,10 @@ import {
   ENHANCER_POSITION,
 } from './constants.js';
 
+export type ClearInputEventT<T> =
```
chasestarr: what benefit do we gain from the generic `T`? will this ever not be called on an `HTMLInputElement`?

src/input/types.js
```diff
@@ -100,8 +106,12 @@ export type BaseInputPropsT<T> = {
   onChange?: (e: SyntheticInputEvent<T>) => mixed,
   onKeyDown?: (e: SyntheticKeyboardEvent<T>) => mixed,
   onKeyPress?: (e: SyntheticKeyboardEvent<T>) => mixed,
-  onKeyUp?: (e: SyntheticKeyboardEvent<T>) => mixed,
+  onKeyUp: (e: SyntheticKeyboardEvent<T>) => mixed,
```
chasestarr: this should be left unchanged. only relax types if possible

sandgraham:
I'm going to add a few e2e tests for this.

src/input/types.js
```diff
@@ -143,6 +153,10 @@ export type StatefulContainerPropsT<T> = {
   /** A state change handler. Used to override default state transitions. */
   stateReducer: StateReducerT,
   onChange: (e: SyntheticInputEvent<T>) => mixed,
+  /** A handler for when a user clicks the clear value icon or presses Escape.  */
+  onClear: (e: SyntheticMouseEvent<T> | SyntheticKeyboardEvent<T>) => mixed,
```
chasestarr: `ClearInputEventT` can likely be reused. Why is `onClear` callback required?

chasestarr: also applies to `onClear` above

src/phone-input/constants.js
```diff
@@ -11,6 +11,7 @@ export const DEFAULT_MAX_DROPDOWN_HEIGHT = '400px';
 export const STATE_CHANGE_TYPE = {
   textChange: 'textChange',
   countryChange: 'countryChange',
+  clear: 'clear',
```
chasestarr: more abstractly, this overlaps the `textChange` event right? do consumers need to have a more refined change type? What use-case do we expect?

src/input/styled-components.js
```diff
@@ -9,6 +9,31 @@ import {styled} from '../styles/index.js';
 import type {ThemeT} from '../styles/types.js';
 import {ADJOINED, ENHANCER_POSITION, SIZE} from './constants.js';
 import type {SharedPropsT} from './types.js';
+import {DeleteAlt} from '../icon/index.js';
+
+export const StyledClearIconContainer = styled(
```
chasestarr: please type this - I understand it's not showing errors now, but it will [show errors](https://github.com/uber-web/baseui/pull/1417) in consumer code. make [this](https://github.com/uber-web/baseui/pull/1390/files#diff-a9d0201e8c5bdc946e54218bb9206174R2) change to test locally

src/phone-input/constants.js
sandgraham: i guess this can just be covered by `textChange` - not sure why folks would need to know about an explicit clear event - just defaulted to fine grained control

sandgraham: * the stateful version doesn't really need the event - but i think we do need on the stateless input/base input so that whatever controls the state of the input can clear the value string. does seem right?

src/input/types.js
sandgraham: tbh i am not sure 🤔 it seems like we default to a generic `T` for most event handlers though. `onChange?: (e: SyntheticInputEvent<T>) => mixed` for example. maybe there is a reason for those handlers but not this case? or perhaps it is just a legacy thing that is being repeated for no good reason.

sandgraham: actually it could be called with `HTMLTextAreaElement` as well

sandgraham: actually our textarea is just a big input so i guess not

src/input/base-input.js
```diff
@@ -55,12 +58,39 @@ class BaseInput<T: EventTarget> extends React.Component<
...
   }
 
+  componentWillUnmount() {
+    const {inputRef, clearable} = this.props;
+    if (clearable && inputRef.current) {
+      inputRef.current.removeEventListener('keyup', this.onInputKeyUp);
+    }
+  }
+
+  onInputKeyUp = (e: SyntheticKeyboardEvent<T>) => {
```
tajo: doing this on `keyDown` might be better so it's same as the native search input?

sandgraham: oh good point - i assumed it was keyup

src/input/styled-components.js
sandgraham: i _think_ i resolved the type issue here, but changing the config didn't make vscode report any problems with the un-typed version so not sure.

src/utils/trigger-input-change-event.js
```diff
@@ -0,0 +1,28 @@
+/*
+Copyright (c) 2018-2019 Uber Technologies, Inc.
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+// @flow
+/* global window */
+
+import createEvent from './create-event.js';
```
chasestarr: please co-locate these utility functions until they are used in more than one place in the codebase. the indirection is unnecessary when coupled to the input component

src/select/select.js
```diff
@@ -8,7 +8,7 @@ LICENSE file in the root directory of this source tree.
 import * as React from 'react';
 
 import {getOverrides} from '../helpers/overrides.js';
-import DeleteIcon from '../icon/delete.js';
+import {DeleteAlt} from '../icon/index.js';
```
chasestarr: I believe we need to import directly to get tree-shaking on unused icons

src/input/styled-components.js
```diff
@@ -8,7 +8,32 @@ LICENSE file in the root directory of this source tree.
...
+    display: 'flex',
+    alignItems: $alignTop ? 'flex-start' : 'center',
+    paddingLeft: sidePadding,
+    paddingRight: sidePadding,
+    paddingTop: $alignTop ? sidePadding : '0px',
+    color: $theme.colors.foregroundAlt,
+  };
+});
+
+export const StyledClearIcon = styled(DeleteAlt, {
```
chasestarr: please type this function

src/input/__tests__/input.e2e.js
```diff
@@ -38,4 +39,33 @@ describe('input', () => {
...
+      await mount(page, 'input-clearable');
+      await page.waitFor(selectors.input);
+
+      let inputValue = await page.$eval(selectors.input, input => input.value);
+      expect(inputValue).toBe('Something');
+
+      await page.click(selectors.deleteIcon);
+
+      inputValue = await page.$eval(selectors.input, input => input.value);
+      expect(inputValue).toBe('');
```
chasestarr: can we assert that the icon is no longer visible?

src/input/styled-components.js
chasestarr: I suppose we can always push a quick patch if necessary

src/utils/trigger-input-change-event.js
```diff
@@ -5,9 +5,18 @@ This source code is licensed under the MIT license found in the
 LICENSE file in the root directory of this source tree.
 */
 // @flow
-/* global window */
+/* global window document */
```
chasestarr: Sorry I was not clear earlier. Both functions should be encapsulated in the input file since it's not used anywhere else

sandgraham: how about moving the trigger input change event into the base-input but leaving `createEvent` as a util? it isn't used anywhere else but it is very general as a safe way to create event objects down to ie11. or we could polyfill `Event` during build and not worry about it

PR #1448, opened 2019-06-24
feat(screener): add tests for the iphone

gergelyke:


chasestarr:
duplicate images seem to be stacked? not quite sure what's going on there

gergelyke:
> duplicate images seem to be stacked? not quite sure what's going on there

yeah, I am writing to support

gergelyke:
it seems like with this change we've lost FF and IE11 screenshots - I am looking into that

PR #1449, opened 2019-06-24
feat(side-navigation): improve item rendering performance

chasestarr:
#### Description

Please refer to the scenario I built for some considerations. This pr attempts to provide reasonable memoization to the `NavItem` component by adding a props comparison function. note that not all props are included for comparison. Changes nested deep in `overrides` will not queue a re-render. For example if overridden and a state change updates a variable passed to style overrides no change will be visible. Is this acceptable behavior by default? I'm concerned that consumers will not be aware of this - leading to bug reports _and_ this may introduce functionality regressions for users that depend on dynamic overrides. Performing a deep equality comparison on overrides shakes out to similar time perf to always re-rendering items.

Tangentially related, I updated storybook configs so that hooks can be used there.

#### Scope

- [x] Minor: New Feature


src/side-navigation/nav-item.js
```diff
@@ -56,3 +56,28 @@ export default class NavItem extends React.Component<NavItemPropsT> {
...
+
+  for (var key in a) {
+    if (a[key] !== b[key]) {
+      return false;
+    }
+  }
+  return true;
+}
+
+function compare(prevProps, nextProps) {
```
chasestarr: I've considered adding a new `memoComparison` prop to give consumers the ability to bail out of default comparison. What do others think about that approach?

tajo: The optimization should be probably opt-in? It's not important for most of applications and brings potential bugs?

tajo:
I played with the example. It still seems re-rendering everything. Or what is testing?

updates

src/side-navigation/nav-item.js
```diff
@@ -56,3 +56,28 @@ export default class NavItem extends React.Component<NavItemPropsT> {
...
 }
+
+function areEqualShallow(a, b) {
+  if (!a || !b) return false;
+  if (typeof a !== 'object' || typeof a !== 'object') return false;
+
+  if (Object.keys(a).length !== Object.keys(b).length) return false;
+
+  for (var key in a) {
+    if (a[key] !== b[key]) {
```
tajo: should be this recursive / check if `a[key]` is an object?

src/side-navigation/nav-item.js
chasestarr: 👍that's a good idea

src/side-navigation/__tests__/nav-long.scenario.js
```diff
@@ -0,0 +1,86 @@
...
+      title: `item ${i + floor + 1}`,
+      itemId: String(i),
+    });
+  }
+  return nav;
+}
+
+function areEqualShallow(a, b) {
+  if (!a || !b) return false;
+  if (typeof a !== 'object' || typeof a !== 'object') return false;
```
raymondwang: I think there's a typo here and the second condition should be `typeof b`

chasestarr: thank you!

```suggestion
  if (typeof a !== 'object' || typeof b !== 'object') return false;
```

PR #1451, opened 2019-06-25
test: regression tests for datepicker highlight

jh3y:
## Description
- Add regression tests for `Datepicker` random highlight.

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change


src/datepicker/__tests__/datepicker-range-highlight.scenario.js
```diff
@@ -0,0 +1,34 @@
...
+import {StatefulDatepicker} from '../index.js';
+
+export const name = 'datepicker-range-highlight';
+
+export const component = () => {
+  const rightArrow = '[aria-label="Next month"]';
+  return (
+    <Screener
+      steps={new Steps()
+        .wait(rightArrow)
```
gergelyke: @jh3y this line times out in screener: 
`[Error] waitFor('[aria-label="Next month"]') timed-out` - probably the selector is off?

PR #1453, opened 2019-06-25
fix(icon): Filter styletron props to global icon overrides

chasestarr:
### Description

resolves warnings like below when icons are globally overridden

```
Warning: Invalid attribute name: `$size`
```

- [x] Patch: Bug Fix


chasestarr:
I'll include it

Issue #188, opened 2018-09-05
Testing components in Node environments


schnerd:
In #186 we ran into a case where a component was throwing errors server-side because it tried to reference browser specific globals.

Our current unit tests and e2e tests make sure that our components work in browser environments, but we don't have any tests to ensure that errors aren't thrown when rendering on the server.

We've historically used [unitest](https://github.com/rtsao/unitest) for this, which requires you to bifurcate your tests into `/node` and `/browser` folders. It seems like unitest is more focused on `tap` for tests than `jest`, so maybe it's not the best tool for us this time around.

Other options:

### Use jest "projects" option

I haven't 100% validated that this would work, but it seems like we may be able to use [Jest projects](https://jestjs.io/docs/en/configuration.html#projects-array-string-projectconfig) to run in multiple environments in parallel.

```
{
  "projects": [
    {
      "displayName": "browser",
      "testMatch": ["<rootDir>/**/*.browser.test.js"],
      "testEnvironment": "jsdom"
    },
    {
      "displayName": "node",
      "testMatch": ["<rootDir>/**/*.node.test.js"]
      "testEnvironment": "node"
    }
  ]
}
```

### Separate Jest configs

Create separate jest configs `jest.browser.config.js` `jest.node.config.js` with different environments and globs, and run them as separate unit test commands.

Thoughts? Other ideas?

kahwee:
Interesting use case. So we are discussing two issues (I think):

1. jest vs tape
2. how do we make a system that test in node and browser env

I prefer `jest` personally, it's a plus that fusionjs is using it too.

For 2, is it possible to run the same test over both node and browser configurations?

gergelyke:
I like the Jest projects option - do you want me to give it a try @schnerd , or will you?

schnerd:
> jest vs tape

To be clear, this issue is not about jest vs tape. We're sticking with jest.

> For 2, is it possible to run the same test over both node and browser configurations?

I don't think so unfortunately, many of our current unit tests invoke component functionality that is dependent on a DOM–or a fake DOM like jsdom–being present. Enzyme's `mount()` isn't intended to work in a node environment and we use that in plenty of our tests. 

@gergelyke it'd be awesome if you or someone else on the team wanted to volunteer to try it out.

gergelyke:
yeah, I'll do it!

jh3y:
Opened a PR for this #1464 👍

PR #1464, opened 2019-06-26
feat: split testing for node/browser

jh3y:
#### Description
- Makes use of `jest` projects to resolve [#188](https://github.com/uber-web/baseui/issues/188)
- Simple avatar snapshot test as POC

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change


gergelyke:
should we run all the tests on Node.js to make sure they work? I think we want to do that to make sure server-side rendering will work?

what do you think @chasestarr @tajo @sandgraham ?

jh3y:
The majority of tests will fail with `node` as the `testEnvironment`.
The main reason being that `mount` relies on a loaded `document` which is available in `jsdom` environment.

gergelyke:
As we are doing server-side rendering for the doc site, we've decided to close this, as it's already covered there

PR #1480, opened 2019-06-27
feat(babel): removes some experimental proposals

chasestarr:


tajo:
whoa, I didn't even know have those

PR #1481, opened 2019-06-27
feat(table): export SORT_DIRECTION const for SortableHeadCell

raymondwang:
#### Description

Exports a constant for 'ASC' and 'DESC' strings for use with SortableHeadCell.

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change

(This is really small, but _technically_ isn't fixing a bug so I'm not sure how bullish we want to be on versioning a change like this.)


src/table/types.js
```diff
@@ -8,13 +8,16 @@ LICENSE file in the root directory of this source tree.
...
 import type {OverrideT} from '../helpers/overrides.js';
 
+export type SortDirectionT = ?$Keys<typeof SORT_DIRECTION>;
+
 export type HeadCellPropsT = {|
   /** Additional content to fill head cell. Most likely used for a filter button. */
   children?: React.Node,
   /** Visually indicates sort direction. Displays a chevron next to column title. */
-  direction: 'ASC' | 'DESC' | null,
+  direction?: SortDirectionT,
```
chasestarr:
```suggestion
  direction: SortDirectionT,
```

Let's maintain required on this field. It ensures that direction is being used. if not passed, the sortable component should not be used at all

PR #1482, opened 2019-06-27
fix(notification): support for longer messages

gergelyke:


src/notification/__tests__/notification.scenario.js
```diff
@@ -15,6 +15,10 @@ export const name = 'notification';
 export const component = () => (
   <React.Fragment>
     <Notification>Default info notification</Notification>
+    <Notification>
+      Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod
+      tempor incididunt ut labore et dolore magna aliqua
```
chasestarr: does this handle the original request where there is a `<strong>` element in children?

gergelyke: yeah, it does - do you want me to add it?

jasonbarone:
LGTM 👍

src/notification/__tests__/notification.scenario.js
chasestarr: no big deal - just double-checking

PR #1484, opened 2019-06-27
Docs minor design fixes

tajo:
- footer is now less dominant (light background instead of black) and centered, Uber logo used, tuned for the dark theme too
- "What's new in v8" was breaking the mobile view, it's hidden earlier
- max width of the main content was decreased so the readability is better (shorter rows read better)
- responsiveness was slightly broken for some resolutions, fixed
- replaced some internal links with next.js routing so there is not a full reload
- blog posts were using one-off top heading that was not aligned with the sidebar and used different font size / weights than all other pages, fixed


documentation-site/components/footer.js
```diff
@@ -65,32 +76,10 @@ export default () => (
       >
         Changelog
       </StyledLink>
+      <Link href="/blog">
+        <StyledLink>Blog</StyledLink>
```
gergelyke: as is, the link is not tab-able - can we add the href prop to the styled link too?

documentation-site/components/footer.js
```diff
@@ -8,48 +8,59 @@ LICENSE file in the root directory of this source tree.
...
-    ...typography.font300,
-    backgroundColor: colors.foreground,
-    color: colors.background,
+  ({$theme: {typography, sizing, colors, name}}) => ({
+    ...typography.font400,
+    position: 'sticky',
+    bottom: 0,
+    color: colors.foreground,
+    backgroundColor:
+      name === 'light-theme' ? colors.mono200 : colors.headerNavigationFill,
```
gergelyke: the original idea behind the darker footer was that we are going to change the whole system to use black - should we change it back to something darker?

documentation-site/components/footer.js
tajo: fixed, same with the `What's new in v8`

tajo:
Addition: Remove double `<a>` in the sidebar navigation.

documentation-site/components/footer.js
tajo: > we are going to change the whole system to use black

What do you mean? I thought the default theme will be always the light theme?

The design team was once discussing to invert the background of the top navigation to make it black for the light theme but it was ruled out since it overpowered the rest of the page. I think the same logic applies for footer that's even less important?

The current version creates a big light space in the dark theme. Although, that looks less distracting than pure black in on the white page.

Screen Shot 2019-06-27 at 3 16 28 PM

This PR:

Screen Shot 2019-06-27 at 3 30 42 PM

gergelyke: that makes sense - please disregard my comment :)

PR #1487, opened 2019-06-28
fix(phone-input): add disabled property to country select

### LINKED ISSUE #1483: [PhoneInput] `disabled` state only partly disables the component
## Current Behavior

Setting `disabled` on the PhoneInput disables the text field part of the component, but the country icon with flags still opens on click. Selecting a country does not change the selected option though.

## Expected Behavior

The entire component to be in disabled/readonly mode. There should be no pointer on the flag and clicking should not be possible.

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v7.8.0  |
| React   |    16.8.6     |
| browser |    Chrome 75.x     |

sandgraham:
Fixes #1483 

#### Description

This PR makes the Country Select in the Phone Input respect a top level `disabled` prop.

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/src/phone-input/types.js
+++ b/src/phone-input/types.js
@@ -63,6 +63,7 @@ export type CountrySelectPropsT = {
   maxDropdownWidth?: string,
   maxDropdownHeight?: string,
   mapIsoToLabel?: mapIsoToLabelT,
+  disabled?: boolean,
   overrides: {
     DialCode?: OverrideT<*>,
     CountrySelect?: OverrideT<*>,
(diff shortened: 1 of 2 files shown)


src/phone-input/types.js
```diff
@@ -63,6 +63,7 @@ export type CountrySelectPropsT = {
   maxDropdownWidth?: string,
   maxDropdownHeight?: string,
   mapIsoToLabel?: mapIsoToLabelT,
+  disabled?: boolean,
```
gergelyke: should we add a short description for it?

PR #1489, opened 2019-06-28
feat(table-grid): creates table components using css-grid

chasestarr:
### Description

example of row-span
Screen Shot 2019-06-28 at 3 10 57 PM

example of nested table
Screen Shot 2019-06-28 at 3 10 46 PM

#### Scope

- [x] Minor: New Feature


documentation-site/pages/components/table-grid.mdx
```diff
@@ -0,0 +1,37 @@
...
+
+<Example title="Basic usage" path="examples/table-grid/basic.js">
+  <Basic />
+</Example>
+
+<Example title="Nested table" path="examples/table-grid/nested.js">
+  <Nested />
+</Example>
+
+<Example title="Table with pagination" path="examples/table/pagination.js">
```
gergelyke: should we update this?

.storybook/load-stories.js
```diff
@@ -1,22 +1,21 @@
...
 
-scenarios.reduce(
-  (stories, scenario) => stories.add(scenario.name, scenario.component),
-  storiesOf('baseui-dark', module),
-);
+scenarios.forEach(scenario => {
+  const Component = scenario.component;
+  light.add(scenario.name, () => <Component />);
+  dark.add(scenario.name, () => <Component />);
+});
```
gergelyke: do we want to include this changeset in this PR?

documentation-site/static/examples/table-grid/nested.js
```diff
@@ -0,0 +1,265 @@
...
+      return 'neutral';
+    }
+  }
+}
+
+// sticky elements create a new stacking context. because parent table head cells are _siblings_
+// of these visually child elements, stacking is determined by order in html. this leads to
+// unintended layering upon scrolling the outer table. this table should not need to scroll
+// therefore chose to remove sticky rather than provide relative zindex values.
+const NotStickyHeadCell = withStyle(StyledHeadCell, {
```
sandgraham: could this be a prop or configured somehow? seems like it might be common enough use case

.storybook/load-stories.js
chasestarr: I don't see an issue with doing so.. I wanted to get hooks working with storybook so that the scenarios in this pr can run load

documentation-site/pages/components/table-grid.mdx
chasestarr:
```suggestion
<Example title="Cell spanning" path="examples/table-grid/cell-span.js">
```

documentation-site/static/examples/table-grid/nested.js
chasestarr: I hesitate to prematurely add features that have not been necessary in-practice. I'd like to ship as-is and improve where requested

.storybook/load-stories.js
gergelyke: sounds good, I wasn't sure why we added it here - thanks for the clarification Chase!

chasestarr: shows up in this pr because I added it to the pr :)

chasestarr:
fyi this is how the css-grid renders in IE:
Screen Shot 2019-07-01 at 12 57 10 PM

chasestarr:
added `unstable` prefix to components

documentation-site/static/examples/table-grid/nested.js
chasestarr: I think I was too quick to write this off. I'm going to make it sticky by default with the option to turn off

tajo:
Is there some way how to make these examples nicer in the context of our documentation layout? Even with a very wide viewport (1400px) it still breaks the page. Could we have some responsiveness built in the table itself? Or maybe wrap the whole example with scrollable container?

chasestarr:
> Is there some way how to make these examples nicer in the context of our documentation layout? Even with a very wide viewport (1400px) it still breaks the page. Could we have some responsiveness built in the table itself? Or maybe wrap the whole example with scrollable container?

thank you - fixed

documentation-site/static/examples/table-grid/nested.js
chasestarr: updated 👍

tajo:
> > Is there some way how to make these examples nicer in the context of our documentation layout? Even with a very wide viewport (1400px) it still breaks the page. Could we have some responsiveness built in the table itself? Or maybe wrap the whole example with scrollable container?
> 
> thank you - fixed

Looks great! Thanks!

chasestarr:
merged this component as 'unstable'. please post here or in a new issue if there are any comments

PR #1490, opened 2019-06-29
feat(phone-input): flow & prop enhancements

sandgraham:
#### Description

This PR reorganizes the flow types for the phone input a bit.

It _shouldn't_ be breaking even though I've made some properties required. I added defaults in defaultProps to make up for those but there may be some other issues I've overlooked.

After hooking up the disabled property in #1487, I noticed that the types were a bit messy and inconsistent with some patterns elsewhere. I also wanted to make the types a bit more explicit about which props were actually accepted and used.

Previously I was just spreading everything the phone input didn't use over the inner input element- but it feels better to be intentional about what is passed through and makes for better documentation. I don't think the phone input should just proxy every prop input accepts - that is what overrides are for.

#### Scope

- [ ] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/src/phone-input/constants.js
+++ b/src/phone-input/constants.js
@@ -5,7 +5,6 @@ This source code is licensed under the MIT license found in the
 LICENSE file in the root directory of this source tree.
 */
 // @flow
-// These are not used as of #1490. Leaving here until a future major so as not to break consumer code.
 export const DEFAULT_MAX_DROPDOWN_WIDTH = '400px';
 export const DEFAULT_MAX_DROPDOWN_HEIGHT = '400px';
 
--- a/src/phone-input/constants.js
+++ b/src/phone-input/constants.js
@@ -5,9 +5,6 @@ This source code is licensed under the MIT license found in the
 LICENSE file in the root directory of this source tree.
 */
 // @flow
-export const DEFAULT_MAX_DROPDOWN_WIDTH = '400px';
-export const DEFAULT_MAX_DROPDOWN_HEIGHT = '400px';
-
 export const STATE_CHANGE_TYPE = {
   textChange: 'textChange',
   countryChange: 'countryChange',
--- a/src/phone-input/constants.js
+++ b/src/phone-input/constants.js
@@ -5,6 +5,10 @@ This source code is licensed under the MIT license found in the
 LICENSE file in the root directory of this source tree.
 */
 // @flow
+// These are not used as of #1490. Leaving here until a future major so as not to break consumer code.
+export const DEFAULT_MAX_DROPDOWN_WIDTH = '400px';
+export const DEFAULT_MAX_DROPDOWN_HEIGHT = '400px';
+
 export const STATE_CHANGE_TYPE = {
   textChange: 'textChange',
   countryChange: 'countryChange',
--- a/src/phone-input/country-select-dropdown.js
+++ b/src/phone-input/country-select-dropdown.js
@@ -9,7 +9,7 @@ LICENSE file in the root directory of this source tree.
 import React from 'react';
 import {List, AutoSizer} from 'react-virtualized';
 
-import {DEFAULT_MAX_DROPDOWN_HEIGHT} from './constants.js';
+import defaultProps from './default-props.js';
 import {
   StyledFlag,
   StyledCountrySelectDropdownContainer as DefaultContainer,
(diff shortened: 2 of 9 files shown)


src/phone-input/country-select-dropdown.js
```diff
@@ -22,87 +22,97 @@ import {getOverrides} from '../helpers/overrides.js';
...
+        }}
+      </AutoSizer>
+    </Container>
+  );
+}
+
+export default React.forwardRef<
+  CountrySelectDropdownPropsT,
+  typeof CountrySelectDropdown,
+>((props, ref) => <CountrySelectDropdown {...props} containerRef={ref} />);
```
chasestarr: I usually call this `forwardedRef` to avoid ambiguity and to explicitly call-out that it's an implementation detail required for `React.forwardRef`

src/phone-input/stateful-phone-input-container.js
```diff
@@ -58,9 +57,10 @@ export default class StatefulPhoneInputContainer extends React.Component<
 
   render() {
     const {children, initialState, stateReducer, ...restProps} = this.props;
+    // $FlowFixMe
```
chasestarr: please remove this - the flow refactor is unsafe if we start introducing ignores where previously not required

src/phone-input/constants.js
```diff
@@ -5,9 +5,6 @@ This source code is licensed under the MIT license found in the
 LICENSE file in the root directory of this source tree.
 */
 // @flow
-export const DEFAULT_MAX_DROPDOWN_WIDTH = '400px';
```
chasestarr: breaking change since it was exported https://github.com/uber-web/baseui/blob/master/src/phone-input/index.js#L17

src/phone-input/country-select-dropdown.js
```diff
@@ -22,87 +22,97 @@ import {getOverrides} from '../helpers/overrides.js';
...
-    );
-  },
-);
+CountrySelectDropdown.defaultProps = {
+  maxDropdownHeight: defaultProps.maxDropdownHeight,
+  overrides: {},
+};
 
-export default CountrySelectDropdown;
+function CountrySelectDropdown(props: CountrySelectDropdownPropsT) {
```
chasestarr: may want to inline the union of `PropsT & {containerRef: ref}` here since it's an encapsulated implementation detail. no-one should be calling this function without the forwardRef wrapping

src/phone-input/types.js
```diff
@@ -40,95 +37,102 @@ export type StateReducerT = (
...
+  /** Sets the size of the component. */
+  size: SizeT,
+  /** Defines the value of the input element. */
+  text: string,
 };
 
-export type StatefulPhoneInputContainerPropsT = {
-  /** Sets aria-label attribute. */
-  'aria-label': ?string,
+export type StatefulContainerPropsT = {
```
chasestarr: may want to revert the naming changes since consumers may have been importing them?

gergelyke:
generally speaking, it seems like there quite some potential breaking changes in this PR, as Chase pointed out too - can we find a middle ground where we still improve, but without any breaking changes?

chasestarr:
@gergelyke we had a conversation in-person the other day about flow prop types + defaultProps. flow _does_ explicitly recommend to mark fields with associated default prop as required https://flow.org/en/docs/react/components/#toc-using-default-props

The challenge here comes from what we understand the types to represent. Are they public documentation for consumers or do they primarily help us internally ensure the code works? We re-use types for our docs site, but flow takes the stance that these types don't necessarily act as documentation. react-docgen seems to be able to mark fields with associated defaultProps as optional https://github.com/reactjs/react-docgen

sandgraham:
yeah i got a little carried away friday evening - i'll dial it back a bit. i do think there are a few useful changes.

src/phone-input/country-select-dropdown.js
sandgraham: oh that's a great point

sandgraham:
One potential issue with these changes - I removed the spreading of props from the top level `PhoneInput` API onto the internal input element. I think this spreading, while convenient, was a mistake in the original code. We should encourage people to use overrides for setting props on nested elements, not just proxy them from the top level. It can become confusing how the props will get used if they are not explicit in the documentation/types.

I've only seen the spreading used in a couple ways - mainly setting the `placeholder`, `id`, or `name` of the input element. Those would still work as well as a few other `Input` props supported in this PR, but it is possible that someone could have added an `onFocus` or something else `Input` supports and this PR would break their code. It isn't very likely - but it is possible, so I am thinking that perhaps I should wait until a future major to make these changes.

One option is to leave out the spread change and make this 100% non-breaking, but I think removing the spread is kind of the heart of the PR so I don't know what would be left that is worth merging in.

PR #1501, opened 2019-07-02
feat(versions): show minor and patch versions

gergelyke:


documentation-site/components/version-selector.js
```diff
@@ -0,0 +1,107 @@
...
+                  size: 'compact',
+                  getChildMenu: item => {
+                    if (
+                      semver.satisfies(semver.coerce(item.label), '>=8.0.0')
+                    ) {
+                      return (
+                        <StatefulMenu
+                          size="compact"
+                          items={versionsToShow}
+                          onItemSelect={({item}) => {
```
chasestarr: I thought this was not working?

gergelyke: it doesn't work if I remove `dismissOnClickOutside={false}`

gergelyke:
> works well - will this break when a v9 is published?

We'll only have to add it the `MAJOR_VERSIONS` constant, or we can grab the versions from the npm registry

PR #1506, opened 2019-07-02
fix(build): esm mapping

gergelyke:
Currently, our mapping use absolute paths instead of relative, and the keys do not match the import statements:

Screen Shot 2019-07-02 at 1 01 01 PM

This PR changes it to be like this:

Screen Shot 2019-07-02 at 12 57 21 PM


gergelyke:
I am in the process of verifying if this really works

PR #1510, opened 2019-07-03
fix(select): always highlight the first item initially

tajo:
When the select menu is open and mounted, always highlight the first item initially. There is no reason to do filtering based on the input value because the options are naturally filtered so the first item should be highlighted.

This was causing a bug with multi select when on the second input, the menu was unfocused.


tajo:
> I tried for a while to break it but seems to work well! Was it possible to reproduce the reported error with a e2e test?

e2e test added

PR #1511, opened 2019-07-03
fix(version-selector): minor improvements

gergelyke:


tajo:
Btw, are we planning to use `baseweb.design` for old versions?

PR #1512, opened 2019-07-03
feat(seo): add keywords and description for components

gergelyke:
for now, this is just a PoC - if we agree that this is a good approach, I'll add information for all the rest of the components


documentation-site/components/meta-favicons.js
```diff
@@ -6,23 +6,11 @@ LICENSE file in the root directory of this source tree.
 */
 
 /* eslint-disable flowtype/require-valid-file-annotation */
-/* global process */
```
gergelyke: these were duplicated in other parts of the doc site

PR #1513, opened 2019-07-03
feat(versions): alias tags

gergelyke:


gergelyke:
I ran into issues currently trying to alias in the now platform - will merge once resolved

gergelyke:
> I ran into issues currently trying to alias in the now platform - will merge once resolved

this is resolved

PR #1525, opened 2019-07-09
feat: update eslint-plugin-flowtype + enable require-exact-type

### LINKED ISSUE #529: Component props should use exact types
This will help prevent regressions for deprecated/renamed props, like `rows` which was renamed `maxDropdownHeight`.

jh3y:
#### Description
- Should fix #529 
- Linting warning aren't firing locally although they should

#### Scope
- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -25,8 +25,8 @@ module.exports = {
   rules: {
     // Enforce flow file declarations
     'flowtype/require-valid-file-annotation': ['error', 'always'],
-    // allow mixed type: https://github.com/uber-web/baseui/issues/626
-    'flowtype/no-weak-types': ['error', {mixed: false}],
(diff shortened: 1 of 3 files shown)


gergelyke:
@jh3y any chance you could spend some time on it? do you think we should spend some time on making it work, or wait for flow to release exact types?

chasestarr:
We may want to consider making this change along with baseui consumers, rather than before. I'm unsure if disrupting consuming code (once when baseui releases and again when flow changes) will be worth the migration effort - while doing it in coordination with the flow change may have devs update their application types along with fixing flow check errors resulting from baseui changes (one time effort)

gergelyke:
sounds good - should we just close this one and the issue?

PR #1526, opened 2019-07-09
 feat: propagate overrides to children in tabs & accordion

anantkamath:
Allows overriding Tab from the Tabs override& allows passing overrides other than Root through the Accordion override prop.


chasestarr:
I would ideally like to keep the overrides specific to the component it's overriding. For example, given a `Tabs` component with no `Tab` component children, the `Tab` override does not do anything.

PR #1530, opened 2019-07-10
fix(screener): remove flaky modal test

gergelyke:


gergelyke:
> fine to remove the test - i'm going to add a task to replace it since this was a large pain amongst consumers of baseui

thanks Chase!

PR #1536, opened 2019-07-11
fix(types): fix styled() type so it can take a single string arg

### LINKED ISSUE #1535: Seemingly invalid errors when using `styled('div')`
## Current Behavior

```
Cannot call `styled` because: Either  string [1] is incompatible with  `React.ComponentType` [2] in type argument `Base`. Or  string [1] is incompatible with  `React.ComponentType` [3] in type argument `Base`.Flow(InferError)
```

## Expected Behavior

no error

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v8.4.0  |
| React   |    v16.8.2     |
| browser |    Chrome     |

vicapow:
Fixes #1535

#### Description

Changed the type of `styled()`

#### Scope

- [x] Patch: Bug Fix

### DIFF
--- a/src/styles/__tests__/styled.test.js
+++ b/src/styles/__tests__/styled.test.js
@@ -42,6 +42,13 @@ test('styled', () => {
   wrapper.unmount();
 });
 
+test('styled can be called with single string argument', () => {
+  const ADiv = styled('div');
+  expect(ADiv).toBeTruthy();
+  const wrapper = mount(<ADiv />);
+  wrapper.unmount();
+});
+
 test('styled override prop', () => {
   const StyledMockButton = styled('button', {
     color: 'red',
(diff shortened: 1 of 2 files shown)


src/styles/__tests__/styled.test.js
```diff
@@ -42,6 +42,13 @@ test('styled', () => {
   wrapper.unmount();
 });
 
+test('styled can be called with single string argument', () => {
+  const ADiv = styled('div');
```
chasestarr: Is this pattern useful? It's not something that [styletron-react](https://github.com/styletron/styletron/blob/master/packages/styletron-react/src/types.js#L52-L63) supports either.

I believe that failing to flow check in this case promotes simpler practices. It can be replaced with a `<div>` tag in this case, rather than calling to styletron.

vicapow: It may not be useful but it's valid and we have existing code that uses this. If we don't want to support this, styletron should error instead.

chasestarr: sgtm - i'm going to consider adding a warning in this case to suggest a simpler approach

PR #1538, opened 2019-07-11
Move Typescript definitions into our repo

tajo:


scripts/ts-copy.js
```diff
@@ -0,0 +1,24 @@
...
+const fs = require('fs-extra');
+const globby = require('globby');
+const path = require('path');
+
+(async () => {
+  const files = await globby(['src/**/*.d.ts']);
+  files.forEach(async file => {
+    const from = path.join(__dirname, '../', file);
+    const to = path.join(__dirname, '../', file.replace('src', 'dist'));
+    await fs.copy(from, to);
```
gergelyke: node now has a copy method, so yo don't need this dep : https://nodejs.org/api/fs.html#fs_fs_copyfilesync_src_dest_flags

scripts/ts-copy.js
```diff
@@ -0,0 +1,24 @@
...
+*/
+
+/* eslint-env node */
+// @flow
+
+const fs = require('fs-extra');
+const globby = require('globby');
+const path = require('path');
+
+(async () => {
```
gergelyke: as we use this for build-time only, I think it's safe to use sync apis - I'd do either that, or add a catch to the async function to make sure we see errors if any

scripts/ts-copy.js
tajo: Do we use Node v12 in our CI?

tajo: Oh, it's already in 10

scripts/ts-copy.js
tajo: It still throws errors

```
(node:35662) UnhandledPromiseRejectionWarning ... specific error message
```

`flow-copy-src` is using the same pattern

PR #1539, opened 2019-07-11
fix(codemods): do not apply props type to styled if obj style arg

chasestarr:
#### Description

- v7 codemod more intelligently handles prop types if the style arg is an object vs function
- adds an empty object as second argument if not present

```js
// style arg is an object
const Component = styled('div', {color: 'red'});`;
const Component = styled<any>('div', {color: 'red'});`; // before
const Component = styled('div', {color: 'red'});`; // new

// style arg is a function 
const Component = styled('div', p => ({color: 'red'}));`;
const Component = styled<any>('div', p => ({color: 'red'}));`; // operates as before

// applies empty object if no style arg is provided
const Component = styled('div');
const Component = styled('div', {}); // new
```


packages/baseui-codemods/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@uber-web-ui/baseui-codemods",
-  "version": "1.0.6",
+  "version": "1.0.8",
```
sandgraham: should this jump from 6 to 8?

chasestarr: i already published the change :) (took 2 tries)

PR #1549, opened 2019-07-12
Update slider flow types to mark optional params optional

### LINKED ISSUE #1548: Incorrect Flow type for Slider props
Slider prop flow types inconsistent with TypeScript types and documentation.

## Current Behavior

Slider Flow type for props requires `min`, `max`, `step`, and `disabled` to be defined.

## Expected Behavior

Slider Flow type for props allows `min`, `max`, `step`, and `disabled` to be optional.

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v8.4.1 |
| Flow | v0.102.0 |

shannonmoeller:
Fixes #1548

#### Description

Changes Flow type for Slider props to match TypeScript types and documentation.

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/src/slider/types.js
+++ b/src/slider/types.js
@@ -28,16 +28,16 @@ export type PropsT = {
   /** Position of the thumbs. It can be a single point (one thumb) or 2 points array (range thumbs). */
   value: Array<number>,
   /** The minimum allowed value of the slider. Should not be bigger than max. */
-  min: number,
+  min?: number,
   /** The maximum allowed value of the slider. Should not be smaller than min. */
-  max: number,
+  max?: number,
   /** The granularity the slider can step through value. Default step is 1. */
-  step: number,
+  step?: number,
   overrides?: OverridesT,
   /** Disable control from being changed. */
-  disabled: boolean,
+  disabled?: boolean,
   /** Handler for events on trigger element, each time thumbs change selection, which is passed in `value`. */
-  onChange: ({
+  onChange?: ({
     ...ParamsT,
   }) => mixed,
 };


chasestarr:
There's been an ongoing debate on how best to type the top-level interface of components. Flow [recommends](https://flow.org/en/docs/react/components/#toc-using-default-props) marking the fields as `required` if there are `defaultProps` [provided](https://github.com/uber-web/baseui/blob/af59afe69002620935037b9b30063837001e67c1/src/slider/slider.js#L37)

In what situation is this causing errors in your code? Do you mind sharing a reproduction? That would help provide better evidence for coming to a conclusion on how to approach required vs optional

shannonmoeller:
I don't understand the rationale behind marking things required when they have a default value. To me, that's exactly what optional means. As it stands, when I use `<Slider />` I have to specify `disabled={false}` to prevent Flow from complaining. That doesn't seem desirable.

```
Cannot create Slider element because property disabled is missing in props [1] but exists in PropsT [2].

     src/components/map/datasources-modal.js
     103│           />
     104│         </FormControl>
     105│         <FormControl label="Slider">
 [1] 106│           <Slider
     107│             value={[datasource.opacity || 0]}
     108│             onChange={handleSliderChange('opacity')}
     109│           />
     110│         </FormControl>
     111│         <FormControl label="Environment">
     112│           <Input

     node_modules/baseui/slider/slider.js.flow
 [2]  36│ class Slider extends React.Component<PropsT> {
```

shannonmoeller:
Oh, wait I see what you're saying now. Flow would stop complaining if `disabled: false` was added to `defaultProps`. So, two questions:

1. Do Flow and TypeScript differ on this, or should the `?` be removed from index.d.ts?
2. Would you like me to update this PR to add `disabled: false` to `defaultProps`?

chasestarr:
> 1. Do Flow and TypeScript differ on this, or should the `?` be removed from index.d.ts?

I'm not personally experienced with typescript, but they [seemingly are aligned](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#support-for-defaultprops-in-jsx)

> 2. Would you like me to update this PR to add `disabled: false` to `defaultProps` instead of changing the types?

If adding to default props works, that would be preferable.

gergelyke:
@shannonmoeller @chasestarr should we close this PR?

shannonmoeller:
Sorry for the delay. Updated PR with default value for `disabled` coming up.

shannonmoeller:
Nevermind. Looks like Vojtech already added the missing `defaultProps.disabled` value.

PR #1554, opened 2019-07-12
feat: resolve #473 with aria-controls support for accordion

### LINKED ISSUE #473: [Accordion] TODOs
- add support for the `aria-controls` attr
- add unit tests for the accordion component

jh3y:
#### Description
- Adds support for `aria-controls` by passing an `id` prop to `Panel` #473

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change


src/accordion/panel.js
```diff
@@ -61,7 +61,14 @@ class Panel extends React.Component<PanelPropsT> {
   }
 
   render() {
-    const {expanded, disabled, overrides = {}, children, title} = this.props;
+    const {
+      expanded,
+      id,
```
gergelyke: should we call this `aria-controls`, if used for that? I believe we follow a practice like that in other components too

jh3y: Updated 👍

src/accordion/__tests__/panel.test.js
```diff
@@ -37,6 +37,27 @@ describe('Panel', () => {
     );
   });
 
+  test('aria-controls populated when ID provided', () => {
```
gergelyke: can you update this too with the aria prop?

src/accordion/panel.js
```diff
@@ -116,7 +125,11 @@ class Panel extends React.Component<PanelPropsT> {
                 overrides={toggleIconOverrides}
               />
             </Header>
-            <Content {...sharedProps} {...contentProps}>
+            <Content
+              {...sharedProps}
+              {...contentProps}
+              {...(ariaControls ? {id: ariaControls} : {})}
```
gergelyke: do we have to provide the id here, or `aria-controls`?

Issue #208, opened 2018-09-08
[Proposal] Consolidate all override usages to using `getOverrideObject`


abmai:
Replace all instances where we use `getOverride` + `getOverrideProps` with `getOverrideObject`. I added the latter helper with the most recent `Menu` changes and looks like we would always want to pass-thru component + props + style overrides together anyway?

Example code snippet:

```js
export default function OptionList({
  item,
  getItemLabel = (item: *) => (item ? item.label : ''),
  overrides = {},
  ...restProps
}: OptionListPropsT) {
  const {component: ListItem, props: listItemProps} = getOverrideObject(
    overrides.ListItem,
    StyledListItem,
  );
  return (
    <ListItem {...restProps} {...listItemProps}>
      {getItemLabel(item)}
    </ListItem>
  );
}
```

**Pros**
* Consolidate everything into 1 function call
* Less chances of forgetting to do props overriding

**Cons**
* Callsite of `getOverrideObject` can be long

Thoughts? cc @nadiia @schnerd @blackswanny

blackswanny:
ok, good idea

chasestarr:
somewhat related - i've been using [this pattern](https://github.com/uber-web/baseui/blob/master/src/file-uploader/file-uploader.js#L34-L58) in components. feels like it cleans up the render function a bit by reducing the override props spreading

schnerd:
> somewhat related - i've been using [this pattern](https://github.com/uber-web/baseui/blob/master/src/file-uploader/file-uploader.js#L34-L58) in components. feels like it cleans up the render function a bit by reducing the override props spreading

One thing to be wary of is that your logic is creating new react components on every single render. I'm pretty sure this could cause the same bug [encountered here](**url**) with inlined override functions – where focus or other state may not be maintained properly.

That said, wiring up the override logic can certainly feel repetitive/cumbersome, I share your desire to somehow simplify it. A custom babel transform might be one option, but that obviously has its own trade-offs. Happy to discuss idea further.

chasestarr:
good point - i just got bit by that override bug yesterday after i posted that comment

PR #1560, opened 2019-07-13
feat(overrides): prop overrides accept a function too

### LINKED ISSUE #1552: Prop overrides should accept a function
Currently, you can pass a function into `style` and `component` overrides but not props. For example:

<Component overrides={{  props: { 'data-test': 'foo' } }} />

works but

<Component overrides={{  props: props => ({ 'data-test': 'foo' }) }} />

doesn't.

This makes it inconsistent with `style` and `component` overrides and prevents some use-cases like adding BEM class names for different internal states of the component.

qqingvarqq:
Resolves #1552 

#### Description
Make it possible to pass a function into props overrides. 
Rename `mergeStyleOverrides` to `mergeSpecialOverrides` since the same logic apply to `styles` and `props` overrides.
One important thing, since `mergeStyleOverrides` is only used inside `overrides.js` file I suggest that it is an internal function, so renaming it, is not breaking changes. (If it is breaking changes, we can define an alias for `mergeSpecialOverrides` as  `mergeStyleOverrides` and remove him when the new major version released)
Also, updated flow types, updated unit tests for `getOverrideProps` and  `getOverrides`.

#### Scope
- [X] Minor: New Feature
- [X] Major: Maybe Breaking Change


src/helpers/__tests__/overrides.test.js
```diff
@@ -12,7 +12,7 @@ import {
   toObjectOverride,
   mergeOverrides,
   mergeOverride,
-  mergeStyleOverrides,
+  mergeSpecialOverrides,
```
chasestarr: can we rename the utility to something like 'mergeConfigurationOverrides'? the fields props and style are not particularly special (they are two-thirds of the available fields). the difference i see between these fields and 'component' is that these provide small configurations while the other fully replaces

src/helpers/__tests__/overrides.test.js
```diff
@@ -35,6 +35,10 @@ describe('Helpers - Overrides', () => {
       props: {propName: 'propsValue'},
       style: {color: 'blue'},
     };
+    const overridePropsAsFunction = {
+      props: () => ({propName: 'propValue'}),
```
chasestarr: can we also get a test where it reads from its function arguments?
```js
{
  props: p => ({/* ...do something with p */}),
}
```

src/helpers/overrides.js
```diff
@@ -14,9 +14,14 @@ export type StyleOverrideObjectT = {};
 
 export type StyleOverrideT = StyleOverrideObjectT | StyleOverrideFunctionT;
 
+export type PropsOverrideFunctionT = ({}) => ?{};
```
chasestarr: can we get a test asserting what happens when the return value of a props function is null/undefined?

src/helpers/overrides.js
```diff
@@ -14,9 +14,14 @@ export type StyleOverrideObjectT = {};
...
+export type PropsOverrideObjectT = {};
+
+export type PropsOverrideT = PropsOverrideObjectT | PropsOverrideFunctionT;
```
chasestarr: This type (and its sub-types) exactly duplicate `StyleOverrideT`. Can it be de-duplicated into a `ConfigurationOverrideT` and alias `export type StyleOverrideT = ConfigurationOverrideT`?

qqingvarqq:
updated,
renamed `mergeStyleOverrides` to `mergeConfigurationOverrides`,
added more tests,
updated types, `PropsOverrideT` and `StyleOverrideT` replaced by `ConfigurationOverrideT` and alias  `StyleOverrideT`

src/helpers/__tests__/overrides.test.js
```diff
@@ -138,20 +165,46 @@ describe('Helpers - Overrides', () => {
...
-    const overrideObject1 = {color: 'red', textTransform: 'uppercase'};
-    const overrideObject2 = {color: 'blue'};
+  test('mergeOverride can compose props functions', () => {
+    const override1 = {
+      props: () => ({prop1: 'val1', prop2: 'val2'}),
+    };
+    const override2 = {props: () => ({prop1: 'newValue'})};
+    const result = mergeOverride(override1, override2);
+    expect(typeof result.props).toBe('function');
+    // $FlowFixMe style should be a function here
```
chasestarr: what does this mean?

jh3y:
Great stuff! This will close #1552 👍 
On a side note. There is also a proposal (#208) to consolidate the `overrides` utils into one function. I haven't fully dug into this yet, but wondering how much this PR steps into that territory.

src/helpers/__tests__/overrides.test.js
qqingvarqq: Sorry, I forgot to change the comment. Since `mergeOverrideds` now can compose props which are functions I added test case for that. Similar to `mergeOverride can compose style functions`
And since props can be undefined flow warn us in this particular scenario when we call props.

qqingvarqq:
@jh3y This PR does not touch this proposal #208, it just makes it possible to pass a function into props overrides.

PR #1561, opened 2019-07-14
fix(multiselect): set the same gap between the Tags (vertical axis)

### LINKED ISSUE #1550: [Select] fix tag overflow for multiselect
horizontal spacing should be the same between rows

qqingvarqq:
Fixes #1550 

#### Description
Make the same gap between Tags components along the vertical axis. Now gaps along vertical and horizontal axes are the same.

#### Scope

- [X] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/src/select/multi-value.js
+++ b/src/select/multi-value.js
@@ -19,10 +19,10 @@ export default function MultiValue(props: any) {
       overrides={{
         Root: {
           style: ({$theme: {sizing}}) => ({
+            marginTop: sizing.scale0,
             marginRight: sizing.scale0,
-            marginBottom: 0,
+            marginBottom: sizing.scale0,
             marginLeft: sizing.scale0,
-            marginTop: 0,
           }),
         },
       }}


src/select/multi-value.js
```diff
@@ -19,10 +19,7 @@ export default function MultiValue(props: any) {
       overrides={{
         Root: {
           style: ({$theme: {sizing}}) => ({
-            marginRight: sizing.scale0,
-            marginBottom: 0,
-            marginLeft: sizing.scale0,
-            marginTop: 0,
+            margin: sizing.scale0,
```
chasestarr: can you please write out all directions? this is so that we have proper layout with styletron's atomic classes: https://www.styletron.org/concepts/#shorthand-and-longhand-properties

when giving examples we like to be more verbose, someone referencing this code may choose to use shorthand property names if they wish

```suggestion
            marginTop: sizing.scale0,
            marginRight: sizing.scale0,
            marginBottom: sizing.scale0,
            marginLeft: sizing.scale0,            
```

qqingvarqq: Sure
updated

chasestarr: thanks!

jh3y:
Closing in favor of #1158 as this will not cover all cases.

The issue isn't the `Tag` styling, it's actually the fact that the `InputContainer` in the render is causing the flex flow to break. If a user overrides the margins to `0` for all directions, the issue will be seen again.

PR #1563, opened 2019-07-14
feat(table-semantic): add table-semantic components

### LINKED ISSUE #1562: Semantic Table
# Semantic Table

The existing table components use non-standard HTML elements to emulate tables. This is overkill for many simple cases that would benefit from standard semantic HTML tables.

shannonmoeller:
Fixes #1562

#### Description

This PR tracks the addition of a table component that uses semantic tag names compared to `table` and `table-grid` which use divs.

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change


src/styles/types.js
```diff
@@ -605,6 +605,7 @@ export type IconT = {
   ArrowLeft?: ComponentType<IconPropsT>,
   ArrowRight?: ComponentType<IconPropsT>,
   ArrowUp?: ComponentType<IconPropsT>,
+  Blank?: ComponentType<IconPropsT>,
```
chasestarr: why is the blank icon required? does not having it cause the ui to jump?

src/table-semantic/__tests__/table-builder.e2e.js
```diff
@@ -0,0 +1,18 @@
...
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+/* eslint-env node */
+/* eslint-disable flowtype/require-valid-file-annotation */
+
+const {mount, analyzeAccessibility} = require('../../../e2e/helpers');
+
+describe('checkbox', () => {
```
chasestarr:
```suggestion
describe('table-semantic-builder', () => {
```

src/table-semantic/__tests__/table.e2e.js
```diff
@@ -0,0 +1,18 @@
...
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+/* eslint-env node */
+/* eslint-disable flowtype/require-valid-file-annotation */
+
+const {mount, analyzeAccessibility} = require('../../../e2e/helpers');
+
+describe('checkbox', () => {
```
chasestarr:
```suggestion
describe('table-semantic', () => {
```

src/table-semantic/table-builder-column.js
```diff
@@ -0,0 +1,16 @@
...
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+// @flow
+import * as React from 'react';
+
+import type {TableBuilderColumnPropsT} from './types.js';
+
+export default class TableBuilderColumn<T> extends React.Component<
```
chasestarr: What do you think are the trade-offs between this pattern versus a `columns` prop on the parent component? Why did you choose this approach?

src/styles/types.js
shannonmoeller: There is a concept of "unsorted" for a column. Base Web doesn't visualize this, but it may be helpful for projects that override the theme to provide an icon for this state. I'm using an empty icon as a placeholder. This prevents the UI from jumping when the sort state changes and serves as the base for the `SortNoneIcon` override.

src/table-semantic/__tests__/table-builder.e2e.js
shannonmoeller: Whoops!

src/table-semantic/__tests__/table.e2e.js
shannonmoeller: Whoops!

src/table-semantic/table-builder-column.js
shannonmoeller: I wanted to provide a declarative means of defining columns. Using an array of configuration feels out of place when composing a React component and scatters the rendering logic throughout the component.

Consider the common case of rendering a link in a table:

```js
function Users({ data }) {
  return (
    <TableBuilder data={data}>
      <TableBuilderColumn header="Name" sortable>
         {user => <a href={user.profileUrl}>{user.name}</a>}
      </TableBuilderColumn>
      <TableBuilderColumn header="Age" numeric sortable>
         {user => user.age}
      </TableBuilderColumn>
    </TableBuilderColumn>
  );
}
```

vs

```js
function Users({ data }) {
  const columns = [
    {
      header: 'Name',
      renderCell: user => <a href={user.profileUrl}>{user.name}</a>,
      isSortable: true,
    },
    {
      header: 'Age',
      renderCell: user => user.age,
      isNumeric: true,
      isSortable: true,
    },
  ];

  return <TableBuilder columns={columns} data={data} />;
}
```

shannonmoeller: Also, I believe providing the component api provides the best of both worlds as you can still rather easily use an array if that's your preference:

```jsx
function Users({ data }) {
  const columns = [
    {
      header: 'Name',
      children: user => <a href={user.profileUrl}>{user.name}</a>,
      sortable: true,
    },
    {
      header: 'Age',
      children: user => user.age,
      numeric: true,
      sortable: true,
    },
  ];

  return (
    <TableBuilder data={data}>
      {columns.map(col => <TableBuilderColumn {...col} />)}
    </TableBuilder>
  );
}
```

chasestarr: I can agree with the desire to consolidate rendering logic. My concern is that it diverges from how the 'basic' table component consumes an array of column titles. Your last example seems like a reasonable option for anyone prefers an array-based approach. Let's check in on the API before 'stabilizing' the component

shannonmoeller: I totally get that. I thought about not having a columns attribute on the basic table at all, but have a header boolean which assumes the first row of the data set IS the headers like a CSV file. But I wanted to keep the basic api of table-semantic 1:1 with the div-based basic table.

The divergent API of table-builder is the reason I created the additional components instead of baking it all into a single component.

documentation-site/examples/table-semantic/builder.js
```diff
@@ -0,0 +1,36 @@
...
+import {
+  Unstable_TableBuilder,
+  Unstable_TableBuilderColumn,
+} from 'baseui/table-semantic';
+
+const DATA = [
+  {
+    foo: 10,
+    bar: 'banana',
+    url: 'https://example.com/b',
```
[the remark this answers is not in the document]
shannonmoeller: Ha. You bet.

sandgraham:
@shannonmoeller here is a pr against your fork with updated snapshots: https://github.com/shannonmoeller/uber-web-baseui/pull/1 - merge that in when ready (it will update with new commits) and the `vrt` check should pass on this pr.

PR #1564, opened 2019-07-15
fix: dont allow InputContainer to break flex flow

### LINKED ISSUE #1550: [Select] fix tag overflow for multiselect

jh3y:
Fixes #1550
Ensure that the `InputContainer` can't be greater in height than the visible tags 👍 

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


src/select/select.js
```diff
@@ -649,7 +649,7 @@ class Select extends React.Component<PropsT, SelectStateT> {
         />
       );
     }
-
+    console.info(sharedProps);
```
chasestarr:
```suggestion
```

chasestarr:
This pr does a bit more than https://github.com/uber-web/baseui/pull/1561 - can you two collaborate/close one?

qqingvarqq:
`The issue isn't the Tag styling, it's actually the fact that the InputContainer in the render is causing the flex flow to break. If a user overrides the margins to 0 for all directions, the issue will be seen again.`
@jh3y Thanks for the explanation :)
Yes, you are correct on this that `InputContainer` causes the issue.
But since `Tag` has 20px in height and `InputContainer` has 24px those bug occur. And fix is to simply make the `InputContainer` and `Tag` height dimension the same. (Which actually my PR do by adding margins).
Your changes to InputContainer is not a really good idea because, it causes visual regression which you probably will see when running CI (height of multipick-select is less by 4px, you can see regression on a doc site).
And it does not solve the problem when you override `Tag` since if you override the `Tag` you should sure that `tag` and `InputContainer` have the same height. To make it happen you should override `InputContainer` and `Tag`.
@chasestarr Ok, Let's keep this PR. If @jh3y sure that he covers more cases, maybe I miss something.

jh3y:
@qqingvarqq The `Tag` is `24px` in `height`. It's the label `span` inside that's `20px` tall. You can see this by inspecting the elements in any of the storybook scenarios.

Also, `margin` doesn't add to the `height` of the element. This is another issue.

The `Tag` elements used in a multi assume the height of their font `lineHeight`. In this case `24px`. But, the `InputContainer` sets a `height` based on the `font.lineHeight`(24px) and then additionally adds `padding`. Because the `InputContainer` wasn't using `box-sizing: border-box`, it throws things off and the padding increases the `height`. As a precaution, that's why the `InputContainer` is also setting a `maxHeight` based on the font that will be used in it.

houmark:
This change made the input for a multi select with tags less tell in Chrome, which now means it is as misaligned with a normal input as Safari were before. Just thought I'd report it here. I reported this in #1576

PR #1567, opened 2019-07-15
feat(select): add startOpen property

### LINKED ISSUE #744: [Select]  Allow component to mount with dropdown rendered
## Feature description

Either by altering the implementation of autoFocus or by adding a new prop, allow the user to pass in a prop to Select/StatefulSelect that allows it to render the dropdown upon mounting.

Currently, autoFocus does focus the component when it mounts, but it does not trigger a dropdown render. Feel free to play with this code sandbox https://codesandbox.io/s/8k682q3nrl

sandgraham:
Fixes #744 

#### Description

This PR adds an `openOnMount` prop to the `Select` component. If true, the dropdown will open upon mounting. By default the prop is `false`.

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/documentation-site/static/examples/select/open.js
+++ b/documentation-site/static/examples/select/open.js
@@ -1,0 +1,18 @@
+import * as React from 'react';
+import {StatefulSelect} from 'baseui/select';
+
+export default () => (
+  <StatefulSelect
+    options={[
+      {id: 'AliceBlue', color: '#F0F8FF'},
+      {id: 'AntiqueWhite', color: '#FAEBD7'},
+      {id: 'Aqua', color: '#00FFFF'},
+      {id: 'Aquamarine', color: '#7FFFD4'},
+      {id: 'Azure', color: '#F0FFFF'},
+      {id: 'Beige', color: '#F5F5DC'},
+    ]}
+    labelKey="id"
+    valueKey="color"
+    startOpen
+  />
+);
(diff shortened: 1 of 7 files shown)


documentation-site/static/examples/select/open.js
```diff
@@ -0,0 +1,24 @@
...
+      {id: 'Aqua', color: '#00FFFF'},
+      {id: 'Aquamarine', color: '#7FFFD4'},
+      {id: 'Azure', color: '#F0FFFF'},
+      {id: 'Beige', color: '#F5F5DC'},
+    ]}
+    labelKey="id"
+    valueKey="color"
+    overrides={{
+      Root: {
+        // Push the next example down so the dropdown does not block it.
```
chasestarr: let's remove this from the example, and add it to the markdown file. i think you can add an arbitrary div with a style prop

src/select/select.js
```diff
@@ -110,6 +110,9 @@ class Select extends React.Component<PropsT, SelectStateT> {
     if (this.props.autoFocus) {
       this.focus();
     }
+    if (this.props.openOnMount) {
+      this.setState({isOpen: true});
```
tajo: Can this be done in `constructor()`? This is adding a second re-render.

sandgraham: oh right - i knew this felt wrong

PR #1571, opened 2019-07-16
feat: remove input wrapper when no enhancers

### LINKED ISSUE #1551: [Input] remove the extra wrapper around the input if it's not using enhancers

jh3y:
- Fixes #1551 
- Not 100% sure on how to update the focus tests with `ReactWrapper`.
- `e2e` tests added to make sure the `parentElement` is the `Root` 👍 

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/src/input/__tests__/__snapshots__/input.test.js.snap
+++ b/src/input/__tests__/__snapshots__/input.test.js.snap
@@ -1,55 +1,5 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`Input - basic functionality: input has correct props 1`] = `
-Object {
-  "aria-describedby": null,
-  "aria-invalid": false,
-  "aria-label": null,
-  "aria-labelledby": null,
-  "aria-required": false,
-  "autoComplete": "on",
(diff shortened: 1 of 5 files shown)


jh3y:
For me personally, I'm not sure it's that valuable. The main reason in my mind would be that the `Root` container was interfering with styles but `overrides` combats this.

I'm happy to close if you are 👍

gergelyke:
> how valuable do you consider this change? does it improve developer experience in-practice? i think there is often a logical fallacy of `less things === less complexity` but this change in particular introduces varying behavior dependent on configuration. A concrete example is how if there is no parent element, `rootProps` overrides are redirected to the `input` element. How should a developer know of this functionality besides looking into the source code?

I think this makes perfect sense - alternative to this PR, should we simply add an example to the doc site saying that if you don't want the wrapper around the input, use the BaseInput component directly?

jh3y:
That would likely make more sense 👍 

I'll close this one.

PR #1572, opened 2019-07-16
fix(phone-input): correct flow type for inputRef

sandgraham:
#### Description

Fix a flow type error where `inputRef` cannot be `null`.

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/src/phone-input/country-select.js
+++ b/src/phone-input/country-select.js
@@ -21,7 +21,7 @@ import type {CountryT, CountrySelectPropsT} from './types.js';
 
 CountrySelect.defaultProps = {
   disabled: defaultProps.disabled,
-  inputRef: null,
+  inputRef: {current: null},
   maxDropdownHeight: defaultProps.maxDropdownHeight,
   maxDropdownWidth: defaultProps.maxDropdownWidth,
   overrides: {},


src/phone-input/country-select.js
```diff
@@ -21,7 +21,7 @@ import type {CountryT, CountrySelectPropsT} from './types.js';
 
 CountrySelect.defaultProps = {
   disabled: defaultProps.disabled,
-  inputRef: null,
```
chasestarr: I believe if we cast this object to the `$Shape<PropsT>` this bug would have been caught earlier. ex:
```js
CountrySelect.defaultProps = ({
  disabled: defaultProps.disabled,
  inputRef: null,
  maxDropdownHeight: defaultProps.maxDropdownHeight,
  maxDropdownWidth: defaultProps.maxDropdownWidth,
  overrides: {},
  size: defaultProps.size,
}: $Shape<CountrySelectPropsT>);
```

sandgraham: I added this explicit casting for all the other components in phone-input. It brought up a couple other places that were misaligned so I am including the fixes in this PR.

PR #1585, opened 2019-07-17
feat: new pin-code component

sandgraham:
#### Description

A new component: the `PinCode`!

#### The shpeel

flow

As you type in codes, focus transfers to the next input. When you delete, focus transfers backward. You can also copy + paste (or autocomplete on some mobile browsers):

pasta

Convenient on mobile too:

mobile-flow

#### API

`PinCode` takes a `values` prop- an array of strings and creates inputs for each value in the array. When there is a change in one of the inputs, some logic runs to determine what `PinCode` thinks the new set of values should be. It calls the `onChange` prop with this new value array. The consumer can determine if the new values look legit and update the state or do something else.

Right now you can't really access the individual inputs from a consuming app. I've added an `ids` and `names` prop that takes an array of string to map over the inputs, but otherwise, the inputs are out of reach. I'm not sure if consumers will need a way to control or interact with the individual inputs. If we see that is the case, then we could change the implementation to allow consumers to render the individual inputs themselves. Maybe something like this: 
```jsx
<PinCode.Container>
  <PinCode.Input />
</PinCode.Container>
```

Please let me know if you think the API is clunky or anything useful is missing. Thank you!

#### Maybe

Here are a few things I am considering. I don't think they need to make it into this version though, so we can gauge demand later on.

- Generalize the ability to pass props to individual inputs? (id, name, value)
- What else should `onChange` pass? (refs, event)
- Add more "shared" handlers?

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change


documentation-site/pages/components/pin-code.mdx
```diff
@@ -0,0 +1,112 @@
...
+  <PinCodeBasic />
+</Example>
+
+Each input collects one digit at a time.
+When a digit is entered, focus transfers to the next input in the sequence, until every input is filled in.
+A few things to note:
+
+1. You can't enter anything other than a digit.
+2. You can only enter one character per input.
+3. To overwrite the value of an input, you don't need to select the text in the input. If you type a digit while the input is focused, the input will update. This is an optimization for quickly editing the inputs.
```
leyanlo:
```suggestion
3. To overwrite the value of an input, you don’t need to select the text in the input. If you type a digit while the input is focused, the input will update. This is an optimization for quickly editing the inputs.
```

documentation-site/pages/components/pin-code.mdx
```diff
@@ -0,0 +1,112 @@
...
+
+<Example title="Basic Usage" path="examples/pin-code/basic.js">
+  <PinCodeBasic />
+</Example>
+
+Each input collects one digit at a time.
+When a digit is entered, focus transfers to the next input in the sequence, until every input is filled in.
+A few things to note:
+
+1. You can't enter anything other than a digit.
```
leyanlo:
```suggestion
1. You can’t enter anything other than a digit.
```

documentation-site/pages/components/pin-code.mdx
```diff
@@ -0,0 +1,112 @@
...
+<Example title="Passing id and name" path="examples/pin-code/id-and-name.js">
+  <PinCodeIdAndName />
+</Example>
+
+In a simliar fashion you can pass individual `id` and `name` props to each input.
+Just provide an array of strings to `ids` or `names` and each string will be passed through to the respective input.
+
+These are the only props that will target individual inputs.
+Every other prop that `PinCode` accepts is "shared" between the inputs.
+Hopefully you don't have many reasons to target the inputs individually.
```
leyanlo:
```suggestion
Hopefully you don’t have many reasons to target the inputs individually.
```

documentation-site/pages/components/pin-code.mdx
```diff
@@ -0,0 +1,112 @@
...
+
+<Example title="Passing id and name" path="examples/pin-code/id-and-name.js">
+  <PinCodeIdAndName />
+</Example>
+
+In a simliar fashion you can pass individual `id` and `name` props to each input.
+Just provide an array of strings to `ids` or `names` and each string will be passed through to the respective input.
+
+These are the only props that will target individual inputs.
+Every other prop that `PinCode` accepts is "shared" between the inputs.
```
leyanlo:
```suggestion
Every other prop that `PinCode` accepts is “shared” between the inputs.
```

documentation-site/pages/components/pin-code.mdx
```diff
@@ -0,0 +1,112 @@
...
+You can control the size of the inputs with the `size` prop.
+
+<Example
+  title="Changing the placeholder"
+  path="examples/pin-code/placeholder.js"
+>
+  <PinCodePlaceholder />
+</Example>
+
+If you don't like the default placeholder (`○`), you can change that too.
```
leyanlo:
```suggestion
If you don’t like the default placeholder (`○`), you can change that too.
```

documentation-site/pages/components/pin-code.mdx
```diff
@@ -0,0 +1,112 @@
...
+Notice how each placeholder disappears when any of the inputs have focus.
+
+<Example title="Disable auto-focusing" path="examples/pin-code/no-tab.js">
+  <PinCodeNoTab />
+</Example>
+
+You can control the automatic focusing logic built into `PinCode` using the `tabOnChange` prop.
+It defaults to `true`, which makes it so that when an input recieves a digit, focus is automatically changed to the next input.
+Set it to `false` if, for some reason, you don't want this behavior.
```
leyanlo:
```suggestion
Set it to `false` if, for some reason, you don’t want this behavior.
```

documentation-site/pages/components/pin-code.mdx
```diff
@@ -0,0 +1,112 @@
...
+
+The code defaults to a length of 4, but you can change that.
+`PinCode` expects an array of strings for the `values` prop.
+The number of inputs rendered will match the length of the array.
+
+<Example title="Passing id and name" path="examples/pin-code/id-and-name.js">
+  <PinCodeIdAndName />
+</Example>
+
+In a simliar fashion you can pass individual `id` and `name` props to each input.
```
leyanlo:
```suggestion
In a similar fashion you can pass individual `id` and `name` props to each input.
```

documentation-site/pages/components/pin-code.mdx
```diff
@@ -0,0 +1,112 @@
...
+
+If you don't like the default placeholder (`○`), you can change that too.
+Notice how each placeholder disappears when any of the inputs have focus.
+
+<Example title="Disable auto-focusing" path="examples/pin-code/no-tab.js">
+  <PinCodeNoTab />
+</Example>
+
+You can control the automatic focusing logic built into `PinCode` using the `tabOnChange` prop.
+It defaults to `true`, which makes it so that when an input recieves a digit, focus is automatically changed to the next input.
```
leyanlo:
```suggestion
It defaults to `true`, which makes it so that when an input receives a digit, focus is automatically changed to the next input.
```

sandgraham:
Who needs a spell-checker when I have @leyanlo? 😜

But for real - I'll do a pass on all the text before I un-WIP it.

gergelyke:
This looks great! A few things to consider/look into:

- using type=number on the input, so on mobile you'll get the mobile keyboard
- for some reason, Chrome tries to autocomplete the inputs with my ZIP code
- "Keep in mind that folks may have mistyped a digit so auto-submitting might be a bit aggressive." - I believe lots of apps work that way - should we change the wording here / add an example for that use case too?

sandgraham:
- `type=number` results in the little up and down ticks on the input and scrolling through numbers which i think would be a little annoying. i'm trying out `type=tel` with a `pattern="\d*"` to get the number pad on mobile. will see how that goes
- autocomplete - not sure about this - i recreated an input in codesandbox with same type/pattern/autocomplete/etc and it does not appear - so i'm still working this out. i think in my case i clicked on the zip code at one point and i think chrome will never let me get away from it now.
- so that last example is to showcase that you can do something when all the inputs are filled in - whether its submitting or focusing something else is kind of up to the app. i was going to add an `onCompletion` to encourage this but figured this example did the trick. i could change the example to prompt an alert instead of focusing a button but i feel like either way the point is made?

I also spent a while trying to add an e2e test for the copy/paste & autocomplete handling but puppeteer doesn't make it obvious how to do pasting - i may have to just generate an input event directly on the element but that is pretty hacky and getting close to implementation. not sure its worth it at that point.

gergelyke:
> * `type=number` results in the little up and down ticks on the input and scrolling through numbers which i think would be a little annoying. i'm trying out `type=tel` with a `pattern="\d*"` to get the number pad on mobile. will see how that goes
> * autocomplete - not sure about this - i recreated an input in codesandbox with same type/pattern/autocomplete/etc and it does not appear - so i'm still working this out. i think in my case i clicked on the zip code at one point and i think chrome will never let me get away from it now.
> * so that last example is to showcase that you can do something when all the inputs are filled in - whether its submitting or focusing something else is kind of up to the app. i was going to add an `onCompletion` to encourage this but figured this example did the trick. i could change the example to prompt an alert instead of focusing a button but i feel like either way the point is made?
> 
> I also spent a while trying to add an e2e test for the copy/paste & autocomplete handling but puppeteer doesn't make it obvious how to do pasting - i may have to just generate an input event directly on the element but that is pretty hacky and getting close to implementation. not sure its worth it at that point.

cool, that sounds great - yeah, feel free to ignore the 3rd point

sandgraham:
ok - i think i've got the silly autocomplete and keyboard issues under control

package.json
```diff
@@ -152,6 +152,7 @@
     "react-input-mask": "^2.0.4",
     "react-is": "^16.8.6",
     "react-movable": "^2.2.0",
+    "react-multi-ref": "^1.0.0",
```
chasestarr: is this package required? seems like it's only a couple lines and is not widely used. writing our own code to do this will provide faster service to consumers in the case of bugs possibly resulting from it

tajo:
Looks very nice and the examples are great!

PR #1586, opened 2019-07-18
fix(avatar): make styled component prop types less restrictive

### LINKED ISSUE #1582: [Avatar] styled component prop types are too restrictive
StylePropsT ([link](https://github.com/uber-web/baseui/blob/master/src/avatar/types.js#L11-L14)) is an 'exact' type and will not allow for children to be provided to the component. My suspicion as to why this error is not surfaced in baseui source code is because of wrapping in overrides code

```js
import { StyledInitials } from 'baseui/avatar';
const Initials = withStyle<typeof StyledInitials>(StyledInitials, {color: 'blue'});

function MyAvatar() {
  return <Initials>name</Initials>; // flow errors here
}
```

qqingvarqq:
Fixes #1582    

#### Description
Make `StylePropsT` not 'exact' type.
Add `AvatarStylePropsT RootStylePropsT InitialsStylePropsT` types for `Avatar Root,Initials ` styled components.

#### Scope

- [X] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


gergelyke:
thanks @qqingvarqq ! 🤗

qqingvarqq:
added tests
@gergelyke @chasestarr  you are welcome :)

chasestarr:
thanks!

PR #1589, opened 2019-07-18
Add flow to examples

tajo:
Sorry for a massive PR but it didn't make sense to do piecemeal it. It almost exclusively changes `documentation-site` and `examples`, not `src` (I did a few flow fixes tho, so you should definitely check those out). The risk for `baseui` consumers should be minimal.

### Changes
- Add flow types to all examples, sometimes I took a shortcut with `any` and `$FlowFixMe` when dealing with local state in some of these more complex examples (like in tables); however, we now "flow test" all our APIs. There were some errors but the number was small. That should not be surprising since `baseui` is now used in hundreds of projects. I expect this process will be way more painful when adding TypeScript examples. 
- Examples are now eagerly loaded with Webpack instead `fetch()`ing them through XHR:

```js
const codeFlow = await import(/* webpackMode: "eager" */ `!!raw-loader!../examples/${
  this.props.path
}`);
const codeJs = await import(/* webpackMode: "eager" */ `!!raw-loader!remove-flow-types-loader?pretty!../examples/${
  this.props.path
}`);
```

You can see it's loaded twice, the second time `remove-flow-types-loader` is applied, so we can keep showing/using vanilla JS examples as well. `eager` mode means that examples are bundled with the parent bundle instead of loaded in separate requests (examples are typically very small, so saving some extra requests per page).

With that, `examples` folder is moved up since it doesn't make sense to keep it as a static asset.

- Show both JS and Flow source code. We will add TS as a third option.
- Prettier's `printWidth` set to 68 for examples so they better fit our documentation website.

sc


documentation-site/components/example.js
```diff
@@ -62,22 +63,31 @@ type PropsT = {
...
     source: null,
+    sourceFlow: null,
   };
 
   async componentDidMount() {
-    const sourcePath = `${String(process.env.STATIC_ROOT)}${this.props.path}`;
-    const res = await fetch(sourcePath);
-    const source = await res.text();
-    this.setState({source});
+    //const sourcePath = `${String(process.env.STATIC_ROOT)}${this.props.path}`;
```
chasestarr:
```suggestion
```

documentation-site/components/example.js
```diff
@@ -107,24 +117,48 @@ class Example extends React.Component<PropsT, StateT> {
...
+              }
             }}
           >
-            {this.state.isSourceOpen ? 'Hide' : 'Show'} Source
-          </Button>
+            <Button
+              kind={KIND.secondary}
+              startEnhancer={() => <CodeIcon />}
+              onClick={() => {
+                trackEvent('show_js_source', this.props.title);
```
chasestarr: is it worth maintaining the `show_source` event? i'm not sure if metrics are measured here cc @gergelyke

src/block/types.js
```diff
@@ -334,7 +334,7 @@ export type BlockPropsT = {
   placeItems?: ResponsiveT<string>,
   /** available values: https://developer.mozilla.org/en-US/docs/Web/CSS/place-self */
   placeSelf?: ResponsiveT<string>,
-  flexWrap?: ResponsiveT<boolean>,
+  flexWrap?: ResponsiveT<string | boolean>,
```
chasestarr: I would reference the block styled component. this prop should be pinned to boolean if any provided string will coerce the result to `wrap`

src/dnd-list/types.js
```diff
@@ -79,4 +81,6 @@ export type SharedStylePropsArgT = {
   $isRemovable: boolean,
   $isRemovableByMove: boolean,
   $isOutOfBounds: boolean,
+  // $FlowFixMe
+  $value: string,
```
chasestarr: why the suppression here?

src/locale/index.js
```diff
@@ -13,7 +13,10 @@ import en_US from './en_US.js';
 
 export const LocaleContext: React.Context<LocaleT> = React.createContext(en_US);
 
-const LocaleProvider = (props: {locale: LocaleT, children: ?React.Node}) => {
+const LocaleProvider = (props: {
+  locale: $Shape<LocaleT>,
```
chasestarr: shape marks all fields as optional, which is not correct. type spread will coerce an exact type to inexact `{...LocaleT}`

documentation-site/components/example.js
tajo: I've added these metrics originally. I think we can just sum up `show_js_source` and `show_flow_source` if we ever want to compare it with `show_source`.

gergelyke: would it be possible to always fire the show_source event, to make it easier?

src/locale/index.js
tajo: This is intentional. We are doing a deep merge with the defaults (English). You should be able to pass only a part of locale file (like a single component). This was breaking the doc example.

gergelyke:
👏

src/dnd-list/types.js
tajo: fixed, update to to `React.Node`, flow was being annoying when doing this

```js
`${$value} label`
```

needs to be changed to

```js
`${typeof $value === 'string' ? $value : ''} label`
```

since it can't coerce `React.Node` subtypes to string

documentation-site/components/example.js
chasestarr: summing should be fine - i just wasn't aware if the data was being used

src/locale/index.js
chasestarr: ah right 👍

documentation-site/components/example.js
tajo: > would it be possible to always fire the show_source event, to make it easier?

I am quite interested to see what's the favorite language of our users.

We could fire `show_source` in parallel but I felt it's not needed if we have a more fine-grained metric and we can sum it up.

qqingvarqq:
Hi @tajo, I've noticed that this PR brings flow warnings https://github.com/uber-web/baseui/blob/65b2b6b6355fb57d730819030a3f15fc05e7fff2/scripts/build-docs.js#L146
Can you look at this?

PR #1592, opened 2019-07-19
feat: improve SEO markup for breadcrumbs

### LINKED ISSUE #1556: [SEO Improvements] Breadcrumbs

jh3y:
### Description
- Resolves #1556 

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change


documentation-site/static/examples/breadcrumbs/pseudo.js
```diff
@@ -0,0 +1,31 @@
...
+
+export default () => (
+  <Breadcrumbs
+    overrides={{
+      ListItem: {
+        style: ({$itemIndex, $theme}) => {
+          if ($itemIndex === 0) return {};
+          return {
+            ':before': {
+              content: "'>'",
```
gergelyke: should we make this the default behaviour?

src/breadcrumbs/breadcrumbs.js
```diff
@@ -10,43 +10,35 @@ LICENSE file in the root directory of this source tree.
...
   const {overrides = {}} = props;
-  const numChildren = Children.count(props.children);
   const childrenWithSeparators = [];
 
   const [Root, baseRootProps] = getOverrides(overrides.Root, StyledRoot);
-  const [Icon, baseIconProps] = getOverrides(overrides.Icon, ChevronRight);
-  const [Separator, baseSeparatorProps] = getOverrides(
-    overrides.Separator,
-    StyledSeparator,
+  const [List, baseListProps] = getOverrides(overrides.List, StyledList);
```
gergelyke: this way, it would be a breaking change, isn't it? do you think we can get away without one?

src/breadcrumbs/types.js
```diff
@@ -13,12 +13,14 @@ import type {OverrideT} from '../helpers/overrides.js';
 
 export type OverridesT = {
   Root?: OverrideT<*>,
-  Separator?: OverrideT<*>,
-  Icon?: OverrideT<*>,
```
chasestarr: We can't remove the current functionality, this would likely break many applications. Can this change be implemented in a way that supports what currently exists?

src/breadcrumbs/breadcrumbs.js
jh3y: I'm not sure how we can do that in a way that won't break it. That's why the PR was how it was originally. It shows how you can use the best practice for markup if you want it. You could pass through `Separator` overrides only to the pseudo element but `component` and `props` would be ignored. Or you stick with how it is and override the `Separator` to be `null` and override the `style` on the `ListItem`.

The issue is that currently the separators are injected and they aren't the "best practice" for SEO/accessibility.

src/breadcrumbs/types.js
jh3y: I did have an example of how to use this and implement the "best practice" markup by overriding the `Separator` to `null` but @gergelyke wondered if the pseudo behavior could be implemented as the default.

documentation-site/pages/components/breadcrumbs.mdx
```diff
@@ -27,6 +28,10 @@ the user’s location in an application.
   <BasicBreadcrumbs />
 </Example>
 
+<Example title="Breadcrumbs with pseudo separators" path="examples/breadcrumbs/pseudo.js">
```
gergelyke: can you add some description here that we recommend using this for apps where SEO is important?

documentation-site/pages/components/breadcrumbs.mdx
```diff
@@ -27,6 +28,15 @@ the user’s location in an application.
...
 </Example>
 
+<Example
+  title="Breadcrumbs with pseudo separators"
+  path="breadcrumbs/pseudo.js"
+>
+  <PseudoBreadcrumbs />
+</Example>
+
+This is the recommended approach where SEO/Accessibility is important.
```
gergelyke:
```suggestion
This is the recommended approach where SEO/Accessibility is important. Later on, we may make this the default functionality, but for now, we did not want to introduce a breaking change.
```

PR #1595, opened 2019-07-22
fix(popover): make styled-components prop types less strict

### LINKED ISSUE #1559: [Popover] SharedStylePropsArgT type is too strict
```js
import {StyledArrow, PLACEMENT} from 'baseui/popover';

<StyledArrow
  $arrowOffset={{top: 0, left: 0}}
  $placement={PLACEMENT.right} 
/>
```

I should only need to specify the fields in the example since others specified are not required by the styled function.

qqingvarqq:
Fixes #1559 

#### Description
1. Add `ArrowStylePropsArgT, BodyStylePropsArgT, InnerStylePropsArgT` types which represent the required properties for corresponding components.
2. Add flow tests for those types.
----
Two more things. 
1. [Here](https://github.com/qqingvarqq/baseui/blob/1c9ff18a174067a7edc35968b7c971290c7b57c4/src/popover/types.js#L180) not possible to use the intersection of types because of https://github.com/facebook/flow/issues/7946 you can reproduce this bug by uncommenting the line.
And you get flow errors in
https://github.com/qqingvarqq/baseui/blob/fix-popup-styled-prop-types/src/popover/__tests__/styled-component.test.js
and 
https://github.com/qqingvarqq/baseui/blob/fix-popup-styled-prop-types/src/tooltip/__tests__/tooltip.test.js
2. If you try to override the Body of popover as the entire component. You need to pass ref to Body to get the expected behavior, and without going to codebase it's not possible to figure out.
I can add an [example](https://codesandbox.io/s/popover-opens-on-click-5qmmg?fontsize=14) to popover doc if we need this?

#### Scope

- [X] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


src/popover/types.js
```diff
@@ -161,6 +161,24 @@ export type PopoverPrivateStateT = {
...
+  $isAnimating: boolean,
+  $isOpen: boolean,
+  $popoverOffset: OffsetT,
+  $placement: TetherPlacementT,
+  $showArrow: boolean,
+};
+
+export type InnerStylePropsArgT = {};
+/*
+* Can't use Intersection types because of https://github.com/facebook/flow/issues/7946
```
chasestarr: spreading with exact types may work for you https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVBLAtgBzgJwBdkwBDAZzACUBTUgY2KnzizAHJ87H2BudGgA88RMIQCeOGmCgZ85YgF4wAbwA+qMGADWNcQEYAXGAX4MAOwDmAGlRqAvv1RCRxCVJM16ccwBMwyuqaOnoATMamFjZ2jgLCBG6S0gBGcIQAFgGqwQB0ebLyhLZaeTnkXj6+trGgYM4AbjT44hlRYBiUcNpgCBgZcACuxBaETeWMGD45zvGi7ilpmYGoAJC6BoaRVrZrYZuEZtuosShxrmJJYN64PjTmStTchDkAwqx45neEACpJADypDJgABkqkWTQA0npjKRzOJ7AA+JxQAbmCY+MCWGiEN43T73AAUAEpjNcPl9slouIQBvhzGAcCwcJRFAiwH9fBh6giVIy4MycuD8FD4bUmix8GB0k1pAhpfTAZkOu17mMvIRJvT5ig-sBOdz+PZUN5zAowLicFksTj3rdCUSnC4EmBfDQoKQBjBiMSAgiCX8LSEDIoAET6ENB0Kh0IRoUi0Ph4AIh1AA

chasestarr: we will need to maintain `SharedStylePropsArgT` before merging this change

qqingvarqq: @chasestarr What you mean 'maintain SharedStylePropsArgT'. `SharedStylePropsArgT` is present here https://github.com/qqingvarqq/baseui/blob/1c9ff18a174067a7edc35968b7c971290c7b57c4/src/popover/types.js#L182
if need i can replace 
```
export type SharedStylePropsArgT = {
  $arrowOffset: OffsetT,
  $isAnimating: boolean,
  $isOpen: boolean,
  $popoverOffset: OffsetT,
  $placement: TetherPlacementT,
  $showArrow: boolean,
};
```
with 
```
export type SharedStylePropsArgT = {
...$Exact<ArrowStylePropsArgT>,
...$Exact<BodyStylePropsArgT>,
};
```

chasestarr: ah sorry - i saw it commented and thought it was removed

chasestarr: the second option you have may be ideal

qqingvarqq: @chasestarr updated

PR #1598, opened 2019-07-22
feat(icon): add mask toggle when type is password

sandgraham:
#### Description

This PR updates `Input` to add a "mask" toggle when the `type` prop is set to `"password"`.

togglemask

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change


tajo:
Looking good!

> automatically doing this when type="password"

I like this idea. I think that "show/hide" should be the default option when using `type="password"` and you can use override if you want to disable it.

documentation-site/examples/input/password.js
```diff
@@ -0,0 +1,38 @@
...
+      overrides={{
+        After: () => {
+          return (
+            <Button
+              kind={KIND.minimal}
+              title={title}
+              onClick={() => setHidden(h => !h)}
+            >
+              {hidden ? (
+                <ViewShow title={title} color="black" />
```
tajo: Screen Shot 2019-07-22 at 5 31 28 PM

The color should be dynamic.

sandgraham:
I'm going to try out the `type="password"` approach. Will update PR in a bit.

sandgraham:
I've updated the PR to bake in the logic when `type` is `"password"`.

src/input/base-input.js
```diff
@@ -128,6 +132,78 @@ class BaseInput<T: EventTarget> extends React.Component<
...
+      Button,
+    );
+    const baseButtonOverrides = {
+      BaseButton: {
+        style: ({$theme}) => ({
+          color: $theme.colors.foreground,
+        }),
+      },
+    };
+    // $FlowFixMe
```
tajo: what's up with those?

sandgraham: `mergeOverrides` has some flow type deficiencies atm - pretty much anywhere we use it we have to add a fix me comment because referencing the specific override property doesn't pass flow checks.

we should look into a better solution probably.

PR #1599, opened 2019-07-23
TypeScript Examples

tajo:
All flow (.js) examples now have .tsx counterparts. Visible to users and also integrated with CodeSandbox.

Screen Shot 2019-07-22 at 4 59 18 PM


documentation-site/examples/side-navigation/basic.js
```diff
@@ -1,3 +1,4 @@
+// @flow
```
chasestarr:
```suggestion
```

scripts/validate-examples.js
```diff
@@ -0,0 +1,37 @@
...
+*/
+
+/* eslint-env node */
+// @flow
+const fs = require('fs');
+const fsPromises = require('fs').promises;
+const globby = require('globby');
+const path = require('path');
+
+(async () => {
```
chasestarr: 🙇

documentation-site/examples/side-navigation/basic.js
chasestarr: saw a couples instances of this

tajo: yea, it should be in every single `.js` file since we want to enable flow typecheck?

tajo:
> getting `Cannot find module 'styletron-engine-atomic'.ts(2307)` errors when running the code sandbox. is that to be expected?

Not expected. What example is failing?

documentation-site/examples/side-navigation/basic.js
chasestarr: hah oh right

chasestarr:
this link - https://codesandbox.io/s/0swb2?module=/example.tsx (first button example) it took about 15 seconds to display the error

tajo:
> this link - https://codesandbox.io/s/0swb2?module=/example.tsx (first button example) it took about 15 seconds to display the error

Oh right, added missing `@types/styletron-foo` packages. Good catch.

PR #1604, opened 2019-07-23
slider flow types

chasestarr:
@jh3y You can use these changes as a reference. The import fails flow check because it's not being exported from the react-range package index file. See the contents I find in node_modules locally below

```js
// @flow
import Range from './Range';
import { getTrackBackground } from './utils';
import { Direction } from './types';

export { Range, getTrackBackground, Direction };
```


jh3y:
I did already open a fix for the missing `flow` type yesterday 👍 https://github.com/tajo/react-range/pull/25

gergelyke:
closing this, as it seems to be incorporated in an other changeset

PR #1606, opened 2019-07-23
fix(toast): position close icon to the first line of text

gergelyke:
Closes https://github.com/uber-web/baseui/issues/1601

### DIFF
--- a/src/toast/styled-components.js
+++ b/src/toast/styled-components.js
@@ -149,5 +149,6 @@ export const CloseIconSvg = styled<SharedStylePropsArgT>('svg', props => {
     cursor: 'pointer',
     width: '16px',
     flexShrink: 0,
+    paddingTop: '2px',
   };
 });


src/toast/styled-components.js
```diff
@@ -149,5 +149,6 @@ export const CloseIconSvg = styled<SharedStylePropsArgT>('svg', props => {
     cursor: 'pointer',
     width: '16px',
     flexShrink: 0,
+    paddingTop: '2px',
```
chasestarr: arbitrary padding feels a bit restrictive, consider the situation if one overrides with a larger icon. does centering vertically work well?

nguyensomniac: Yeah, that's a good point. I don't think there's a good way to do this scalably.

If we're choosing between vertically centering and keeping it as is, I think I'd rather do the latter.

gergelyke: sounds good - I'll close this, and let's leave things as they are

PR #1619, opened 2019-07-23
fix(breakpoints): stable breakpoint keys

alexzywiak:
#### Description

Javascript objects do not guarantee stable property order.  When generating media queries, the breakpoints could be referenced in the wrong order and the resulting media queries would apply to the wrong style values.

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


gergelyke:
@alexzywiak thanks for the contribution! is this something we could add test coverage for?

alexzywiak:
@gergelyke I added a test in there.

PR #1625, opened 2019-07-24
feat: seo improvements for pagination

### LINKED ISSUE #1608: [SEO Improvements] Pagination
## Feature description
- Convert `Button` to anchors with `$as="a"` potentially

jh3y:
#### Description
- Fixes #1608 

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/src/pagination/__tests__/pagination.e2e.js
+++ b/src/pagination/__tests__/pagination.e2e.js
@@ -10,8 +10,8 @@ LICENSE file in the root directory of this source tree.
 const {mount, analyzeAccessibility} = require('../../../e2e/helpers');
 
 const selectors = {
-  prevButton: 'button[data-test="prev-button"]',
-  nextButton: 'button[data-test="next-button"]',
+  prevButton: 'a[data-test="prev-button"]',
+  nextButton: 'a[data-test="next-button"]',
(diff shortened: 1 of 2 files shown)


gergelyke:
can we keep the original look and feel?

Screen Shot 2019-07-25 at 9 46 57 AM

jh3y:
Maybe we should close this one. It's to do with the `disabled` attribute. I didn't notice in the scenario that we are using `disabled` on the `Button` dependent on the `currentPage`.

The alternative is to style the anchor tag based on the `currentPage` and conditionally fire the `onPrevClick` handler. We don't have styling in place for a `disabled` button though so it would be a case of trying to match it up.

gergelyke:
> Maybe we should close this one. It's to do with the `disabled` attribute. I didn't notice in the scenario that we are using `disabled` on the `Button` dependent on the `currentPage`.
> 
> The alternative is to style the anchor tag based on the `currentPage` and conditionally fire the `onPrevClick` handler. We don't have styling in place for a `disabled` button though so it would be a case of trying to match it up.

Sounds good - I am okay with closing this. I am following up with the SEO consultant because some of the recommendations seem really hard to follow.

jh3y:
Thanks 👍

PR #1626, opened 2019-07-24
feat: seo improvements for card

### LINKED ISSUE #1615: [SEO Improvements] Card
## Feature description
- Text content within `p` rather than `div`.
- Heading should be at most `h2`.
- If the CTA is a navigational link, don't make it a `Button`. It should be an anchor.

jh3y:
#### Description
- Resolves #1615 
- Switch out containing body to `div` whilst normal text content body should be `p`.
- Switch out header to `h2` 

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change


src/card/card.js
```diff
@@ -83,7 +83,9 @@ function Card(props: CardsPropsT) {
             {title}
           </Title>
         )}
-        <Body {...getOverrideProps(BodyOverride)}>{children}</Body>
+        <Body $as="div" {...getOverrideProps(BodyOverride)}>
```
chasestarr: why as div?

jh3y: According to SEO docs, only the text content goes in a `p` but the card examples wrap a `Body` in a `Body`. I guess it can be done as `p` inside a `p`.

src/card/styled-components.js
```diff
@@ -13,7 +13,7 @@ export const Action = styled<{}>('div', ({$theme: {typography}}) => ({
   ...typography.font450,
 }));
 
-export const Body = styled<{}>('div', ({$theme}) => {
+export const Body = styled<{}>('p', ({$theme}) => {
```
gergelyke: I am a bit worried, that with this if someone wants to add multiple paragraphs for the card, we'll end up with the same issues as with the modal. I've just sent an email to the SEO consultant to ask for recommendation

gergelyke:
we've discussed this offline, and we decided not to work on these - will close all the SEO related PRs/issues

PR #1630, opened 2019-07-24
feat(datepicker): horizontal calendar layout for multiple months

### LINKED ISSUE #1628: [Datepicker] Support a horizontal layout in the datepicker calendar
Would love it if the calendar in the datepicker supported a horizontal layout

jonsadka:
Fixes #1628

#### Description

Control the layout of the multiple month view by specifying an orientation of `horizontal` or `vertical` using `ORIENTATION`

#### Scope

- Add example to documentation
- Unit test
- Only display the Previous and Next month buttons at the bookend of the component

Kapture 2019-07-24 at 11 57 35


gergelyke:
Thanks @jonsadka !

src/datepicker/calendar.js
```diff
@@ -70,6 +72,7 @@ export default class Calendar extends React.Component<
     onMonthChange: () => {},
     onYearChange: () => {},
     onChange: () => {},
+    orientation: ORIENTATION.vertical,
```
chasestarr: let's make the default horizontal

src/datepicker/calendar.js
```diff
@@ -369,8 +377,11 @@ export default class Calendar extends React.Component<
           />
         </CalendarContainer>,
       );
+      monthList.push(<div key={`month-${i}`}>{monthSubComponents}</div>);
```
chasestarr: is the wrapping div required? we try to allow all elements to be overridable

chasestarr: `monthKey` on L345 is the same as this line's key prop

src/datepicker/types.js
```diff
@@ -163,6 +166,7 @@ export type CalendarPropsT = {
 
 export type HeaderPropsT = CalendarPropsT & {
   date: Date,
+  order: number,
```
chasestarr: can you describe the `order` prop's purpose? I have a hard time reasoning about what it does by reading the type definition

chasestarr: to put it more clearly, can this prop be named something more descriptive? maybe index?

src/datepicker/calendar-header.js
```diff
@@ -97,6 +102,26 @@ export default class CalendarHeader extends React.Component<
     }
   };
 
+  isHiddenPaginationButton = (direction: $Values<typeof DIRECTION>) => {
```
chasestarr: is this better described as 'disabled` rather than hidden?

src/datepicker/__tests__/datepicker.test.js
```diff
@@ -151,4 +154,28 @@ describe('Datepicker', () => {
 
     expect(renderedInput.props().value).toEqual('2019/01/01 - 2019/01/04');
   });
+
+  test('calendar popover renders multiple months and hides redundant UI', () => {
```
chasestarr: what ui is being hidden and what line here tests that? I would expect to see some tests related to hiding the arrows

src/datepicker/calendar-header.js
jonsadka: I think this might be confusing because on line 138 I also set this value to determine if the button is disabled, but this actually does control if the arrow is hidden

src/datepicker/types.js
jonsadka: This prop [already existed](https://github.com/uber-web/baseui/blob/6389c9b3e3044ba983487e1e22189b05a3f4158c/src/datepicker/calendar.js#L172) but wasn't defined in the props for some reason. I dont mind changing the name but would rather issue it in a separate diff

src/datepicker/calendar.js
jonsadka: > is the wrapping div required? we try to allow all elements to be overridable

Unfortunately yes, it's needed for styling but also makes sense semantically since it groups each month within it's own block. Thoughts?

> `monthKey` on L345 is the same as this line's key prop

Good call out, will make changes

src/datepicker/calendar.js
jonsadka: FYI this would be a breaking change

src/datepicker/__tests__/datepicker.test.js
jonsadka: This test includes the multiple-month feature test and the hiding arrow test into one test, but based on this feedback I separated them into two distinct tests

src/datepicker/calendar.js
gergelyke: oh wow, I had no idea this is already supported https://codesandbox.io/s/datepicker-t6rh2 🤦‍♂

jonsadka:
@chasestarr @gergelyke addressed code review comments and left some follow up questions

Screen Shot 2019-07-24 at 4 51 03 PM

src/datepicker/calendar.js
chasestarr: It's probably fine for now. if overriding this div is necessary for someone, we can add in the override later

src/datepicker/calendar-header.js
chasestarr: Can you share reproduction steps for when the arrow is hidden? I clicked the right arrow until 'December 2030' is on the right-hand side. This displays a grey arrow and disabled button. Based on this [line](https://github.com/uber-web/baseui/pull/1630/files#diff-92ab803c37c9efa8648d2fa9682bc9c8R227), I would expect the arrow to not be rendered.

What I'm trying to get at is - why was this additional check needed in addition to disabling the button?

jonsadka: Following up with a discussion I had with @chasestarr, we decided that the current implementation is the correct solution for the use case here where each month has it's own dedicated header

PR #1638, opened 2019-07-25
fix: add breadcrumbs pseudo tsx example

jh3y:
- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


tajo:
Hm, strange that you were able to pass the CI yesterday with this missing .tsx example. It should fail the build.

PR #1640, opened 2019-07-25
fix(ci): add buildkite tasks for typescript and example validation

tajo:
We are not currently enforcing `tsc` and `validate:examples` scripts through buildkite.


chasestarr:
this line should change i think https://github.com/uber-web/baseui/blob/master/.flowconfig#L14

tajo:
> this line should change i think https://github.com/uber-web/baseui/blob/master/.flowconfig#L14

Replaced with `examples`.

PR #1652, opened 2019-07-27
fix(docs): document override prop function

### LINKED ISSUE #1587: Document the override prop function
Add documentation for this feature: https://github.com/uber-web/baseui/pull/1560

Probably, this should live here: https://baseweb.design/theming/understanding-overrides/

IgorWilbert:
Closes #1587 .

#### Description

Updates documentation with information about how to use a function to override `style`, `component` and `props`.

#### Scope

- [X] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


IgorWilbert:
@gergelyke sorry, I couldn't think of an example that is more clear than the one you gave in #1552 . Let me know if you want me to change anything. Thank you!

documentation-site/pages/theming/understanding-overrides.mdx
```diff
@@ -79,6 +79,14 @@ But as you might guess, there are multiple React components under the hood, comp
 />
 ```
 
+Overrides for `style`, `component` and `props` can be passed as functions too. For example, the overrides below are valid and equivalent.
+
+```jsx
+<Component overrides={{ props: { 'data-test': 'foo' } }} />
+
+<Component overrides={{ props: props => ({ 'data-test': 'foo' }) }} />
```
tajo:
```suggestion
<Component
  active
  overrides={{
    props: props => ({ 'data-test': props.active ? 'foo-active' : 'foo' })
  }}
/>
```

tajo: Just adding some random prop usage so the example makes more sense.

IgorWilbert:
@tajo did it just now, thank you!

PR #1661, opened 2019-07-31
feat(yard): introducing interactive documentation Yard, button component only

tajo:
This PR introduces a new interactive documentation playground nicknamed **Yard**. For now, it's a very beta version and used only on the button page (most visited doc page). It integrates together code live editing (using [react-live](https://github.com/FormidableLabs/react-live)), API documentation and overrides inspector.

## Some features:

- All props can be updated through "knobs", string is matched to input, boolean to a checkbox etc
- Once you change a knob it re-renders the component and updates the source code
- All knobs have tooltips with an additional description and type
- You can also live edit the source code and that triggers re-render of the component and also knobs to keep everything in a consistent state
- All changes are immediately pushed into the URL as a query param, so at any time you can copy the URL (or refresh) and your setting will get preserved, this should be useful for a quick share with somebody
- You can reset to the initial state
- The source code is formatted by prettier, you can also hit the "Format" button
- Overrides get a special treatment, there is a second tab with an unique UI where you can expand each override which sets it to the "highlighted styles" default, you can change the override and keep it open to make it active
- There is a special feature "Toggle shared props" that will add all shared props into your override, also the button itself has a tooltip describing all shared props
- Overrides can be reseted (highlight state), emptied (empty style object) or formatted (prettier)
- Most of actions will be tracked by Google Analytics so we can see how people use it
- Overrides inspector and API documentation removed, I want to see if anybody complains about that or this UI is a sufficient replacement

## Some limitations
- Overrides allow only to update the style, not props or component. Arguably, it would clutter the UI too much and it's not that useful in this visual playground. You should probably use CodeSandbox anyway if you are doing component overrides. 
- this is still a beta version and source code is not that pretty (a lot of `any`s, lack of tests, components too big) but better to launch it asap to get some feedback before polishing it more

## Main prop tab

Screen Shot 2019-07-31 at 2 18 01 PM

## Style overrides

Screen Shot 2019-07-31 at 2 18 26 PM

## Other changes

- prettier was bumped to the latest version, Yard uses its bundled babel parser for AST things
- syntax highlighting was reworked for the whole doc site (part of react-live integration), it now uses more modern `prism-react-renderer`, JS themes and inline styles, no global CSS pollution
- written in typescript since we are planning to use it for the rest of doc site as well

## Things to do

- Theme UI (third tab), should provide a list of values related to the specific component and allow you to do these updates
- Better convey the error state, probably red borders around the inputs / textareas when there is a syntax error
- With that, the source code generation is now too naive (gluing strings together), needs to be replaced by more robust AST transforms
- Use Yard for more components (more complex, with states or multiple components)
- Improve the performance, it feels pretty snappy in the prod mode, a bit sluggish in the dev mode, we can probably `React.memo` a lot of things and optimize some transforms
- Finalize the format of the configuration file (source of truth for this whole setup)

P.S. It exploded a bit in the scope. 🤷So many cool things this thing can be!


documentation-site/components/yard/code-box.tsx
```diff
@@ -0,0 +1,25 @@
+import * as React from 'react';
+import {useStyletron} from 'baseui';
+
+const CodeBox = ({children}: {children: React.ReactNode}) => {
+  const [css, theme] = useStyletron();
+  const isLight = theme.name === 'light-theme';
```
gergelyke: I think we should check for if the name starts with `light-theme`, as the user can toggle the font too. I think we call that `light-theme-move`

documentation-site/components/yard/const.ts
```diff
@@ -0,0 +1,101 @@
...
+  Number = 'number',
+  Enum = 'enum',
+  Array = 'array',
+  Object = 'object',
+  Function = 'function',
+  Overrides = 'overrides',
+}
+
+export const COMPONENTS = {
+  Button: {
```
gergelyke: should we move this part to a new file for the button component? so for the next component, we'd show the implementer the structure we'd like to have, instead of making this file really long?

tajo: Yep, this is a temporary location, still in the process of figuring out of the ideal structure. Once this works for multiple components + theming values and we are ready to roll it out for all components, we can colocate it with the components. 

I was even thinking that it could be exported so userland could build their own tooling / documentation based on this data-structure.

PR #1686, opened 2019-08-02
feat: rtl support progress steps

### LINKED ISSUE #1682: [RTL] Progress steps

AC:

* in rtl, the progress indicator should be on the right side

Screen Shot 2019-08-01 at 12 14 07 PM
Screen Shot 2019-08-01 at 12 14 03 PM

jh3y:
#### Description
- Resolves #1682 
- Requires a new `string` key on `Theme` for `direction` in order to get around SSR issues and no `document` in scope.
#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/src/progress-steps/__tests__/progress-steps-rtl.scenario.js
+++ b/src/progress-steps/__tests__/progress-steps-rtl.scenario.js
@@ -8,54 +8,13 @@ LICENSE file in the root directory of this source tree.
 
 import * as React from 'react';
 
-import {Block} from '../../block/index.js';
-import {Button} from '../../button/index.js';
+import {component as ProgressSteps} from './progress-steps.scenario.js';
 import {ThemeProvider, LightTheme} from '../../index.js';
-import {ProgressSteps, Step} from '../index.js';
 
 export const name = 'progress-steps-rtl';
 
-export const component = () => <ProgressStepsContainer />;
-
-class ProgressStepsContainer extends React.Component<{}, {current: number}> {
-  state = {current: 0};
-  render() {
-    return (
-      <ThemeProvider theme={{...LightTheme, direction: 'rtl'}}>
-        <ProgressSteps current={this.state.current}>
-          <Step title="Create Account">
-            <Block data-e2e="content-1" font="font400">
-              Here is some step content
-            </Block>
-            <Button
-              data-e2e="button-next"
-              onClick={() => this.setState({current: 1})}
-            >
-              Next
-            </Button>
-          </Step>
-          <Step title="Verify Payment">
-            <Block data-e2e="content-2" font="font400">
-              Here is some more content
-            </Block>
-            <Button
-              data-e2e="button-previous"
-              onClick={() => this.setState({current: 0})}
-            >
-              Previous
-            </Button>
-            <Button onClick={() => this.setState({current: 2})}>Next</Button>
-          </Step>
-          <Step title="Add Payment Method">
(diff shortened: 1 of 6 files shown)


jh3y:
Build fails on linting. `document` is not defined. Run locally with `env.browser: true` and you get errors about `document` already being defined. Without it's `undefined` 😕

src/progress-steps/styled-components.js
```diff
@@ -55,6 +57,12 @@ export const StyledIcon = styled<StylePropsT>(
 
     const marginTop = `calc((${font.lineHeight} - ${size}) / 2)`;
 
+    if ($theme.direction && $theme.direction === 'rtl') {
```
gergelyke: should we just set the default in the theme to `ltr`, so this conditional is simpler?

src/progress-steps/styled-components.js
```diff
@@ -20,6 +20,8 @@ export const StyledProgressSteps = styled<{}>('ol', ({$theme}) => {
     paddingRight: $theme.sizing.scale500,
     paddingLeft: $theme.sizing.scale500,
     paddingBottom: $theme.sizing.scale300,
+    textAlign:
```
gergelyke: do we need this line? shouldn't it be handled by the browser automatically?

src/progress-steps/styled-components.js
```diff
@@ -91,8 +99,11 @@ export const StyledInnerIcon = styled<StylePropsT>('div', ({$theme}) => {
 });
 
 export const StyledContent = styled<StylePropsT>('div', ({$theme}) => {
+  console.info($theme);
```
gergelyke: i think we can remove this

src/progress-steps/styled-components.js
jh3y: It isn't.

src/progress-steps/styled-components.js
jh3y: Done.

src/progress-steps/styled-components.js
jh3y: `auto` would make more sense if using a default value.

src/progress-steps/__tests__/progress-steps-rtl.scenario.js
```diff
@@ -0,0 +1,61 @@
...
+export const name = 'progress-steps-rtl';
+
+export const component = () => <ProgressStepsContainer />;
+
+class ProgressStepsContainer extends React.Component<{}, {current: number}> {
+  state = {current: 0};
+  render() {
+    return (
+      <ThemeProvider theme={{...LightTheme, direction: 'rtl'}}>
+        <ProgressSteps current={this.state.current}>
```
gergelyke: can we require the original story, and export the wrapped version from here, so we don't duplicate the code?

src/progress-steps/styled-components.js
```diff
@@ -55,6 +56,12 @@ export const StyledIcon = styled<StylePropsT>(
 
     const marginTop = `calc((${font.lineHeight} - ${size}) / 2)`;
 
+    if ($theme.direction === 'rtl') {
+      let oldMarginLeft = marginLeft;
```
gergelyke: `[marginLeft, marginRight] = [marginRight, marginLeft]` would do the trick too - feel free to disregard

PR #1692, opened 2019-08-03
docs: prefer stateless components and hooks

sandgraham:
#### Description

Changes should only touch the `INPUTS` component pages and examples, except for a couple small fixes I found along the way. This is a pretty big diff with just this one section so before spending a lot of time updating things I wanted to get a PR in for this section of the components and start the discussion. I think there may be some areas where folks take issue.

At a high level here are the changes:
- Replace all stateful examples with stateless versions
  - Leave one stateful example for each component
- Replace all class components with hooks
- Make examples more consistent
  - Titles are more consistent (title-case, drop redundant component name from each example)
  - Coding style is more consistent (exporting a lambda function, etc)
  - Break up examples that were "cluttered" (this is subjective but hopefully people agree)
  - Move out unnecessary styles and comments
- Some grammar/copy changes
- A couple minor bug fixes

This snow-balled a bit, but as I was editing things I kept noticing inconsistencies introduced over time and changing them in little sweeps. There are definitely inconsistencies left and areas where things could be more abstracted/DRY- but before getting carried away I wanted to get this PR in.

If folks like the changes I can start to introduce them to the other sections of the site. In terms of keeping things consistent, we can add some tooling as well create some doc site contribution guidelines (for examples/naming/style).

I also still need to create a page explaining differences between stateful and stateless components.


tajo:
Ha, wish we did these changes before forking everything as `.tsx`.

tajo:
> we can add some tooling as well create some doc site contribution guidelines (for examples/naming/style).

This would be great! Some rules that come to mind:

- no classes, use hooks only
- no `styled`, use `useStyletron` only

PR #1701, opened 2019-08-06
feat: input rtl support

### LINKED ISSUE #1671: [RTL] Input

AC:

Screen Shot 2019-08-01 at 11 21 38 AM
Screen Shot 2019-08-01 at 11 21 33 AM
Screen Shot 2019-08-01 at 11 22 13 AM
Screen Shot 2019-08-01 at 11 22 06 AM

jh3y:
Resolves #1671 

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change


gergelyke:
it seems like the enhancer's padding needs some change too? in rtl, it seems like it should has more spacing

Screen Shot 2019-08-06 at 10 15 24 AM
Screen Shot 2019-08-06 at 10 14 39 AM

jh3y:
This is part of the `Button` code therefore it's updated in the PR related to enhancers.

gergelyke:
> This is part of the `Button` code therefore it's updated in the PR related to enhancers.

👍

jh3y:
Ahh actually it isn't part of the `Button` code. The enhancers are inline coded. But as we are now reusing the scenario the padding is that for `ltr`. Prior to this I was setting the appropriate padding in the `rtl` scenario. It's down to the user and the way they implement the enhancers.

PR #1707, opened 2019-08-06
feat: select rtl support

### LINKED ISSUE #1678: [RTL] Select

AC:

Screen Shot 2019-08-01 at 12 04 58 PM
Screen Shot 2019-08-01 at 12 04 51 PM

jh3y:
Resolves #1678 

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change


gergelyke:
should the dropdown's content be right aligned too?
Screen Shot 2019-08-06 at 10 27 23 AM

jh3y:
It will be if the actual `document.dir` is `rtl` but in a scenario it's hard to enforce this.
You can apply `dir` inline to the iframe `html` tag and it will show as working.

PR #1712, opened 2019-08-06
fix: deepMerge array properties correctly

anantkamath:
#### Description
Currently deepMerge incorrectly converts array properties to objects if the key is missing in the target object.

For example `deepMerge({}, {items: ['a', 'b', 'c']})` 
incorrectly returns  `{items: {0: 'a', 1: 'b', 2: 'c'}}`

**After this fix:**
`deepMerge({}, {items: ['a', 'b', 'c']})` 
returns  `{items: ['a', 'b', 'c']}` as expected
#### Scope

- Patch: Bug Fix


anantkamath:
> lgtm - can you please add a test?

added

PR #1714, opened 2019-08-07
feat(yard): add input

tajo:
Screen Shot 2019-08-07 at 11 51 31 AM

### DIFF
--- a/documentation-site/components/yard/code-generator.ts
+++ b/documentation-site/components/yard/code-generator.ts
@@ -51,6 +51,7 @@ export const getCode = (
       }
     }
   });
+  const hasChild = children && children.value;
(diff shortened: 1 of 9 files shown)


tajo:
Hm, there is some bug when you first change any knob (value disappears), works afterwards. Will look into that.

tajo:
> looks great! the only thing I couldn't make work is the clearable prop - if I click it, it does not clear the input

Yea, it's because the component is stateless with a set `value` (you can't change the input value directly either). Not sure what to do about it.

In a theory, we could add some state handling into the code snippet but then it would be competing with the yard's state and that needs to be somehow reconciled. This could get pretty complex. We can think about it more.

Issue #1680, opened 2019-08-01
[RTL] dnd


gergelyke:
To reproduce, open the dev tools, and run `document.dir = 'rtl'`

AC:

* the component should have the same spacings and structure in `rtl` too.

<img width="237" alt="Screen Shot 2019-08-01 at 12 09 36 PM" src="https://user-images.githubusercontent.com/2174968/62320540-46874200-b455-11e9-9dee-f93a08d4f155.png">
<img width="255" alt="Screen Shot 2019-08-01 at 12 09 23 PM" src="https://user-images.githubusercontent.com/2174968/62320541-46874200-b455-11e9-9d1c-cf477d82bdea.png">

Issue #1679, opened 2019-08-01
[RTL] Pagination


gergelyke:
To reproduce, open the dev tools, and run `document.dir = 'rtl'`

AC:

* the component should have the same spacings and structure in `rtl` too.

<img width="394" alt="Screen Shot 2019-08-01 at 12 08 00 PM" src="https://user-images.githubusercontent.com/2174968/62320442-0aec7800-b455-11e9-9341-1ddae8bc58e3.png">
<img width="390" alt="Screen Shot 2019-08-01 at 12 07 54 PM" src="https://user-images.githubusercontent.com/2174968/62320443-0aec7800-b455-11e9-80a2-d9b3bb170d33.png">

Issue #1677, opened 2019-08-01
[RTL] Menu


gergelyke:
To reproduce, open the dev tools, and run `document.dir = 'rtl'`

AC:

* the component should have the same spacings and structure in `rtl` too.
* menu should open to the left

<img width="577" alt="Screen Shot 2019-08-01 at 12 03 15 PM" src="https://user-images.githubusercontent.com/2174968/62320167-6b2eea00-b454-11e9-996e-b91605739e5a.png">
<img width="585" alt="Screen Shot 2019-08-01 at 12 03 04 PM" src="https://user-images.githubusercontent.com/2174968/62320169-6b2eea00-b454-11e9-9c6e-f98b6b6f9adb.png">

PR #1715, opened 2019-08-07
feat(rtl): enable rtl layout for components

### LINKED ISSUE #1669: [RTL] Button enhancers

AC:

Screen Shot 2019-08-01 at 11 13 46 AM
Screen Shot 2019-08-01 at 11 13 38 AM

jh3y:
#### Description
Resolves #1669 #1676 #1680 #1671 #1677 #1679 #1672 #1673 #1678 

⚠️ Be mindful that in some storybook scenarios, you need to explicitly set `dir="rtl"` on the containing `iframe` `html` tag. Otherwise, you won't see `rtl` in effect. This is true with scenarios such as that for `modal`.

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change


gergelyke:
@jh3y can we close the rest of the PRs re RTL?

jh3y:
Yes. I just didn't want to in case this one wasn't right.

gergelyke:
looks great! few minor comments:

* once the dnd is grabbed, it jumps back to ltr
* for the menu, can you change to example to include nested menus?

jh3y:
> once the dnd is grabbed, it jumps back to ltr

It will do because the actual `document.dir` hasn't been set to `rtl` and it's being rendered in a portal.

I will add the nested version for menu.

gergelyke:
> > once the dnd is grabbed, it jumps back to ltr
> 
> It will do because the actual `document.dir` hasn't been set to `rtl` and it's being rendered in a portal.
> 
> I will add the nested version for menu.

sounds great, thank you

one last thing - could you also add direction switcher to the doc site (and fix any issues that look odd), something like material-ui has in a follow-up pr?

jh3y:
I can't promise that being a short task. Do you have an icon or design for this?

gergelyke:
> I can't promise that being a short task. Do you have an icon or design for this?

Probably we could use something like this https://www.iconfinder.com/icons/103727/align_right_text_icon

gergelyke:
It seems like in a few cases the LTR versions changed too - can you take a look at these ones?

Screen Shot 2019-08-07 at 11 20 30 AM
Screen Shot 2019-08-07 at 11 20 26 AM

gergelyke:
One last thing - it seems like the pagination changed a bit too

Screen Shot 2019-08-08 at 9 29 26 AM

PR #1727, opened 2019-08-08
feat(rtl): components & docs

gergelyke:


chasestarr:
For some reason screener is not capturing the full-width: https://screener.io/v2/states/SJA6Etm6Q.uber-web-baseui/rtl/1024x768/Chrome/baseui-input-rtl-0 or maybe the spacing is incorrect on inputs

gergelyke:
that's a good catch, totally missed that: https://deploy-preview-1727--baseui.netlify.com/legacy/?selectedKind=baseui&selectedStory=input-rtl&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Frtl%2Frtl-panel

let me send a quick fix for that

PR #1732, opened 2019-08-09
fix(menu): set the initial state correctly

nadiia:
#### Description

When a partial initialState value is passed in props the rest of the initial state was set incorrectly.

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


gergelyke:
is it expected that the screener screenshots changed?

nadiia:
> is it expected that the screener screenshots changed?

Yes, the menu's example height was increased.

PR #1740, opened 2019-08-13
fix(docs): get rid of dev error messages

gergelyke:
It's getting rid of dev error messages - I'll take a look at prod errors in a follow-up PR


documentation-site/examples/datepicker/datepickers-color-states.js
```diff
@@ -3,6 +3,7 @@ import React from 'react';
 
 import {useStyletron} from 'baseui';
 import {FormControl} from 'baseui/form-control';
+import {Block} from 'baseui/block';
```
chasestarr: do you mind replacing `Block` with `useStyletron`? I went through and removed all block calls from examples last week

gergelyke: thanks @chasestarr - just fixed it

documentation-site/examples/use-styletron/overrides.js
```diff
@@ -12,14 +12,16 @@ export default () => {
         BaseButton: props => {
           return (
             <button
-              {...props}
```
nadiia: I think we still need to spread the props here to pass handlers if any.

documentation-site/examples/use-styletron/overrides.tsx
```diff
@@ -11,14 +11,16 @@ export default () => {
         BaseButton: props => {
           return (
             <button
-              {...props}
```
nadiia: same here

documentation-site/examples/use-styletron/overrides.js
gergelyke: we cannot spread them, as they are invalid on the `button` component - as it serves more of an example, I think it should be up for the consumer to do that

documentation-site/components/markdown-elements.js
```diff
@@ -114,9 +113,9 @@ export const DocLink = ({children, href}: {children: string, href: string}) => {
     (parts[0] === '' && parts[1] !== '') || !href.includes('http');
   return (
     <Link href={href}>
-      <StyledLink href={href} {...(internal ? {} : {target: '_blank'})}>
+      <a href={href} {...(internal ? {} : {target: '_blank'})}>
```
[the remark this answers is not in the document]
gergelyke: I removed it, because of two reasons:

* causes `ref` errors
* the bold links look a bit off - I think the links without the heavier weight look better in the body of the main content

PR #1741, opened 2019-08-13
fix(bundle): make sure only the needed icons are bundled

gergelyke:


chasestarr:
What options do we have on catching bundle size increases before publishing? I could imagine either a lint rule to ensure we are importing icons correctly, as well as running a job on prs to show bundle size deltas. Any other ideas?

gergelyke:
> What options do we have on catching bundle size increases before publishing? I could imagine either a lint rule to ensure we are importing icons correctly, as well as running a job on prs to show bundle size deltas. Any other ideas?

I think I'd rather add a build job to show bundle size - as a baseline, we could them to a JSON file, and a build would fail if they grow significantly. Probably we should do for each component in isolation.

PR #1744, opened 2019-08-14
feat(bundle): show component size on the docs site

gergelyke:


chasestarr:
can the task print the output? if I click on the details of 'guard-component-sizes' I have no clue what the sizes are

gergelyke:
> can the task print the output? if I click on the details of 'guard-component-sizes' I have no clue what the sizes are

that makes sense - added some console.logs

PR #1748, opened 2019-08-14
fix(flex-grid): tweak column width values for IE/Edge

leyanlo:
#### Description

Tweak column width values by -0.5px for IE/Edge. IE/Edge optimistically rounds fractional pixels, which breaks FlexGrid every so often if the widths round up to more than 100%. By subtracting 0.5px from the width and letting flexbox handle growing to fill available space, we avoid possible overflows. Also went ahead and moved some of the math to JS to shrink the resulting CSS.

Before:
flex-grid-broken

After:
flex-grid-fixed

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


leyanlo:
Ah shoot don’t merge. The math is not quite right for the gaps..

leyanlo:
@chasestarr how do I get an account with screener? I don’t have access..

In general, this is the result of how IE/Edge rounds fractional pixels. If you were to just set three inline-blocks to `width: calc(100% / 3)`, it would overflow half of the time due to rounding. Not sure if there’s a good way to test for that..

gergelyke:
> @chasestarr how do I get an account with screener? I don’t have access..
> 
> In general, this is the result of how IE/Edge rounds fractional pixels. If you were to just set three inline-blocks to `width: calc(100% / 3)`, it would overflow half of the time due to rounding. Not sure if there’s a good way to test for that..

@leyanlo just invited you

chasestarr:
> it would overflow half of the time due to rounding. Not sure if there’s a good way to test for that..

could we set the width of a container element to trigger this effect?

leyanlo:
@chasestarr added test for 800px container, which should fail for the previous version in IE/Edge.

Demo:
https://codesandbox.io/s/flexgrid-basic-usage-tbr95
Screen Shot 2019-08-15 at 12 01 31 PM

PR #1753, opened 2019-08-15
fix(docs): various fixes for main page

leyanlo:
#### Description

* Remove `flex: 2` from ContentWrapper which breaks IE.
* Use FlexGrid for adopters section to work on IE.
* Fix button links not actually being the button, which cause confusing keyboard tab behavior.
* Use marginTop for blog link to avoid asymmetric tap target
* Fix typo 'noWrap' -> 'nowrap'
* Fix apostrophe and em dash punctuation
* Use png instead of svg for IE support, since it needs to scale

IE/Edge before: 
ie-edge-before

IE/Edge after:
ie-edge-after

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


documentation-site/pages/index.js
```diff
@@ -147,30 +192,12 @@ const Index = (props: {
...
-      >
-        <Button size="compact" kind="secondary">
-          Please let us know if you are using Base Web!
-        </Button>
-      </Block>
-    </Block>
+    <Adopters
+      logoSrcs={[
+        '/static/images/uber-logo.png',
+        '/static/images/broadcom-logo.png',
```
gergelyke: what's the reason for changing the broadcom logo to png?

leyanlo: @gergelyke 	SVGs don’t scale in IE. See screenshot of IE on the left:
ie-edge-after-svg

PR #1755, opened 2019-08-15
fix(timepicker): ensure onchange queues rerender

chasestarr:
#### Description

the value in the `onChange` callback was re-using the `value` date object. because there is referential equality, react does not queue a re-render. this change creates a new date object to ensure the values are not 'equal'. also adding some formatting changes to the scenario files, so there will be screener diffs flagged

this example shows how it was broken: https://codesandbox.io/s/timepicker-jis9b

#### Scope

- [x] Patch: Bug Fix


src/datepicker/timepicker.js
```diff
@@ -149,7 +149,7 @@ class TimePicker extends React.Component<TimePickerPropsT, TimePickerStateT> {
     const date = this.props.value || new Date();
     const [hours, minutes] = secondsToHourMinute(seconds);
     date.setHours(hours, minutes, 0);
-    this.props.onChange && this.props.onChange(date);
+    this.props.onChange && this.props.onChange(new Date(date));
```
nadiia: How about we work with a newly created date starting the line `149`? So on the `152` we modify that newly created one.

chasestarr: updated

PR #1756, opened 2019-08-15
fix(bundle-size): update build script

gergelyke:
* adds an option to trigger the bundle analyzer
* build the main component file, instead of the index, so webpack can treeshake it properly

### DIFF
--- a/scripts/component-sizes.js
+++ b/scripts/component-sizes.js
@@ -16,6 +16,8 @@ const fs = require('fs');
 const webpack = require('webpack');
 const MemoryFS = require('memory-fs');
 const Terser = require('terser');
+const BundleAnalyzerPlugin = require('webpack-bundle-analyzer')
+  .BundleAnalyzerPlugin;
 
 function modulePath(identifier) {
   // the format of module paths is
@@ -212,10 +214,17 @@ function getDependencySizes(stats) {
 }
 
 function makeWebpackConfig({entryPoint}) {
+  const plugins = [];
+
+  if (process.env.ANALYZER) {
+    plugins.push(new BundleAnalyzerPlugin());
+  }
+
   return {
     entry: {
       main: entryPoint,
     },
+    plugins,
     mode: 'production',
     optimization: {
       namedChunks: true,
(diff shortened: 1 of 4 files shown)


scripts/component-sizes.js
```diff
@@ -278,8 +287,23 @@ function compile({path}) {
...
+    'progress-bar': 'progress-bar/progressbar',
+    rating: 'rating/index',
+    'side-navigation': 'side-navigation/nav',
+    'table-grid': 'table-grid/index',
+    typography: 'typography/index',
+  };
+
+  const sourcePath = exceptions[component]
+    ? path.join(__dirname, `../src/${exceptions[component]}.js`)
+    : path.join(__dirname, `../src/${component}/${component}.js`);
```
nadiia: Shouldn't we do `../src/${component}/index.js` for all?

gergelyke: The problem with that is that it becomes a bit misleading imho. For example, if you take a look at the `Input` component, the `index.js` file imports the MaskedInput too. If we use the index.js file as the entrypoint, it will include this, but if we use `input/input.js`, that's removed because of tree-shaking. 

My thinking was that the most common use-cases are for the components themselves, and `index.js` files that sometimes contain these extra exports. What do you think?

gergelyke: I am not sure what would be the best - what's your take on it @nadiia  @sandgraham @chasestarr ?

chasestarr: in practice, does importing `{Input} from 'baseui/input';` vs `{MaskedInput} from 'baseui/input';` result in different bundle sizes? I think most accurate is to whitelist files that we want to keep track of

gergelyke: yeah, it does - the first one does not include the package `react-input-mask`, while the import of `MaskedInput` is, so the bundle sizes differ

nadiia: We could calculate the size of every export from the `index.js`. But exclude styled component exports.

gergelyke: For now, I'd rather just show users a ballpark number, and wouldn't overcomplicate it - is everyone okay with the current approach, or do folks feel strongly either way?

chasestarr: I'm fine as-is for public docs, it may be useful to us to get the more precise info but not required right now

PR #1761, opened 2019-08-16
feat(docs): api cheat sheet

chasestarr:
This is an mvp of an [api cheat sheet page](https://5d56e874d68d0500088a8edf--baseui.netlify.com/cheat-sheet/). Please share feedback on the approach before I spend more time dialing in the styling. Target audience here is someone more familiar with baseui and needs quick access to the finer details about api types and the source code referenced.

- grabs all `types.js` files from src directory
- transforms that file into a list of exported types
- if an exported type is an object type definition, include its properties
- includes line numbers for each type/property definition. rendered page links directly to the line
- docs site reads `COMMIT_REF` from netlify build env
- reorganizes testing to include scripts directory
- bundles documentation site file generation into a single script


gergelyke:
I think it makes perfect sense 👍

chasestarr:
This is how the page looks like with better styling: https://deploy-preview-1761--baseui.netlify.com/cheat-sheet/

gergelyke (approved):
looks amazing! should we still link this from the sidebar?

chasestarr:
There's a link called 'API cheat sheet', but I removed the sidebar from this page to allow for faster ctrl + f functionality

PR #1763, opened 2019-08-16
fix(slider): bidi support

gergelyke:


chasestarr:
looks like time-picker screener diffs were included in this pr... approved them, hopefully it doesn't cause issues

PR #1768, opened 2019-08-19
style(theme): update large breakpoint

chrisrauh:
#### Description

Update the large breakpoint value to match the latest base web design spec.

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


gergelyke:
should we release this as part of https://github.com/uber-web/baseui/pull/1697?

sandgraham:
Yeah - @chrisrauh, mind if we put this in our `v9` which has a lot of style updates. This might be breaking for some folks.

gergelyke:
I've just changed the base to be against the v9 branch

chrisrauh:
> Yeah - @chrisrauh, mind if we put this in our `v9` which has a lot of style updates? This might be breaking for some folks.

That's fine if it's not a long delay, we are also anxiously waiting for those other ones to land. :-)

PR #1785, opened 2019-08-21
fix(docs): show all gh contributors

chasestarr:
#### Description

We were previously not fetching all pages of contributors
Screen Shot 2019-08-21 at 3 38 05 PM

#### Scope

- [x] Patch: Bug Fix


gergelyke:
it seems like a few additional commits landed in this PR?

PR #1717, opened 2019-08-07
enhancement(phone-input) implement type to search and bugfixes

qqingvarqq:
Resolves #1462
#### Description
PhoneInput:
1. Implement filter functionality.
2. Propagate down missing property (maxDropdownHeight).
3. Enable adjusting of dropdown height when we have a small number of countries (mean when countries are filtering).
4. ScrolltoIndex now works correctly when we set different `maxDropdownHeight` property.
5. Add types (TS, Flow) (Also update incorrect ts types).
6. Update screener and e2e tests.
7. Add examples to docs for phone input with filter.
----
Note:
1. To phone input component I add new property `enableFiltering` to show or not show filtering, so existing users wouldn't see any changes until they do not enable filtering explicitly.
(Potentially this PR does not introduce any breaking changes.)
2. I don't know if I add correct styles, because you don't provide any sketches. let me know if need change styles.
3. Changes to ` documentation-site/components/yard/override.tsx,  documentation-site/examples/popover/placements.js` added automatically by husky prehook (It's just prettier changes but if you need I can revert them)
#### Scope

- [X] Patch: Bug Fix
- [X] Minor: New Feature
- [ ] Major: Breaking Change

PR #1791, opened 2019-08-23
feat(phone-input): replace flag svg with emoji

nadiia:
#### Description

Replaces default flag svgs usage with emojis. It results in the phone-input size reduction from 136.17 KB to 10.79 KB.

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/src/phone-input/styled-components.js
+++ b/src/phone-input/styled-components.js
@@ -5,8 +5,6 @@ This source code is licensed under the MIT license found in the
 LICENSE file in the root directory of this source tree.
 */
 // @flow
-
-import Flag from './flag.js';
 import {SIZE} from './constants.js';
 import {styled, withStyle} from '../styles/index.js';
 import {StyledList} from '../menu/index.js';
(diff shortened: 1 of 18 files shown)


src/phone-input/styled-components.js
```diff
@@ -15,22 +13,23 @@ import {
...
 import defaultProps from '../select/default-props.js';
+import type {SizeT} from './types.js';
 
 type SizeStyleProps = {
-  $size?: $Keys<typeof SIZE>,
+  $size?: SizeT,
 };
 type HeightStyleProps = {$height: string};
 
-export const StyledFlag = styled<typeof Flag, SizeStyleProps>(
```
chasestarr: `StyledFlag` export needs to be maintained

chasestarr: otherwise, lgtm

houmark:
This is great, but it's gonna conflict with #1717 which is near done also! I hope it doesn't slow that one down and I hope you core guys will put an effort into reviewing and helping to get that one merged as it will really improve the usability of this component.

src/phone-input/styled-components.js
nadiia: I moved it to the `flag.js` and still export from `index`

mihaisavezi:
Just reopening this old issue since this is not supported on windows devices. Would be good to have a headsup in the documentation, or revert to old version.

"Microsoft doesn't support any country flags on Windows, instead showing the two-letter country codes". - https://newbedev.com/flag-emojis-not-rendering

PR #1792, opened 2019-08-23
feat(data-table): prototype the typed column approach

chasestarr:
A WIP at the moment, but wanted to get some code up for others to look at


src/data-table.js/data-table.js
```diff
@@ -0,0 +1,338 @@
...
+    if (sortIndex !== -1) {
+      const sortFn = sortFnByColumn(props.columns[sortIndex]);
+      nextRows.sort((a, b) => sortFn(a.data[sortIndex], b.data[sortIndex]));
+
+      if (sortDirection === 'ASC') {
+        nextRows = nextRows.reverse();
+      }
+    }
+
+    nextRows = Array.from(filters, f => f).reduce((rows, [title, filter]) => {
```
chasestarr: some shower thoughts:
it may be more performant to separate out filter/sort transformations. where sorting can be modeled as an array of row indices and filtering as a set of row indices. what's nice is that changing a filter/sort only affects it's corresponding array/set, rather than needing to perform both transforms when either change.

src/data-table.js/data-table.js
```diff
@@ -0,0 +1,338 @@
...
+      return props.rows;
+    }
+
+    let nextRows = [].concat(props.rows);
+    if (sortIndex !== -1) {
+      const sortFn = sortFnByColumn(props.columns[sortIndex]);
+      nextRows.sort((a, b) => sortFn(a.data[sortIndex], b.data[sortIndex]));
+
+      if (sortDirection === 'ASC') {
+        nextRows = nextRows.reverse();
```
chasestarr: `reverse` could probably be factored out if it more exactly defines DESC vs ASC sort by reversing provided a and b data values

```js
if (sortDirection === 'DESC') {
  nextRows.sort((a, b) => sortFn(a.data[sortIndex], b.data[sortIndex]));
} else if (sortDirection === 'ASC') {
  nextRows.sort((a, b) => sortFn(b.data[sortIndex], a.data[sortIndex]));
}
```

chasestarr:
categorical/boolean sort can have better performance than the `.sort` method by 'bucketing' the categories and counting the number of instances.

Doing so would reduce time from O(n log n), where n is the number of array elements to O(n + m log m), where m is the number of unique categories. At worst (when all values are separate category) we will end up slightly longer than a standard comparison sort, and at best (true/false values) we'll have very close to linear time.

```
// {"red": 0, "green": 0, "blue": 0} (initial dictionary state)
for each value in array:
  increment count in dictionary for corresponding key
sort keys by comparison sort
// we now know the count of each category and can deduce the position of each
// category's first element in the return array. if 10 of each color: blue starts
// at 0, green at 10, red at 20.
loop through original array again:
  place element at starting index plus count of instances in that category already seen
```

In practice with 2000 rows with comparison sort ~= 6600 array element visits
2000 rows, 2 options ~= 4002 array element visits
2000 rows 100 options ~= 4200 array element visits

I started this idea thinking that it would be exactly linear about 1/3 or comparison sort, but may not be as worth the added complexity at 2/3 comparison sort. At smaller row counts the benefits is mostly negligible... There's likely a better way to do this using a variation on [counting sort](https://en.wikipedia.org/wiki/Counting_sort) or [pidgeonhole sort](https://en.wikipedia.org/wiki/Pigeonhole_sort)

chasestarr:
i'm going to change the folder name in my next diff - have windowing features incoming fairly soon

chasestarr:
closing because of https://github.com/uber-web/baseui/pull/1835

PR #1806, opened 2019-08-29
fix(tag component): removing role and aria-label attrib if disabled

rohit012:
Fixes https://github.com/uber-web/baseui/issues/1788 
Added unit test for the change
Ran e2e tests in local mentioned in TESTING.md

#### Description
If a Tag component is disabled, removing aria-label and role from it. If disabled Tag has close button then it should not have these attributes as well.

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


gergelyke:
thank you @rohit012 !

PR #1810, opened 2019-08-29
fix(modal): add contrasting color for close button in modal

### LINKED ISSUE #1809: Close button on modal in dark theme is not visible
## Current Behavior
The close button on modal in dark theme is not visible and only appears on hover

Aug-29-2019 22-46-17

## Expected Behavior
The close button should be clearly visible and have a contrast with default background colour

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v8.16.6  |
| React   |   16.3.2  |
| browser |  Chrome Version 76.0.3809.132 (Official Build) (64-bit)   |

akshatnitd:
Fixes #1809 

#### Description
- add contrasting color for close button in modal
- handle case of light and dark theme

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/src/modal/styled-components.js
+++ b/src/modal/styled-components.js
@@ -130,17 +130,17 @@ export const Close = styled<SharedStylePropsArgT>('button', props => {
     paddingBottom: 0,
 
     // colors
-    color: $theme.colors.mono700,
+    color: $theme.colors.modalCloseColor,
     transitionProperty: 'color, border-color',
     transitionDuration: $theme.animation.timing100,
     borderWidth: '1px',
     borderStyle: 'solid',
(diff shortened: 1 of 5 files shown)


gergelyke:
@akshatnitd can we fix it in a way that it won't effect the light theme? probably we should add a component-specific color to the theme, something like this https://github.com/uber-web/baseui/blob/e3b10ce54ed7e12acf8f0b17d39a8842592def83/src/themes/creator.js#L261

akshatnitd:
@gergelyke Yes, we should do that. Making the change.

PR #1812, opened 2019-08-29
fix(link): forward the link ref so it works together with the popover

tajo:


documentation-site/examples/popover/dismiss.js
```diff
@@ -7,18 +7,17 @@ import {Paragraph1} from 'baseui/typography';
 
 export default () => {
   const [useCss, theme] = useStyletron();
+  const contentCx = useCss({
```
tajo: not related to this PR but this example was broken bc of the placement of this hook

PR #1814, opened 2019-08-29
feat(datepicker): dates can be typed in using the keyboard

gergelyke:
[wip]: tests still need to be added


tajo:
It would be nice if the datepicker got updated when the input value is changed and date is valid. Now you have to re-open it to see the selected date.

tajo:
Seems not be working for the ranged selection.

range

PR #1825, opened 2019-08-30
fix(drawer): some cleanup for the drawer and modal

nadiia:


src/modal/types.js
```diff
@@ -26,7 +26,8 @@ export type OverridesT = {
   Close?: OverrideT<SharedStylePropsArgT>,
 };
 
-export type ElementRefT = {current: React.ElementRef<'div'> | null};
```
chasestarr: Was this failing flow check? I don't see a related change in the modal implementation code

nadiia: Yes, the only change here is to the element ref type, so no additional related changes in the modal.

PR #1834, opened 2019-09-04
fix(typescript): custom theme type support

alexzywiak:
#### Description

Allow typescript to use custom theme types when using the `createTheme` and `styled` functions.

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/documentation-site/examples/styled/basic.tsx
+++ b/documentation-site/examples/styled/basic.tsx
@@ -1,8 +1,17 @@
 import * as React from 'react';
 import {styled} from 'baseui';
+import {Theme} from 'baseui/theme';
 
-const BlueDiv = styled('div', ({$theme}) => ({
-  color: $theme.colors.primary,
+type CustomTheme = Theme & {extraProp: string};
+
+const BlueDiv = styled<
+  {$color: keyof Theme['colors']},
+  'div',
+  CustomTheme
+>('div', ({$color, $theme}) => ({
+  color: $theme.colors[$color],
 }));
 
-export default () => <BlueDiv>This is a blue div</BlueDiv>;
+export default () => (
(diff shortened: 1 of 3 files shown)


alexzywiak:
Two questions:

Previously there was a typescript type definition test file, but I was unable to find it.  I am happy to write tests if you point me in the right direction.  

Should I expand the custom theme typing to other files that hardcode the baseui theme type such as `WithStyleFn` or `BaseProvider`?

tajo:
> Previously there was a typescript type definition test file, but I was unable to find it. I am happy to write tests if you point me in the right direction.

Yea, that was deleted when we moved typings to this repo. Our primary way of testing TS types is through examples `./documentation-site/examples/*/**.tsx`.

> Should I expand the custom theme typing to other files that hardcode the baseui theme type such as WithStyleFn or BaseProvider?

That would be nice. Although, we tend to not use `styled` and `withStyle` helpers anymore and prefer `useStyletron` API instead. So for the start, I would prefer an addition of `createUseStyletron` that would let you to pass the custom theme type. I'm surprised we don't have this function already (so the implementation in flow + its TS typedefs should be added).

alexzywiak:
@tajo `createThemedUseStyletron<CustomThemeType>()` works as expected already.  Didn't see that you have recommended to move away from `styled` in favor of `useStyletron`.  I'll start doing that, too.  

I updated `WithStyleFn` for completeness and updated the types in the documentation site example.

gergelyke:
@alexzywiak can you help us resolve the merge conflicts?

alexzywiak:
@gergelyke Fixed merge conflicts

gergelyke:
thank you @alexzywiak !

PR #1835, opened 2019-09-05
feat(data-table): column measurement, table virtualization, cell formatting

chasestarr:
improves on https://github.com/uber-web/baseui/pull/1792 based on further prototyping

Screen Shot 2019-09-04 at 4 45 21 PM

#### Description

Adds virtualization to only render rows/columns that are within the visible viewport. Before, when the data set contains ~2000 rows performing actions like sorting/filtering was slow (1 - 2 seconds). By adding virtualization, we pay a bit more initial rendering time to trade for more responsive reoccurring actions (quickly click header cell twice to sort column ascending).

Required for virtualization is knowing absolute cell width/height values to properly layout the grid. I've chosen to set rows to always be 40px tall, based on design mockups. Column widths are measured dynamically because this has been a major pain point from developers using baseui in the past. Cells in the column are rendered to the page and their widths are measured. The component sets all cells widths within that column to the widest value. Measuring each cell would take quite a long time. In this case, I've opted for sampling a random subset of cells to get a best approximation. There are likely improvements to be made in this area, but what's implemented feels fine for now.

All columns render cells appropriately based on design specs. The most notable here are `numerical` and `boolean`. View the screenshot above for `numerical` examples. Numerical columns may be rendered as 'accounting' or 'percentage' and may have a highlight effect depending on a filter function.

Optionally renders filter/sorting icons dependent on if the column may be filtered/sorted respectively. For example, in the screenshot above, a custom column without a filter component provided will not give the end-user an option to filter the column.

Columns passed to the `DataTable` components are now created with factory functions for each data option and return a consistent type. It streamlines the code so that we don't need lookup tables for map from column `kind` to sorting function or filter ui for example. What's also nice about this approach is we will get better type inference, even for custom columns. see screenshot below for example.

Screen Shot 2019-09-04 at 11 31 13 AM

Cleaned up the scenario file so that it produces consistent results. Before, it was randomly generating values that would always fail visual regression testing. 

#### Scope

- [x] Minor: New Feature


tajo:
This is a really impressive work! It looks and works great. I like the column API, it makes the user configuration pretty concise and simple.

What is the plan with this PR? Are you going to keep adding more features (like implementing other filter UIs?).

Also, what's the best way to test this? e2e?

chasestarr:
> What is the plan with this PR? Are you going to keep adding more features (like implementing other filter UIs?).

I'd like to land this pr, and merge smaller feature sets incrementally so that changes will be more tightly constrained than this one. It's gotten quite large in scope already. Here I've solved the reproducible vrt problems, and because of those issues was not able to merge the previous pr.

> Also, what's the best way to test this? e2e?

Possibly. There may be issues with selecting elements related to cell virtualization. For example, I want to filter a column from 2000 rows to 200. Puppeteer will not be able to effectively count the remaining rows since they won't all be present in the dom. I'll look into how to test the features included here

documentation-site/examples/data-table/basic.tsx
```diff
@@ -0,0 +1,118 @@
...
+import {
+  Unstable_DataTable,
+  BooleanColumn,
+  CategoricalColumn,
+  CustomColumn,
+  NumericalColumn,
+  StringColumn,
+  COLUMNS,
+  // @ts-ignore
+} from 'baseui/data-table';
```
tajo: Btw, it would be nice to add `index.d.ts` with `any` . It's fine not having types for unstable component but would be nice to avoid `// @ts-ignore` for consumers. They might miss errors when the component gets stable, typed but they don't remove the ignore.

chasestarr: 👍 will update

chasestarr:
will send a follow-up pr with tests

PR #1836, opened 2019-09-05
feat(v9): black is the new blue

sandgraham:
#### Description

This PR updates styles across Base Web components.

- Update base color primitives
- Update `primary` to black
- Add new `accent` for blue
- Update buttons and button group to latest designs
- Update typography in theme and components
- Add a codemod to `baseui-codemods` that maps typography values and components
- Add a new module `baseui/tokens` with Base tokens (colors only for now)


documentation-site/pages/components/tokens.mdx
```diff
@@ -0,0 +1,94 @@
...
+});
+
+// some component later on...
+import React from 'react';
+import {useStyletron} from 'baseui';
+
+export default () => {
+  const [useCss, theme] = useStyletron();
+  return (
+    <div className={useCss({color: theme.colors.blue400})}>
```
chasestarr: I recommend not showing this example, but rather show an example of how to import from the tokens directly. This will not pass flow-check due to `colors.blue400` not being present in the theme type. It will require adding a `themedUseStyletron` call

packages/baseui-codemods/src/v9-theme-scale.test.js
```diff
@@ -0,0 +1,256 @@
...
+      expect(await transform(content)).toMatchInlineSnapshot(`
+        "const Foo = styled('div', ({$theme}) => ({
+            ...$theme.typography.font200
+          }));"
+      `);
+    });
+
+    it('direct usage of values with destructured value', async () => {
+      const content = `const Foo = styled('div', ({$theme: {typography: {font300}}}) => ({
+    ...$theme.typography.font300
```
chasestarr: is this testing what you intend? looks like it's referencing `$theme` which should throw if it was in source code, right?

```suggestion
    ...font300
```

packages/baseui-codemods/src/v9-theme-scale.test.js
```diff
@@ -0,0 +1,256 @@
...
+
+      expect(await transform(content)).toMatchInlineSnapshot(`
+        "const Foo = styled('div', ({$theme: {typography: {font200}}}) => ({
+            ...$theme.typography.font200
+          }));"
+      `);
+    });
+
+    it('override style function', async () => {
+      const content = `const Foo = <Block overrides={{ Block: { style: ({$theme}) => ({ ...$theme.typography.font200 })}}} />`;
```
chasestarr: does this work for other baseui components besides block?

packages/baseui-codemods/src/v9-theme-scale.test.js
```diff
@@ -0,0 +1,256 @@
...
+            );
+          }
+          export default Foo;"
+        `);
+      });
+    });
+  });
+
+  describe('does not update', () => {
+    it('variables', async () => {
```
chasestarr: does something like below stay unchanged?

```js
const $theme = {typography: {font100: {}}};
```

packages/baseui-codemods/src/v9-theme-scale.test.js
chasestarr: A similar, but equally important test would be `...typography.font300`

packages/baseui-codemods/src/v9-theme-scale.test.js
```diff
@@ -0,0 +1,256 @@
...
+              <Block font=\\"font150\\" />
+              <Block font=\\"font100\\" />
+              <Block font=\\"font100\\" />
+            </React.Fragment>
+          );"
+      `);
+    });
+
+    it('direct usage of font values', async () => {
+      const content = `const Foo = styled('div', props => ({
```
chasestarr: it may be unnecessary, since I'm sure you've tested this across many sample applications. But I worry about how the codemod does not select for the context in which to perform these changes. For example, only Block and related typography components should have their `font` props shifted. And similarly, only baseui's styletron wrappers have access to the theme we want to change. Some limitations/complexity of this approach, would be defining style definitions outside of where we expect or even worse importing from another file(s). Just interested to know if you've weighed these tradeoffs between more code vs the possibility or code-modding false positive matches

sandgraham: this is a very valid concern. i went back and forth on how much to include. here is where i am now: i worry more about _missed_ updates (false negatives), so i made the rules for triggering an update a bit lax.

if i update something that should not be updated, you might catch that when looking over your diff. but if i don't update something that _should_ be updated, there is no easy way to catch that other than a type error or visual regression test (manual or auto) flagging sizing changes.

what really sold me was running the mod on actual repos. i found they worked quite well in practice. in the projects i sampled, devs were not really creating or importing variables with names like `font100`. i didn't encounter any false positives but 4 or 5 repos may be too small a sample to have a lot of confidence.

if there are ways to cut down potential false positives without introducing false negatives (or ~99% still pass) that would be a clear win. for example, i considered only running these changes on files that imported baseui, only updating Blocks and typography components, etc. I'd be open to adding things like that in again if they keep that negative/positive ratio.

chasestarr: yeah i agree that not-found instances are much worse. this should be fine, especially given what you wrote in the blog post

documentation-site/pages/components/tokens.mdx
sandgraham: yeah i didn't try it out - figured it probably would throw a type error. i wanted to address why this isn't just a part of the theme. maybe ill just make it more pseudo-code-ish.

documentation-site/pages/components/tokens.mdx
```diff
@@ -0,0 +1,94 @@
...
+Some teams may decide to alias tokens on their theme object for slightly easier access when styling components. Here is how that might look in pseudo-code, adapting the above example:
+
+```js
+// theme setup...
+import {createTheme, lightThemePrimitives} from 'baseui';
+import {colors as colorTokens} from 'baseui/tokens';
+
+export default createTheme(lightThemePrimitives, {
+  colors: {
+    ...colorTokens,
```
nadiia: We shouldn't recommend extending theme like this, how about an example of theme primitives override instead?

packages/baseui-codemods/.npmrc
```diff
@@ -0,0 +1,2 @@
+registry=https://registry.npmjs.org
```
nadiia: Was it committed intentionally?

packages/baseui-codemods/README.md
```diff
@@ -0,0 +1,159 @@
...
+
+For some more info on developing CLI tools in node, check out [this article](https://medium.com/netscape/a-guide-to-create-a-nodejs-command-line-package-c2166ad0452e#c2d2). This project follows a lot of the recommendations in the article.
+
+A very useful tool for iterating on AST related logic is https://astexplorer.net/. Set the compiler to `Babylon7` and enable `Transform`, which will open two new panes. The bottom left is a babel plugin you can fill in and the bottom right is the output of that plugin on the source in the top left pane. The top right pane is for exploring the source code's AST, which will prove very useful if doing any sort of AST traversal in your codemod.
+
+For more information on AST traversal, you will want to become familiar with [how Babel plugins work](https://github.com/jamiebuilds/babel-handbook/blob/master/translations/en/plugin-handbook.md), specifically **Traversal** via the **Visitor** pattern.
+
+#### Publishing
+
+Make sure you are logged in to the npm account which is part of the `@uber-web-ui` org.
```
nadiia: Should replace it with `reach out to us for publishing`?

packages/baseui-codemods/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@uber-web-ui/baseui-codemods",
-  "version": "1.0.9",
+  "version": "2.0.2",
```
nadiia: Why `2.0.2`?

packages/baseui-codemods/src/v9-theme-scale.test.js
sandgraham: ah yeah updated this and added the `typography` case

packages/baseui-codemods/src/v9-theme-scale.test.js
sandgraham: yeah this should be picked up by the identifier traversal - but add change it for another component

packages/baseui-codemods/src/v9-theme-scale.test.js
sandgraham: yeah - added a test case

packages/baseui-codemods/src/v9-theme-scale.js
```diff
@@ -0,0 +1,160 @@
...
+        });
+      }
+    },
+  });
+}
+
+function updateTypographyComponents(path) {
+  path.traverse({
+    JSXIdentifier(path) {
+      if (Object.keys(COMPONENT_TYPOGRAPHY_MAP).includes(path.node.name)) {
```
nadiia: Can be just `if (COMPONENT_TYPOGRAPHY_MAP [path.node.name])`, no need to extract keys and loop through it.

packages/baseui-codemods/src/v9-theme-scale.js
```diff
@@ -0,0 +1,160 @@
...
+        path.node.name = COMPONENT_TYPOGRAPHY_MAP[path.node.name];
+      }
+    },
+  });
+}
+
+function updateFontIdentifiers(path) {
+  path.traverse({
+    Identifier(path) {
+      if (Object.keys(THEME_TYPOGRAPHY_MAP).includes(path.node.name)) {
```
nadiia: Same here

packages/baseui-codemods/src/v9-theme-scale.js
```diff
@@ -0,0 +1,160 @@
...
+function updateTypographyComponentImports(path) {
+  path.traverse({
+    ImportDeclaration(path) {
+      path.traverse({
+        StringLiteral(path) {
+          if (path.node.value === 'baseui/typography') {
+            path.parentPath.traverse({
+              ImportSpecifier(path) {
+                if (
+                  Object.keys(COMPONENT_TYPOGRAPHY_MAP).includes(
```
nadiia: Same here and for the rest of similar checks.

packages/baseui-codemods/package.json
sandgraham: i published 2.0.0 and 2.0.1 in the original pr - so i had to skip a couple numbers here.

packages/baseui-codemods/src/v9-theme-scale.js
sandgraham: i had this originally. false positives would occur due to a property being on in the prototype chain. although looking at this again it would be better to use `hasOwnProperty`.

packages/baseui-codemods/README.md
sandgraham: yeah i'll remove this- going to leave in the version policy just for clarification though

packages/baseui-codemods/.npmrc
sandgraham: yeah - i couldn't publish until this was specified. it would try to publish to private npm. figure this will help anyone else who tries to publish in future.

src/themes/creator.js
```diff
@@ -412,100 +477,126 @@ export default function createTheme(
...
       },
+      // deprecated, maintained for backward compatibility, may be removed in the future
       font470: {
         fontFamily: primitives.primaryFontFamily,
         fontSize: '18px',
         fontWeight: 500,
         lineHeight: '24px',
       },
-      font500: {
+      font550: {
```
nadiia: Why the `X00` >= 500 were removed? We should keep those and set to the `normal` weight of the fonts in `X50`

src/themes/creator.js
```diff
@@ -412,100 +477,126 @@ export default function createTheme(
...
+        lineHeight: '28px',
+      },
+      font450: {
+        fontFamily: primitives.primaryFontFamily,
+        fontSize: '18px',
+        fontWeight: 500,
+        lineHeight: '28px',
+      },
+      // deprecated, maintained for backward compatibility, may be removed in the future
       font460: {
```
nadiia: It doesn't make sense to keep `460` and `470` while we removing `500`, `600` and so on that is `X00 >=500`.

src/themes/dark-theme-primitives.js
```diff
@@ -5,59 +5,75 @@ This source code is licensed under the MIT license found in the
...
   mono700: '#1F1F1F',
   mono800: '#141414',
   mono900: '#111111',
-  mono1000: '#000000',
+  mono1000: colors.black,
 
-  rating200: '#FFE1A5',
-  rating400: '#FFC043',
+  rating200: colors.yellow200,
+  rating400: colors.yellow400,
```
nadiia: Since we already adding breaking changes to the theme, maybe we should get rid of (remap to semantically named color scale) the `ratingX00` colors and add something like `accentSecondary`.

nadiia:
Great job, @sandgraham!
Some additional comments:
- Button group: selected buttons don’t have a visual indication of focus state (a11y)
- Default inputs and buttons seem to have different height
- Radio buttons don’t have a visual indication of focus state (a11y)
- Buttons focus/hover color changes may not be a11y friendly
- Breadcrumbs are still blue
- Header with blue links looks a slightly strange 
- Side navigation items don’t have any hover style
- Basic icons example is orange, should we make it default or blue?
- Warning and Negative notifications are not far different in color
- Will default info notification color stay blue?
- Default info Toast still blue (well, dark blue).
- Spinner with overrides doesn’t look different from the default (and it's default blue).

src/datepicker/styled-components.js
```diff
@@ -112,9 +115,8 @@ function getArrowBtnStyle({$theme, $disabled}) {
   return {
     boxSizing: 'border-box',
     color: $disabled
-      ? $theme.colors.datepickerDayFontDisabled
-      : $theme.colors.white,
-    cursor: $disabled ? 'default' : 'pointer',
```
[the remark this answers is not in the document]
sandgraham: i think i removed it because it was not actually applying the style - but i agree it isn't really related and ill take it out of this pr

src/themes/creator.js
sandgraham: they aren't used anywhere and aren't in DS mockups, similar to new colors, so i figured to leave them out.

src/themes/creator.js
sandgraham: yeah i agree - i think at some point there was a request to leave them in but if we are breaking it this much it may be confusing to leave those.

sandgraham:
@nadiia thanks for all the feedback - lemme address some of the points:

- button group: in v8 selected buttons do not have a focus or hover state
- input and button size: again, in v8 these are not aligned
- radio button: again, in v8 there is no focus state for radio group

for these first three, should i fix in this v9 major or in a subsequent patch?

- a11y friendly button changes: i agree the contrast is not very great. we could make the color separation between each state greater but need to check with design
- breadcrumbs still blue: designs still have them blue
- header with blue links looks strange: i think these should be black - i will change
- side nav has no hover state: could show the light background color, but there are no designs for side nav, perhaps i should just change to use accent (blue)?
- basic icon example is orange: that is the current example - prob could change in a diff pr?
- warning and negative notification look alike: the shades have changed slightly but they are pretty close in v8 as well. not sure what to do about this.
- default notification is blue: latest designs in figma have it still blue - not sure if that is intended.
- default toast is blue: toast and notification should stay in sync i think. no design updates here at least.
- spinner override example looks the same: ill update that example to be noticeable

nadiia:
> for these first three, should i fix in this v9 major or in a subsequent patch?
> 
Let's do a patch after this large change is merged.

I agree a lot of those questions are for the designers to answer. We can create a ticket to follow up on it once this PR is merged.

Thanks for addressing the rest of the feedback! 🚀

src/themes/creator.js
nadiia: Sounds good. It adds up to the migration but the codemod should help here 👍

PR #1841, opened 2019-09-05
feat(yard): defer loading

tajo:
Yard is a pretty giant component so it makes sense to load it after the rest of page is done. In the dev mode, it's not being loaded by default at all - you can "click to load" or use an env flag. This  speeds up doc site deving. 

Also, the config is split by components so each yard instance loads only config (and components) that are really needed.


documentation-site/components/yard/index.tsx
```diff
@@ -0,0 +1,103 @@
+import * as React from 'react';
+import {Card, StyledBody} from 'baseui/card';
+import {Button} from 'baseui/button';
+import {Paragraph2} from 'baseui/typography';
+import getConfig from 'next/config';
```
chasestarr: TIL

documentation-site/pages/components/input.mdx
```diff
@@ -23,16 +23,17 @@ import Clearable from 'examples/input/clearable.js';
...
 
+import Yard from '../../components/yard/index';
+import inputYardConfig from '../../components/yard/config/input';
+
 export default Layout;
 
 # Input
 
-<Yard componentName="Input" />
+<Yard componentName="Input" defaultHeight={726} {...inputYardConfig} />
```
chasestarr: is there a default 'defaultHeight'? 😆I would be nice to not specify this prop if it looks fine by default, but maybe the TS error will be enough

tajo: Well, this is used when the yard content is not loaded yet, it preallocates the vertical space and prevents the page to jump when loaded.

PR #1842, opened 2019-09-06
feat(datepicker): update its look to the latest design with circles

nadiia:
#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change


src/datepicker/utils/day-state.js
```diff
@@ -0,0 +1,61 @@
...
+  rL == range-on-left - the range-highlighed case with the highlighed date is before the selected one
+  sM == start-of-month - the first day of the month
+  eM == end-of-month - the last day of the month
+  oM == outside-month - date outside of currently disaplayed month (when peekNextMonth is set to true)
+ */
+
+//  r  d  h  s  rS sD eD pS rH pH rR rL sM eM oM
+//  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1
+
+export default function getDayStateCode(props: SharedStylePropsT) {
```
gergelyke: At first, this seems a bit complex for me - but I don't know how we could simplify this. But as it's not an external  API, I guess we can do this whenever if we come up with another solution

nadiia:
@sandgraham Thanks for the feedback! Good catch on the styles of a single date selected in a range, I totally missed it due to the outline and not very large color difference 😬  Will have it fixed. With the outline removal, it's a little uncertain since we keep the highlighted state on the selected date while focus can be moved to a different element in the calendar so there won't be any visual feedback when the focus is actually on the day. I'll work with designers to figure it out. This change will probably be coming as a follow-up PR.

chasestarr:
Woah super complex. Great work on wrapping your head around this. The bitfield could probably be better utilized with bitwise comparisons rather than enumerating all of the different options, but I'm not an expect on the subject. I found a couple bugs when testing out with keyboard controls. Screenshots below.

Screen Shot 2019-09-13 at 2 19 31 PM
Screen Shot 2019-09-13 at 2 17 02 PM

PR #1858, opened 2019-09-09
feat(data-table): adds unit and e2e tests

chasestarr:
- adds a `useStyletron` mock so that tests can assert on the css properties rather than the generated class names
- unit tests explore using `act` from `react-dom/test-utils` rather than `enzyme`.
- improves header cell hovering + open filter popover (hovering out of cell will not close the filter)

### DIFF
--- a/src/data-table/__tests__/column-boolean.test.js
+++ b/src/data-table/__tests__/column-boolean.test.js
@@ -1,0 +1,105 @@
+/*
+Copyright (c) 2018-2019 Uber Technologies, Inc.
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+// @flow
+
+import React from 'react';
+import ReactDOM from 'react-dom';
+import {act} from 'react-dom/test-utils';
+
+import {BooleanColumn} from '../index.js';
+
+let container: HTMLDivElement;
+
+describe('boolean column', () => {
+  beforeEach(() => {
+    if (__BROWSER__) {
+      container = document.createElement('div');
+      if (document.body) {
+        document.body.appendChild(container);
+      }
+    }
+  });
+
+  afterEach(() => {
+    if (__BROWSER__) {
+      if (document.body && container) {
+        document.body.removeChild(container);
+      }
+    }
+  });
+
+  it('is sortable by default', () => {
+    const column = BooleanColumn({title: 'column'});
+    expect(column.sortable).toBe(true);
+  });
+
+  it('is filterable by default', () => {
+    const column = BooleanColumn({title: 'column'});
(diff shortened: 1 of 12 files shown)


src/data-table/__tests__/column-boolean.test.js
```diff
@@ -0,0 +1,105 @@
...
+  it('cell renders T if true value provided', () => {
+    const column = BooleanColumn({title: 'column'});
+    const Cell = column.renderCell;
+
+    act(() => {
+      ReactDOM.render(<Cell value={true} />, container);
+    });
+
+    const cell = container.querySelector('div');
+    if (cell) {
```
gergelyke: how is it possible to that sometimes `cell` is defined, but sometimes not? maybe I am missing something here

src/data-table/__tests__/column-categorical.test.js
```diff
@@ -0,0 +1,163 @@
...
+    const column = CategoricalColumn({title: 'column'});
+    const Cell = column.renderCell;
+
+    act(() => {
+      ReactDOM.render(<Cell value="A" />, container);
+    });
+
+    const cell = container.querySelector('div');
+    if (cell) {
+      expect(cell.textContent).toBe('A');
```
gergelyke: same questions as above

src/data-table/__tests__/data-table-columns.e2e.js
```diff
@@ -0,0 +1,205 @@
...
+describe('data table columns', () => {
+  it('passes basic a11y tests', async () => {
+    await mount(page, 'data-table-columns');
+    const accessibilityReport = await analyzeAccessibility(page);
+    expect(accessibilityReport).toHaveNoAccessibilityIssues();
+  });
+
+  it('renders expected number of cells', async () => {
+    await mount(page, 'data-table-columns');
+    await page.waitFor('div[data-baseweb="data-table"]');
```
gergelyke: nit: may make sense to move this to a const to the top of the file, as it used a few times

src/data-table/__tests__/column-boolean.test.js
chasestarr: it's because `querySelector` can return a dom node or `null`. trying to access `textContent` on the object causes a flow error

gergelyke: oh I see - so the else clause would fail the tests?

chasestarr: correct, I suppose I could have added a flowfixme comment. maybe that's clearer?

gergelyke: either that or a comment would help

PR #1859, opened 2019-09-09
feat(table): enable header cell click to sort

rohit012:
Fixes https://github.com/uber-web/baseui/issues/1830

#### Description
Adding additional boolean prop extendClick on SortableHeadCell to enable click to sort functionality throughout the header.

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change


src/table/types.js
```diff
@@ -28,6 +28,8 @@ export type HeadCellPropsT = {|
   },
   /** Column title. */
   title: string,
+  /** Extend click enable click to sort on whitespace in a header cell. */
+  extendClick?: boolean,
```
chasestarr: let's call this `fillClickTarget` to convey that it fills the available negative space rather than extends to an arbitrary point.

src/table/sortable-head-cell.js
```diff
@@ -28,25 +28,33 @@ function SortDirectionIcon({direction}: {direction: SortDirectionT}) {
...
   );
 
+  const onClick = () => {
+    props.onSort && props.onSort();
+  };
+
   return (
-    <HeadCell role="columnheader" {...headCellProps}>
+    <HeadCell
+      role="columnheader"
```
chasestarr: Can you please create an e2e tests that validates this has no accessibility issues? Here is an example: https://github.com/uber-web/baseui/blob/master/src/checkbox/__tests__/checkbox.e2e.js#L17-L21 I'm believe applying an `onClick` property to an element without `role="button"` will not properly translate the meaning to screen readers

src/table/sortable-head-cell.js
```diff
@@ -28,25 +28,33 @@ function SortDirectionIcon({direction}: {direction: SortDirectionT}) {
...
+      $cursor={extendClick ? 'pointer' : undefined}
+      onClick={extendClick ? onClick : undefined}
+    >
       <SortableLabel
         aria-label={`sorts table by ${props.title} column`}
         disabled={props.disabled}
-        onClick={() => {
-          props.onSort && props.onSort();
-        }}
+        onClick={!extendClick ? onClick : undefined}
```
chasestarr: please in-line the function and check for clarity

src/table/__tests__/sortable-head-cell.test.js
```diff
@@ -20,6 +21,16 @@ describe('sortable-head-cell', () => {
     expect(down).toHaveLength(1);
   });
 
+  it('prop extendClick enable click on the complete cell', () => {
```
chasestarr: Please duplicate 'calls provided onSort' and 'does not call if disabled' tests to ensure it works as-expected with this new flag

documentation-site/examples/table/sortable.js
```diff
@@ -115,6 +115,7 @@ export default class SortableTable extends React.Component<
               onSort={() =>
                 this.handleSort('name', this.state.nameDirection)
               }
+              extendClick
```
chasestarr: Please don't change the example. I'd like to leave it with default properties

src/table/sortable-head-cell.js
chasestarr:
```
onClick={() => {
  if (props.extendClick && props.onSort) {
    props.onSort();
  }
}}
```

chasestarr: Ah just noticed that `onClick` function is used twice. Here and on L52. This is fine

src/table/sortable-head-cell.js
rohit012: Added e2e tests which didn't show any accessibility issues. You are right, adding `onClick` does not translate that to screen readers but screenreader behaves similar to the case with no fillClickTarget which is accessible. Validated the behavior with voiceover on chrome. Please suggest.

rohit012:
@chasestarr thanks for review, tried to resolve all your comments. Please let me know if anything is missing.

src/table/__tests__/sortable-head-cell.test.js
```diff
@@ -20,6 +21,31 @@ describe('sortable-head-cell', () => {
...
+
+  it('without fillClickTarget prop, no sort on cell click', () => {
+    const spy = jest.fn();
+    const wrapper = mount(
+      <SortableHeadCell onSort={spy} direction="ASC" title="test" />,
+    );
+    expect(wrapper.find(StyledHeadCell).prop('$cursor')).toBeUndefined();
+    wrapper.simulate('click');
+    expect(spy).not.toHaveBeenCalled();
+  });
```
chasestarr: what happens if the `disabled` prop on the component is true? does it allow clicks?

rohit012: Yes, it allow clicks. `<SortableLabel /> ` handles `disabled` prop really well but not the `<HeadCell />`. Tried to fix it with a check. Let me know what you think in latest commit.

chasestarr:
Looking good @rohit012. Do you mind resolving the merge conflicts. Then I think this will be ready to land

rohit012:
@chasestarr resolved the conflicts

PR #1865, opened 2019-09-09
docs(yard): adds button-group yard

chasestarr:


documentation-site/components/yard/ast.ts
```diff
@@ -193,7 +193,7 @@ export function parseCode(code: string, elementName: string) {
             propValues[name] = value;
           });
           propValues['children'] = generate(
-            (path.node as any).children[0],
```
chasestarr: @tajo I'm not aware of why this was restricted to first element. I needed to make this change so that all button elements were parsed

tajo: not sure, it seems to work fine for all children (arrays) too

PR #1868, opened 2019-09-10
fix(version-selector): group major versions properly

gergelyke:


documentation-site/components/version-selector.js
```diff
@@ -19,48 +19,50 @@ import {Button, KIND} from 'baseui/button';
...
-  {label: 'v4'},
-  {label: 'v3'},
-  {label: 'v2'},
-  {label: 'v1'},
-].map(version => {
+const majorVersions = Array.from(
+  versions.reduce((set, version) => {
+    return set.add(semver.major(version.tag_name));
+  }, new Set()),
+).map(version => ({
```
sandgraham: i just learned the other day that the second argument to `Array.from` is a map function to run over the new array. not sure it makes a difference though.

PR #1869, opened 2019-09-10
fix(accordion-panel): use background-color instead of background

MythlinMasterdo:
#### Description

- Use `background-color` instead of `background` in accordion-panel Header to avoid React warnings when setting `background-color` via overrides. `background` should only be used in conjunction with multiple `background` properties but in this case, we are only setting color so we should use `background-color` similar to what we do on line 73 for the content styles. 

#### Scope

- [X] Patch: Bug Fix


gergelyke:
thank you @MythlinMasterdo !

gergelyke:
@MythlinMasterdo can you update the snapshot tests please? they seem to be failing

MythlinMasterdo:
@gergelyke Sure no problem

gergelyke:
closing this in favor of https://github.com/uber-web/baseui/pull/1873

PR #1871, opened 2019-09-10
fix(buttons): update loading spinner colors

sandgraham:
#### Description

Added new theme variables for styling the loading state spinner for each button kind and for disabled buttons. This fixes some lingering blue styles and some visual issues in dark mode. I also updated the button storybook scenario to show more loading states.

There were some issues with button loading states post v9:

Screen Shot 2019-09-10 at 10 56 34 AM

Screen Shot 2019-09-10 at 10 56 24 AM

Here is what they look like now:

lightspinners

darkspinners

Not sure how screener will react to loading spinners - we shall see.

#### Scope

- [x] Patch: Bug Fix


chasestarr:
Screen Shot 2019-09-10 at 11 09 17 AM

I'm assuming this was an issue in IE 11 before this change? I don't remember seeing changes to the spinner in the v9 diff. Thanks for adding loading state to the VRTs so that we catch it.

PR #1874, opened 2019-09-10
fix(typography): forwards the refs

chasestarr:
#### Scope

- [x] Patch: Bug Fix


tajo:
I wonder, does the forwardRef HOC prevents outer withStyle to work?

chasestarr:
It already was unable to be used with `withStyle` since it is a wrapper to the block component right?

PR #1877, opened 2019-09-10
chore(build): build docs site image too

gergelyke:


chasestarr:
This is sweet. Each of the docker images take ~6 mins to build. Not needed in this change, but I imagine that this could be further improved by building what's shared by the three separately?

gergelyke:
yeah, I'll continue on this tomorrow

Issue #1462, opened 2019-06-26
[PhoneInput] Type to search not working


houmark:
## Current Behavior

When opening the phone input one would expect it being possible to type either part of the country name or the dial code to filter or be moved to the best match. Though, typing does nothing.

## Expected Behavior

Typing should either filter or move the user to the nearest match for added usability. This list is LONG and scrolling through the list is exhausting.

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v7.8.0  |
| React   |   16.8.6      |
| browser |   Chrome 75.x      |

- [x] I have searched the [issues](https://github.com/uber-web/baseui/issues) of this repository and believe that this is not a duplicate.

houmark:
So after submitting this I saw a very fresh [blog post](https://baseweb.design/blog/phone-input/) about this exact field. Quoting from that post:
> Of course, because there are almost 300 country codes to choose from, we want our list to be filterable. Typing in your country name should quickly filter the list down to something more manageable. Sometimes country names can vary slightly so it is helpful to use fuzzy filtering for our list rather than following the default browser select filtering behavior. This means a user can open the dropdown, type america and quickly select United States of America.

So it seems it's been on the radar when developing this component, but neither of the included examples actually has the filtering working. Weird?

gergelyke:
Hi @houmark !

Thanks for bringing this up! This is something we want to address soon and is on our roadmap.

houmark:
> Of course, because there are almost 300 country codes to choose from, we want our list to be filterable. Typing in your country name should quickly filter the list down to something more manageable. Sometimes country names can vary slightly so it is helpful to use fuzzy filtering for our list rather than following the default browser select filtering behavior. This means a user can open the dropdown, type america and quickly select United States of America.
> 
> Thankfully, Base Web’s default Select component has fuzzy filtering and keyboard accessibility built in!

@gergelyke I appreciate you guys putting all these components out for the world to use as open source, but I think you should be more careful about announcing so proud features that are not present in the current released version in the future. It confuses us, developers, trying to use the components, and I would personally appreciate a more honest approach where you acknowledge in the blog post that, this feature is not yet ready, but that it's something you want to address in soon.

I tried this component with the `disabled` property and while it does make the text field itself disabled, the flag/country dropdown is still clickable and opens up, though it's not possible to change the country. Should I report that in a new bug?

gergelyke:
that makes perfect sense - we should have communicated more clearly on what is still needs to be done

[...]

Issue #1457, opened 2019-06-26
[TimezonePicker] Warning: Cannot update during an existing state transition


houmark:
When opening, clicking and generally using the TimezonePicker the console throws the following error:
`Warning: Cannot update during an existing state transition (such as within `render`). Render methods should be a pure function of props and state.`

I thought it may be something with my local setup (just testing out BaseWeb and TimezonePicker was my first component to test), but following the link from the docs page to CodeSandbox, the error is also reported in the console there:
https://codesandbox.io/s/pulvl?module=/example.js

## Expected Behavior

No errors in console.

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v7.8.0  |
| React   |  16.8.6       |
| browser |   Chrome 75.x      |

- [x] I have searched the [issues](https://github.com/uber-web/baseui/issues) of this repository and believe that this is not a duplicate.

jh3y:
I've narrowed down the source of the issue here 👍 
#1291 introduced a conditional inside `getRequiredItemProps` of [`src/menu/stateful-container#L210`](https://github.com/uber-web/baseui/blob/master/src/menu/stateful-container.js#L210) that triggers a `setState`. 

This `setState` is updating the `aria-activedescendant` on the `Menu` being used in any `Select` component. This includes `TimezonePicker` that uses `Select`.

This happens on first render as I believe every option is mapped invoking `getRequiredItemProps` at [`src/menu/menu#L64`](https://github.com/uber-web/baseui/blob/master/src/menu/menu.js#L64).

A solution could potentially be made inside `Menu` by explicitly updating an element ref attribute instead of triggering an entire state change. This would require an element ref for the `List` component and potentially a ref for the current `activedescendant`.

houmark:
I'm not sure this a good first issue anymore :)

I have a tight deadline, but do plan to collaborate in the future with fixes such as for example the missing error / positive state on some components. Those seem like good first issues, while this one seems to be potentially very breaking for several components that depend on it.

houmark:
This seems to still be an issue and affects quite a lot of components that use `Select`. Should this not be a pretty high priority bug?

chasestarr:
This shouldn't have been closed as stale. I'm looking into the problem now. @houmark do you use `activedescendantId` in your application?

houmark:
> @houmark do you use `activedescendantId` in your application?

[...]

PR #1892, opened 2019-09-12
feat(actions): add stale issue action

gergelyke:


gergelyke:
> @gergelyke Not sure if this is working 100% as expected. I commented on a few issues to remove the stale option, but nothing happening. In other repos, the stale tag would normally remove instantaneously, but it's been some minutes and nothing has happened on either issue #1462 or #1457

Thanks @houmark - just opened a GH issue against the stale actions repo, I am not sure what the issue is.

PR #1898, opened 2019-09-13
fix(e2e-tests): improve datepicker e2e test reliability

chasestarr:
This was kind of wacky. Two test runs back to back had dramatically different results. To improve, resets the page after every test run, and adds 3 max retries.

Screen Shot 2019-09-13 at 3 43 50 PM


scripts/runn.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
```
gergelyke: do we want to include this file in the repo?

chasestarr: removed

Issue #1491, opened 2019-06-29
TimezonePicker is too opinionated about "noisy" timezones


houmark:
## Current Behavior

I am situated in timezone "America/Guayaquil". After scrambling over my timezone not being default selected and reading the TimezonePicker code, it looks like it's filtering "noisy" timezones (and mine seems to be such). But that has the unwanted side effect of removing the option that could match with `Intl.DateTimeFormat().resolvedOptions().timeZone;` which makes it either complicated or impossible to have the current user timezone default selected in the dropdown.

## Expected Behavior

Filtering should consider the current user timezone and **not** filter that timezone to allow a default selection. In addition, it should either be possible to disable the filtering completely with a prop to get a complete unfiltered list.

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v8.0.2  |
| React   |    16.8.6     |
| browser |     Chrome 75.x    |

- [x] I have searched the [issues](https://github.com/uber-web/baseui/issues) of this repository and believe that this is not a duplicate.

houmark:
I dug into this more with the intent to attempt a fix.

I first thought that my timezone was just being filtered and I would be able to not filter it, if this is the current user timezone. So I removed the filter in timezone-picker.js to see my timezone default selected. That did not happen. It looks like my timezone `America/Guayaquil` is a "sub named zone" below `America/Lima` which gets filtered away in the timezone support package.

That means this fix seems even more complex than I first thought and while this may affect only a small percentage of users, it's a bit of a dealbreaker for me.

Would it somehow be possible to tweak the code so one can pass along the current offset when loading up the `<TimezonePicker />` and then the code would match that and display the first found timezone that is with that offset? For me that could be either Lima or Panama and while neither are the exact correct in naming, at least my UI would be loading the correct timezone and the rest of my component would be able to use the timezone to show the correct date/time.

At this time it does not seem possible to pass a property to set the default selected timezone.

jh3y:
Hey @houmark 👋 

I took a look into this. If you know the `offset` and the `label` you want for a specific timezone that isn't included, could you not add this via `overrides` of the `Select`? `overrides` allows you to hook into the timezone options and add anything you like. You could then also set a default via the `Select` props.

```js
<TimezonePicker
  overrides={{
    Select: {
      component: props => (
        <Select
          {...props}
          options={[
            ...props.options,
            {
              id: 'JHEY/TIME',
              label: 'JHEY TIME!',
              offset: 100
             },
[...]

PR #1900, opened 2019-09-16
fix(timezone-picker): list all IANA timezone options + prioritize controlled value

chasestarr:
#### Description

User in Amsterdam reported that the timezone picker did not auto-select their zone. This was due to the component using a subset of timezone options that remove aliases. I've edited the code to import the larger list (17kb bundle increase), but it should now work for a larger audience. 

Another bug highlighted in [this code sandbox](https://codesandbox.io/s/timezone-picker-5z10v) shows that the auto-selected timezone was overwriting a developer/state defined value prop. To fix, it only auto-selects if the `value` prop is falsy. Added a test to confirm this works.

#### Scope

- [x] Patch: Bug Fix


houmark:
I reported this in #1491 - good to see it's fixed, here's to hoping it does not break my custom override which auto adds the time zone if not found (which Amsterdam is not).

PR #1901, opened 2019-09-16
feat(data-table): boolean filter ui

chasestarr:
#### Description

branches from https://github.com/uber-web/baseui/pull/1894 and should be merged after that pr

Screen Shot 2019-09-16 at 11 34 14 AM

Boolean column filter UI is a small wrapper around the categorical filter that limits the available options and does some string<>boolean coercion to satisfy the categorical filter type requirements. Relevant code changes are in this [commit](https://github.com/uber-web/baseui/pull/1901/commits/f265e0eb182e86768f117479cece89143d3bab73).

#### Scope

- [x] Minor: New Feature


src/data-table/__tests__/column-categorical.test.js
```diff
@@ -14,6 +14,31 @@ import {CategoricalColumn} from '../index.js';
 
 let container: HTMLDivElement;
 
+// at this point it's likely worth pulling in react-testing-library, but
+// it seems to require the latest React@16.9 with async act. this maybe a
```
tajo: We can use 16.9 in Base and still allow 16.8 for consumers? There are not really any new features (except testing and profiling) that we could use in components, so it should be safe.

tajo:
boolean

It's a bit strange that after I reopen the filter, the checkbox is empty. I would expect to match the filter state and be selected.

src/data-table/__tests__/column-categorical.test.js
chasestarr: Cool, I agree that should be fine. I can put up another pr that removes this util and adds `react-testing-library`

chasestarr:
> It's a bit strange that after I reopen the filter, the checkbox is empty. I would expect to match the filter state and be selected.

I've held off on defining the 'update' user story. It's unclear at the moment if clicking the button again will create a new filter or update an existing filter.

PR #1911, opened 2019-09-17
fix(documentation): changed css to useCss

MythlinMasterdo:
#### Description
- replace `css` with `useCss` in documentation 

#### Scope

- [X] Patch: Bug Fix


MythlinMasterdo:
> I might be missing some context- but don't we want to "use" `useCss` in the examples to emphasize that the return value is a hook itself?
> 
> I think there is confusion around "rules of hooks" when it is named `css`.

There was a conversation in the BaseUI uChat channel earlier today where it was agreed that `useCss` should change to `css` because `useCss` is a hook and if used with conditional rendering/loop you will have a broken hook.

gergelyke:
> > I might be missing some context- but don't we want to "use" `useCss` in the examples to emphasize that the return value is a hook itself?
> > I think there is confusion around "rules of hooks" when it is named `css`.
> 
> There was a conversation in the BaseUI uChat channel earlier today where it was agreed that `useCss` should change to `css` because `useCss` is a hook and if used with conditional rendering/loops you will have a broken hook.

I believe the conclusion was the opposite - that we want it to change to `useCss`. @tajo can you confirm?

MythlinMasterdo:
Any updates on this PR? I can change `css` to `useCss` if that is what we want to move forward with.

tajo:
> I believe the conclusion was the opposite - that we want it to change to useCss. @tajo can you confirm?

Yep. We should be using `useCss` since it is a React hook.

MythlinMasterdo:
@gergelyke @tajo Updated with `useCss`

gergelyke:
thank you @MythlinMasterdo

PR #1918, opened 2019-09-18
feat(data-table): numerical filter

chasestarr:
#### Description

Extends from https://github.com/uber-web/baseui/pull/1901, please review that pr first.

Screen Shot 2019-09-18 at 9 27 29 AM
Screen Shot 2019-09-18 at 9 28 18 AM

Some complexity required for this to have a good user experience is rounding on each filter based on the precision specified. For example, by default the numerical-column rounds to the closest whole number. If a user wants to match exactly `300`, but cells are rounded from 299.9 the filter would not match.

#### Scope

- [x] Minor: New Feature


tajo:
Looks nice!

Maybe it would be nice to allow applying multiple range filters? For example, creating a multi-filter like `GT 10 AND LT 100`.

chasestarr:
> Maybe it would be nice to allow applying multiple range filters? For example, creating a multi-filter like `GT 10 AND LT 100`.

🤷‍♂ Design explicitly did _not_ want between filters. I plan to send some questions over related to 'updating an existing filter', and seeing if they want to support more than one filter on a single column

PR #1919, opened 2019-09-18
fix(build): changes to es modules mapping

nadiia:
#### Description

Creates modules mapping only for the `baseui/[COMPONENT]/index.js` files as we don't want to have any file available to be imported by its file path.
Adds `package.json` file inside every built component directories with the `module` value pointing to the `esm` build.

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


gergelyke:
depends on https://github.com/uber/fusionjs/pull/532

chasestarr:
Noticed that some code was duplicated in the bundle; `creator.js` for example. To resolve, I replaced the `module` field in `package.json` with `"module": "/esm/index.js"` rather than the file-path mapping. This should save about ~14kb after gzip from an application's bundle size. See before/after screenshots below:

before:
Screen Shot 2019-09-30 at 4 43 54 PM

after:
Screen Shot 2019-09-30 at 4 43 30 PM

nadiia:
> Noticed that some code was duplicated in the bundle; `creator.js` for example. To resolve, I replaced the `module` field in `package.json` with `"module": "/esm/index.js"` rather than the file-path mapping. This should save about ~14kb after gzip from an application's bundle size. See before/after screenshots below:
> 
That is so strange because it didn't give the expected result when I tried to replace the obj we build as a `module` value to just point to `esm/index` 🤔

chasestarr:
Yeah, it was definitely not what I was expecting. I explored this option because even if I removed the `module` field completely, we would still import esm modules into the bundle. This lead me to believe that the filepath mapping did not have an effect on code resolution.

PR #1922, opened 2019-09-18
fix(screener): update

gergelyke:


package.json
```diff
@@ -20,9 +20,8 @@
     "test": "yarn validate:examples && yarn lint && yarn flow check --max-warnings 0 && yarn tsc && yarn unit-test",
     "storybook": "start-storybook -p 6006",
     "build-storybook": "build-storybook -o public/legacy && build-storybook",
-    "storybook:screener:wrapper": "./scripts/run-screener.sh",
-    "storybook:screener": "yarn run storybook:screener:wrapper --conf screener.config.js",
-    "storybook:screener:dark": "THEME=dark yarn run storybook:screener:wrapper --conf screener.config.js",
+    "storybook:screener": "yarn run screener-storybook --conf screener.config.js --debug",
```
nadiia: Do we want to keep the `--debug` flag here?

gergelyke: ops, no - thanks Nadiia

PR #1926, opened 2019-09-19
docs(grid): add API docs

gergelyke:


sdkennedy2:
A component like this would be super useful! 

The existing FlexGrid uses an array to set values for different breakpoints. I kind of like what you purposed better, but it would also introduce inconsistencies with the FlexGrid component.

image

tajo:
> MVP API docs for the future Grid component
> 
> I wonder if we should add a prop to the `Col` to enable customers to set span/offset based on the current viewport, so something like this:
> 
> ```js-jsx
> <Col small={{ span: 5, offset: 1 }} medium={{ span: 6, offset: 2 }}>
>   Content
> </Col>
> ```
> 
> 🤔

Yea, otherwise I am not sure what would happen if you set `span` to 6 and on mobile you have only 4 columns.

How about this API though:

```js-jsx
<Col span={[2, 4, 5]} offset={[1, 2, 3]}>
  Content
</Col>
```

So it resembles the Block responsive API.

src/grid/README.md
```diff
@@ -0,0 +1,48 @@
+# Grid
+
+## Grid values
+
+| Breakpoints   | Viewport family | Columns | Margins | Gutters |
+| ------------- | --------------- | ------- | ------- | ------- |
+| 320 — 599     | Small           | 4       | 16px    | 16px    |
```
nadiia: What about `<320` case?

gergelyke:
@tajo that makes perfect sense - I'll update the PR

src/grid/README.md
gergelyke: good question - let me clarify with design

sdkennedy2:
> How about this API though:
> 
> ```js-jsx
> <Col span={[2, 4, 5]} offset={[1, 2, 3]}>
>   Content
> </Col>
> ```
I like it.

If I didn't explicitly specify the span for each breakpoint and instead specified a constant would the number of grid columns still change based on the viewport size or would it be fixed at the 12 column grid?

If it does change based on the viewport what happens when I specify the following and the grid becomes 4 columns?
```
<Row>
  <Col span={1}>...</Col>
  <Col span={11}>...</Col>
</Row>
```

If the columns is variable in this case, we could clamp the span where if the span is larger than the number of columns is becomes the number of columns. If we are on a small viewport that has 4 columns this would mean `<Col span={1}>...</Col>` would be in the first row taking 25% of the row width and `<Col span={11}>...</Col>` would be on a second row taking 100% of the row width.

Alternatively we could use a fixed 12 column grid if a constant is used. You can then also add useful typing such as
```
type ValidMobileSpan = 1;
type ValidSmallSpan = 1 | 2 | 3 | 4;
type ValidMediumSpan = 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8;
type ValidLargeSpan = 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12;

type Props = {|
  span: ValidLargeSpan
    | [ValidMobileSpan, ValidSmallSpan]
    | [ValidMobileSpan, ValidMediumSpan]
    | [ValidMobileSpan, ValidMediumSpan, ValidMediumSpan]
    | [ValidMobileSpan, ValidMediumSpan, ValidMediumSpan, ValidLargeSpan]
|};
function Col(props: Props) {...}
```

gergelyke:
that's an excellent question - we'll coordinate with design to figure that out

src/grid/README.md
```diff
@@ -0,0 +1,69 @@
...
+## Design considerations
+
+* To support Internet Explorer, `Grid` will use `flex` layout under the hood
+* Grid values will live in the `Theme`, and the component will pick them up from there
+
+## API
+
+### Row
+
+* `gutter` - optionally, it can override the margin between columns for a specific `Row`
```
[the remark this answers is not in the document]
gergelyke: I believe the only css-grid prop that IE11 does not support is grid-gap - so the idea was that we'd apply margins for the grid items

chasestarr:
This feature is still intended to be worked on, but may go in a different direction from the initial commit. Feel free to continue conversation on the direction of responsive helpers in baseui

chasestarr:
some evolution of the above design below

```js
import {useStyletron} from 'baseui';

function Foo() {
  const [useCss, theme] = useStyletron();
  return (
    <div
      className={useCss({
        sm: {backgroundColor: 'blue'},
        md: {backgroundColor: 'red'},
        lg: {backgroundColor: 'green'},
      })}
    >
      hello world
    </div>
  )
}
```

References this fela plugin: https://www.npmjs.com/package/fela-plugin-native-media-query

sandgraham:
taking inspiration from grid area templates - what about something like this:
```jsx
function Footer() {
  const [css] = useStyletron();
  const [grid, q1, q2, q3, q4] = useGrid(['aabb', 'aabbccdd', 'aa_bbbcccddd']);
  return (
    <div
      className={css({
        color: 'pink',
        ...grid,
      })}
    >
      <div className={css(q1)}>logo</div>
      <div className={css(q2)}>some links</div>
      <div className={css(q3)}>other links</div>
      <div className={css(q4)}>copyright stuff</div>
    </div>
  );
}
```
The different chars correspond to column spacing. Each string corresponds to a media break-point. Kind of fun-ky.

PR #1928, opened 2019-09-19
feat(list): adds list item component

chasestarr:
#### Description

localhost_6006_iframe html_selectedKind=baseui selectedStory=list-item full=0 addons=1 stories=1 panelRight=0 addonPanel=storybook%2Frtl%2Frtl-panel

#### Scope

- [x] Minor: New Feature


tajo:
Looks good!

I wonder if it made more sense to have `startEnhancer` instead of `artwork`?

chasestarr:
> I wonder if it made more sense to have `startEnhancer` instead of `artwork`?

I was torn on choosing `artwork` vs `startEnhancer`, and eventually chose artwork because it reinforces the idea that the left-hand side should be reserved only for an icon. But I could generalize it. Do you think the `artworkSize` prop should be renamed as well?

tajo:
> it reinforces the idea that the left-hand side should be reserved only for an icon

Yea, but `baseui` is usually very permissive and we already use this pattern across multiple components so I guess that's what most of our user will expect?


I think the API could be

```
<ListItem startEnhanced={<Search size={ARTWORK_SIZES.SMALL} />}>
  <ListItemLabel>Label Two</ListItemLabel>
</ListItem>
```

where `ARTWORK_SIZES` is just a recommended subset of icon sizes.

chasestarr:
There's some varying padding that changes between each artwork size. Should that be left up to someone implementing this? I think at that point this component is fairly trivial for someone to implement depending on their use-case

tajo:
> There's some varying padding that changes between each artwork size.

Could the `ArtworkContainer` apply vertical and horizontal centering to its children (the icon itself) instead of setting the padding?

chasestarr:
Added a `MenuAdapter` component so that the two can be easily connected

PR #1932, opened 2019-09-20
fix(data-table): remove scroll jump and ensure no remounts

chasestarr:
#### Description

When opening a filter, the page 'jumps' to focus the first element in that filter. I've added a config to the focus lock so it does not cause unintended scrolling.

Also, I've removed the `InnerTableElement` component from outside of the `DataTable` component to avoid unnecessary re-mounts (I originally thought this was the issue). It was convenient to read the state values from within the component scope, but it was re-creating the component and therefore remounting it on each parent update. Initially, I naively in-lined the component to inject extra props into the component as well as read props coming from the react-window grid component. That essentially left it in the same situation where it's re-creating the component each time. Therefore, I created a header context to avoid. 

before
Screen Shot 2019-09-19 at 4 52 34 PM
after
Screen Shot 2019-09-19 at 5 33 22 PM

This PR also includes some 'rules-of-hooks' lint fixes. I chose to not land that addition into the project eslint configuration because it raised maybe 30 problems (all related to examples and scenarios), but it does fix any cases related to the data-table

#### Scope

- [x] Patch: Bug Fix


src/data-table/data-table.js
```diff
@@ -84,6 +83,128 @@ function useSortParameters() {
   return [sortIndex, sortDirection, handleSort];
 }
 
+const HeaderContext = React.createContext<{
+  addFilter: (*, string, string) => void,
```
tajo: Afaik `*` are now just aliased to `any`, so we should use `any`s?

chasestarr: updated

PR #1938, opened 2019-09-24
feat(yard): add transforms and code generation tests, fix some bugs

tajo:
-  most of changes is the addition of unit tests for code transformations and code generation
-  state hydration from URL didn't work in prod (static) build, it needs to be done client-side only since the page is built statically (w/o query params)
- some state (react hooks) types were not handled properly
- updated the ButtonGroup setup so it has state, removed the `mode` property since it has no meaning in the controlled version of this component
- some styles fixes


documentation-site/components/yard/yard.tsx
```diff
@@ -175,10 +138,50 @@ export default withRouter(
...
+          dispatch({
+            type: Action.Update,
+            payload: {
+              code: router.query.code,
+              updatedPropValues: propValues,
+              theme: parsedTheme,
+            },
+          });
+        } catch (e) {
+          console.log(e);
```
gergelyke: should we use `console.error`?

tajo: yea, I'll probably remove those completely

PR #1947, opened 2019-09-25
bugfix: disable outline for tab

simPod:
#### Description

I have created two tabs:

image

And there's outline when clicked (focus state). So I was thinking to remove it. Or is it intentional?

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


chasestarr:
Yes, it has not been removed intentionally to indicate browser focus. This is best for accessibility purposes and is the default, but you can explicitly override to remove it.

PR #1957, opened 2019-09-25
fix(breadcrumbs): add missing List override property

alexzywiak:
#### Description

The Breadcrumbs component was missing `List` property in its overrides interface.

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


sandgraham:
Thank you @alexzywiak !

PR #1958, opened 2019-09-25
fix(data-table): ensure width dim exists before setting

chasestarr:
#### Description

strangely this bug only occurs in the docs site, not in the scenarios. that's why CI allowed prior diffs to land. to be honest i'm not sure what was causing the width dimensions to be missing in the docs site, but i've updated the code to only update the widths if a value is present

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/src/data-table/measure-column-widths.js
+++ b/src/data-table/measure-column-widths.js
@@ -96,7 +96,7 @@ export default function MeasureColumnWidths(props: MeasureColumnWidthsPropsT) {
   }, [props.columns, props.rows, props.widths, sampleSize]);
 
   function handleDimensionsChange(columnIndex, rowIndex, dimensions) {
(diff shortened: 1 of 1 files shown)


src/data-table/measure-column-widths.js
```diff
@@ -100,6 +96,8 @@ export default function MeasureColumnWidths(props: MeasureColumnWidthsPropsT) {
   }, [props.columns, props.rows, props.widths, sampleSize]);
 
   function handleDimensionsChange(columnIndex, rowIndex, dimensions) {
+    if (!dimensions.width) return;
```
chasestarr:
```suggestion
    if (dimensions.width === undefined) return;
```

PR #1961, opened 2019-09-26
feat(yard): add the Tabs component (nested components support)

tajo:
- **added the Tabs component**
- `PropTypes.ReactNode` is now properly parsed and formatted (works great for nested components). There is no need inventing some complicated cross-reference mechanism at this point. Nested components can be just edited through the code (they usually don't have many options anyway).
- all knob textareas switched to the `Editor` component to get that nice highlighting
- better popup babel parse errors (LoC numbers were off)
- added an API for extra imports and supporting default imports as well

Screen Shot 2019-09-26 at 6 02 35 PM


tajo:
Just noticed that the knob values lose formatting when editing the main code. Will fix that.

PR #1963, opened 2019-09-26
fix(dark-theme): export dark theme overrides for the dark theme creation

nadiia:


nadiia:
> should we add this to flow and TS definitions? also, I'd argue that it's a feature, and not a bug

Do you mean type for the export? Hm, it's basically the type of overrides in `createTheme` type. Not sure it needs a separate type.

PR #1970, opened 2019-09-29
fix(phone-input): fix the behavior of highlighted item

qqingvarqq:
Part 1 of #1717

* feat(menu): add the possibility to show highlighted item in the center
 I added new property `initialScrollAllignInView` to StateFullMenu which represents the position of highlighted item in the view on initial render. So now we can use this feature in phone input to show highlighted country in the center of dropdown.
  Added unit tests and updated flow and TS types too.

* fix(phone-input): highlight correct dropdown item
* fix(phone-input): show highlight item in the center of the view

Previous behavior
prev
Current behavior
current

Keyboard navigation now works well(Pressing enter when open dropdown menu)

* fix(phone-input): propagate maxDropdownHeight maxDropdownHeight properties to the dropdown
 Previously `maxDropdownHeight` and  `maxDropdownHeight` wasn't propagated to dropdown so there are not possible to use those properties. Now we can (even use relative dimensions mean `vh/vw` `rem`...). And if we set some specific value to `maxDropdownHeight` highlighted country would occur in the center of a dropdown. (Added Screener test for this behavior).

#### Scope

- [X] Patch: Bug Fix
- [X] Minor: New Feature
- [ ] Major: Breaking Change


src/menu/__tests__/utils.test.js
```diff
@@ -155,4 +155,106 @@ describe.only('Menu Utils - scrollItemIntoView', () => {
...
+  // | |-------------------| |
+  // | |       50 ht       | | <-- child to scroll into view
+  // | |-------------------| |
+  // | |-------------------| |
+  // | |       50 ht       | |
+  // | |-------------------| |
+  // | |-------------------| |
+  // | |       50 ht       | |
+  // | |-------------------| |
+  // |_______________________| <-- bottom of parent scroll height (at 350)
```
nadiia: 👏

src/menu/types.js
```diff
@@ -127,6 +127,10 @@ export type StatefulContainerPropsT = {
   getChildMenu?: (ref: {current: HTMLElement | null}) => ?{
     current: HTMLElement | null,
   },
+  /* Indicates where the highlighted item would show in a view.
+  If `auto`, we scroll the least amount possible to ensure that the highlighted item is visible.
+  If `center`, highlighted item would occur in the center of a view. */
+  initialScrollAllignInView?: 'auto' | 'center',
```
nadiia: What do you think about not exposing this setting for customization? 🤔

chasestarr: yeah - let's just replace the default behavior

qqingvarqq:
updated

chasestarr:
Thanks for updating, can we remove the `'auto' | 'center'` configuration? I don't see a reason why an application would need to choose an option other than 'center'

qqingvarqq:
@chasestarr 
We need those 2 options, because
1. `'center'` used on initial render, so, for example, an initial country would be in the center of the view.
current
2.` 'auto'` used when we use the keyboard navigation, so when we scroll to the next item which is not in the view. (it would be confusing if we scroll to the next item and it would render in a center instead of in the top or bottom)
w1

chasestarr:
Ah good point. This LGTM then. Started CI and will merge once it passes

PR #1978, opened 2019-09-30
docs(yard): add pin

gergelyke:


tajo:
btw you can create a draft PR, might be better than adding a WIP label since it will not prompt people for code reviews

https://github.blog/2019-02-14-introducing-draft-pull-requests/

gergelyke:
> btw you can create a draft PR, might be better than adding a WIP label since it will not prompt people for code reviews
> 
> https://github.blog/2019-02-14-introducing-draft-pull-requests/

sounds good, I'll do that moving forward

gergelyke:
> Looks pretty good. A couple bugs below:
> 
> * typing into the component does not work.
> * selecting a 'size' option causes an error `Identifier 'SIZE' has already been declared`
> * `clearable` is not a prop on `PinCode`
> * rename `value` to `values`
> * is it possible to increase the height of the `onClick` text area?

yeah, it's still a work in progress, sorry about the misleading PR - moving forward, I'll create draft Rs

PR #1979, opened 2019-10-01
feat(yard): cleaning up the code, config shape changes, better types

tajo:
- abstracting some code / removing repetition from the main file
- more comments
- `meta` config object is gone, all options are now on the same (first) level
- `extraImports` config option renamed to `importsConfig` to match `propsConfig` etc
- some internal renaming to better describe what is going on
- improving typescript types
- set theme values were not preserved when user typed into the rendered input (changing the internal state)

**I avoided further steps to decompose the whole project for OSS purposes.** Can be done in the future when we are happy with the overall functionality and have time (it will be a significant effort).

### DIFF
--- a/documentation-site/components/yard/config/button-group.ts
+++ b/documentation-site/components/yard/config/button-group.ts
@@ -1,8 +1,9 @@
 import {ButtonGroup, MODE} from 'baseui/button-group';
 import {Button} from 'baseui/button';
 import {PropTypes} from '../const';
+import {TConfig} from '../types';
 
-export default {
+const ButtonGroupConfig: TConfig = {
   importsConfig: {
     'baseui/button': {
       named: ['Button'],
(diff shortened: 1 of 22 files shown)


documentation-site/components/yard/config/button-group.ts
```diff
@@ -1,9 +1,10 @@
 import {ButtonGroup, MODE} from 'baseui/button-group';
 import {Button} from 'baseui/button';
 import {PropTypes} from '../const';
+import {TConfig} from '../types';
 
-export default {
-  extraImports: {
+const ButtonGroupConfig: TConfig = {
+  importsConfig: {
```
chasestarr: nit: I don't think it's necessary to label this as a `config` since this whole file is a config already

documentation-site/components/yard/types.ts
```diff
@@ -1,41 +1,69 @@
...
+export type TPropValue =
+  | undefined
+  | boolean
+  | string
+  | {
+      [key: string]: {
+        active: boolean;
+        style: string;
+      };
+    };
```
chasestarr: It took me a moment to realize what this represents. Could the override definition be extracted to its own named type?

PR #1983, opened 2019-10-01
docs(yard): add radio

gergelyke:


documentation-site/components/yard/config/radio.ts
```diff
@@ -0,0 +1,30 @@
+import {Radio, RadioGroup} from 'baseui/radio';
+import {PropTypes} from '../const';
+
+export default {
+  scopeConfig: {
```
tajo: All `fooConfig` were renamed to `foo:` and TConfig type cast should be applied to the whole config object.

// just mentioning because it's not currently causing a build break / conflicts

tajo:
Update: Reimplemented the imports (global vs prop level), they always need to be explicit now, no magic. 

Added some basic props to the Radio.

TODO
- add the rest of radio props

PR #1987, opened 2019-10-01
docs(yard): remove smart imports

gergelyke:


tajo:
One downside of this approach is that it will always show all enum imports even when they are not used. So you would always have to purge imports after copy&pasting the source code.

I think for now, we should preserve that part of "smartness" or make it explicit but through introducing `imports` scoped to the prop level (a new prop next to the `enumName`)

tajo:
Will implement this as a part of yard/radio-group.

PR #1988, opened 2019-10-01
fix(phone-input): removes block component

chasestarr:
#### Description

Suppose an application only imports `PhoneInput`. That application would be roughly 3kb larger (gzipped) because this component imports `Block`. Applications should be able to make the choice to opt-in to block's features in-exchange for the bundle size increase

#### Scope

- [x] Patch: Bug Fix


src/phone-input/country-select.js
```diff
@@ -186,7 +185,7 @@ export default function CountrySelect(props: CountrySelectPropsT) {
   );
 
   return (
-    <Block display="flex" alignItems="center">
+    <div style={{display: 'flex', alignItems: 'center'}}>
```
nadiia: Styled shouldn't be inlined. Could you create a styled component?

chasestarr: updated

PR #1960, opened 2019-09-26
[WIP] visual regression tests with differencify

sandgraham:
This is a POC for moving visual regression tests to a custom solution based on differencify, puppeteer, jest and git/github.

I will be trying out CI and workflow related things on this branch so odd commits may appear here or there. Do not be alarmed 🙏.

todo:
- [x] basic jest abstractions
- [x] cover interactions from previous screener tests
- [ ] add docker image for local runs 
- [ ] add step to ci with simple pass/fail
- [ ] add all components

PR #2001, opened 2019-10-02
Visual Regression Tests POC

sandgraham:
This is a POC for moving visual regression tests to a custom solution based on jest-image-snapshot.

#1960 is prior art, but issues with `differencify` around puppeteer versions have led me to try something a bit more flexible. We should be able to use almost the same config and setup as e2e tests to run this new approach.

So here is what this PR adds:

- [x] basic jest abstractions for running snapshot unit tests
- [x] ci pass/fail based on snapshot tests
- [x] ci uploading diffs as artifacts
- [x] running locally with docker (check README for more)
- [x] covers current scenarios and interactions (no dark theme/mobile passes though)
- [x] covers some common utilities like passing selectors, skipping tests, etc. 

Here is what this does not have:

- A nice UI for viewing diffs
- A way to update snapshots without running locally
- Dark theme
- Mobile sizing
- Flakiness-proof (wait times mainly)

The flow here going forward is, if CI visual test job fails, run local vrt tests in Docker to update snapshots, push new commit, then wait for CI to run again and pass.

Things that aren't great:

Currently there is some wait time padding required for things that are not snappy, such as the phone input dropdown. I'm a little concerned about flakiness on some of the interaction tests for this reason, but hopefully once the right `waitFor` is added, interactions should run pretty consistently.

Another annoying bit here is how slow Docker is running this stuff locally. I made a few adjustments to my local machine (adding more RAM mainly) to make things faster but still a gripe.

Lastly, the current snapshots are about ~~3MB~~ **2MB** total. Assuming we add dark theme and mobile passes that goes to around 8MB. That could be cut down a bit by snapshotting only the component (not whole screen), and perhaps running some optimizations on images (maybe use JPEG instead). But still, it may be worthwhile in the future to host images in a bucket or something so we don't have all that in git.

Next steps:

- Add dark theme and mobile passes.
- Reduce size of images
- Simplify local runs
- Add debugging utilities (filter tests, run in browser, etc)
- Save DOM snapshot on failure for inspection? Could overlay diff with DOM snapshot to make it easier to inspect change.
- Post diffs to PR thread?
- Add a nice UI for viewing/updating/approving diffs?
- Move away from global config object (define in scenario file?)
- Remove name equality requirement

I think a future version of this could have a simple UI for viewing diffs and approving them, without having to run things locally. I also don't love having 10MB of images in git, but still not sure if moving out is worth it.


jest.vrt.config.js
```diff
@@ -0,0 +1,18 @@
...
+
+// @flow
+/*eslint-env node*/
+
+module.exports = {
+  testPathIgnorePatterns: ['/node_modules/', '/dist/', './babel/cup.js'],
+  modulePathIgnorePatterns: ['<rootDir>/dist/'],
+  preset: 'jest-puppeteer',
+  testRunner: 'jest-circus/runner',
+  testRegex: 'vrt.js$',
```
gergelyke: are you planning to change this to `scenario.js`? that's what screener is using currently

jest-puppeteer.config.js
```diff
@@ -18,6 +18,12 @@ module.exports = {
   launch: {
     headless,
     slowMo: process.env.PUPPETEER_SLOWMO, // slow down tests with slowMo ms
+    args: [
```
gergelyke: are you sure we need them? the e2e tests seem to run perfectly without these

sandgraham:
So I got a local docker running in sync with CI - though the method is odd.

```
$ docker-compose build
$ docker-compose run e2e-test bash

# in docker bash...

$ yarn add puppeteer
$ yarn vrt -u
```

The updated snapshots will persist to your local directory due to the `volumes` mapping. The intermediary `bash` step is pretty odd though. I don't know why this doesn't work:

```
$ docker-compose run e2e test yarn add puppeteer && yarn vrt -u
```

jest.vrt.config.js
sandgraham: The `vrt.js` files are intended for specifying jest unit tests vs the actual storybook scenarios.

jest-puppeteer.config.js
sandgraham: These weren't needed - just some trial & error based on some thread somewhere.

sandgraham:
To run locally you can do:
```
$ docker-compose run e2e-test bash -c "yarn add puppeteer && yarn vrt"
```

jest.vrt.config.js
chasestarr: Could this be collapsed into the `.e2e.js` tests? What benefit do we gain from running the vrts separately? Both rely on the static server infrastructure. I'm not particularly excited about adding another file if it's not strictly necessary

sandgraham: i think this is mostly resolved by moving all tests to the top level - all scenarios are tested by default and you only need to specify interactions.

moving the interactions to e2e tests worries me a bit due to the reliance on docker for visual consistency. local test runs would require running in docker, at least to avoid snapshots being flagged/updated. it may not be that bad in practice though, so worth a try.

chasestarr: fine with me, in this case. what I most worry about is having code in one part of the codebase, that is not near dependent code

sandgraham: i've added a future todo for this.

either move the interactions into e2e or into scenario files as exports. i kind of like the latter pattern, which is similar to the screener steps api thing, and has the best co-location. i think this means adding a build step for vrt to access exports from es modules.

for now there is a top level config object that is kinda ugly but not terrible.

vrt/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+We use `jest-image-snapshot` for a simple suite of visual regression tests (VRT). Note, that the current system is a work in progress and may change in the future.
+
+## Set up
+
+The suite will pick up and run any test file with the `*.vrt.js` extension. By default we turn every storybook scenario file into a single snapshot test. To add a new test, simply create a new `.scenario.js` file in the `__tests__` directory of a component.
+
+> Note, a temporary requirement is that the export `name` and file name of the scenario are identical.
+
+You can configure each scenario by adding a property matching the scenario name to the vrt config object in `vrt/config.js`.
```
gergelyke: should we co-locate this with the tests? it seems a bit unintuitive to edit a configuration file to change a test case

vrt/README.md
```diff
@@ -0,0 +1,114 @@
...
+  }
+}
+```
+
+The possible properties are:
+
+1. `skip`: Skip this scenario. Default is `false`.
+2. `selector`: A selector to target for the snapshot image. Default is `null`. Takes precedence over `fullPage`
+3. `fullPage`: If `true`, takes a picture of entire page. Useful for tests with absolutely positioned elements. Default is `false`.
+4. `interactions`: An array of interaction configuration objects. Default is `[]`.
```
gergelyke: nice 👍

vrt/config.js
```diff
@@ -0,0 +1,184 @@
...
+LICENSE file in the root directory of this source tree.
+*/
+
+/* eslint-disable flowtype/require-valid-file-annotation */
+/* eslint-env node */
+
+module.exports = {
+  'country-select-dropdown': {
+    fullPage: true,
+    interactions: [
```
gergelyke: when will we take screenshots? after each command? or before and after?

vrt/tests.vrt.js
```diff
@@ -0,0 +1,112 @@
...
+const {getAllScenarioNames} = require('./utils.js');
+
+const toMatchImageSnapshot = configureToMatchImageSnapshot({
+  customDiffDir: '__vrt_diffs__',
+  diffDirection: 'vertical',
+});
+
+expect.extend({toMatchImageSnapshot});
+
+describe('visual regression tests', () => {
```
gergelyke: what's the plan for handling dark theme / LTR? or did I just miss them?

vrt/README.md
sandgraham: yeah i don't think the config object is a great solution. i went back and forth on a couple different approaches and ended up here.

i mentioned to chase in a previous comment how i'd like to move this type of configuration into the scenario files or into a new `.vrt.js` file for each component (similar to how i originally had it but only need to add if overriding default config or adding interactions).

the config is pretty simple for now though so i figured i could come back to that as a follow up effort in the interest of getting this out.

vrt/config.js
sandgraham: each interaction `behavior` should prepare the page for a screenshot. the actual screenshot is taken **after** the behavior function runs. we kind of capture the before with the default scenario snapshot. if you really need a before/after where there are interactions before each you can add an interaction to the array for both before and after (and name appropriately).

originally i passed in a function so you could take snapshots yourself in the `behavior` function but you can only take one snapshot per test due to jest limitations. i thought maybe with some kind of generator function thing we could do tests dynamically but jest doesn't seem to allow for run time test creation.

vrt/tests.vrt.js
sandgraham: i haven't added any of the alternate passes yet (dark, LTR, mobile). i thought i would do that in a subsequent pr (which is why i didn't remove screener).

i figure they will be additional `describe` blocks for iterating over scenarios. i was going to add a couple new url params to the e2e static site app and wrap the scenarios in rtl/dark theme as needed. It'll probably add a few minutes to build.

i could add it to this pr, it'll just take a lil more time- what do you think? then i could also remove screener 🤔

sandgraham:
I didn't remove Screener because I haven't added the dark theme/RTL/mobile passes yet. See https://github.com/uber-web/baseui/pull/2001#discussion_r333287910 for more context. If i can spend another day on this I can get that in and feel better about removing Screener all together.

vrt/tests.vrt.js
gergelyke: should we merge as-is (keeping screener), so we'd have this running on the other branches too, collecting additional feedback? and we could do that in a follow-up pr

sandgraham: i think that is a good idea - it won't be too hard to add the other passes but i do worry about unknown issues when this new system is run at a larger scale.

i'll merge this in as-is. i made a few follow up tickets for other passes (dark/ltr/mobile), configuration restructuring, and better reporting, etc.

PR #2002, opened 2019-10-02
docs(yard): add payment card

tajo:


gergelyke:
now that we have removed the probot that didnt let us use the "docs" label and commit prefix, can we use that instead of "feat" and "feature"?

PR #2003, opened 2019-10-03
feat(select): add single and multi select as separate components

nadiia:


gergelyke:
I like that you didn't add this to the documentation site - it will enable to incorporate some of the ideas from https://github.com/uber-web/baseui/issues/1821 into it, without a breaking change 👍

nadiia:
> do we want to include this

Not into the scope of this PR.

chasestarr:
Does not being on the docs site indicate api changes are non-breaking? I considered shipping the `PhoneInputLite` component as unstable so that we could have more flexibility in the short-term to fix any issues.

nadiia:
> Does not being on the docs site indicate api changes are non-breaking? I considered shipping the `PhoneInputLite` component as unstable so that we could have more flexibility in the short-term to fix any issues.

Not really, also considering the single select reused and available for overrides in the phone input component.

PR #2004, opened 2019-10-03
fix(select): hightlight index

ashish-kale:
https://github.com/uber-web/baseui/issues/1998

#### Description

Fixed select highlight index

#### Scope

- [x] Patch: Bug Fix


src/select/__tests__/select-highlight.scenario.js
```diff
@@ -0,0 +1,33 @@
...
+
+import * as React from 'react';
+
+import {Select} from '../index.js';
+
+export const name = 'select-highlight';
+
+export const component = () => {
+  const [value, setValue] = React.useState([]);
+  return (
```
gergelyke: I think it would be the best if the dropdown would be opened for this test case - you can learn how to do that here https://screener.io/v2/docs/test-interactions

src/select/dropdown.js
```diff
@@ -80,6 +80,26 @@ export default class SelectDropdown extends React.Component<DropdownPropsT> {
     e.preventDefault();
   };
 
+  getHighlightedIndex = () => {
+    const {value, options, valueKey} = this.props;
+    // Highlight only first value as menu supports only a single highlight index
+    let firstValue: OptionT = {};
+
+    if (Array.isArray(value) && value.length > 0) {
```
chasestarr: I believe the multi select filters options that have already been selected. What situation were you thinking about that required this code path?

src/select/__tests__/select-highlight.scenario.js
ashish-kale: done

src/select/dropdown.js
ashish-kale: this check is inspired from the getItemLabel function in same file

ashish-kale:
@chasestarr @gergelyke how do I fix "screener found visual regression" error in build? I don't have access to screener app to check build logs

PR #2005, opened 2019-10-03
docs(yard): add file-uploader

gergelyke:


documentation-site/components/yard/config/file-uploader.ts
```diff
@@ -0,0 +1,102 @@
+import {FileUploader} from 'baseui/file-uploader';
+import {PropTypes} from '../const';
+import {TConfig} from '../types';
+
+const changleHandlers = [
```
tajo: `changeHandlers`?

documentation-site/components/yard/config/file-uploader.ts
```diff
@@ -0,0 +1,102 @@
...
+      type: PropTypes.Boolean,
+      description: 'Renders component in disabled state.',
+    },
+    errorMessage: {
+      value: '',
+      type: PropTypes.String,
+      description: 'Error message to be displayed.',
+      stateful: true,
+    },
+    ...changleHandlers.reduce((acc, current) => {
```
tajo: I wonder if we should come up with some more robust mechanism to share parts of the documentation. Maybe having `config/common.ts` with all these handlers and then just picking them across other config files. We already do a lot of sharing for input components. Wdyt?

documentation-site/components/yard/config/file-uploader.ts
gergelyke: lol, yes :)

documentation-site/components/yard/config/file-uploader.ts
gergelyke: that makes sense - I'll do that as part of this PR

PR #2007, opened 2019-10-03
docs(yard): adds yard to breadcrumbs page

chasestarr:


chasestarr:
> should the first child be a link, to better showcase real-life use-cases?

updated, thanks!

documentation-site/pages/components/breadcrumbs.mdx
```diff
@@ -56,11 +65,6 @@ This is the recommended approach where SEO/Accessibility is important.
 
 ## API
```
tajo: `Overrides` should be removed too

documentation-site/components/yard/config/breadcrumbs.ts
```diff
@@ -0,0 +1,42 @@
...
+  },
+  scope: {Breadcrumbs, StyledLink},
+  theme: ['breadcrumbsText', 'breadcrumbsSeparatorFill'],
+  props: {
+    children: {
+      value: `
+        <StyledLink href="#basic-parent">Parent Page</StyledLink>
+        <StyledLink href="#basic-subparent">Sub-Parent Page</StyledLink>
+        <span>Current Page</span>
+      `,
```
tajo: Could you break this into more lines and also remove the first and last empty row? Currently, the initial knob value is not going through prettier, so it looks like this.

Screen Shot 2019-10-03 at 2 21 16 PM

I might add prettier even during initialization, was just trying to avoid it to keep it faster.

PR #2011, opened 2019-10-03
docs(yard): adds pagination component

chasestarr:


tajo:
`paginationTriangleDown` doesn't really do anything?

same goes for `NextButton`, `PrevButton` overrides?

Might be a problem with the component. If so, we should just create a ticket and not document it for now.

chasestarr:
I quickly looked into how to fix it, but seems a bit more troubled that I had guessed. Added a task to take care of it

PR #2021, opened 2019-10-04
docs(yard): add accordion

gergelyke:
Also added a few fixes:

- `disabled` prop was not working - now it does
- `Root` expands to the parent's width
- `overrides` are passed down to the child panels


documentation-site/pages/components/accordion.mdx
```diff
@@ -50,48 +59,15 @@ Accordions display a list of high-level options that can expand/contract to reve
...
 **Note:** the `Panel` has the same overrides, with the exception of the `Root` subcomponent.
 
 ## API
 
-<API
-  heading="Accordion API"
-  api={require('!!extract-react-types-loader!../../../src/accordion/accordion.js')}
-/>
-
 <API
```
gergelyke: Kept this on purpose, because I am not sure how we could include this API in Yard. @tajo thoughts?

tajo: I would still remove it. I think the only prop that's being ever used is `title` and that's pretty obvious even from the yard example. You will probably never use `Panel` as a standalone thing.

Anyway, this a bit awkward component because it doesn't have stateless version which would be much better fit yard.

gergelyke: sounds good

PR #2023, opened 2019-10-04
docs(yard): adds yard to rating component page

nadiia:


documentation-site/pages/components/rating.mdx
```diff
@@ -28,6 +30,8 @@ Like star ratings, there five options.
 **Star ratings** are to be used to collect feedback from a user’s experience with a product. The
 stars indicate a performance rating from one to five.
 
+<Yard componentName="StarRating" placeholderHeight={48} {...ratingYardConfig} />
```
tajo: Tiny nitpick, the height should be adjusted (lowered in this case). It should correspond to the actual height so when the spinner disappears the UI doesn't jump and at the same time there is no extra empty space bellow (that happens now since the component is actually smaller than `48px`)

PR #2024, opened 2019-10-04
docs(yard): time slice yard callback for perf

tajo:
Using the new secret React scheduler. Test it for yourself. It actually works nicely for inputs, not that impactful for the slider (but that's an edge case).

Thoughts?


tajo:
I think https://github.com/uber-web/baseui/pull/2027 is way to go (at least in the near future). Closing this.

PR #2025, opened 2019-10-04
chore(buildkite): move documentation site check to zeit

gergelyke:


scripts/link-checker.sh
```diff
@@ -0,0 +1,22 @@
...
+
+set -e
+
+branchUrl=$(echo $BUILDKITE_BRANCH | tr / -)
+url="https://baseui-git-$branchUrl.uber-ui-platform.now.sh/"
+
+attempt_counter=0
+max_attempts=20
+
+until $(curl --output /dev/null --silent --head --fail $url); do
```
chasestarr: what is this protecting against? attempting to reach the base-url before it's ready? I've seen the link checker fail on some thrid-party links while the `blc` command is running. Mostly interested to learn why this was required

gergelyke: yeah - this part just makes sure that the zeit deployment is ready

chasestarr: ok, so the key difference is that we don't run the site locally in CI

gergelyke: yes, that's correct

PR #2027, opened 2019-10-04
docs(yard): debounce component, knobs, editor and theme editor

tajo:
Added debounce to most of places

- rendered component state updates
- all knobs except booleans and enums
- code editor
- theme values

It makes the whole thing snappy (even in the dev mode). Probably a better approach for now. In the future, we can optimize underlying operations and combine it with the time slicing to remove debounce completely.


documentation-site/components/yard/editor.tsx
```diff
@@ -44,6 +45,13 @@ const Editor: React.FC<{
     },
   };
 
+  // debouncing editor code updates
+  const [code, setCode] = React.useState(globalCode);
+  const debouncedOnChange = React.useRef(debounce(onChange, 300)).current;
```
sandgraham: this pattern is interesting. i see it used a few times in this PR. could you describe how it works?

tajo: - it introduces a second parallel local state so as you type it instantly updates the value (code in this case)
- the original global "setter" is debounced, `useRef` is a pattern that kinda replaces `this.` from class components, gives you a way to [keep mutable value around](https://reactjs.org/docs/hooks-faq.html#is-there-something-like-instance-variables) in function components
- `useEffect` is used to synchronize the local state with upstream state since you can't directly change internal state of children components, initially the local state is set to the global state

PR #2028, opened 2019-10-05
chore(eslint): enable object shorthand lint rule

sendilkumarn:
NA to any Issues

#### Description

This enables eslint rule for object shorthand.

#### Scope
 Chore adds a new eslint rule

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -29,6 +29,7 @@ module.exports = {
     'flowtype/space-after-type-colon': 'off',
     'react/jsx-filename-extension': 0,
     'react/prop-types': 0,
+    'object-shorthand': ['error', 'always', {avoidQuotes: true}],
     // Enforces imports of external modules to be declared in the package.json
     'import/no-extraneous-dependencies': [
       'error',
--- a/scripts/component-sizes.js
+++ b/scripts/component-sizes.js
@@ -122,7 +122,7 @@ function getDependencySizes(stats) {
     return {
       path: modulePath(mod.identifier),
       sources: [source],
-      source: source,
+      source,
     };
   };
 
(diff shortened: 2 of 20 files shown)


chasestarr:
Closing, but feel free to continue the conversation here

sendilkumarn:
@chasestarr Thanks for getting back. 

The main objective benefit is the code base looks consistent. We have at some places using short-hand while at the other not using them. I would prefer consistent code across the library (we have 328 places (including methods) in short form currently in master).

> functions are defined on objects to be less readable

That is entirely subjective here. Since it is a JavaScript Standard, I believe most of us (including me) will prefer to have it in this way and it is readable. 

But hey, leaving it up to you to decide on this, since there is no performance or any other benefit w.r.t the change.

chasestarr:
I read recently that 'abstraction is a tool not a goal' and I believe the sentiment applies here. The code looking consistent is not beneficial in and of itself. Does making the change allow for more developers to read the code? Maybe for some and maybe not for others. The particular rule's subjective quality is why I'd rather not be prescriptive in this instance. If we can identify particular bugs/pitfalls that can be avoided by adding I'm for it; otherwise it's an arbitrary roadblock towards contributing code to the project.

sendilkumarn:
> abstraction is a tool, not a goal

I perfectly agree with it. But I don't think a sugar syntax falls into this place. We are not abstracting anything but making sure the code is consistent (within the standards). I will try to apply 
[Stroustrup's Rule](https://thefeedbackloop.xyz/stroustrups-rule-and-layering-over-time/) here.

> The particular rule's subjective quality is why I'd rather not be prescriptive in this instance.

Consistently looking codebase is not a bad target to achieve.

>  arbitrary roadblock towards contributing code to the project.

I would stick to standards and consistent codebase rather than having a personal preference. The latter will always add a roadblock for anyone contributing to the project. It would be even more awesome if we ensure that we have a discussion before making a decision.

PR #2032, opened 2019-10-07
docs(build): build only for the master

gergelyke:


e2e.Dockerfile
```diff
@@ -11,5 +11,4 @@ RUN yarn --ignore-scripts
 COPY . .
 
 # Perform any build steps if you want binaries inside of the image
-RUN yarn build
```
nadiia: Don't we need to build the components source for e2e?

gergelyke: it's being built as part of `e2e:build`

PR #2037, opened 2019-10-08
fix(yard): edits knob debounce and adds form-control

chasestarr:
`state.props` in the knobs set callback was always stale due to assigning the value to a ref, that does not update across re-renders. This pr edits the way that the knob value is debounced and adds the form-control component.

also adds a wrapper in the `Compiler` component so that multi-element components layout nicely in yard

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


documentation-site/components/yard/compiler.tsx
```diff
@@ -106,7 +106,9 @@ const Compiler: React.FC<{
           justifyContent: 'center',
         })}
       >
-        {Element ? <Element /> : <PlaceholderElement />}
+        <div className={css({width: '400px'})}>
```
nadiia: Adding another wrapper here will remove horizontal centering for smaller components. Could you add this style to the existing outer wrapper? What is this width restriction for?

documentation-site/components/yard/config/form-control.ts
```diff
@@ -0,0 +1,69 @@
...
+  props: {
+    children: {
+      value: '<Input />',
+      type: PropTypes.ReactNode,
+      description: 'Content wrapped by FormControl.',
+      imports: {'baseui/input': {named: ['Input']}},
+    },
+    label: {
+      value: 'label',
+      type: PropTypes.String,
```
nadiia: `label` and `caption` are `ReactNode | Function` type. I think we can go with a `Function` type for the yard case and maybe in the description mention that both types are accepted.

documentation-site/components/yard/config/form-control.ts
```diff
@@ -0,0 +1,69 @@
...
+      description: 'A label rendered above the input field.',
+    },
+    caption: {
+      value: 'caption',
+      type: PropTypes.String,
+      description: 'A caption rendered below the input field.',
+    },
+    positive: {
+      value: undefined,
+      type: PropTypes.String,
```
nadiia: Same for the `positive` and `error`. But apparently it can also be a `boolean` based on the code, but it's not reflected in types.

documentation-site/components/yard/config/form-control.ts
```diff
@@ -0,0 +1,69 @@
...
+      description:
+        'Positive state of the input. If an error prop passed it will be rendered in place of positive as an error message.',
+    },
+    error: {
+      value: undefined,
+      type: PropTypes.String,
```
nadiia: The same for the `error` prop.

documentation-site/components/yard/compiler.tsx
tajo: Done. This would be otherwise breaking button's centering.

documentation-site/components/yard/config/form-control.ts
tajo: Done. We might add some hybrid multi-type to yard in the future. For now, function is the most powerful type, so let's use that.

tajo:
- abstracted the debounce fix into a reusable hook so it can be applied to the `Editor` and `ThemeEditor` as well
- fixed the compiler wrapper, the width of `form-control` set to 100% (we should take this approach with other component too)
- PR feedback addressed
- don't display Theme tab when there is nothing to edit, same for Style Overrides

PR #2041, opened 2019-10-08
docs(yard): add popover

gergelyke:


documentation-site/pages/components/popover.mdx
```diff
@@ -37,6 +40,12 @@ content view after being triggered by a click-event (or hover) such as a button,
 icon. Popovers can contain range of content from simple text strings to inputs, or lightly
 structured content and imagery/illustration. It’s usage depends on the context.
 
+<Yard
```
tajo: Can we move it under `# Popover` to be consistent with the rest of components?

Also, `Overrides` and `API` imports should be removed.

documentation-site/pages/components/popover.mdx
```diff
@@ -37,6 +40,12 @@ content view after being triggered by a click-event (or hover) such as a button,
...
+  componentName="StatefulPopover"
+  placeholderHeight={48}
```
tajo: should be updated

PR #2044, opened 2019-10-08
docs(yard): adds yard to notification component page

nadiia:


documentation-site/components/yard/config/notification.ts
```diff
@@ -0,0 +1,83 @@
...
+      placeholder: '() => {}',
+    },
+    autoHideDuration: {
+      value: undefined,
+      type: PropTypes.Number,
+      description: `The number of milliseconds to wait before automatically dismissing a
+        notification. This behavior is disabled when the value is set to 0.`,
+      placeholder: 10000,
+    },
+    ...pick(changeHandlers, ['onMouseEnter', 'onMouseLeave']),
```
gergelyke: should we combine the two `pick`s from the changeHandlers?

nadiia: it's broken for the look 🤷‍♀

tajo:
Hm, do those style overrides actually do anything?

nadiia:
> Hm, do those style overrides actually do anything?

Should be passed as overrides to the toast elements since it's reused.

PR #2045, opened 2019-10-08
docs(yard): add datepicker

tajo:


documentation-site/components/gallery.js
```diff
@@ -81,6 +81,14 @@ const COMPONENTS = {
       href: '/components/select',
       Component: thumbnails.SvgSelect,
     },
+    {
```
gergelyke: should we position these next to the datepicker?

tajo: It's sorted alphabetically so that would be strange. Maybe we could create a new "date & time" category?

gergelyke: yeah, that's true - I was just thinking, that for me they belong to the same logical thing

maybe a new category would make sense, yeah

tajo: Everyone fine with "DATE & TIME"? @nadiia @sandgraham @chasestarr

chasestarr: sure

PR #2047, opened 2019-10-09
docs(yard): adds yard to menu

chasestarr:


documentation-site/pages/components/menu.mdx
```diff
@@ -23,10 +23,15 @@ import * as MenuExports from 'baseui/menu';
 import {Block} from 'baseui/block';
 import Code from '../../components/code';
```
tajo: Should we delete Overrides / API imports and components?

chasestarr: I was thinking that we should at least show that they are there, but I'm fine with removing if you don't agree

tajo: I'm fine keeping them. We can always delete those later.

PR #2053, opened 2019-10-09
docs(list): adds yard to list-item

chasestarr:
#### Description

this component is a bit unique, since it depends on the outer `<ul />` element for defining its width.  I added a `{width: '100%'}` to the list-item to solve this. It also would be nice to display the currently rendered yard component within a wrapping component.

#### Scope

- [x] Minor: New Feature


documentation-site/pages/components/list.mdx
```diff
@@ -50,37 +55,6 @@ List items best used in mobile applications.
   <MenuAdapter />
 </Example>
 
-## Overrides
-
-<Overrides
```
tajo: Overrides / API imports should be deleted

chasestarr: overrides are deleted in this diff, and imports [are maintained](https://baseweb.design/components/button/#api) correct?

tajo: You should delete 

```
import API from '../../components/api';
import Overrides from '../../components/overrides';
```

from this file since it's not used anymore?

chasestarr: oh right

PR #2055, opened 2019-10-09
docs(modal): adds yard to modal page

chasestarr:
modal is a bit special since it renders no content unless `isOpen === true`. In this change, I've set the placeholder height to 0, so that it takes no space. Maybe there is a better way to show this?


tajo:
Oh, this thing is funny. I think it's fine. The only glitch is that if you check `isOpen`, close the modal and check `isOpen` again it doesn't open. I have to double-click that knob, then the modal opens again. Strange.

PR #2056, opened 2019-10-09
docs(header-navigation): adds yard to docs page

chasestarr:


documentation-site/components/yard/config/header-navigation.ts
```diff
@@ -0,0 +1,66 @@
...
+<StyledLink href="#basic-link2">Tab Link Two</StyledLink>
+</StyledNavigationItem>
+</StyledNavigationList>
+<StyledNavigationList $align={ALIGN.right}>
+<StyledNavigationItem>
+<Button>Get started</Button>
+</StyledNavigationItem>
+</StyledNavigationList>
+`,
+      type: PropTypes.String,
```
tajo:
```suggestion
      type: PropTypes.ReactNode,
```

documentation-site/components/yard/config/header-navigation.ts
```diff
@@ -0,0 +1,66 @@
...
+  imports: {
+    'baseui/header-navigation': {
+      named: ['HeaderNavigation', 'ALIGN'],
+    },
+    'baseui/link': {
+      named: ['StyledLink'],
+    },
+    'baseui/button': {
+      named: ['Button'],
+    },
```
tajo: Could be moved to the children prop but doesn't really matter in this case.

documentation-site/components/yard/config/header-navigation.ts
```diff
@@ -0,0 +1,66 @@
...
+</StyledNavigationItem>
+<StyledNavigationItem>
+<StyledLink href="#basic-link2">Tab Link Two</StyledLink>
+</StyledNavigationItem>
+</StyledNavigationList>
+<StyledNavigationList $align={ALIGN.right}>
+<StyledNavigationItem>
+<Button>Get started</Button>
+</StyledNavigationItem>
+</StyledNavigationList>
```
tajo: (welp, we'll probably need to do something about the initial knob formatting through prettier)

chasestarr: maybe we could mark a prop to take up the full width of the container? (span both columns)

tajo: Yea, it's just there's not much value to edit the children knob vs edit the code snippet at that point.

chasestarr: good point

PR #2057, opened 2019-10-09
docs(yard): show the default value of enum type props & skip it in code example

nadiia:


tajo:
This is great! 

We should toggle (hide) imports as well when the default value is used. For example, the default button example does this:

```
import { Button, KIND, SIZE, SHAPE } from "baseui/button";
```

but those enums are not used.

Also, it would be great to add some code-generator unit test for this.

PR #2058, opened 2019-10-09
docs(yard): add progress-steps

gergelyke:


documentation-site/components/yard/config/progress-steps.ts
```diff
@@ -0,0 +1,71 @@
...
+    current: {
+      value: '0',
+      type: PropTypes.Number,
+      description: 'Defines the current active step index.',
+      stateful: true,
+    },
+    children: {
+      value: `<Step title="Verify Address">
+  <p>Address on file: 1455 Market Street</p>
+  <Button size="compact" onClick={() => setCurrent(1)}>
```
gergelyke: in this case, can I update the `current` prop with this value? @tajo

tajo: sure, this works right? (waiting for the build)

tajo:
Wonder if should have two yard examples for this page (Step vs NumberedStep). 

Also, should we remove overrides / API components?

documentation-site/components/yard/config/progress-steps.ts
tajo: Oh I see, it doesn't update the snippet / knob. You could probably try

```jsx
<Button size="compact" onClick={() => {
  setCurrent(1);
  __yard_onChange('ProgressSteps', 'current', 1);
}>
```

It should be taken out (not visible in the code snippet).

tajo: Um, but that would probably work only initially... we could probably enhance the whole `propHook` thing to target arbitrary inner callbacks.

documentation-site/components/yard/index.tsx
```diff
@@ -23,16 +23,18 @@ const YardWrapper: React.FC<
...
+        urlCode={
+          typeof queryStringName !== 'undefined'
+            ? router.query[queryStringName]
+            : router.query.code
+        }
+        queryStringName={queryStringName}
+        placeholderElement={() => (
           if (!placeholderHeight) return null;
-
-          return (
```
tajo: does this `return`  disappeared?

gergelyke: just fixed it - i didn't resolve this merge conflict properly 🤦‍♂

tajo:
Added a new feature - `propHook` now also accepts a function where you can define your own visitor to locate an arbitrary function that should be instrumented.

It solves the use-case of this component where `onClick` is deeply nested in a children prop.

This is a very flexible API since you can arbitrarily traverse the tree.

gergelyke:
> Added a new feature - `propHook` now also accepts a function where you can define your own visitor to locate an arbitrary function that should be instrumented.
> 
> It solves the use-case of this component where `onClick` is deeply nested in a children prop.
> 
> This is a very flexible API since you can arbitrarily traverse the tree. It can cover `renderProps` pattern as well.

this is very cool, thanks Vojtech!

PR #2061, opened 2019-10-10
docs(yard): small fixes for card and checkbox

gergelyke:


documentation-site/components/yard/config/checkbox.ts
```diff
@@ -87,6 +92,11 @@ const CheckboxConfig: TConfig = {
       enumName: 'LABEL_PLACEMENT',
       description:
         'Determines how to position the label relative to the checkbox.',
+      imports: {
```
tajo: Could you also set the `value` and `defaultValue`?

gergelyke: sure, done

PR #2067, opened 2019-10-10
fix(package): remove engine restriction

gergelyke:


tajo:
hm, we actually need Node 10+ for some build tasks?

gergelyke:
> hm, we actually need Node 10+ for some build tasks?

yeah, one of our devDeps need it

PR #2071, opened 2019-10-10
docs(yard): adds yard to dnd list component page

nadiia:


documentation-site/pages/components/dnd-list.mdx
```diff
@@ -79,35 +82,11 @@ This component also supports iOS and Android browsers and screen readers.
...
 
-<API
-  heading="List API"
-  api={require('!!extract-react-types-loader!../../../src/dnd-list/list.js')}
-/>
-
 <API
   heading="Stateful List API"
   api={require('!!extract-react-types-loader!../../../src/dnd-list/stateful-list.js')}
 />
```
tajo: I would delete this too (+ the import). The preferred way is to use stateless API and there is still a stateful example.

(we might be able to delete `extract-react-types-loader` too)

documentation-site/pages/components/dnd-list.mdx
```diff
@@ -17,14 +17,17 @@ import VaryingHeights from 'examples/dnd-list/varyingHeights.js';
...
 import {StatefulList} from 'baseui/dnd-list';
 import * as ListExports from 'baseui/dnd-list';
+import Yard from '../../components/yard/index';
+import dndListYardConfig from '../../components/yard/config/dnd-list';
 
 export default Layout;
 
 # Drag and Drop List
 
+<Yard componentName="List" placeholderHeight={48} {...dndListYardConfig} />
```
tajo: `placeholderHeight` should be updated

documentation-site/components/yard/config/dnd-list.ts
```diff
@@ -0,0 +1,86 @@
...
+  theme: [],
+  props: {
+    items: {
+      value: `[
+        'Item 1',
+        'Item 2',
+        'Item 3',
+        'Item 4',
+        'Item 5',
+        'Item 6',
```
tajo: I think 3 items are good enough to demo it and could save same vertical space here.

documentation-site/pages/components/dnd-list.mdx
nadiia: Since we don't describe the rest of available components API, like stateful when only use stateless in the yeard, and subcomponents like in the Tabs and Accordion cases, I think we should keep the API docs for those not included in yard.

documentation-site/components/yard/config/dnd-list.ts
```diff
@@ -0,0 +1,86 @@
...
+  },
+  scope: {List, arrayMove, arrayRemove},
+  theme: [],
+  props: {
+    items: {
+      value: `[
+        'Item 1',
+        'Item 2',
+        'Item 3',
+      ]`,
```
tajo: It actually fits the single line, so I would use this as the value:

```
["Item 1","Item 2","Item 3"]
```

PR #2075, opened 2019-10-11
docs(yard): adds yard to toast component page

nadiia:


gergelyke:
Should we rather add the Toaster to yard? I think that's a more common use-case

nadiia:
> Should we rather add the Toaster to yard? I think that's a more common use-case

It's not really possible with the current yard configuration. I added a feature to render children passed to a 'ToasterContainer' component to be able to add toaster example to yard. What do you think of this change?

documentation-site/components/yard/config/toast.ts
```diff
@@ -0,0 +1,116 @@
...
+
+const toastConfig: TConfig = {
+  imports: {
+    'baseui/toast': {
+      named: ['toaster', 'ToasterContainer', 'Toast'],
+    },
+    'baseui/button': {
+      named: ['Button', 'SIZE'],
+    },
+    'baseui/button': {
```
tajo:
```suggestion
    'baseui/block': {
```

PR #2076, opened 2019-10-11
feat(data-table): expandable rows prototype

chasestarr:
#### Description

Prototype implementation of row-expand functionality. I'm unsure of the specific needs from design and ultimately end-users, but this is built against the figma specs. Are 'child' rows useful if they must contain the same columns as the parents? I personally cannot think of real world use-cases. Maybe something more along the lines of airtable's [grouping](https://airtable.com/shrboYH2xp41kojwQ) feature is more appropriate? 

Screen Shot 2019-10-11 at 2 46 49 PM

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/package.json
+++ b/package.json
@@ -175,7 +175,7 @@
     "static-server": "^2.2.1",
     "storybook-addon-rtl": "^0.2.2",
     "styletron-engine-atomic": "^1.4.0",
-    "styletron-react": "^5.2.0",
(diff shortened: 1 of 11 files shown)


chasestarr:
going to close this pr for now. not included in initial release of the component

PR #2080, opened 2019-10-14
test(vrt): update snapshots during ci

sandgraham:
This PR allows CI to update snapshots as part of the `vrt` job. When a PR is opened, we run the `vrt` tests with the update flag so that it generates updated snaps. If there are any updated snaps, we open a new branch/PR based off of the original branch/PR. This new snapshot PR will get refreshed as new commits come through the original branch. Note this new logic only applies to commits on branches with open PRs. (non PR commits that run CI will have the old treatment of just a success/failure)

This should remove the need to update things locally in the majority of cases. It lets people easily see what has changed (via the snapshot PR) and pull those changes into their branch/PR. It also keeps a more isolated record of snapshot updates over time.

There is logic for posting comments with relevant links, opening/closing PRs, requesting reviews, etc. I also tried to account for some edge case scenarios (rebasing, merging PR and causing new changes, resolving conflicts on main branch) but there may (likely) be some even edgier stuff i haven't considered.

This PR has some auto-generated comments that link to a [sample snapshot PR](https://github.com/uber/baseweb/pull/2108). There are a lot of new files because i initially removed all of the snapshots on this branch (more on this behavior below). This leads us to...

There a couple things left to do/consider here:

- this does not compare changes against the master branch. so if you have approved/merged snapshot changes into your branch, that will be the base source of truth for snapshots as far as the ci check is concerned. there a couple ways to solve this. we could checkout the master snapshots (`vrt/__image_snapshots__`) in ci before running the tests. this would generate a pr with diffs against master. the downside is you can't gradually approve/merge changes into your branch. i thought the current implementation may be more graceful for more use cases, but it is something to track/look out for. i don't think changing it would require much code change (based on strategy above).
- if the origin pr is closed/branch deleted - the snapshot branch/pr will still be open (if it exists). i will need to add a cleanup job when PRs close or something but prob in a subsequent pr.
- this will not work with forks - i will need to add some more logic in a subsequent pr.


vrt/ci.js
```diff
@@ -0,0 +1,351 @@
...
+  installChromium();
+  if (buildWasTriggeredByPR()) {
+    await runWithUpdates();
+  } else {
+    runWithNoUpdates();
+  }
+}
+
+function installChromium() {
+  _echo(`Add Puppeteer package to trigger Chromium installation script.`);
```
gergelyke: can we simply 'console.log' if we are in a node.js script?

vrt/ci.js
```diff
@@ -0,0 +1,351 @@
...
+  _exec(`yarn add puppeteer`);
```
gergelyke: what's the reasoning behind the underscore? can we use `exec`?

vrt/ci.js
```diff
@@ -0,0 +1,351 @@
...
+  } else {
+    const newSnapshotPullRequest = await openNewSnapshotPullRequest();
+    await addLabelsToNewPullRequest(newSnapshotPullRequest.number);
+    await addCommentToOriginalPullRequest(newSnapshotPullRequest.number);
+    await addOriginalAuthorAsReviewer(newSnapshotPullRequest.number);
+    _echo(
+      `Snapshots on \`${SNAPSHOT_BRANCH}\` must be merged into \`${BUILDKITE_BRANCH}\` before it can be merged into \`master\`.`,
+    );
+  }
+  _exit(1);
```
gergelyke: same comment as with console.log, can we use `process.exit`?

vrt/ci.js
```diff
@@ -0,0 +1,351 @@
...
+  });
+}
+
+// A few wrappers around shell commands
+function _echo(message) {
+  return shell.echo(`❖ VRT: ${message}`);
+}
+
+function _exec(command, callback) {
+  return shell.exec(command, {silent: true}, callback);
```
gergelyke: instead of shelljs, can we use https://nodejs.org/dist/latest-v10.x/docs/api/child_process.html#child_process_child_process_execsync_command_options?

that way, we wouldn't need the extra dependency

vrt/ci.js
```diff
@@ -0,0 +1,351 @@
...
+      `Posted a comment linking to snapshot PR on original PR: ${comment.data.html_url}`,
+    );
+  } catch (er) {
+    _echo(`Error creating comment on original PR.`);
+    _echo(er);
+  }
+}
+
+async function openNewSnapshotPullRequest() {
+  try {
```
gergelyke: I would remove the the try/catch blocks, and let the error propagate back to the main function - that way you only need to do error handling once

I'd do the same for all the utility funcitons

vrt/ci.js
```diff
@@ -0,0 +1,351 @@
...
+  }
+}
+
+async function addLabelsToNewPullRequest(newPullRequestNumber) {
+  try {
+    await octokit.issues.addLabels({
+      owner: `uber`,
+      repo: `baseweb`,
+      issue_number: newPullRequestNumber,
+      labels: [`bugfix`, `ci`],
```
chasestarr: maybe we should create a new label?

vrt/ci.js
gergelyke: so you wouldn't have to do the process.exits either

vrt/ci.js
sandgraham: oh i didn't know about this - yeah i think it can all be replaced then.

sandgraham:
@chasestarr ah forks - i may need to add a little logic to get that working. i will try that out next.

vrt/ci.js
sandgraham: yeah good idea - maybe `snapshots` or `vrt`. although i am starting to not like `vrt` since it is not a commonly known abbreviation.

gergelyke:
> @chasestarr ah forks - i may need to add a little logic to get that working. i will try that out next.

should we land this first, and send a follow up pr?

sandgraham:
i'm pretty sure this won't work as is with forks - i think it will create a branch but i dont think the pull request will work. i will add this in as a subsequent pr.

> will merging from master trigger vrts to update? - @chasestarr 

if the merge triggers CI then it should, yeah. the job itself doesn't care about master. it will take the current commit state and create a new `__image_snapshots__` directory based on that. if there are any differences with the current `__image_snapshots__` directory, it will open a branch/pr and notify you to look at updated files and merge in if you want to update.

i think of it as a dry run of updating snapshots yourself - with the option to accept changes if things seem right.

PR #2081, opened 2019-10-14
fix(tag): do not call onClick for the action

gergelyke:


src/tag/tag.js
```diff
@@ -89,7 +89,11 @@ class Tag extends React.Component<PropsT, {}> {
     const actionHandlers = disabled
       ? {}
       : {
-          onClick: onActionClick,
+          onClick: event => {
+            // we don't want onClick to be called when the close icon is clicked
+            event.stopPropagation();
```
nadiia: Is there any way to solve it without using `stopPropagation` in the base component?

gergelyke: I am not sure - do you have any recommendations?

nadiia: One option can be checking the target in the onClick handler before calling the one from props.

PR #2094, opened 2019-10-15
fix(select): change default width #1577

### LINKED ISSUE #1577: Width of StatefulSelect when inside a Block or FlexGrid
## Current Behavior

`StatefulSelect` is rendering super narrow when placed inside a Block or a FlexGrid.

In the linked demo I have outcommented the override, which makes it a min width, but I think it should not be needed to put an override to get a sensible default width?

Demo: https://codesandbox.io/s/block-with-grid-usage-h4u3j

## Expected Behavior

A better default width.

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v8.6.1  |
| React   |    16.8.2     |
| browser |    Chrome 75.x     |

Hrily:
#### Description

This commit fixes the default width of any select input component.
This also updates tests to incorporate the changes.

Fixes #1577

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/src/select/__tests__/__snapshots__/dropdown.test.js.snap
+++ b/src/select/__tests__/__snapshots__/dropdown.test.js.snap
@@ -47,6 +47,7 @@ Object {
   "List": Object {
     "component": Object {
       "$$typeof": Symbol(react.forward_ref),
+      "displayName": "StyledList",
       "render": [MockFunction],
     },
     "props": Object {
@@ -63,6 +64,7 @@ Object {
         "ListItem": Object {
           "component": Object {
             "$$typeof": Symbol(react.forward_ref),
+            "displayName": "StyledListItem",
             "render": [MockFunction],
           },
           "props": Object {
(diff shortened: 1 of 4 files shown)


Hrily:
The issue was seen because `AutosizeInput` sets `width` to `2px. The fix sets width only when `inputWidth > 2` else sets it to `null`, where the default value is used.

chasestarr:
Thanks for the contribution!
Please see the conversation here: https://github.com/uber/baseweb/issues/1577#issuecomment-542402134. It looks like this issue has been resolved in a non-related diff. Could you please add a `.scenario.js` that replicates the codesandbox reproduction in the issue so that we can protect against a regression?

nadiia:
Hi @Hrily, thanks for looking into this! 
The change seems to break the select's sizing, see the screenshot below: 
Screen Shot 2019-10-15 at 3 13 07 PM

Expected and current master behavior:
Screen Shot 2019-10-15 at 3 16 03 PM

Hrily:
Abandoning this since issue is resolved.

PR #2095, opened 2019-10-15
feat(data-table): adds row/column hover highlights

chasestarr:
#### Description

- Hovering over a cell, highlights all cells in that row. 
- Hovering over a column header, highlights all cells within that column. 

At first this feature was incredibly slow, averaging about `~340ms` per hover event. According to the book [Usability Engineering](https://www.nngroup.com/articles/response-times-3-important-limits/), response times longer than `100ms` causes users to feel as if they cannot manipulate the interface directly. 3x the recommended response-time budget was not ideal.

Initial performance measurement:
Screen Shot 2019-10-15 at 8 55 11 AM

Two performance improvements are introduced in this diff. First, I [fixed](https://github.com/uber/baseweb/commit/1e01fe040a37afb0c7a010c535e24be1656513de) a bug where hidden measurement elements continue to render and measure on each state update. This cut out a massive share of wasted operations which reduced the interaction time to `~118ms`. Still a hair shy of our budget, but still did not feel quite snappy enough for an always-present interaction.

After measurement elements removed:
Screen Shot 2019-10-15 at 8 54 15 AM

Secondly, I've started to [conditionally update](https://github.com/uber/baseweb/compare/data-table-row-column-hover?expand=1#diff-a2301d2239124fb8ee9a9bbe7bbba752R73) cells only when necessary. For example, if a particular cell was neither previously highlighted nor will be highlighted next it will avoid an update. Special attention was made to ensure that sorting, filtering, and updating the row/column data continued to operate as expected. Doing so dropped the interaction time close to `100ms` and was now averaging just about `~25ms`.

After conditionally rendering cells:
Screen Shot 2019-10-14 at 4 30 48 PM

#### Scope

- [x] Minor: New Feature


tajo:
There is a bug with batch actions. When I select the row, I have to first leave it to see the updated checkbox.

toggle-hover

chasestarr:
Good catch, that was tricky since the initial row select will cause the position to jump and change the hovered row

tajo:
> Good catch, that was tricky since the initial row select will cause the position to jump and change the hovered row

Yea, it took me a while to see what's happening. First I thought it's just super slow.

PR #2101, opened 2019-10-16
test(vrt): add mobile screen size

gergelyke:


gergelyke:
> couple comments ~~ seems good otherwise

updated

PR #2117, opened 2019-10-17
fix(vrt): ci snapshots

sandgraham:
Some bug fixes and clean up work for CI snapshot job.

I noticed in #2103 if the snapshot update task fails the process still exits with `0` so the CI job passes. This means you could just throw an error in the test suite to pass the CI snapshot check. [Link to that Buildkite job](https://buildkite.com/uberopensource/baseweb/builds/12093#6744b71f-be4e-448e-8bfe-8d4bdeac3447).

Also fixes buildkite `[skip ci]` issue. i think the problem is that on buildkite sometimes the message is the commit message and sometimes it is the pr name. adding to both means it always gets applied. not beautiful stuff


.buildkite/pipeline.yml
```diff
@@ -17,6 +17,8 @@ steps:
         image-repository: 027047743804.dkr.ecr.us-east-2.amazonaws.com/uber
     agents:
       queue: builders
+    concurrency: 1
+    concurrency_group: 'e2e-and-vrt-test-server'
```
sandgraham: @gergelyke I tried this out as a solution to the port clash race condition- the thinking here being that if we remove the server build race condition if this to being built one at a time.

gergelyke: i think the issue is not with the builders, but when it runs

vrt/ci.js
```diff
@@ -333,6 +314,14 @@ function someSnapshotsWereUpdated() {
   return result;
 }
 
+function handleError() {
+  // Fail the CI job if an error propagates to this level
+  // Any non critical function should handle error and log a useful message
+  // Basically everything other than the test results and pushing a new branch is non critical (PR/Comments/Labels/Reviewers)
+  log(`The job has failed, but it is not a failure.`);
+  process.exit(1);
```
gergelyke: you don't need this line, it will crash anyways

.buildkite/pipeline.yml
sandgraham: yeah this doesn't work

vrt/ci.js
sandgraham: not sure why but it does not fail the ci job. i tried removing the handler, but you get a `UnhandledPromiseRejectionWarning` and then i tried re-throwing the er in the handler but it still doesn't work.

gergelyke:
the race condition is resolved here https://github.com/uber/baseweb/pull/2129

sandgraham:
this should be ready to merge in - it has a fix for ci skip and it fixes a scenario where a run time error allows the job to pass

PR #2118, opened 2019-10-17
feat(data-table): row actions

chasestarr:
#### Description

Screen Shot 2019-10-17 at 1 27 06 PM

#### Scope

- [x] Minor: New Feature


src/data-table/measure-column-widths.js
```diff
@@ -62,7 +62,7 @@ type MeasureColumnWidthsPropsT = {
 };
 
 // sample size could likely be generated based on row count, to have higher confidence
-const MAX_SAMPLE_SIZE = 10;
+const MAX_SAMPLE_SIZE = 50;
```
chasestarr: I changed this to handle the variable movie title lengths in a new example.. Signals to me that it needs to be configurable. Will create a new PR to do so

gergelyke:
Looks great! one thing I found is that when you select a row, the actions are not positioned correctly:

Screen Shot 2019-10-18 at 3 56 39 PM

chasestarr:
> Looks great! one thing I found is that when you select a row, the actions are not positioned correctly:

good catch!

chasestarr:
refactored the code so that overlay position is unrelated to y offset from parent container

PR #2136, opened 2019-10-18
docs(yard): react view proposal

tajo:
**This is not the final README or documentation**. It also doesn't go into a great detail since that will be in flux. However, it should give you the idea about overall architecture and APIs.

[**RENDERED**](https://github.com/uber/baseweb/blob/0c121e7b3e7d15fb50635be3eb4b45ebd62a0649/documentation-site/components/yard/README.md)

### DIFF
--- a/documentation-site/components/yard/README.md
+++ b/documentation-site/components/yard/README.md
@@ -1,0 +1,254 @@
+# react-view
+
+**React View is an interactive playground, documentation and code generator** for your  React components. All that packaged in a single component that can be dropped into any application.
+
+Use it to document your component APIs. Users can then live-edit component props and grab the generated ready-to-use source code.
+
+[**Check this demo.**](https://baseweb.design/components/input/)
+
+All you need to do is describing your component with a configuration file and React View will do the rest.
+
+React View has **three main parts**:
+
+- **rendered component**
+- **knobs** - each prop is mapped into an input, radio group, textarea or checkbox, those values are live-editable
+- **code** - gets updated as you edit the knobs, it can be also edited directly
+
+## Always Synchronized
+
+These three parts are **3-way synchronized**. 😱 What does it mean?
+
+- An interaction with the rendered component updates the knobs and code.
+- Editing the knobs updates the rendered component and code.
+- Editing the code updates the knobs and rendered component.
+
+## Installation
+
+    yarn add react-view
+
+## Building Blocks
+
+- **react-view basic**
+    - Comes with no UIs or styling, focused on the data.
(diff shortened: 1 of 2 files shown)


documentation-site/components/yard/README.md
```diff
@@ -0,0 +1,254 @@
...
+  View,
+  Compiler,
+  Editor,
+  Error,
+  PropTypes,
+} from 'react-view';
+
+<View
+  componentName="Button"
+  componentPlaceholder={<div />}
```
chasestarr: what is this for? render a spinner while initializing?

```suggestion
  componentPlaceholder={<div>Loading...</div>}
```

documentation-site/components/yard/README.md
```diff
@@ -0,0 +1,254 @@
...
+      type: PropTypes.Boolean,
+      description: 'Is checkbox checked?'
+    }
+  }}
+  onUpdate={({ code, props }) => {
+    // useful to update the URL
+    console.log(code, props);
+  }}
+  initialCode="useful to hydrate the state from URL"
+  providerValues={['buttonFillPrimary']}
```
chasestarr: how does this string map to the `ThemeProvider` state? below, `providerProps` are spread onto the `ThemeEditor` component, but i'm missing how it can be manually wired up based on this interface

documentation-site/components/yard/README.md
tajo: Yea!

documentation-site/components/yard/README.md
tajo: This API is a struggle. I can imagine many different provider (theme) APIs and would be cool to cover them all. The `providerValues` example is actually wrong, it should look like this (will update):

```js
providerValues={{
  buttonFillPrimary: '#FCFCFC'
}}
```

So there is no internal assumption about the theme (getting values out of the context).

The overall idea is to pass `value: any` into `providerValues`. Then you have to do four things:

- in `provider.parse`specify how to get the `value` out of the code (string)
- in `provider.generate` specify how to get the AST of provider based on `value`
- define related imports
- build an UI editor based on `providerProps` (values, initialValues, onChange, reset)

I think this should be quite flexible. Wdyt?

chasestarr: Ah ok, I think I misunderstood the interface. This sounds reasonable to me

tajo: alright, updated with a new paragraph about the provider

tajo:
@chasestarr I changed the API to use a hook instead of render props. I think it better separates the data and UI bits + there are no assumptions about the defaults.

PR #2138, opened 2019-10-18
feat(data-table): text search row filter

chasestarr:
#### Description

- filters rows for if a cell in that row contains a string which includes provided search query
- cells with matching substrings are highlighted with bold text
- displays empty results message if no rows to render
- provides a new `minWidth` property on columns

Screen Shot 2019-10-21 at 12 57 09 PM

Screen Shot 2019-10-21 at 12 57 40 PM

#### Scope

- [x] Minor: New Feature


gergelyke:
when selecting a filter through the column headers, the text input disappears, but the text highlight remains - is that the expected behavior? should we always show the text input?

chasestarr:
> when selecting a filter through the column headers, the text input disappears, but the text highlight remains - is that the expected behavior? should we always show the text input?

thanks, I remembered discussing this operation but then I checked the mocks again and it should be like below. will update

Screen Shot 2019-10-21 at 9 16 24 AM

tajo:
Screen Shot 2019-10-21 at 2 14 27 PM

There is some overflowing (source code buttons) going on.

chasestarr:
> There is some overflowing (source code buttons) going on.

Thank you! I was just working on this issue. fix it up now

chasestarr:
gonna merge this... I can make changes in a follow-up pr if warranted

Issue #2144, opened 2019-10-19
[Typescript] Missing align prop in StatefulRadioGroup component


andersonba:
Codesanbox
https://codesandbox.io/s/horizontal-alignment-m1yxf?eslint=1&fontsize=14&module=%2Fexample.tsx

## Current Behavior

I can't use the "align" property with the StatefulRadioGroup.

<img width="606" alt="image" src="https://user-images.githubusercontent.com/1501013/67151361-149bb400-f2bd-11e9-8b90-bc7170362855.png">

## Expected Behavior

Should work properly with the same properties from RadioGroup.

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v9.9.0  |
| React   | 16.10.2 |
| Typescript | 3.6.4 |

- [x] I have searched the [issues](https://github.com/uber/baseweb/issues) of this repository and believe that this is not a duplicate.

PR #2145, opened 2019-10-21
fix(radio): add align prop to ts defs

gergelyke:
Fixes https://github.com/uber/baseweb/issues/2144


gergelyke:
> lgtm, `align` should likely be converted to an enum in flow though

updated - is this what you had in mind?

chasestarr:
technically a breaking change now.. although, if this does break any applications it will be a clear bug so maybe acceptable. Now that it's an enum in flow, it should be an enum in yard

PR #2153, opened 2019-10-21
docs(survey): automated csat collection

gergelyke:
To force the survey to show, visit this url: https://baseweb-git-docs-csat.uber-ui-platform.now.sh?survey

The idea with this is to replace delighted, and survey returning visitors every two months.


documentation-site/components/survey.js
```diff
@@ -0,0 +1,148 @@
...
+          cursor: 'pointer',
+          position: 'absolute',
+          top: theme.sizing.scale400,
+          right: theme.sizing.scale400,
+        })}
+        onClick={() => {
+          setIsOpen(false);
+        }}
+      >
+        <Delete size="24px" />
```
sandgraham: can't see the `x` on dark mode

documentation-site/components/survey.js
```diff
@@ -0,0 +1,148 @@
...
+    Cookies.set(cookies.FIRST_SEEN, now.getTime(), {
+      // by default, cookies are session cookies, so without an expiration they are removed
+      // when the browser window is closed.
+      expires: new Date(now.getFullYear() + 10, now.getMonth()),
+    });
+    return {showSurvey: false, delay};
+  }
+
+  // the person only knows base web for less than a week
+  if (firstSeen > ONE_WEEK_AGO) {
```
sandgraham: should this be `<`?

documentation-site/components/survey.js
```diff
@@ -0,0 +1,148 @@
...
+    return {showSurvey: false, delay};
+  }
+
+  // the person was surveyed less than two months ago, bailing out
+  if (lastSurveyed > TWO_MONTHS_AGO) {
```
sandgraham: same here

documentation-site/components/survey.js
chasestarr: (hypothetical numbers)
if now is `200ms` since epoch, one week ago was `100ms`, and first seen is `120ms`: that means that first seen is within the one week range

this conditional could be modeled closer to the comment description as a delta between now and the measured times

sandgraham: oh gotcha - at a glance it seems backwards. "first seen over a week ago"

gergelyke: hm, let me change things around so it is less confusing

documentation-site/components/survey.js
```diff
@@ -0,0 +1,148 @@
...
+    return {showSurvey: false, delay};
+  }
+
+  Cookies.set(cookies.LAST_SURVEYED, now.getTime(), {
+    expires: new Date(now.getFullYear() + 10, now.getMonth()),
```
chasestarr: is this correct? looks like we set expiration to 10 years in the future. should be be like below?
```suggestion
    expires: new Date(now.getFullYear(), now.getMonth() + 2),
```

gergelyke: the idea was to set it for 10 years, as we use the value of this cookie to decide if we show it to the user - does that make sense?

also, if we'd like, we could set the expiration for 2 months, they way we'd check if the cookie value is undefined

documentation-site/components/survey.js
gergelyke: @chasestarr @sandgraham just updated the conditional - does it make more sense now?

documentation-site/components/survey.js
chasestarr: ha right, more than one datetime going on in the cookie

documentation-site/components/survey.js
chasestarr: don't see an update, push the change?

gergelyke: oh lol, just left my terminal:

```
giut: command not found
```

should be updated now

PR #2155, opened 2019-10-22
chore: bump flow to 0.110

tajo:
Turns out that almost all errors were input related.

Input types need to be exact due to the new way how Flow handles type spreads. This could cause some flow error if people are passing unknown props downstream.

**EDIT: Styletron updated. This PR is ready for merge.**


gergelyke:
from a customer point of view, will it be a breaking change?

src/input/types.js
```diff
@@ -112,21 +112,22 @@ export type BaseInputPropsT<T> = {
...
-  type: string,
+  type?: string,
   /** Input value attribute. */
   value?: string,
   rows?: number,
-};
+|};
 
-export type InputPropsT = {
+export type InputPropsT = {|
```
chasestarr: this will break if additional props are provided to `Input`. While it will flag obvious errors that never affected the component it may be considered a breaking change since consumers will need to update their code.

tajo: Well, we could consider it as a fixing a bug. If you are passing through props that have no effect on the Input component you might have a problem in your application?

tajo: I would not be that strict (breaking change bump) when it comes to improving static types.

chasestarr: hmm 105 also made object types exact by default right? because of that, i'm surprised this was required and it may be more impactful than a single component

tajo: afaik, they just added a config option to enable them by default

tajo:
Styletron bumped. It's ready to go. Just waiting for the Fusion release.

PR #2158, opened 2019-10-22
fix(tether): ensures no flow errors from tether

chasestarr:
#### Description

Applications using `baseui` reported errors coming from the `node_modules/baseui/.../tether` directory. The reason why this was not caught in this repo was that the `yarn.lock` file pinned `popper.js` version to `1.14.3` where applications were installing `1.16.0` which contained new flow type definitions.

#### Scope

- [x] Patch: Bug Fix

### DIFF
--- a/src/layer/tether.js
+++ b/src/layer/tether.js
@@ -7,6 +7,7 @@ LICENSE file in the root directory of this source tree.
 // @flow
 import * as React from 'react';
 import Popper from 'popper.js';
+import type {Instance as PopperInstance} from 'popper.js';
 import {toPopperPlacement, parsePopperOffset} from './utils.js';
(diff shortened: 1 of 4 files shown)


src/layer/tether.js
```diff
@@ -66,6 +67,9 @@ class Tether extends React.Component<TetherPropsT, TetherStateT> {
   initializePopper() {
     const {placement, popperOptions} = this.props;
     const {modifiers, ...restOptions} = popperOptions;
+
+    if (!this.props.anchorRef || !this.props.popperRef) return;
```
gergelyke: why do we need this line?

chasestarr: those two props are optional in the component interface, but required to the popper interface

PR #2164, opened 2019-10-23
feat(yard): hover tokens to show prop type

chasestarr:
#### Description

Screen Shot 2019-10-23 at 2 16 35 PM

Only added to `Accordion` for now. If we agree on this approach I can apply to the rest of components


gergelyke:
this looks really great!

documentation-site/components/yard/yard.tsx
```diff
@@ -277,6 +281,53 @@ const Yard: React.FC<
             updateCode(dispatch, newCode);
           }
         }}
+        transformToken={tokenProps => {
```
tajo: I would move it into a separate file to keep yard.tsx as short as possible

tajo:
Works great! Should be this part of OSS version?

documentation-site/components/yard/type-definition.js
```diff
@@ -0,0 +1,385 @@
+/*
```
tajo: Is this file from some other project? Should we have some tests?

chasestarr: yeah this is from the extract-react-types package. I've vendored the code for now, but going to put up a pr to that package to help out

chasestarr:
hmm not sure if it should be in the initial OSS version we may want to see how stable/useful it is

documentation-site/components/yard/type-definition.js
chasestarr: related pr: https://github.com/atlassian/extract-react-types/pull/101

gergelyke:
👏

Issue #2048, opened 2019-10-09
[Select] multi height changes when 0/1 options selected


chasestarr:
## Current Behavior

visible on the docs site https://baseweb.design/components/select/#select-as-multi-pick-search
- open the dropdown
- select an option
- notice that the input height grows from 40px to 44px
- only affects the default and compact sizes, large is stable

## Expected Behavior

height should be consistent

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v9.7.1  |
| browser |chrome |

- [x] I have searched the [issues](https://github.com/uber-web/baseui/issues) of this repository and believe that this is not a duplicate.

gergelyke:
hi @r4zorgeek - sure, thank you for offering your help!

I've just assigned the issue to you

houmark:
This seems to affect more situations for us. We have 3 Select's next to each other, two of them multiple and the last one is a single select. The multiple renders 4px less tall than the single Select (when multiple are empty as they expand in height with "tags" added).

Our "hotfix" has been to force the `minHeight` of the multiple through overrides. It does not matter what sizes the fields are (but we are using the standard size).

Screenshot with no `minHeight` override:
https://share.getcloudapp.com/o0uxvqNP

With `minHeight` on `ControlContainer` override set to 48px:
https://share.getcloudapp.com/L1uYpOPD

Screenshot taken on a fresh Basweb v9.8.0.

r4zorgeek:
@chasestarr @gergelyke Hey I am new to open source so can you please help me with this. I tried navigating through the files it seems complicated. Please guide me.

My guess is this problem is occurring due to scaling property of CSS.

chasestarr:
I believe it has something to do with this function https://github.com/uber/baseweb/blob/master/src/select/styled-components.js#L25

r4zorgeek:
@gergelyke Hey have you fixed this issue?

gergelyke:
yes, going to merge today - sorry for taking it from you, but we had quite some internal folks bumping the prio on this one

r4zorgeek:
@gergelyke It's okay. Actually it was my mistake I couldn't complete in time because I was busy with college. Let me know any other issues which will be good for me(new to open source).

PR #2166, opened 2019-10-24
fix(select): multi and single select has the same height

gergelyke:
Fixes https://github.com/uber/baseweb/issues/2048


chasestarr:
looks like there are some issues highlighted in the vrts

PR #2171, opened 2019-10-25
feat(checkbox): allow checkboxes to be unlabeled

shannonmoeller:
#### Description

When using `<Checkbox />` with no label, the result is inaccessible and has extraneous padding. This PR does two things:

1. Adds a `title` prop. This provides screen readers with content to read aloud and renders a native tooltip on long-hover for sighted users.
2. ~Prevents rendering of the label when there are no children. This prevents there from being additional padding to the right of the checkbox.~ Reverted change as it could be considered as breaking. Reintroduced in a standalone PR, #2172.

image

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change


gergelyke:
thank you @shannonmoeller ! can you please also add to? https://github.com/uber/baseweb/blob/master/documentation-site/components/yard/config/checkbox.ts

shannonmoeller:
The visual changes vrt caught could be considered by users to be a breaking change. Will remove the label change from this PR to fix the regression and make this safe to merge in v9. Then will create another PR for the label padding targeted for v10.

PR #2172, opened 2019-10-25
fix(checkbox): remove padding from unlabeled checkboxes

shannonmoeller:
#### Description

Reintroduces change 2. from #2171: Remove extraneous padding on right side from unlabeled checkboxes.

image

#### Scope

- [ ] Patch: Bug Fix
- [ ] Minor: New Feature
- [x] Major: Breaking Change


gergelyke:
@shannonmoeller would this still be a breaking change? if so, should we just make a comment in the codebase, and close this pr?

shannonmoeller:
It's a visual change. It impacted work Chase was doing on the Data Table components, so we thought it was worth flagging as a breaking change.

Not sure how these are handled with baseweb's new release schedule. You tell me. :)

PR #2189, opened 2019-10-29
chore: add volta

tajo:


tajo:
> should we downgrade Node in Docker too, to make sure it is the same env?

might be a good idea, but it's in uber/web-base-image?

PR #2192, opened 2019-10-29
chore(project): update deps

gergelyke:


package.json
```diff
@@ -184,23 +184,23 @@
...
+    "card-validator": "^6.2.0",
+    "date-fns": "^2.6.0",
     "just-extend": "^4.0.2",
-    "memoize-one": "^5.0.0",
+    "memoize-one": "^5.1.1",
     "polished": "^3.2.0",
     "popper.js": "^1.16.0",
     "react-dropzone": "^9.0.0",
-    "react-focus-lock": "^1.9.1",
+    "react-focus-lock": "^2.2.1",
```
gergelyke: this should reduce the bundle size of react-focus-lock with 60% - it needs React 16.8, that's why it's a major

PR #2194, opened 2019-10-30
feat(data-table): numberical filter between case

chasestarr:
#### Description

- adds 'between' operation to include values within upper/lower bounds
- adds input auto-focus dependent on operation choice
- applies compact size to inputs and buttons

Screen Shot 2019-10-29 at 5 50 20 PM

#### Scope

- [x] Minor: New Feature


src/data-table/column-numerical.js
```diff
@@ -8,9 +8,9 @@ LICENSE file in the root directory of this source tree.
 
 import * as React from 'react';
 
-import {Button} from '../button/index.js';
+import {Button, SIZE} from '../button/index.js';
```
gergelyke: should we import this as `BUTTON_SIZE`, to be consistent with `INPUT_SIZE`?

PR #2196, opened 2019-10-30
docs(validation): add email validation example

lessenadam:
#### Description

This PR does 2 things.

1. Adds a devDependency on the npm package [email-validator](https://www.npmjs.com/package/email-validator)
2. Adds a validation example on the `FormControl` documentation site that can be easily copied, pasted, and modified as needed in applications using baseui.

Screenshot:
Screen Shot 2019-10-30 at 4 31 49 PM

Screen recording:
validation-2

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change


gergelyke:
Thank you Adam for putting this together! I'd remove the validation logic from the input component, and put together an example using https://www.npmjs.com/package/email-validator (or something similar). That way, we can keep the API surface smaller. Does that make sense?

gergelyke:
@lessenadam - I wonder if it would be a more scalable, and standard-friendly solution to add support for the `:invalid` pseudo selector : https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/email

that way, a simple `type=email` would work, and folks could also leverage the `pattern` prop, and define more complex validations

lessenadam:
Hey @gergelyke, thanks for the feedback. A couple thoughts / questions:

1. My guiding principle was to try and make an example that was easy to copy and paste and "would just work", but was also easy to extend if needed. 
2. Regarding `:invalid`, would the thought be to use a ref and check for the presence of the pseudo -selector using something like [node.matches](https://developer.mozilla.org/en-US/docs/Web/API/Element/matches), or actually use `:invalid` to style the form control and input? 
3. If it's the latter, I'm not exactly sure how we would apply the styles. Plus, we wouldn't be able to use the error, etc properties exposed by form control and input
4. Is the question more about **how** we validate the input? Ie, using an npm package or just some utility js vs using the browser's native email validation? I'm also not 100% confident in the browser's built-in validation. It looks like it passes any `<string>@<string>`, and doesn't validate an actual `.com`, etc domain
5. I can also extend the example to show it in the context of a form with a submit button so that you can see the "this field is required" and "this must be an email" alerts on submit

PR #2199, opened 2019-10-30
docs(walkthrough): add yard info

gergelyke:


documentation-site/components/yard/index.tsx
```diff
@@ -19,28 +19,32 @@ const YardWrapper: React.FC<
...
-        pathname={router.pathname}
-        urlCode={
-          typeof queryStringName !== 'undefined'
-            ? router.query[queryStringName]
-            : router.query.code
-        }
-        queryStringName={queryStringName}
-        placeholderElement={() => {
-          if (!placeholderHeight) return null;
+    <React.Fragment>
```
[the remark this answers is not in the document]
gergelyke: ops, yeah, that's a leftover - thanks Chase

tajo:
I have to refresh the page to not get the modal on every page.

gergelyke:
> I have to refresh the page to not get the modal on every page.

good catch, fixed

documentation-site/components/walkthrough.js
```diff
@@ -0,0 +1,210 @@
...
+} from 'baseui/modal';
+import {StatefulTabs, Tab} from 'baseui/tabs';
+import {withStyle, useStyletron} from 'baseui';
+import {Button} from 'baseui/button';
+
+import {Label3, Label4, Paragraph3} from 'baseui/typography';
+import {StatefulTooltip} from 'baseui/tooltip';
+
+import CookiesConstructor from 'universal-cookie';
+const Cookies = new CookiesConstructor();
```
tajo: This should probably be inside of the component. Otherwise the `cookie` value never changes unless you reload.

documentation-site/components/walkthrough.js
```diff
@@ -0,0 +1,210 @@
...
+  const [useCss, theme] = useStyletron();
+  const [isOpen, setIsOpen] = React.useState(!cookie);
+  const [currentPage, setPage] = React.useState(0);
+  const now = new Date();
+  Cookies.set(COOKIE_NAME, now.getTime(), {
+    // by default, cookies are session cookies, so without an expiration they are removed
+    // when the browser window is closed.
+    expires: new Date(now.getFullYear() + 10, now.getMonth()),
+    path: '/',
+  });
```
tajo: This should be inside of the `React.useEffect` so it's not called on every single re-render?

documentation-site/components/walkthrough.js
gergelyke: this should solve it https://github.com/uber/baseweb/pull/2199/files#diff-12eed97910803029f83c0e98ee55f26aR35

documentation-site/components/walkthrough.js
gergelyke: the thinking here was to keep the code simpler - this is the only code that runs: https://github.com/reactivestack/cookies/blob/master/packages/universal-cookie/src/Cookies.ts#L64-L76

should I still put this into a `useEffect`?

gergelyke: I'll just quickly do this change

gergelyke: @tajo should be ready for another pass

documentation-site/components/walkthrough.js
```diff
@@ -0,0 +1,212 @@
...
+          [`@media screen and (max-width: ${theme.breakpoints.medium}px`]: {
+            display: 'none',
+          },
+        })}
+      >
+        <Label3
+          overrides={{Block: {style: {fontSize: '20px', cursor: 'pointer'}}}}
+        >
+          <span role="img" aria-label="open announcement modal">
+            🎁
```
tajo: I first missed this. That's a little cute touch!

gergelyke:
I am thinking if we should not show this on the blog pages?

tajo:
> I am thinking if we should not show this on the blog pages?

Yea, we should show it only for components.

tajo:
Why not just display it when url has `/component/`? Is there a reason why to show it on the landing page or guides?

PR #2204, opened 2019-10-30
docs(yard): fix mobile view

gergelyke:
Fixes https://twitter.com/andreujuanc/status/1185471571248459776


tajo:
Should we apply some media queries? Now it always wraps everything into a single column even on a wide screen.

tajo:
Also, the same fix should be used for the button group in "Style Overrides" panels and "Theme" inputs. Thanks for fixing this!

gergelyke:
@tajo updated

PR #2206, opened 2019-10-31
fix(modal): make modal scrollable while cursor is on the backdrop

nadiia:


documentation-site/examples/modal/basic.js
```diff
@@ -23,8 +23,65 @@ export default () => {
         <ModalHeader>Hello world</ModalHeader>
         <ModalBody>
           Proin ut dui sed metus pharetra hend rerit vel non mi.
+          <br />
```
gergelyke: do we need these additions? I think we already have a modal example with lots of text, called `Dealing with a buried interactive element `

src/modal/modal.js
```diff
@@ -254,12 +254,12 @@ class Modal extends React.Component<ModalPropsT, ModalStateT> {
               {...sharedProps}
               {...rootProps}
             >
-              <Backdrop
-                onClick={this.onBackdropClick}
-                {...sharedProps}
-                {...backdropProps}
-              />
               <DialogContainer {...sharedProps} {...dialogContainerProps}>
+                <Backdrop
```
chasestarr: moving this element is a bit dangerous considering if anyone overriding the modal relies on it, but i suppose it's fixing a bug therefore justified

PR #2216, opened 2019-11-01
fix(polished): import darken and lighten only

gergelyke:


kahwee:
It's 33.01 KB for importing using `import {darken} from polished` via Fusion. 

Importing via darken and lighten directly is 14.88 KB.

Screen Shot 2019-11-01 at 4 43 46 PM

Issue #2208, opened 2019-10-31
[Accordion] toggle icon overrides not properly applied


chasestarr:
The `ToggleIcon` overrides on `Panel` is not properly configured. When the icons are 'globally' overridden. See https://codesandbox.io/s/pagination-icon-overrides-yt38g for reproduction

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v9.12.0  |

- [x] I have searched the [issues](https://github.com/uber/baseweb/issues) of this repository and believe that this is not a duplicate.

PR #2228, opened 2019-11-04
fix(accordion): handle panel icon overrides

gergelyke:
Fixes https://github.com/uber/baseweb/issues/2208


gergelyke:
> This lgtm. I tried to test some other solutions to no avail. Do we think this affects all icon overrides across the source code?

I don't think so - I think it was a special case because we change the icon based on if the accordion is expanded or not

PR #2238, opened 2019-11-05
test(vrt): fix some vrt with interactions tests

nadiia:


src/datepicker/__tests__/datepicker-range.e2e.js
```diff
@@ -54,22 +54,6 @@ describe('Datepicker, Range', () => {
...
-    await page.waitFor(selectors.calendar);
-    await page.click(selectors.rightArrow);
-    await page.waitFor(selectors.day4);
-    const value = await page.$(selectors.day4);
-    const highlighted = await page.evaluate(
-      element => element.getAttribute('data-highlighted'),
-      value,
-    );
-    // Check the value of a custom data-highlighted attribute
-    expect(highlighted).toBe('false');
```
nadiia: This test didn't test what it was intended for but just checked a single day to be not highlighted. There is a better screenshot test for it.

src/datepicker/__tests__/datepicker.scenario.js
```diff
@@ -20,7 +20,7 @@ export const component = () => {
     <React.Fragment>
       <Button
         onClick={() => {
-          setDate(new Date('2019/03/18'));
+          setDate(new Date('2019/07/01'));
```
nadiia: changed the selected date to be in a different month than the default initially highlighted day

src/drawer/__tests__/drawer-select.scenario.js
```diff
@@ -7,16 +7,18 @@ LICENSE file in the root directory of this source tree.
...
 import {StatefulSelect} from '../../select/index.js';
 import {Drawer} from '../index.js';
 
 export const name = 'drawer-select';
 
 const Example = () => {
+  const [isOpen, setIsOpen] = React.useState(true);
   return (
     <div>
-      <Drawer closeable={false} isOpen>
```
nadiia: The drawer was not even closable so the check for it to stay open when an option from select's dropdown clicked didn't test anything.

PR #2245, opened 2019-11-06
feat(modal): move the backdrop bg and onClick handler to DialogContainer

nadiia:


gergelyke:
@nadiia should we close this PR for now?

Issue #2233, opened 2019-11-04
[DataTable] Filters can be positioned outside of table context


chasestarr:
Consider the situation where I want to build and edit filters outside of the `Add Filter` button menu. What's the best way to send that information from a separate react tree to the table contents? One suggested option was to manage this information through react context providers/consumers. We should put together a prototype to test the ergonomics/pitfalls of doing so.

PR #2246, opened 2019-11-06
feat(data-table): refactor to enable uncontrolled usage

chasestarr:
Prerequisite to https://github.com/uber/baseweb/issues/2233

#### Scope

- [x] Minor: New Feature


src/data-table/column-categorical.js
```diff
@@ -28,8 +28,9 @@ type OptionsT = {|
 |};
 
 type FilterParametersT = {|
-  selection: Set<string>,
+  description: string,
```
gergelyke: why do we need the additional `description` param?

chasestarr: it comes from the complexity around supporting 'custom' columns. How can we translate a custom set of filter properties into a string that is rendered into a 'tag' component

PR #2249, opened 2019-11-06
docs(seo): add guide & small fixes

gergelyke:


documentation-site/examples/accordion/renderpanelcontent.js
```diff
@@ -6,7 +6,7 @@ const content =
   'Praesent condimentum ante ac ipsum aliquam, ac scelerisque velit sagittis. Ut sit amet libero scelerisque, accumsan ante vitae, hendrerit tellus. Nullam metus est, vehicula a aliquet id, lobortis in mauris.';
 
 export default () => (
-  <Accordion renderPanelContent>
+  <Accordion renderAll>
```
gergelyke: it's not a breaking change, but now it supports both, to have the same API as the Tab comonent

src/accordion/accordion.js
```diff
@@ -27,6 +27,7 @@ export default class Accordion extends React.Component<
       expanded: [],
     },
     renderPanelContent: false,
```
nadiia: Could you add a `TODO` to remove the older `renderPanelContent ` prop in the next major?

PR #2255, opened 2019-11-06
fix(build): cjs, es and esm

gergelyke:


rtsao:
What's the difference between `es` and `esm`?

Also, should an env preset target be specified? Or is the intent only to transform modules (in that case maybe just use https://babeljs.io/docs/en/babel-plugin-transform-modules-commonjs)

gergelyke:
> What's the difference between `es` and `esm`?
> 
> Also, should an env preset target be specified? Or is the intent only to transform modules (in that case maybe just use https://babeljs.io/docs/en/babel-plugin-transform-modules-commonjs)

The idea with `es` is that we don't do any transpilation, just ship the code as is - similarly to [material-ui](https://github.com/mui-org/material-ui/blob/master/babel.config.js)

`esm` and `commonjs` is used to determine if modules should be transpiled https://babeljs.io/docs/en/babel-preset-env#modules

it seems like `babel-plugin-transform-modules-commonjs` does the same thing?

Issue #2258, opened 2019-11-07
Docs: Overriding password masking mechanism


LXSMNSYC:
## Current Behavior

<img width="616" alt="image" src="https://user-images.githubusercontent.com/4783372/68373130-329e5b00-017d-11ea-9919-2874231f1276.png">

According to the docs, the way to override the mask toggle button feature is to pass an Override object which replaces the mask toggle button component.

I tried using the same shape of the Override object as given in the examples, but it seems that the field does not exist.

Contrary to the docs, `<Password>` component does not exist in the library, which makes it impossible to override it.

## Expected Behavior

`<Input>` component accepts the `MaskToggleButton` property from the Override object to replace the mask toggle button feature.

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v9.14.0  |
| React   |  v16.11.0  |
| Typescript | v3.6.3  |

- [x] I have searched the [issues](https://github.com/uber/baseweb/issues) of this repository and believe that this is not a duplicate.

gergelyke:
You can do something like this : https://codesandbox.io/s/input-for-passwords-44emc

LXSMNSYC:
the problem is that I am using TypeScript and it isn't the type definitions.

PR #2260, opened 2019-11-07
docs(input): password example to use input component

gergelyke:
closes https://github.com/uber/baseweb/issues/2258


LXSMNSYC:
What about the type definitions?

gergelyke:
@LXSMNSYC - I'll take a look. 

Next time you encounter a bug, and if you have the time, I'd like to invite you to send a PR - we always welcome contributions, and it helps our team to keep up with the open tickets. Thanks a lot in advance! 🤗

PR #2262, opened 2019-11-07
docs(versioning): update the policy

gergelyke:


documentation-site/pages/discover-more/versioning-policy.mdx
```diff
@@ -11,22 +11,27 @@ export default Layout;
...
+## Gradual Upgrades via Warnings
+
 _Development builds of Base Web include many helpful warnings._
 
-In the event of a future breaking change, we will add warnings which indicate the feature is scheduled for deprecation and will provide migration instructions. Deprecation warnings may be added in a minor version. Following major versions will remove the deprecated feature.
+In the event of a future breaking change, we will add warnings that indicate the feature is scheduled for deprecation and will provide migration instructions. Deprecation warnings may be added in a minor version. The following major versions will remove the deprecated feature.
+
+## Breaking Changes
+
+Breaking changes are inconvenient for everyone, so we thrive to **limit the number of breaking changes by releasing only one major version a year, around September.** To help with the migration, we provide codemods for every major release. We may release major versions at any time in cases of unforeseen emergencies.
```
gergelyke: @chasestarr @tajo @nadiia @sandgraham what do you think about this? do you think it's realistic?

chasestarr: This sounds good to me

documentation-site/pages/discover-more/versioning-policy.mdx
```diff
@@ -6,27 +6,51 @@ LICENSE file in the root directory of this source tree.
...
+<Notification
+  overrides={{
+    Body: {style: {width: 'auto'}},
+  }}
+>
+  The release schedule is offered as general guidance and may be adjusted by us
+  when necessary.
+</Notification>
+
+Breaking changes are inconvenient for everyone, so we thrive to **limit the number of breaking changes by releasing only one major version a year, around September.** To help with the migration, we provide codemods for every major release.
```
sandgraham:
```suggestion
Breaking changes are inconvenient for everyone, so we strive to **limit the number of breaking changes by releasing only one major version a year, around September.** To help with the migration, we provide codemods for every major release.
```

documentation-site/pages/discover-more/versioning-policy.mdx
sandgraham: I'm a little worried that the one major a year will end up as a [christmas tree major](https://en.wikipedia.org/wiki/Christmas_tree_bill), with lots of unrelated breakages included. Is that better than several smaller majors with specific changes?

gergelyke: it seems like most people prefer that - doing one bigger migration a year, then a few smaller

at least, that's what the feedback we got was

gergelyke: with that said, I've added a disclaimer, that we may release major versions anytime, if we think it's needed

PR #2263, opened 2019-11-07
docs(yard): link to API

gergelyke:


gergelyke:
> The direction looks good to me, but it fails on some components; `button-group` for example

oh, I missed that - looking into that now, thanks!

PR #2264, opened 2019-11-07
docs(yard): add icons for actions

gergelyke:
With the addition of a [new button](https://github.com/uber/baseweb/pull/2263), the space really got limited. This PR changes the 3 edit buttons to icons

Screen Shot 2019-11-07 at 3 22 04 PM

### DIFF
--- a/documentation-site/components/yard/yard.tsx
+++ b/documentation-site/components/yard/yard.tsx
@@ -12,6 +12,12 @@ import copy from 'copy-to-clipboard';
 import {trackEvent} from '../../helpers/ga';
 import debounce from 'lodash/debounce';
 
+import {
+  MdContentCopy,
+  MdFormatIndentIncrease,
+  MdRotateRight,
+} from 'react-icons/md';
+
 // code sandbox stuff
 //@ts-ignore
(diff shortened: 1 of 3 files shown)


documentation-site/components/yard/yard.tsx
```diff
@@ -314,6 +320,9 @@ const Yard: React.FC<
             updateCode(dispatch, formatCode(state.code));
           }}
         >
+          <MdFormatIndentIncrease
+            style={{paddingRight: theme.sizing.scale100}}
+          />{' '}
           Format
```
nadiia: Did you mean to remove the button text? If so, could you add a tooltip instead?

gergelyke: Currently, it doesn't remove the text, just adds the icon - in L326, you can see the "Format" text. Is that what you meant? Maybe I misunderstood

nadiia: The screenshot is confusing then since 3 first buttons only contain icons and no text. But preview (and code) has text next to icons for those 3 buttons.

gergelyke: oh, I forgot to update those - sorry about that

PR #2266, opened 2019-11-07
docs(yard): make it visible that props are hoverable

gergelyke:


sandgraham:
i think i agree with using an underline like @chasestarr suggested. i can't really make out the question mark in the circle- but then again i'm not wearing my glasses.

gergelyke:
@chasestarr @sandgraham makes perfect sense, just updated the PR

Issue #2269, opened 2019-11-08
[TimePicker] Does not support null values


chasestarr:
## Feature description

Consider the situation where we have a form that has an optional time selection. Right now, the timepicker always selects the closest time to 'now' but in that case we will always have a value set.

See here for a hack around it, but less that ideal: https://codesandbox.io/s/timepicker-placeholder-hack-bfiyi Nullable values should be built into the component

PR #2275, opened 2019-11-08
fix(timepicker): enable null value

gergelyke:
Fixes https://github.com/uber/baseweb/issues/2269


src/timepicker/timepicker.js
```diff
@@ -130,7 +130,6 @@ class TimePicker extends React.Component<TimePickerPropsT, TimePickerStateT> {
       });
       this.setState({
         steps: steps,
-        value: {id: closestStep, label: secondsToLabel(closestStep)},
```
chasestarr: this would be a breaking change fyi, may need to add some configuration to opt-out of this behavior

src/timepicker/__tests__/time-picker.scenario.js
```diff
@@ -79,6 +79,10 @@ export const component = () => {
           creatable
         />
       </div>
+      <div data-e2e="12-hour">
+        Without a value
+        <TimePicker nullable />
```
chasestarr: Can you add a custom placeholder so that it can be included in the VRT?

PR #2293, opened 2019-11-11
fix(progress-steps): pass down overrides to children

gergelyke:


src/progress-steps/types.js
```diff
@@ -12,6 +12,12 @@ import type {OverrideT} from '../helpers/overrides.js';
 
 export type ProgressStepsOverridesT = {
   Root?: OverrideT<*>,
+  Icon?: OverrideT<*>,
+  InnerIcon?: OverrideT<*>,
+  Tail?: OverrideT<*>,
+  Content?: OverrideT<*>,
+  Title?: OverrideT<*>,
+  Description?: OverrideT<*>,
```
nadiia: It looks like `ProgressSteps` doesn't pass down overrides to `Step` children. Also `Root` override name for `ProgressSteps` and `Step` relate to different elements.

src/progress-steps/types.js
```diff
@@ -40,20 +36,10 @@ export type StepPropsT = {
   isActive: boolean,
   /** Defines if the step is the last item displayed. */
   isLast: boolean,
-  overrides?: StepOverridesT,
+  overrides?: ProgressStepsOverridesT,
```
nadiia: It should still be `StepOverridesT` since `ProgressStepsOverridesT ` should only have `Root` in it.

src/progress-steps/types.js
```diff
@@ -63,7 +49,7 @@ export type NumberedStepPropsT = {
   isActive: boolean,
   /** Defines if the step is the last item displayed. */
   isLast: boolean,
-  overrides?: NumberedStepOverridesT,
+  overrides?: ProgressStepsOverridesT,
```
nadiia: Same here, shouldn't change.

src/progress-steps/types.js
nadiia: Sorry, missed the overrides addition above. We shoud remap the `Step`'s `Root` override  so it's not overlapped with `ProgressSteps`' `Root` override that maps to a different element.

PR #2294, opened 2019-11-12
feat(layout-grid): add initial grid and cell components

sandgraham:
This is a v1 for a new module, `layout-grid`. Currently it exports `Unstable_Grid` and `Unstable_Cell`.

Implemented in flexbox, it essentially calculates padding and margin for various cell elements. It uses padding for gutters and subtracts the first and last gutters from the container's left and right margins. This allows everything to line up even when things wrap to a new line.

I'm curious if anything seems unintuitive to folks. One example might be that `Cell` spans one column by default. Another is that `skip` can shrink a cell to fit on one line.

I also went kind of restrictive on the API. No `overrides` and the styled components have some light "wrappers". The `Grid` component makes sense as it shares grid properties with the `Cells` passed to it - but the `Cell` component mainly serves to discourage passing custom styles to the cell.

Check the doc site preview for in depth examples of how the grid functions. ✌️


src/grid/styled-components.js
```diff
@@ -0,0 +1,220 @@
...
+  StyledCellPropsT,
+} from './types.js';
+
+// defaults
+const GRID_COLUMNS = [4, 8, 12];
+const GRID_GUTTERS = [16, 36, 36];
+const GRID_MARGINS = [16, 36, 64];
+const GRID_MAX_WIDTH = 1280;
+
+export function Grid({
```
nadiia: Could we move out the `Grid` component into a separate file since it's not a styled component?

src/grid/styled-components.js
```diff
@@ -0,0 +1,220 @@
...
+          ? `${$gridMaxWidth +
+              2 * getResponsiveNumber($gridMargins, Infinity) -
+              1}px`
+          : null,
+      ...gridStyles,
+    };
+  },
+);
+
+export function Cell({
```
nadiia: Same for the `Cell` component, please move it out of the `styled-components.js`.

src/grid/types.js
```diff
@@ -0,0 +1,61 @@
...
+export type SoftResponsiveT<T> = ResponsiveT<T | null>;
+
+export type AlignmentT =
+  | typeof ALIGNMENT.start
+  | typeof ALIGNMENT.center
+  | typeof ALIGNMENT.end;
+
+export type BehaviorT = typeof BEHAVIOR.fixed | typeof BEHAVIOR.fluid;
+
+export type GridPropsT = {
```
nadiia: Please add a description for all the Grip and Cell props.

tajo:
Very cool documentation! Looks like a great and flexible API to me.

sandgraham:
@gergelyke yeah I'd love to see it used a few times in the real world before locking in the API/implementation.

I don't anticipate it being "unstable" for very long - I based the API and implementation on a couple very popular grid systems (Bootstrap, Material) but I wouldn't be surprised if a couple novel use cases came up in the next few weeks.

This just provides a little insurance against breaking anything if something needs to be updated.

PR #2297, opened 2019-11-12
docs(select): fix virtualized example

leyanlo:
- Use `box-sizing: border-box` so list items fit in list. Also avoid `item="[object Object]" overrides="[object Object]"` from appearing in DOM.
Before:
Screen Shot 2019-11-11 at 4 43 15 PM
After:
Screen Shot 2019-11-11 at 5 00 48 PM

- Avoid mis-sized lists by resizing the height based on the number of children. E.g. try searching for `1111` and see the list height resize now.
Before:
baseweb design_components_select_

After:
localhost_3000_components_select


gergelyke:
thank you Leyan!

PR #2303, opened 2019-11-13
docs(select): use react-window for virtualized example

leyanlo:
Use react-window for virtualized example, which has a 6.2kb gzipped footprint vs a 35.1kb gzipped footprint for react-virtualized.


gergelyke:
👏 
would you mind also removing this line https://github.com/uber/baseweb/blob/33967cb22561bcf37a33d7b2cd26dda0ac4ec1ce/documentation-site/pages/components/select.mdx#L100?

nadiia:
I think we should keep both examples. If react-virtualized is already used in an app there is a good reason to stick with it for select. What do you think @leyanlo, @gergelyke?

gergelyke:
> I think we should keep both examples. If react-virtualized is already used in an app there is a good reason to stick with it for select. What do you think @leyanlo, @gergelyke?

for the record, we've discussed this offline - and we'll move the phone input to use react-window too in a future PR

leyanlo:
> 👏
> would you mind also removing this line
> 
> https://github.com/uber/baseweb/blob/33967cb22561bcf37a33d7b2cd26dda0ac4ec1ce/documentation-site/pages/components/select.mdx#L100
> 
> ?

Updated to `react-window`

PR #2314, opened 2019-11-14
feat(media-query): add mediaQuery object to theme and update docs

leyanlo:
#### Description

* Add mediaQuery object to theme based on breakpoints
* Update docs to use mediaQuery instead of media
* Follow mobile-first pattern in docs when possible
* Fix Block documentation which said there was a media query for below the first breakpoint

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/documentation-site/pages/_app.js
+++ b/documentation-site/pages/_app.js
@@ -19,7 +19,6 @@ import {
   LightTheme,
   LightThemeMove,
 } from 'baseui';
-import {getMediaQuery} from 'baseui/helpers/responsive-helpers';
 import type {BreakpointsT, ThemeT} from 'baseui/styles/types';
 
 import App from 'next/app';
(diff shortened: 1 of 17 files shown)


documentation-site/pages/blog/responsive-web/index.mdx
```diff
@@ -64,28 +64,26 @@ In many cases, you would want to override Base Web’s breakpoints with your own
 To override these, we recommend creating your own mapping, as well as generating the respective `min-width` media query strings so we can target different screen widths for styling:
 
 ```js
-const MyBreakpoints = {
+import {getMediaQuery} from 'baseui/helpers/responsive-helpers';
```
chasestarr: I removed this from the blog post recently: https://github.com/uber/baseweb/pull/2159 because we don't want to direct towards importing from files deeper than one level. Those files are not considered part of the versioned api. Can you please either leave this unchanged or export the function at a directory level? Can you share why this warrants it's own function since it mostly prepends `@media screen and`? Maybe I'm missing some of it's usefulness

leyanlo: I guess it’s not super useful. Before we standardized our breakpoints on our team, we had a few different variants and it was a more useful helper function then, and I just copied it over to Base Web. I’ll remove from the blog and maybe send another PR to remove it in general.

documentation-site/pages/_app.js
```diff
@@ -34,50 +34,33 @@ import {styletron, debug} from '../helpers/styletron';
...
 };
 
-const ResponsiveTheme = Object.keys(Breakpoints).reduce(
+const ResponsiveTheme = Object.keys(breakpoints).reduce(
   (acc, key) => {
-    acc.breakpoints[key] = Breakpoints[key];
-    acc.media[key] = getMediaQuery({
-      'min-width': `${Breakpoints[key]}px`,
-    });
+    acc.mediaQuery[key] = getMediaQuery({'min-width': `${breakpoints[key]}px`});
```
chasestarr: please revert this line as well - LGTM other than that

PR #2316, opened 2019-11-14
feat(theme): refactor theming and add semantic tokens

nadiia:
Moved the change for adding semantic tokens in a separate PR #2391


chasestarr:
There seems to have been a slight change in the green and red colors, was that intentional?

src/themes/creator.js
```diff
@@ -9,696 +9,45 @@ LICENSE file in the root directory of this source tree.
 import type {ThemeT} from '../styles/types.js';
```
jasonbarone: I don't see createTheme being used anywhere else in this diff. Is it only here for backward compatibility?

src/themes/creator.js
```diff
@@ -9,696 +9,45 @@ LICENSE file in the root directory of this source tree.
...
+import lighting from './shared/lighting.js';
+import borders from './shared/borders.js';
+import animation from './shared/animation.js';
 
 export default function createTheme(
+  // overrides for colorTokens + primatyFontFamily
   primitives: PrimitivesT,
   overrides?: {},
 ): ThemeT {
+  const {primaryFontFamily, ...rest} = primitives;
```
jasonbarone: Just calling out that it was a bit difficult to determine what's considered a **primitive** and what's considered a **token**.

src/themes/creator.js
```diff
@@ -9,696 +9,45 @@ LICENSE file in the root directory of this source tree.
...
-      timing700: '0.6s',
-      timing1000: '1s',
-      easeOutCurve: 'cubic-bezier(.2, .8, .4, 1)',
-      easeInCurve: 'cubic-bezier(.8, .2, .6, 1)',
-      easeInOutCurve: 'cubic-bezier(0.4, 0, 0.2, 1)',
+      ...colorTokensWithOverrides,
+      ...getColorComponentTokens(colorTokensWithOverrides),
+      ...getDeprecatedSemanticColors(colorTokensWithOverrides),
     },
+    typography: primaryFontFamily
```
jasonbarone: It took me a bit to figure out what was happening in some of these functions, and I've been working with the themes for a bit now. It might be difficult for users to understand how to theme Base if they reference this. I wonder if there is a way to make these more readable.

src/themes/dark-theme/dark-theme-with-move.js
```diff
@@ -0,0 +1,25 @@
...
+  fontTokens as moveFontTokens,
+} from '../move-theme/typography.js';
+import getTypography from '../shared/typography.js';
+
+import type {ThemeT} from '../../styles/types.js';
+
+export const DarkThemeMove: ThemeT = deepMerge(
+  {
+    name: 'dark-theme-with-move',
+    typography: deepMerge(getTypography(moveFontTokens), moveTypography),
```
jasonbarone: Same call out as before. This pattern with typography seems a little confusing to understand, but it might be the best options for now to reduce redundancy.

src/themes/dark-theme/dark-theme-with-move.js
```diff
@@ -0,0 +1,25 @@
...
+import {DarkTheme} from './dark-theme.js';
+import {
+  typography as moveTypography,
+  fontTokens as moveFontTokens,
+} from '../move-theme/typography.js';
+import getTypography from '../shared/typography.js';
+
+import type {ThemeT} from '../../styles/types.js';
+
+export const DarkThemeMove: ThemeT = deepMerge(
```
jasonbarone: Any reason this uses deepMerge instead of createTheme?

src/themes/dark-theme/dark-theme-with-move.js
```diff
@@ -0,0 +1,25 @@
+/*
```
jasonbarone: "dark-theme-with-move" contains an Uber-specific font name. I recommend reorganizing the themes to distinguish between the default themes and the themes specific to Uber. I also recommend renaming the theme folder/files to "uber-theme-light" and "uber-theme-dark" considering we want to eventually add the brand colors to the theme.

src/themes/dark-theme/dark-theme.js
```diff
@@ -0,0 +1,41 @@
...
+import breakpoints from '../shared/breakpoints.js';
+import mediaQuery from '../shared/media-query.js';
+import sizing from '../shared/sizing.js';
+import lighting from '../shared/lighting.js';
+import borders from '../shared/borders.js';
+import animation from '../shared/animation.js';
+
+import type {ThemeT} from '../../styles/types.js';
+
+export const DarkTheme: ThemeT = {
```
jasonbarone: I like that these do not use the theme creator and are just plain objects.

src/themes/index.js
```diff
@@ -6,25 +6,33 @@ LICENSE file in the root directory of this source tree.
...
+import lightColorTokens from './light-theme/color-tokens.js';
+import {fontTokens} from './shared/typography.js';
+import {DarkTheme} from './dark-theme/dark-theme.js';
+import {DarkThemeMove} from './dark-theme/dark-theme-with-move.js';
+import darkColorTokens from './dark-theme/color-tokens.js';
 
-export const LightTheme = createTheme(lightThemePrimitives, {
-  name: 'light-theme',
-});
+const lightThemePrimitives = {
```
jasonbarone: Similar to my comments above. I think we're using the names **primitives** and **tokens** loosely, which would cause confusion for users.

src/themes/move-theme/typography.js
```diff
@@ -0,0 +1,31 @@
+/*
```
jasonbarone: As mentioned above, I recommend creating an "uber-theme" folder, with both the light and dark themes being maintained there. This file could be moved into that folder. Although we're calling this "move theme", it's an Uber theme. We're probably OK to rename it.

src/themes/shared/animation.js
```diff
@@ -0,0 +1,20 @@
+/*
```
jasonbarone: The shared folders seem like a great update for organization.

src/themes/shared/typography.js
```diff
@@ -0,0 +1,125 @@
...
+// @flow
+import type {FontTokensT, TypographyT} from '../types.js';
+
+export const fontTokens: FontTokensT = {
+  primaryFontFamily:
+    'system-ui, "Helvetica Neue", Helvetica, Arial, sans-serif',
+};
+
+// color constants
+export default (primitives: FontTokensT = fontTokens): TypographyT => ({
```
jasonbarone: Had some trouble understand the argument here similar to above— primitives vs tokens. Maybe there's a clearer way handle the typography and overrides?

src/tokens/colors.js
```diff
@@ -10,6 +10,7 @@ import type {ColorTokensT} from './types.js';
 const colors: ColorTokensT = {
```
jasonbarone: Any reason why we're keeping the tokens folder at all? The tokens are part of the theming, so it seemed out of place that some items would be in tokens and some in themes. Also, this particular file contains the primitives and are not tokens. Considering the refactor going on in this PR, it might be worth us better organizing and naming some of these files.

src/tokens/dark-theme-color-semantic.js
```diff
@@ -0,0 +1,83 @@
+/*
```
jasonbarone: I recommend all tokens moving to a single folder instead of spreading them across tokens and themes.

src/tokens/dark-theme-color-semantic.js
```diff
@@ -0,0 +1,83 @@
...
+  borderNegative: colors.red500,
+  borderWarning: colors.yellow500,
+  borderPositive: colors.green500,
+};
+
+const semantic = {
+  ...foundation,
+  ...core,
+  ...coreExtensions,
+};
```
jasonbarone: In the Uber theme we have brand specific tokens. We'll need to determine how to extend these semantic tokens so we can pass values from the Uber theme.

src/themes/creator.js
nadiia: For backward compatibility as it's part of the current API. But there is still value in it in term of a custom theme creation, the only thing would need to be changed in the future is the structure of the primitive.

src/themes/creator.js
nadiia: I agree, although changing the structure of primitives would be a breaking change so we're keeping it as is for now.

src/themes/dark-theme/dark-theme-with-move.js
nadiia: It's due to the font family being included in primitives and used to make the typography overrides much simpler rather than passing the whole `theme.typograhy` object. I'll add some comments to clarify it.

src/themes/dark-theme/dark-theme-with-move.js
nadiia: It's identical to the dark theme with exception of typography so no reason to duplicate dark theme build on top of the light theme values that are default in `createTheme`.

src/themes/dark-theme/dark-theme-with-move.js
nadiia: I see your point, and it would be nicer to have a single theme with that includes custom font, icons, and maybe additional props, not sure if we want to have it as a part of public package tho. @gergelyke what do you think about it?

src/themes/index.js
nadiia: The only place we use `primitives` currently is theme creator `createTheme`. We could start deprecating that one and create a new version of creator with a clearer definition and usage.

nadiia:
> With current PR state, a developer has to import the new semantic tokens every time they need to use them and then check if they should use the light or dark values based on `theme.name`.
> 
> Perhaps that is for the best - our components do not use any of these tokens, so in some ways they don't belong on the `baseui` theme (or even in `baseui` for that matter).
> 
> If an application really wants to use them in a way that is "hooked" up to the theme, they could extend our theme and create a custom theme with them attached. That might be worth showing an example of.
> 
> We should probably update the tokens page on the doc site with some information and usage examples for the new `baseui/tokens` exports.
> 
> The theme page might need some updating too - the refactor/reorganization has me a little bit confused by all the concepts and layers involved (colors/tokens/primitives/overrides/semantic/deprecated/etc).

Thank you for summarizing yesterday's discussion in a comment so the rest of the reviewers are aware of the details.
I'll update the theme documentation.

nadiia:
> There seems to have been a slight change in the green and red colors, was that intentional?

Yes, those are intentional.

nadiia:
@gergelyke Moved the change for adding semantic tokens in a separate PR #2391

PR #2320, opened 2019-11-15
chore(media-query): simplify getMediaQuery to min-width case

leyanlo:
#### Description

Simplify getMediaQuery to the min-width case used by getMediaQueries and throughout the app.

### DIFF
--- a/src/helpers/responsive-helpers.js
+++ b/src/helpers/responsive-helpers.js
@@ -9,27 +9,14 @@ LICENSE file in the root directory of this source tree.
 import type {BreakpointsT} from '../styles/types.js';
 
 /**
- * Helper function that generates media queries based on given parameters
- *
- * E.g.
- * getMediaQuery({'max-width': '1280px', 'min-height': '720px'}, 'and') =>
- *   '@media screen and (max-width: 1280px) and (min-height: 720px)'
+ * Helper function that generates media queries based on breakpoint, e.g.
+ * getMediaQuery(720) => '@media screen and (min-width: 720px)'
  */
-export const getMediaQuery = (
-  options: {[string]: string},
-  booleanOperator: 'AND' | 'OR' = 'OR',
-): string => {
-  const mediaFeatureSeparator = booleanOperator === 'OR' ? ', ' : ' and ';
-
-  const mediaFeatures = Object.keys(options).map(
-    key => `(${key}: ${options[key]})`,
-  );
-
-  return `@media screen and ${mediaFeatures.join(mediaFeatureSeparator)}`;
-};
+export const getMediaQuery = (breakpoint: number): string =>
+  `@media screen and (min-width: ${breakpoint}px)`;
 
 export const getMediaQueries = (breakpoints: BreakpointsT): string[] =>
   Object.keys(breakpoints)
     .map(key => breakpoints[key])
     .sort((a, b) => a - b)
-    .map(size => getMediaQuery({'min-width': `${size}px`}));
+    .map(getMediaQuery);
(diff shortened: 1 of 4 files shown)


leyanlo:
@chasestarr what do you think about exporting the responsive helpers as part of the baseui root? I find it nice to have a dedicated method for generating the string to prevent typos and guarantee consistency. Also it seems there are a few internal repositories in sourcegraph that are using the export from `baseui/helpers/responsive-helpers` already..

leyanlo:
Ugh, maybe not a good idea to expose more APIs. Let me think about this some more.

chasestarr:
> @chasestarr what do you think about exporting the responsive helpers as part of the baseui root? I find it nice to have a dedicated method for generating the string to prevent typos and guarantee consistency. Also it seems there are a few internal repositories in sourcegraph that are using the export from `baseui/helpers/responsive-helpers` already..

This sounds fine to me. I can make a note to move applications to this new export at some point

src/helpers/responsive-helpers.js
```diff
@@ -9,27 +9,14 @@ LICENSE file in the root directory of this source tree.
...
-  const mediaFeatureSeparator = booleanOperator === 'OR' ? ', ' : ' and ';
-
-  const mediaFeatures = Object.keys(options).map(
-    key => `(${key}: ${options[key]})`,
-  );
-
-  return `@media screen and ${mediaFeatures.join(mediaFeatureSeparator)}`;
-};
+export const getMediaQuery = (breakpoint: number): string =>
+  `@media screen and (min-width: ${breakpoint}px)`;
```
nadiia: I think we should keep it a media query builder vs a single `min-width` media query getter.

src/index.js
```diff
@@ -26,6 +26,7 @@ export {
 } from './themes/index.js';
 export {default as LocaleProvider} from './locale/index.js';
 export {default as BaseProvider} from './helpers/base-provider.js';
+export {getMediaQueries, getMediaQuery} from './helpers/responsive-helpers.js';
```
nadiia: To be honest I'm on a fence regarding including it as a part of `baseui` API. 
Also, if we do think there is a value in providing `mediaQuery` maybe we should combine and ship all the helpers from `baseui/helpers` or `baseui/utils`.

src/helpers/responsive-helpers.js
leyanlo: My intention was to keep these two helpers consistent. Since getMediaQueries returns the media queries for breakpoints, which are min-width, then getMediaQuery should also return min-width for a single breakpoint.

Also, with the original implementation, it’s not guaranteed that the input is valid, unless we type the input options to valid CSS rules. And to make this truly generic, we would need to account for `all`/`print`/`screen`/`speech` media types, all media features, e.g. `any-hover`, `any-pointer`, `aspect-ratio`, etc., `not` and `only` logical operators, and any arbitrary combination of media types, features, and logical operators. Rather than try to implement the spec for media queries, I thought it would be simpler to limit the helper to supporting only breakpoints.

Let me know if this is ok!

src/index.js
leyanlo: Reverted exposing the helper to the API

src/helpers/responsive-helpers.js
nadiia: How about renaming it to `getMinWidthMediaQuery`? But as long as it's not a part of the public API it's ok to keep as is.

leyanlo: My only issue with `getMinWidthMediaQuery` would be that we would also need to update the other method to `getMinWidthMediaQueries` to match.

I feel like the fact that it takes in a `breakpoint` argument already suggests that the return would be a min-width, since that is how we deal with breakpoints in the app..

PR #2325, opened 2019-11-15
chore(component-sizes): update to use the new script

gergelyke:
The previous approached created a virtual file system, so the gzip values were just estimations. This new setup uses the real file system, so the values will be more precise.


documentation-site/components/layout.js
```diff
@@ -195,35 +193,10 @@ class Layout extends React.Component<PropsT, {sidebarOpen: boolean}> {
...
-                      </div>
-                    }
-                  >
-                    <Block font="font100">
-                      [?] Component size, gzipped: {componentSizeKb}kb
-                    </Block>
-                  </StatefulTooltip>
+                {componentSizeKb ? (
+                  <Block font="font100">
+                    [?] Component size, gzipped: {componentSizeKb}kb
```
chasestarr:
```suggestion
                    Component size, gzipped: {componentSizeKb}kb
```

scripts/component-sizes.js
```diff
@@ -8,355 +8,99 @@ LICENSE file in the root directory of this source tree.
...
+    'button-group': 'ButtonGroup',
+    button: 'Button',
+    card: 'Card',
+    checkbox: 'Checkbox',
+    'data-table': 'Unstable_StatefulDataTable',
+    datepicker: 'Datepicker',
+    'dnd-list': 'List',
+    drawer: 'Drawer',
+    'file-uploader': 'FileUploader',
+    'flex-grid': 'FlexGrid',
```
chasestarr:
```suggestion
    'flex-grid': 'FlexGrid',
    'form-control': 'FormControl',
```

chasestarr: It's a bit unfortunate that this manually created map is required. Leads to missing data like above and adds another requirement on top of the many things necessary to contribute a new component. I suppose the problem is difficult though, since you need to select an array of exports in some situations

gergelyke: Yeah, I agree, and I am not sure how to better automate it for now :/

PR #2341, opened 2019-11-19
fix(datepicker): use the locale object for the input too

gergelyke:


src/datepicker/__tests__/datepicker-int.e2e.js
```diff
@@ -13,8 +13,8 @@ const {mount} = require('../../../e2e/helpers');
 const selectors = {
   input: 'input',
   calendar: '[data-baseweb="calendar"]',
-  day: '[aria-label="Choose Sunday, March 10th 2019. It\'s available."]',
-  day2: '[aria-label="Choose Thursday, March 28th 2019. It\'s available."]',
+  day: '[aria-label="Choose vasárnap, március 10. 2019. It\'s available."]',
```
gergelyke: I'll open a github issue to track work on making this translatable too

PR #2342, opened 2019-11-19
feat(modal): add unstable prop for the backdrop scroll functionality

nadiia:


gergelyke:
can you also add an example to the docs page and to react-view?

nadiia:
> can you also add an example to the docs page and to react-view?

I wasn't sure about publicizing it as it's a temporary prop solution.

src/modal/modal.js
```diff
@@ -186,14 +197,22 @@ class Modal extends React.Component<ModalPropsT, ModalStateT> {
...
+      unstable_ModalBackdropScroll,
+    } = this.props;
     return {
       $animate: animate,
       $isVisible: this.state.isVisible,
       $isOpen: !!isOpen,
       $size: size,
       $role: role,
       $closeable: !!closeable,
+      $unstable_ModalBackdropScroll: unstable_ModalBackdropScroll,
```
chasestarr: Do you mind adding a dev warning in `componentDidMount` saying that this will be removed? Hard to find that information without reading the `types.js` file

src/modal/modal.js
```diff
@@ -56,6 +57,21 @@ class Modal extends React.Component<ModalPropsT, ModalStateT> {
 
   componentDidMount() {
     this.setState({mounted: true});
+    if (__DEV__) {
+      if (!this.props.unstable_ModalBackdropScroll) {
```
gergelyke: should we add a few words on what this property does?

nadiia: the is a description in the types and yard

PR #2357, opened 2019-11-20
fix(datepicker): mask argument accepts null

uParkA:
`Fix`

#### Description
The `mask` property within `Datepicker` should be able to take in the argument of `null`. When doing so flow gives an error, only allowing `mask` to take in strings. 

Added null type into mask property arguments.
#### Scope

- [ ] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


gergelyke:
Thank you Alex!

PR #2359, opened 2019-11-21
chore(flow): update flow-bin to version 0.112.0

chasestarr:
updates based on flow-bin changes described here: https://medium.com/flow-type/spreads-common-errors-fixes-9701012e9d58


documentation-site/examples/select/label.js
```diff
@@ -6,13 +6,13 @@ import type {ValueT} from 'baseui/select';
 
 const ColorSwatch = styled<{$color: string}>('div', props => {
   return {
+    ...props.$theme.borders.border400,
```
tajo: wonder if we should start turning everything into exact objects instead

chasestarr: 👍 updated

gergelyke:
do we want to keep this PR alive? it seems there are a ton of merge conflcits

tajo:
> do we want to keep this PR alive? it seems there are a ton of merge conflcits

I am going to close it. There are probably two better options now:

A. Keep using the same version of flow internally and make changes only to satisfy our consumers (seems much easier).
B. Explore the TypeScript path.

PR #2367, opened 2019-11-22
fix(datepicker): aria-label i18n through the localeContext (#2349)

### LINKED ISSUE #2349: Make Datepicker aria labels i18-n compliant
Currently, the labels are hardcoded: https://github.com/uber/baseweb/blob/838965328fcf286a2d014871b513bcd72e1fd1a7/src/datepicker/day.js#L207-L227

We should expose them through the LocaleContext

jafaryfarzaneh:
Fix #2349‍‍
aria-label i18n through the localeContext

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


src/datepicker/locale.js
```diff
@@ -45,6 +51,12 @@ const locale = {
   timeSelectStartLabel: 'Start time',
   timePickerAriaLabel: 'Select a time',
   timezonePickerAriaLabel: 'Select a timezone',
+  selectedStartDate: 'Selected start date.',
+  selectedEndDate: 'Selected end date.',
+  notAvailable: 'Not available.',
```
chasestarr: do you mind making the property names more specific? For example, here I would call it `dateNotAvailable`

PR #2372, opened 2019-11-22
fix(phone-input): make the countries searchable

gergelyke:


src/phone-input/__tests__/phone-input.e2e.js
```diff
@@ -34,7 +34,14 @@ describe('PhoneInput', () => {
   });
 
   it('passes basic a11y tests', async () => {
-    const accessibilityReport = await analyzeAccessibility(page);
+    const accessibilityReport = await analyzeAccessibility(page, {
```
gergelyke: I am not sure why it started failing - does anyone have any ideas? 🤔

chasestarr: can you please share the axe error or remove the rule option so I can view it in CI?

src/phone-input/__tests__/phone-input.e2e.js
```diff
@@ -63,6 +70,23 @@ describe('PhoneInput', () => {
     });
   });
 
+  it('allows a user select a country using the keyboard', async () => {
+    // click select
+    await page.click(selectors.phoneInputSelect);
+    // verify dropdown is open
+    await page.waitFor(selectors.phoneInputSelectDropdown);
+
+    await page.keyboard.type('United Kingdom');
```
chasestarr: can you please add a test showing that if I type in `United` and press arrow down, I reach `United Arab Emirates`?

src/phone-input/__tests__/phone-input.e2e.js
gergelyke: it was complaining because of this: https://dequeuniversity.com/rules/axe/3.0/label

chasestarr: I believe you can add an aria-label to the `Input` through overrides

houmark:
This seems to be a boiled down / minimalistic version of the earlier proposed, but never merged PR. This one does not show the text being searched, is that intended? I personally think the other one was much nicer and if someone wants the minimalistic edition like this one, then they could either override or maybe it could be configurable to show or not show a search text field.

I absolutely do not want to slow the merging of this one down as any search improvement is an improvement over the current state of the phone input component.

PR #2373, opened 2019-11-22
fix(data-table): removes duplicate sample cells from measurement

chasestarr:
#### Description

To measure cell widths, if the row count is large, the data-table gets a random subset of cells in each column. Prior to this change it had the possibility to choosing duplicate cells to sample. Now, it keeps track of 'seen' indices and avoids choosing duplicates.

#### Scope

- [x] Patch: Bug Fix


src/data-table/measure-column-widths.js
```diff
@@ -90,8 +90,23 @@ export default function MeasureColumnWidths(props: MeasureColumnWidthsPropsT) {
...
+      const seen = {};
       for (let i = 0; i < sampleSize; i++) {
-        indices.push(Math.floor(Math.random() * props.rows.length));
+        // this seems to handle large sets fine, but there's a pathological case where
+        // the total number of rows is only slightly larger than the MAX_SAMPLE_SIZE (50).
+        // Consider the situation where the total row count is 51. This while loop just
+        // repeats until it has randomly chosen 50 unique integers. In this hypothetical,
+        // the last index chosen has a 1:51 chance of proceeding and the prior choice has 2:51.
+        // This is a lot of wasted cycles. I imagine there is a better way to handle this...
+        // but nothing jumps out to me that can be generalized between small and large sets.
```
chasestarr: it could special case the `MAX_SAMPLE_SIZE + 1` through `MAX_SAMPLE_SIZE + 100` and generate a deterministic distribution of indices

PR #2374, opened 2019-11-22
fix(button): remove ModalButton margin outside button group

leyanlo:
#### Description

Remove ModalButton margin outside button group so margin does not extend past the buttons. This was especially wrong in RTL languages.

Before:
modal-button-before

After:
modal-button-after

RTL before:
modal-rtl__mobile-snap

RTL after:
modal-rtl__mobile-snap (1)

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


gergelyke:
@leyanlo can you help us look into the unit tests?

leyanlo:
> @leyanlo can you help us look into the unit tests?

Oops, sorry about that. Updated!

gergelyke:
thank you

src/modal/modal-button.js
```diff
@@ -15,7 +15,9 @@ import {mergeOverrides} from '../helpers/overrides.js';
 const overrides = {
   BaseButton: {
     style: ({$theme}) => ({
-      marginLeft: $theme.sizing.scale500,
+      ':nth-last-child(n+2)': {
+        marginRight: $theme.sizing.scale500,
```
gergelyke: @leyanlo ops, sorry for the delay in noticing this, but I think we should update it to handle RTL - in dir=rtl, it should be marginLeft, if I am correct

probably something like this would do the trick https://github.com/uber/baseweb/blob/cb9adaa3cad6b344c18ac05fc57b5f10e7ba2464/src/payment-card/styled-components.js#L20

gergelyke: you can repro the issue on the documentation site, by switch it to RTL using the icon in the top right corner

leyanlo: Ok will update! How come the rtl snapshots didn’t catch this..?

gergelyke: yeah, that's an excellent question - I'll look into that

PR #2378, opened 2019-11-25
fix(tabs): space key a11y on Tabs and Accordion component.

lucasterra:
#### Description

[Tabs](https://baseweb.design/components/tabs/) and [Accordion](https://baseweb.design/components/accordion/) components makes the scroll jump when navigating using Space key. This is really annoying for accessibility.

The fix is really easy, just need to `e.preventDefault()` the event when it's triggered via space key.

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


gergelyke:
Thank you @lucasterra !

PR #2397, opened 2019-11-28
feat(select): separate overrides for SearchIcon and SearchIconContainer

leyanlo:
#### Description

Separate overrides for SearchIcon and SearchIconContainer. Before, SearchIcon overrides would set the props on the container, but set the component as the SearchIcon. Now we separate into two overrides SearchIconContainer and SearchIcon, so users can target overrides for one or the other.

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change


gergelyke:
@leyanlo for now, can we do this in a non-breaking manner?

leyanlo:
> @leyanlo for now, can we do this in a non-breaking manner?

I am open to suggestions. The issue lies in the original implementation:
```js
const [SearchIcon, searchIconProps] = getOverrides(
  overrides.SearchIcon,
  SearchIconComponent,
);
return (
  <StyledSearchIcon {...sharedProps} {...searchIconProps}>
    <SearchIcon size={16} title={'search'} />
  </StyledSearchIcon>
);
```
This used the SearchIcon overrides for the child component and the parent props. My solution separated these two concerns:
```js
const [SearchIconContainer, searchIconContainerProps] = getOverrides(
  overrides.SearchIconContainer,
  StyledSearchIconContainer,
);
const [SearchIcon, searchIconProps] = getOverrides(
  overrides.SearchIcon,
  SearchIconComponent,
);
return (
  <SearchIconContainer {...sharedProps} {...searchIconContainerProps}>
    <SearchIcon size={16} title={'search'} {...searchIconProps} />
  </SearchIconContainer>
);
```
If the goal would be to be able to target the child props and the parent component, how should we do it in a non-breaking manner?

gergelyke:
@leyanlo what if we would do something like this

```js

const [SearchIcon, searchIconProps] = getOverrides(
  overrides.SearchIcon,
  SearchIconComponent,
);

const [SearchIconContainer, searchIconContainerProps] = getOverrides(
  overrides.SearchIconContainer,
  StyledSearchIconContainer,
);

// TODO(v10): remove searchIconProps
return (
  <SearchIconContainer {...sharedProps} {...searchIconProps} {...searchIconContainerProps}>
    <SearchIcon size={16} title={'search'} {...searchIconProps} />
  </SearchIconContainer>
);
```

this way, we would keep the current functionality, while enabling the overrides - and we could remove this hack in the next major

leyanlo:
> @leyanlo what if we would do something like this
> 
> ```js
> const [SearchIcon, searchIconProps] = getOverrides(
>   overrides.SearchIcon,
>   SearchIconComponent,
> );
> 
> const [SearchIconContainer, searchIconContainerProps] = getOverrides(
>   overrides.SearchIconContainer,
>   StyledSearchIconContainer,
> );
> 
> const sharedProps = this.getSharedProps();
> // TODO(v10): remove searchIconProps
> return (
>   <SearchIconContainer {...sharedProps} {...searchIconProps} {...searchIconContainerProps}>
>     <SearchIcon size={16} title={'search'} {...searchIconProps} />
>   </SearchIconContainer>
> );
> ```
> 
> this way, we would keep the current functionality, while enabling the overrides - and we could remove this hack in the next major

Great idea! Updated :-)

src/select/select-component.js
```diff
@@ -740,16 +740,25 @@ class Select extends React.Component<
...
+    );
     const [SearchIcon, searchIconProps] = getOverrides(
       overrides.SearchIcon,
       SearchIconComponent,
     );
     const sharedProps = this.getSharedProps();
 
     return (
-      <StyledSearchIcon {...sharedProps} {...searchIconProps}>
-        <SearchIcon size={16} title={'search'} />
```
nadiia: It looks like an error in the code.

src/select/index.js
```diff
@@ -29,7 +29,7 @@ export {
   StyledIconsContainer,
   StyledSelectArrow,
   StyledClearIcon,
-  StyledSearchIcon,
+  StyledSearchIconContainer,
```
nadiia: We should keep the `StyledSearchIcon` export that equals `StyledSearchIconContainer`, to keep it as non-breaking.

nadiia:
Minor comment about `StyledSearchIcon`, but good to be merged after the fix 👍

src/select/index.js
leyanlo: Good catch. Updated!

gergelyke:
thank you Leyan!

PR #2409, opened 2019-12-02
fix(phone-input): account for filtering in country-select-dropdown

### LINKED ISSUE #2401: Phone Input component is breaking the app when filtering country flags and search dont find results
## Current Behavior
If you set up custom flags on phone input component and try to filter on countries and the search don't find any result on country list, throws an error and break the app.

## Expected Behavior
You should be able to filter countries (with custom flags) and don't get an error.

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v9.23.1  |
| React   |  16.12.0  |
| browser | Chrome, Firefox        |

sandgraham:
Fixes #2401 

The `country-select-dropdown` component required some guards for when no `item` is passed through, as with a list filtered down to nothing.


gergelyke:
should we implement the same behavior as how the default component works, with showing the no results message?

gergelyke:
> should we implement the same behavior as how the default component works, with showing the no results message?

probably we could do that in the follow up PR

PR #2413, opened 2019-12-03
fix(package): update module field to relative path

jcemer:
Parcel looks to ignore the absolute path.

#### Description

The `package.json` included in subfolder has a relative path on `module` field. Parcel ignores the absolute path on the main `package.json` for `import { LightTheme, DarkTheme, BaseProvider } from "baseui"`. It leads the bundle to include two different ThemeProvider files breaking theming for example.

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
   "main": "index.js",
   "types": "index.d.ts",
   "sideEffects": false,
-  "module": "/esm/index.js",
+  "module": "./esm/index.js",
   "repository": "uber/baseweb",
   "scripts": {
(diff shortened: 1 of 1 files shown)


gergelyke:
so we have to also update the package json generator?

tajo:
Tested with https://github.com/tajo/fusion-baseui (both dev/prod builds) and seems to work fine.

jcemer:
> so we have to also update the package json generator?

The package json generator already includes a relative path.

PR #2419, opened 2019-12-04
feat(typography): rename components to match Base typography style names

jasonbarone:
#### Description
This PR updates the typography component names to match the Base design system typography names, making it easier for engineers to find the typography component that matches the text styles found in Figma.

The existing names are moved to aliases to keep backward compatibility.

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/src/typography/index.js
+++ b/src/typography/index.js
@@ -254,6 +254,7 @@ export const ParagraphXSmall = React.forwardRef<BlockPropsT, HTMLElement>(
 );
 ParagraphXSmall.displayName = 'ParagraphXSmall';
 
+// TODO(v10): remove
 // Aliases for backwards compatability
 export const Display = DisplayLarge;
 export const Display1 = DisplayLarge;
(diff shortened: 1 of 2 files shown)


src/typography/index.js
```diff
@@ -240,10 +252,29 @@ export const Paragraph4 = React.forwardRef<BlockPropsT, HTMLElement>(
     />
   ),
 );
-Paragraph4.displayName = 'Paragraph4';
+ParagraphXSmall.displayName = 'ParagraphXSmall';
 
 // Aliases for backwards compatability
-export const Display = Display1;
+export const Display = DisplayLarge;
```
gergelyke: can you add a comment like `TODO(v10): remove`, so we won't miss it when releasing the next major?

jasonbarone: Got it, thx!

gergelyke:
Thank you Jason

PR #2422, opened 2019-12-04
feat(typography): add Base design system typography names to theme

jasonbarone:
#### Description

This PR adds the Base design system typography names to the theme, allowing users to work with the typography names they will see in design files instead of the existing font primitive names (font150, font1150, etc).

#### Example Usage

```
<Block font="DisplayMedium">Lorem ipsum</Block>
```

```
const Container = styled('div', ({$theme}) => ({
  height: $theme.typography.DisplayMedium.fontSize,
}));
```

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change


gergelyke:
should we first land @nadiia 's theme changes?

jasonbarone:
> should we first land @nadiia 's theme changes?

We can if you want, although @nadiia split the token update into its own PR:https://github.com/uber/baseweb/pull/2391. I assumed the theme update would take a bit more time to work out. @nadiia what do you think?

nadiia:
Wasn't it the goal of updating `typography` components to reference the design system names? Why add it in addition to the current scale in the theme object?

src/themes/shared/typography.js
```diff
@@ -122,4 +122,112 @@ export default (themePrimitives: FontTokensT = fontTokens): TypographyT => ({
...
+    fontWeight: 500,
+    lineHeight: '52px',
+  },
+  DisplayMedium: {
+    fontFamily: themePrimitives.primaryFontFamily,
+    fontSize: '52px',
+    fontWeight: 500,
+    lineHeight: '64px',
+  },
+  DisplayLarge: {
```
gergelyke: instead of duplicating the values, could we simply

```
DisplayLarge: {
  ...this.font1450,
},
```

gergelyke:
Merging as is, and will send a follow up to make these changes

PR #2424, opened 2019-12-05
fix(datepicker): clicking scroll in month selection closes datepicker

norenma:
#### Description

When clicking the popover which contains `calendar-header` in the datepicker, `onClickOutside` in the popover that contains the `calendar` is called. This is because the calendar-header is not rendered _in_ the parent popover, but in a separate node. The closing of the parent popover is prevented with a hack, that checks if the `role` of the clicked element is "option". However, if the scroll bar is clicked, the `menu`-element is the target, and the component closes(even though the user wanted to scroll). 

This PR renders calender-header popover as a child of the popover that renders the calendar, so that clicking calendar-header is not considered "outside" of the calendar. 

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/src/datepicker/calendar.js
+++ b/src/datepicker/calendar.js
@@ -81,7 +81,6 @@ export default class Calendar extends React.Component<
     trapTabbing: false,
   };
 
-  root: React.ElementRef<*>;
   calendar: React.ElementRef<*>;
 
   constructor(props: CalendarPropsT) {
@@ -94,6 +93,7 @@ export default class Calendar extends React.Component<
       focused: false,
       date: this.getDateInView(),
       quickSelectId: null,
+      rootElement: null,
     };
   }
 
@@ -183,6 +183,7 @@ export default class Calendar extends React.Component<
         order={order}
         onMonthChange={this.changeMonth}
         onYearChange={this.changeYear}
+        popoverMountNode={this.state.rootElement}
       />
     );
   };
@@ -287,8 +288,8 @@ export default class Calendar extends React.Component<
         const activeElm = document.activeElement;
         // need to look for any tabindex >= 0 and ideally for not disabled
         // focusable by default elements like input, button, etc.
-        const focusable = this.root
-          ? this.root.querySelectorAll('[tabindex="0"]')
+        const focusable = this.state.rootElement
+          ? this.state.rootElement.querySelectorAll('[tabindex="0"]')
           : null;
         const length = focusable ? focusable.length : 0;
         if (event.shiftKey) {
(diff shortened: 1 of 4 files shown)


gergelyke:
@norenma - can you please provide a gif of what's the current / fixed behaviour? I am not sure if I could reproduce

src/datepicker/calendar.js
```diff
@@ -552,7 +552,9 @@ export default class Calendar extends React.Component<
           <Root
             data-baseweb="calendar"
             ref={root => {
-              this.root = root;
+              if (!root && this.state.rootElement) {
```
nadiia: Why `!root`?

src/datepicker/calendar.js
```diff
@@ -287,7 +287,7 @@ export default class Calendar extends React.Component<
         const activeElm = document.activeElement;
         // need to look for any tabindex >= 0 and ideally for not disabled
         // focusable by default elements like input, button, etc.
-        const focusable = this.root
+        const focusable = this.state.rootElement
```
nadiia: Does it need to be in the state? Also, the state type and initial value would need to be updated.

nadiia:
In the preview it doesn't work and the calendar gets closed on a click on a month select option.

chasestarr:
before (clicking the scrollbar closes the calendar): 
ezgif-4-593c33a186cf

preview: https://baseweb-git-fork-norenma-fix-calender-header-scroll-click.uber-ui-platform.now.sh/components/datepicker

On latest chrome, it looks like this change does not quite fix the issue, and also regressed on the case the original code meant to solve. I can't click on another month, nor use the scroll bar

norenma:
Sorry I broke this just before committing, should be fixed now.

src/datepicker/calendar.js
norenma: Yea sorry, this was incorrect and it was what caused the bug. This check is needed because this function is called null every render until the element is rendered to the dom. 
Since I keep the element in state I also need to check if it has been set before, otherwise it will cause a recursive loop.

src/datepicker/calendar.js
norenma: I think it needs to be in state. If this was a ref, I would send the same ref to CalendarHeader. Initially the ref would be empty, and CalendarHeader would not render on the correct node. When the ref is set, it would not trigger a re-render and CalendarHeader would still be rendered on the wrong node. 

BUT, when I try to make this a ref, it still works. I think it works because other changes trigger re-renders after the ref has gotten the correct element, but I could be wrong. If that is the case, it seems like a fragile solution IMO. What do you think?

PR #2429, opened 2019-12-06
fix(select): use quotation marks for creatable options

leyanlo:
#### Description

Use quotation marks instead of straight quotes for creatable options.
See https://practicaltypography.com/straight-and-curly-quotes.html

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


gergelyke:
thank you Leyan - can you help us update end to end tests too?

PR #2431, opened 2019-12-06
fix(select): merge StatefulMenu overrides

leyanlo:
#### Description

Merge StatefulMenu overrides instead of replacing. This fixes the issue of getOptionLabel not working when you try overriding the StatefulMenu’s EmptyState.

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


gergelyke:
@leyanlo could you add a regression test for this?

PR #2432, opened 2019-12-06
fix(modal): fix autofocus deprecation

gergelyke:


src/modal/modal.js
```diff
@@ -33,8 +33,8 @@ import type {
 class Modal extends React.Component<ModalPropsT, ModalStateT> {
   static defaultProps: $Shape<ModalPropsT> = {
     animate: true,
-    // remove in a future major version
-    autofocus: false,
+    // TODO(v10): remove
+    autofocus: null,
```
nadiia: 👍 Could you also add a warning in componentDidmount if it's passed through props?

gergelyke: sure - it was already there, I've just updated it to all trigger if it's false

Issue #2436, opened 2019-12-09
Datepicker with positive prop


Mater:
## Current Behavior
`Datepicker` with `positive` prop, error in console.
> Warning: Received `true` for a non-boolean attribute `positive`.

Same problem with `MaskedInput` (`Datepicker` uses it)

## Expected Behavior
No errors

## Your Environment

https://codesandbox.io/s/datepicker-ciotw

| Tech    | Version |
| ------- | ------- |
| Base UI | v9.27.0  |
| React   | 16.8.6  |
| browser |  Google Chrome Version 78.0.3904.108 (Official Build) (64-bit) |

- [x] I have searched the [issues](https://github.com/uber/baseweb/issues) of this repository and believe that this is not a duplicate.

PR #2438, opened 2019-12-09
fix(masked-input): do not pass the prop positive

gergelyke:
Fixes https://github.com/uber/baseweb/issues/2436

### DIFF
--- a/src/input/masked-input.js
+++ b/src/input/masked-input.js
@@ -16,9 +16,13 @@ import type {MaskedInputPropsT} from './types.js';
 const MaskOverride = React.forwardRef<MaskedInputPropsT, HTMLElement>(
   (
     {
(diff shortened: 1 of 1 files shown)


src/input/masked-input.js
```diff
@@ -24,6 +24,9 @@ const MaskOverride = React.forwardRef<MaskedInputPropsT, HTMLElement>(
       onBlur,
       value,
       disabled,
+      // do nothing with it - we just don't want to pass it to the InputMask, as
+      // it does not have a property called positive
+      positive,
```
chasestarr: do you mind co-locating this with the other 'ignored' props slightly above?

gergelyke: great catch, thank you @chasestarr - just fixed it

PR #2445, opened 2019-12-09
fix(modal): properly render modal rtl vrt

chasestarr:
#### Description

Screen Shot 2019-12-09 at 12 51 12 PM

#### Scope

- [x] Patch: Bug Fix


sandgraham:
I think this should have included the updated snapshot as well?

PR #2454, opened 2019-12-11
feat(theme): refactor themes

nadiia:
#### Description

The change is broken down in 3 separate commits:
- Part#1 refactors theme
- [Part#2](https://github.com/uber/baseweb/pull/2474) add semantic color tokens
- [Part#3](https://github.com/uber/baseweb/pull/2473) applies newly added semantic color tokens in place of older to be deprecated semantic tokens 

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change


src/card/styled-components.js
```diff
@@ -50,7 +50,7 @@ export const Root = styled<{}>('div', ({$theme}) => {
     borderTopRightRadius: borders.surfaceBorderRadius,
     borderBottomLeftRadius: borders.surfaceBorderRadius,
     borderBottomRightRadius: borders.surfaceBorderRadius,
-    backgroundColor: colors.backgroundAlt,
+    backgroundColor: colors.backgroundSecondary,
```
gergelyke: shouldn't it stay with in the light theme?

gergelyke:
looks great!

chasestarr:
many of the 'surface' backgrounds have changed from white to light grey in the vrts, was that intentional? Otherwise, the change lgtm

src/themes/dark-theme/color-semantic-tokens.js
```diff
@@ -0,0 +1,88 @@
...
+  CoreExtensionSemanticColorTokensT,
+  SemanticColorTokensT,
+} from '../types.js';
+import {hexToRgb as hexToRgba} from '../../styles/util.js';
+import colors from '../../tokens/colors.js';
+
+const foundation: FoundationSemanticColorTokensT = {
+  primaryA: colors.gray200,
+  primaryB: colors.gray900,
+  accent: colors.blue500,
```
sandgraham: if people have these already set as primitives in their current theme will they be overwritten by these keys with the same name?

nadiia: You mean if they create a custom theme and extend it with custom props? Then it depends on how they extend the current theme, should their custom props come last then it overrides what's in the default theme.

sandgraham: What would `accent` end up as on the final theme object if I did the following?

```js
createTheme({
  // other primitives...
  accent: 'pink',
});
```

In `createTheme` we combine a bunch of colors:

```js
const theme = {
    colors: {
      ...colorTokensWithOverrides, // user defined primitives
      ...getColorComponentTokens(colorTokensWithOverrides),
      ...getDeprecatedSemanticColors(colorTokensWithOverrides),
      ...semanticColors, // has `accent`, `positive`, etc.
    },
};
```

`semanticColors` has keys for `accent`, `positive`, etc. so wouldn't they override the primitives on the final theme object?

nadiia: I see. The `accent`, etc, will be overriden but not the other props that have value set from the theme primitives. I guess foundation semantic colors can be included in the theme building primitives and the sert of semantic values also built it taking overrides into account. What do you think?

sandgraham: yeah, maybe add `primaryA` and `primaryB` to `ColorTokensT`.

then we can use a getter like `getSemanticColors(colorTokensWithOverrides)` which will override the `foundation` values with the passed in primitives.

nadiia: Addressed it in [here](https://github.com/uber/baseweb/pull/2474/commits/4ebed3e526f0c349f7c64e8ebc178d335a6e159c)

nadiia:
> many of the 'surface' backgrounds have changed from white to light grey in the vrts, was that intentional? Otherwise, the change lgtm

You're right, I think the legacy `backgroundAlt` was the same as primary but in the new semantic values they are different. I changed it to use new `backgroundPrimary` in [here](https://github.com/uber/baseweb/pull/2473). See the rendered docs [here](https://baseweb-git-wpt-5198-3.uber-ui-platform.now.sh).

PR #2457, opened 2019-12-11
fix(various): do not pass item as prop to styled component

leyanlo:
#### Description

Do not pass item as prop to styled components to avoid appearing in the DOM.

Select before: 
Screen Shot 2019-12-11 at 3 09 37 AM
Select after: 
Screen Shot 2019-12-11 at 3 17 22 AM

Phone Input before: 
phone-input-before
Phone Input after: 
phone-input-after

Side Navigation before: 
side-navigation-before
Side Navigation after: 
side-navigation-after

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


chasestarr:
thanks for improving this! can you resolve the regression shown in the vrt?

leyanlo:
@chasestarr I added a `$` prefix to item, which should resolve the vrt regression. Not sure if this makes the most sense, since `$` prefixes are designed for sending props to styled components, but we’re using it instead to send to an override in your menu adapter example..

chasestarr:
we'll need to hold off on merging to the next major because of the prop name change

tajo:
> we'll need to hold off on merging to the next major because of the prop name change

Hm, about doing a bit workaround in meantime and wrapping `StyledItemList` with another React component just to filter our the `item` prop?

chasestarr:
that sounds fine to me

leyanlo:
Update with workaround

leyanlo:
Is this snapshot correct? Not sure if it is supposed to look like this but updated anyway..

PR #2460, opened 2019-12-11
feat(select): enables grouped options in dropdown menu

chasestarr:
#### Description

Screen Shot 2019-12-11 at 1 03 05 PM

Screen Shot 2019-12-11 at 1 03 15 PM

I personally like the updated `OptionT` type with a new `optgroup` property to define grouping, since it requires almost no existing select code refactoring, but I understand that applications may have their data in a {[string]: OptionT[]} shape when fetching from a backend. For now, I've implemented as described above and will explore how much code needs to change to enable an object provided to the `options` prop on the component. The new tests will provide security that any changes will work appropriately.

https://github.com/uber/baseweb/blob/8f485c330332d2c0465926d94642d02a6a6de9ac/src/select/__tests__/select-option-group.scenario.js#L15-L23

#### Scope

- [x] Minor: New Feature


tajo:
> I personally like the updated OptionT type with a new optgroup property to define grouping, since it requires almost no existing select code refactoring

I would not mind this that much but you already used the other API (`{[string]: OptionT[]}`) for Menu? So might keep the same?

> but I understand that applications may have their data in a {[string]: OptionT[]} shape when fetching from a backend.

This. Most APIs probably always avoid duplications such as `optgroup`.  

> I've implemented as described above and will explore how much code needs to change to enable an object provided to the options prop on the component

This would be a breaking change (or we would have to keep supporting the optgroup and the implementation would get more confusing?)

PR #2462, opened 2019-12-11
feat(select): accept option group as object

chasestarr:
#### Description

[relevant commit](https://github.com/uber/baseweb/commit/2a4f52ac73786c34f8f000a60251725041538d39) from transitioning to `optgroup` property to accepting `options` as an object. I find the conversions needed to be quite messy, but maybe it would be a better developer experience?

#### Scope

- [ ] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


src/select/types.js
```diff
@@ -95,7 +98,7 @@ export type PropsT = {
   /** Defaults to filterOptions that excludes selected options for
    * multi select. A custom method to filter options to be displayed in the dropdown. */
   filterOptions: ?(
-    options: ValueT,
+    options: OptionsT,
```
chasestarr: is this a breaking change now that flow will complain that consumers are not properly handling the case where `OptionsT` could be an object shape?

tajo: probably not since we are making the type less strict?

tajo:
Does this PR support both APIs (object vs `optgroup` param)? If so, we should do just one?

src/select/types.js
chasestarr: right, but if a developer does not handle the non-array case flow will complain

chasestarr:
> Does this PR support both APIs (object vs `optgroup` param)? If so, we should do just one?

it's half-baked at the moment. the idea would be to only include one pattern

PR #2463, opened 2019-12-11
fix(select): apply input style overrides to autosizer

sandgraham:
#2434 

Apply override styles for `Input` to the `StyledInputSizer` which determines the width of the `Input`. If they are out of sync then text can get cut-off.


src/select/autosize-input.js
```diff
@@ -76,7 +76,12 @@ export default class AutosizeInput extends React.Component<
       <React.Fragment>
         <Input {...componentInputProps} ref={inputRef} {...inputProps} />
         {/* a hidden helper element to calculate the size of the input */}
-        <StyledInputSizer $size={this.props.$size} ref={this.sizerRef}>
+        <StyledInputSizer
+          $size={this.props.$size}
+          ref={this.sizerRef}
+          // $FlowFixMe checking for $style before use
+          $style={inputProps.$style ? inputProps.$style : null}
```
chasestarr: does undefined rather than null pass flow check?

sandgraham: `inputProps`, which is returned by a `getOverrides` call, is typed as `{}` so `inputProps.$style` isn't allowed. not sure if we can alter the return type for props on `getOverrides`?

chasestarr: i think it's safe to change that return type to {$style?: any} but may be better to save for another pr

PR #2467, opened 2019-12-11
fix(dnd-list): handle interactive elements better, role checkbox

tajo:
The fix itself is in [react-movable](https://github.com/tajo/react-movable/pull/32).

Added the `role="checkbox"` to our checkbox so it is recognized as an interactive elements.


chasestarr:
> must also include the aria-checked attribute to expose the checkbox's state to assistive technology

fyi https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/checkbox_role

Issue #2435, opened 2019-12-07
[Modal] Closing nested modals with ESC leaves body scroll locked


kirbycool:
If you open nested modals and close them both by pressing ESC, the body scroll lock doesn't get reset properly.
This is reproducible in the nested modal example from the docs (I tried in both Chrome and FF).
1. Go to https://baseweb.design/components/modal/#adding-nested-modals
2. Open the modal and the nested modal
3. Press ESC to close the modals
4. The body isn't scrollable

I dug in a bit, and the reason this is happening is because the bottom modal gets unmounted before the top modal. Each modal caches its mount node's `overflow` style attribute and replaces it on close. The top modal caches `overflow: hidden`, which is set by the bottom modal, but since it closes last, it resets to `overflow: hidden` after the bottom modal resets to `overflow = ''`.
https://github.com/uber/baseweb/blob/a82b504c9408f0659755075029f32ab65e61ae90/src/modal/modal.js#L119

In terms of solutions, ideally this would work with external components that modify the body overflow (I'm using a separate Drawer component that has this same bug when including a nested modal).
One option would be to not reset the reset the mountNode's `overflow` if it's been changed from `hidden` by some other component.
```js
  resetMountNodeScroll() {
    const mountNode = this.getMountNode();
    const lastStyle = this.lastMountNodeOverflowStyle;
    // If mountNode.style.overflow !== 'hidden', it's modified externally and we shouldn't reset it
    if (mountNode && lastStyle !== null && mountNode.style.overflow === 'hidden') {
      mountNode.style.overflow = lastStyle || '';
      this.lastMountNodeOverflowStyle = null;
    }
  }
```

## Current Behavior

Closing nested modals with ESC leaves the mount node (usually <body>) scroll locked.

## Expected Behavior

Closing nested modals does not cause scroll lock

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v9.27.0  |
| React   |         |
| browser |         |

- [ ] I have searched the [issues](https://github.com/uber/baseweb/issues) of this repository and believe that this is not a duplicate.

gergelyke:
@kirbycool thank you for the bug report! any chance you could send a PR fixing it?

kirbycool:
Sure! I'll take a stab at it when I get some free time.

gergelyke:
thank you! 🤗

PR #2472, opened 2019-12-12
fix(modal): handle nested modal overflow reset

kirbycool:
Resolves https://github.com/uber/baseweb/issues/2435

### Problem
When closing nested modals simultaneously (e.g. press ESC), the bottom one can unmount before the top one. This causes the top modal to reset the mountNode's overflow to 'hidden', which locks scrolling.
The flow goes like this:
1. Bottom modal mounts, caches last overflow as '', and sets overflow to 'hidden'
2. Top modal mounts, caches last overflow as 'hidden', and sets overflow to 'hidden'
3. Bottom modal unmounts and resets overflow to ''
4. Top modal unmounts and resets overflow to 'hidden'

### Solution
This change bails out of resetting overflow if it's changed since the modal mounted.
So now the flow would be:
1. Bottom modal mounts, caches last overflow as '', and sets overflow to 'hidden'
2. Top modal mounts, caches last overflow as 'hidden', and sets overflow to 'hidden'
3. Bottom modal unmounts and resets overflow to ''
4. Top modal unmounts and bails out because overflow is no longer
'hidden'

I'm not 100% sure this is the best approach, so let me know if you have a better idea.

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/src/modal/__tests__/modal.test.js
+++ b/src/modal/__tests__/modal.test.js
@@ -149,6 +149,53 @@ describe('Modal', () => {
     expect(body.style.overflow).toBe('');
   });
 
+  describe('nested modals', () => {
+    const TwoModals = ({isOpen1 = false, isOpen2 = false}) => (
+      <>
+        <Modal isOpen={isOpen1}>Modal 1</Modal>
+        <Modal isOpen={isOpen2}>Modal 2</Modal>,
+      </>
+    );
+
+    test('resets body scroll when top closes first', () => {
+      wrapper = mount(<TwoModals />);
+
+      const body = ((document.body: any): HTMLBodyElement);
+      expect(body.style.overflow).toBe('');
+
+      wrapper.setProps({isOpen1: true, isOpen2: false});
+      expect(body.style.overflow).toBe('hidden');
+
+      wrapper.setProps({isOpen1: true, isOpen2: true});
+      expect(body.style.overflow).toBe('hidden');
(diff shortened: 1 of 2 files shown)


src/modal/__tests__/modal.test.js
```diff
@@ -149,6 +149,32 @@ describe('Modal', () => {
...
+    const body = ((document.body: any): HTMLBodyElement);
+    expect(body.style.overflow).toBe('');
+
+    wrapper.setProps({isOpen1: true, isOpen2: false});
+    expect(body.style.overflow).toBe('hidden');
+
+    wrapper.setProps({isOpen1: true, isOpen2: true});
+    expect(body.style.overflow).toBe('hidden');
+
+    wrapper.setProps({isOpen1: false, isOpen2: true});
```
chasestarr: Can you write a test that verifies closing the modals in the opposite order? It looks like that case is handled fine in the site preview, but I'd like to ensure we don't cause a regression.

kirbycool: Sure thing. 
I don't think this will happen for the ESC case, since event listeners (keyup) get run in the order they're added, so the bottom modal should always unmount first.

kirbycool: Added an additional test

PR #2476, opened 2019-12-13
feat(select): enables option groups in dropdown menu

chasestarr:
#### Description

Screen Shot 2019-12-11 at 1 03 05 PM

Screen Shot 2019-12-11 at 1 03 15 PM

see https://github.com/uber/baseweb/pull/2460 and https://github.com/uber/baseweb/pull/2462 for some alternative takes on this feature

#### Scope

- [x] Minor: New Feature


src/select/dropdown.js
```diff
@@ -16,171 +16,166 @@ import {StatefulMenu} from '../menu/index.js';
...
-              List: {
-                component: StyledDropdown,
-                style: p => ({
-                  maxHeight: p.$maxHeight || null,
-                }),
-                props: {
-                  $maxHeight: maxDropdownHeight,
-                  'aria-multiselectable': multi,
-                },
+  }, [props.options]);
```
tajo: this makes assumption that `props.options` is immutable but usually it's nots?

tajo:
lgtm except the grouping might not respect `props.options` update by the user since the `useMemo` usage

src/select/dropdown.js
chasestarr: yeah good point, i think i was getting a bit too aggressive on this perf optimization without truly measuring the improvements. i'll run some tests tommorow

PR #2481, opened 2019-12-13
fix(radio): radio's onChange prop typescript definition

### LINKED ISSUE #2469: [Typescript] RadioGroup and Radio's onChange prop is not typed like the native input
First of all : I'd like to say that baseweb is a great project : I have been using it for personal projects for a while now, and we were able to use it in my company since the typescript definitions have been added, so thank you all 🎉 

## Current Behavior

Currently, in typescript we need to cast the `event.target` in the `onChange` event handler, even though it is not needed for a native radio input, see : https://codesandbox.io/s/basic-usage-80wkd

## Expected Behavior

We shouldn't need to cast the event's target. After some digging, I've found that the [onChange prop](https://github.com/uber/baseweb/blob/master/src/radio/index.d.ts#L30 )'s typescript definition could be changed to use the same as the native input element (see the [@types/react definitions](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L2006)

i.e it would require changing the onChange props in all the radio components (Radio, RadioGroup, StatefulRadioGroup...)

from `onChange?: React.FormEventHandler<HTMLInputElement>;`

to `onChange?: React.ChangeEventHandler<HTMLInputElement>;`

I gave it a try and updated the prop in a project :  it worked liked a charm, plus it is backward-compatible and doesn't conflict with code that casts the event's target

If you are ok with this change, I would gladly send a PR to update the prop and the code examples in the doc :)

a-peltier:
Fixes #2469 

#### Description

Update the onChange props typescript definition all radio components (Radio, RadioGroup, StatefulRadioGroup & StatefulContainer)

from `onChange?: React.FormEventHandler<HTMLInputElement>;`

to `onChange?: React.ChangeEventHandler<HTMLInputElement>;`

This removes the need to cast the event's target as a "HTMLInputElement"  inside the onChange callback

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/documentation-site/examples/radio/basic.tsx
+++ b/documentation-site/examples/radio/basic.tsx
@@ -6,7 +6,7 @@ export default () => {
   return (
     <RadioGroup
       name="radio group"
-      onChange={e => setValue((e.target as HTMLInputElement).value)}
+      onChange={e => setValue(e.target.value)}
       value={value}
     >
       <Radio value="1">First</Radio>
--- a/documentation-site/examples/radio/error.tsx
+++ b/documentation-site/examples/radio/error.tsx
@@ -7,7 +7,7 @@ export default () => {
     <RadioGroup
       isError
       name="radio group"
-      onChange={e => setValue((e.target as HTMLInputElement).value)}
+      onChange={e => setValue(e.target.value)}
       value={value}
     >
       <Radio value="1">First</Radio>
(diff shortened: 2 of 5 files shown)


gergelyke:
thank you @a-peltier

gergelyke:
thank you @a-peltier - would you mind signing the contributor agreement?

a-peltier:
Sure thing, I just did :)

chasestarr:
thank you

PR #2483, opened 2019-12-13
feat(theme): use newly added semantic color tokens

nadiia:
#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


sandgraham:
Isn't this a breaking change for projects that have customized any of the replaced theme properties?

nadiia:
> Isn't this a breaking change for projects that have customized any of the replaced theme properties?

Could you clarify which part you referring to?

nadiia:
> > Isn't this a breaking change for projects that have customized any of the replaced theme properties?
> 
> Could you clarify which part you referring to?

But as we state in our [Versioning Policy](https://baseweb.design/discover-more/versioning-policy/) visual changes are not considered breaking.

PR #2486, opened 2019-12-13
feat(data-table): enables column resizing

chasestarr:
#### Description

Screen Shot 2019-12-13 at 1 57 59 PM

- does not handle min/max column widths provided on the column config
- slider should 'clamp' to min/max values in UI

#### Scope

- [x] Minor: New Feature


tajo:
Nice, a few things that would make the interaction a bit nicer: 

1. Keep resizing the column the whole time (like Apple Numbers), or at least make the ruler full height (like Google Sheets).
2. The cursor style flickers because the hover target is too small / lags behind. Might style it through JS instead.
3. When resizing, there is a tiny spot (the sorting arrow) that I can hit and suddenly the ruler  is gone:



4. Don't let user to go behind the minimal column width to prevent this abrupt columns jump:

chasestarr:
resolved the bugs I was aware of (min/max clamping, janky mousemove, chevron hover) and also improved performance by avoiding a cell remeasure. going to get some tests in here soon

chasestarr:
was thinking about keyboard controls as well. would focus and then arrow keys be appropriate? most other tools like google sheets do not have such features

PR #2490, opened 2019-12-15
feat(type): export UseStyletronFn

leyanlo:
#### Description

Export UseStyletronFn so others can use with createThemedUseStyletron.

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change


chasestarr:
I'm fine with exporting the type, but curious why it's needed?

leyanlo:
At the time, I wanted to pass `[css, theme]` as arguments to other functions. Turns out I no longer need this functionality, so ok with closing this too.

PR #2496, opened 2019-12-16
fix(datepicker): fix Datepicker quickSelect

leyanlo:
#### Description

Fix Datepicker quickSelect, which did not work on click because it would close the Calendar. The fix is to mount it to the root element in calendar.js, so click events will be accepted.

#### Test Plan

Try toggling quickSelect in Datepicker playground and see it works on click now.

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


gergelyke:
thank you @leyanlo - would you mind adding a regression test for this?

leyanlo:
> thank you @leyanlo - would you mind adding a regression test for this?

Done!

gergelyke:
@leyanlo I am wondering, should we add an e2e test instead? that would codify the user story instead of the DOM structure

leyanlo:
> @leyanlo I am wondering, should we add an e2e test instead? that would codify the user story instead of the DOM structure

Done

gergelyke:
Thank you Leyan

PR #2497, opened 2019-12-16
fix(build): do not include import of types in the esm target

gergelyke:


todorgrigorov:
Hi, is there a reason you have skipped the fix for component _Tabs_?

chasestarr:
> Hi, is there a reason you have skipped the fix for component _Tabs_?

Thanks, it looks to have been overlooked

PR #2501, opened 2019-12-17
fix(datepicker): clearable fix

gergelyke:


src/datepicker/datepicker.js
```diff
@@ -118,6 +120,25 @@ export default class Datepicker extends React.Component<
...
+    const inputHasLetters = new RegExp('[a-z]', 'i').test(inputValue);
+
+    // without letters and numbers, chances are, the user cleared the input
+    // and only the mask is present
+    if (!inputHasNumbers && !inputHasLetters) {
+      if (this.props.range) {
+        this.props.onChange &&
+          this.props.onChange({
+            // $FlowFixMe
+            date: [null, null],
```
chasestarr: if it's an array, items in it must be defined: https://github.com/uber/baseweb/blob/master/src/datepicker/types.js#L19

chasestarr:
Can you add this prop to the react-live? I couldn't test the functionality

src/datepicker/datepicker.js
```diff
@@ -118,6 +120,24 @@ export default class Datepicker extends React.Component<
 
   handleInputChange = (event: SyntheticInputEvent<HTMLInputElement>) => {
     const inputValue = event.currentTarget.value;
+    const inputHasNumbers = new RegExp('[1-9]').test(inputValue);
+    const inputHasLetters = new RegExp('[a-z]', 'i').test(inputValue);
+
+    // without letters and numbers, chances are, the user cleared the input
+    // and only the mask is present
```
leyanlo: Could we pull out the mask below into a class method and check against that?
```js
  getMask = () => {
    const {formatString} = this.props;
    let mask = null;
    if (this.props.mask !== null) {
      mask =
        // using the mask provided through the top-level API
        this.props.mask ||
        // to make sure it's not a breaking change, we try calculating the input mask
        // from the formatString, if used by the developer

        // 1. mask generation from the formatstring if it's a range input
        (formatString && this.props.range
          ? `${formatString} – ${formatString}`.replace(/[a-z]/gi, '9')
          : null) ||
        // 2. mask generation from the formatstring if it is NOT a range input
        (formatString ? formatString.replace(/[a-z]/gi, '9') : null) ||
        // falling back to the default masks
        (this.props.range ? '9999/99/99 – 9999/99/99' : '9999/99/99');
    }
    return mask;
  };
```
and here 
```js
    // user cleared the input and only the mask is present
    if (inputValue === this.getMask()) {
      if (this.props.onChange) {
        this.props.range
          ? this.props.onChange({
              date: [],
            })
          : this.props.onChange({
              date: null,
            });
      }
      return;
    }
```

gergelyke: yeah, that's what I wanted initially - but it's a bit more tricky

the `getMask` function would return something like this: `9999/99/99 – 9999/99/99`, but the inputvalue for this mask will be: `    /  /   –     /  /  `

this is why I ended up with the current logic - what do you think?

gergelyke:
> Can you add this prop to the react-live? I couldn't test the functionality

Ops, sorry - updated

documentation-site/components/yard/config/datepicker.ts
```diff
@@ -56,9 +56,12 @@ const DatepickerConfig: TConfig = {
...
         // them all
-        what: `Array.isArray(date) && date.length === 2 ?
-    "[new Date('" + date[0].toISOString() + "'), new Date('" + date[1].toISOString() + "')]"
-  : (Array.isArray(date) ? "[new Date('" + date[0].toISOString() + "')]" : "[new Date('" + date.toISOString() + "')]")`,
+        what: `(Array.isArray(date) && (!date[0] || !date[1])) || !date ?
+          "[new Date()]"
+          :
+            Array.isArray(date) && date.length === 2 ?
+              "[new Date('" + date[0].toISOString() + "'), new Date('" + date[1].toISOString() + "')]"
+            : (Array.isArray(date) ? "[new Date('" + date[0].toISOString() + "')]" : "[new Date('" + date.toISOString() + "')]")`,
```
nadiia: Need to check for `Array.isArray(date) ? date.length ? "[new Date('" + date[0].toISOString() + "')]" :  ` here since it can be an empty array.

src/datepicker/datepicker.js
chasestarr: it may be possible to replace all instances of `9` with ` ` and compare like that, but i'm not sure the extent of change required. I'm comfortable seeing what edge cases shake out based on the implementation in this pr

leyanlo: Ohh that’s right. I still think we should check against the mask for exactness, though, in case we have a mask with text like `Pickup: 9999-99-99`. Could we swap out the 9s?
```js
if (inputValue === this.getMask().replace(/9/g, ' ')) {
```

gergelyke: that could work - let me try that

gergelyke: @leyanlo you were right, this works perfectly - thank you

leyanlo: 🎉

documentation-site/components/yard/config/datepicker.ts
gergelyke: I think that's covered in the first part, here: `(Array.isArray(date) && (!date[0] || !date[1]))`

PR #2508, opened 2019-12-17
docs: remove links to cheatsheet when no types available

sandgraham:
Add some flow types so that the cheat sheet picks up the `typography` module.


chasestarr:
maybe special-case the typography directory? I'd rather not make code changes to satisfy the docs site

sandgraham:
@chasestarr what sort of special casing do you have in mind? Can make adjustments at a few places.

Also doing a full QA sweep of other components to make sure things are consistent.

PR #2509, opened 2019-12-18
fix(theme): remove duplicate code

gergelyke:


src/themes/shared/typography.js
```diff
@@ -13,221 +13,152 @@ export const fontTokens: FontTokensT = {
...
+    font650,
+    font750,
+    font850,
+    font950,
+    font1050,
+    font1150,
+    font1250,
+    font1350,
+    font1450,
+    ParagraphXSmall: font100,
```
jasonbarone: I recommend reversing this and using the new names as the default, and setting the fontXXX using the default names. I assumed that in the future the old names might be removed.

gergelyke: just discussed offline - probably we could keep both, and get away without a breaking change here in the future too (these are just aliases, so there is a barely any impact on the bundle size)

PR #2514, opened 2019-12-18
feat(data-table): controlled row highlight

chasestarr:
#### Description

Allows for the row highlight to be controlled from outside the component, with keyboard controls for example

#### Scope

- [x] Minor: New Feature


src/data-table/data-table.js
```diff
@@ -587,24 +570,63 @@ export function Unstable_DataTable(props: DataTablePropsT) {
...
 
+  const [columnHighlightIndex, setColumnHighlightIndex] = React.useState(-1);
+  const [rowHighlightIndex, setRowHighlightIndex] = React.useState(-1);
+
+  function handleRowHighlightIndexChange(nextIndex) {
+    setRowHighlightIndex(nextIndex);
+    if (gridRef.current) {
+      if (nextIndex >= 0) {
+        // $FlowFixMe - unable to get react-window types
+        gridRef.current.scrollToItem({rowIndex: nextIndex});
```
sandgraham: i think this can cause a slight jump when mousing over a partially shown row.

rowslightjump

src/data-table/data-table.js
```diff
@@ -587,24 +570,63 @@ export function Unstable_DataTable(props: DataTablePropsT) {
...
+      }
+      if (props.onRowHighlightChange) {
+        props.onRowHighlightChange(nextIndex, rows[nextIndex - 1]);
+      }
+    }
+  }
+
+  const handleRowMouseEnter = React.useCallback(
```
sandgraham: is there any need for a `handleRowMouseLeave` handler?

chasestarr: yeah it might be worth exploring how to 'unhighlight' the row when mouse leaves the whole table. right now the last highlighted row stays

src/data-table/data-table.js
chasestarr: this is a 'feature' of react-window, i'm afraid i can't turn this off

chasestarr: https://github.com/bvaughn/react-window/blob/51a24c4972c9ca5eee4df5ea3ec25bc4328e40c5/src/__tests__/__snapshots__/VariableSizeGrid.js.snap#L92

sandgraham: Ah okay - not a big issue.

PR #2515, opened 2019-12-18
feat(card): update styles to latest spec

sandgraham:
Update `Card` styles to latest spec:

- Remove box-shadow from StyledRoot
- Add 2px border to StyledRoot
- Increase StyledTitle type from font550 to HeadingSmall (20px to 24px)
- Increase StyledBody type from font200 to ParagraphMedium (14px to 16px)
- Reduce StyledContents padding from scale800 to scale600 (24px to 16px)


gergelyke:
on the homepage, can we horizontally align the buttons? 
Screen Shot 2019-12-18 at 2 02 27 PM

leyanlo:
@sandgraham would it make sense to add a theme value borders.surfaceBorderWidth set to 2px and then use that?

sandgraham:
@leyanlo I didn't add it as a theme property because I figured it was a unique instance of a 2px border. Looking through Figma, this 2px styling does seem to recur more than I thought, but I'm still not sure where else the `surfaceBorderWidth` would be used.

From a quick audit of designs, here are the components I found using the 2px styling for borders:
- `Card`
- `Slider`
- `Input` (on focus)
- `ProgressSteps`

Do you think that all of these should change in tandem if you edit a single theme property? Or would it be useful to have `surfaceBorderWidth` only apply to `Card`? I hesitate to add a property that is only used once, unless it has value as a consistent token for developers to build new things with.

gergelyke:
+1 for what Graham said - first we should get buy-in from design that this is something they treat as a design token before we start reusing it

leyanlo:
:+1:

Issue #2521, opened 2019-12-19
Datepicker with clearable prop


Mater:
## Current Behavior
`Datepicker` with `clearable` prop, error in console.
> Warning: Received `true` for a non-boolean attribute `clearable`.

## Expected Behavior
No errors

## Your Environment

https://codesandbox.io/s/datepicker-s2t2u

| Tech    | Version |
| ------- | ------- |
| Base UI | v9.33.0  |
| React   | 16.8.6  |
| browser |  Google Chrome Version 79.0.3945.79 (Official Build) (64-bit) |

- [x] I have searched the [issues](https://github.com/uber/baseweb/issues) of this repository and believe that this is not a duplicate.

PR #2524, opened 2019-12-19
fix(input): avoid passing clearable prop to base-input

leyanlo:
Closes https://github.com/uber/baseweb/issues/2521

#### Description

Avoid passing clearable prop to base-input, which causes React warnings when using clearable:
```
Warning: Received `true` for a non-boolean attribute `clearable`.

If you want to write it to the DOM, pass a string instead: clearable="true" or clearable={value.toString()}.
```

Also apparently fixes Datepicker issue where clicking input does not set focus border.

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


gergelyke:
Thank you Leyan!

PR #2528, opened 2019-12-19
feat(checkbox): update visuals

sandgraham:
Updates the `Checkbox` component to latest design specs:

- Updated to square corners
- Updated tick marks to new SVGs with... square corners
- Updated `disabled` state colors for light theme
- `theme.borders.useRoundedCorners` no longer applies to `Checkbox`, instead it pulls from `theme.borders.inputBorderRadius`


sandgraham:
@gergelyke So I went back to the lighter style for default checkbox. After updating the radio styles things looked a bit clunky with a dark/thick border. It reads a lot better with lighter gray as default- though a11y wise it is a concern.

We can discuss tomorrow more, but current PR has the lighter values in it for reference.

PR #2531, opened 2019-12-19
docs(sidebar): clean up sidebar links

leyanlo:
Second attempt at #2489, this time keeping Link from 'next/link'.

#### Description

Clean up sidebar links, which were unnecessarily nested inside Labels and had href attributes on divs. Now, cmd+click will properly open link in a new tab.

Before: 
sidebar-before

After: 
sidebar-after3

### DIFF
--- a/documentation-site/components/sidebar.js
+++ b/documentation-site/components/sidebar.js
@@ -7,70 +7,49 @@ LICENSE file in the root directory of this source tree.
 
 /* eslint-disable flowtype/require-valid-file-annotation */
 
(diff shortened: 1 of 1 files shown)


leyanlo:
> I think there is still a full page load - try switching the website to dark mode, and then navigate to a different page. You'll see a flash of the light theme

Ohh! Ok will play around some more

leyanlo:
Ok, I think this one works!

gergelyke:
works perfectly, thank you Leyan!

PR #2535, opened 2019-12-20
feat(spinner): add new div based spinner

sandgraham:
This PR introduces `StyledSpinnerNext`, a new `div` based spinner. 

Currently it is just a styled component. The API differs a bit from the old version. It has 3 set sizes based on latest design specs rather than allowing the developer to pass a `number` or `string`. It also doesn't haven't a `color` prop.


src/spinner/spinner.js
```diff
@@ -26,6 +26,12 @@ class Spinner extends React.Component<SpinnerPropsT> {
   };
 
   render() {
+    // TODO(v10): remove warning when switching default Spinner
+    console.warn(
```
chasestarr: can this move to mount so that it doesn't spam on re-renders?

sandgraham:
gokuthumbsup

src/spinner/spinner.js
```diff
@@ -25,6 +25,15 @@ class Spinner extends React.Component<SpinnerPropsT> {
     overrides: {},
   };
 
+  componentDidMount() {
+    // TODO(v10): remove warning when switching default Spinner
+    console.warn(
```
nadiia: It should log only in a dev env.

documentation-site/examples/spinner/next-custom.js
```diff
@@ -0,0 +1,15 @@
+// @flow
+
+import * as React from 'react';
+import {StyledSpinnerNext} from 'baseui/spinner';
+
+export default () => (
+  <StyledSpinnerNext
+    $style={({$theme}) => ({
```
nadiia: We should use `withStyle` in the example since `$style` is not a recommended way for adding style overrides.

src/spinner/spinner.js
sandgraham: backfilled a few additional missing checks here: https://github.com/uber/baseweb/pull/2542

documentation-site/examples/spinner/next-custom.js
sandgraham: pr to edit this: https://github.com/uber/baseweb/pull/2543

PR #2540, opened 2019-12-21
fix(table): show last column border in RTL

sijad:
#### Description

before:

2019-12-21-12:50:11-screenshot

after:

2019-12-21-12:49:34-screenshot

#### Scope

- [X] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


sandgraham:
Hi @sijad , thank you for the contribution!

Would you please add a visual regression test to cover this `rtl` use case? You can add a new scenario like so:

New file: `src/table/__tests__/table-rtl.scenario.js`
```js
/*
Copyright (c) 2018-2019 Uber Technologies, Inc.

This source code is licensed under the MIT license found in the
*/
// @flow

import {Table} from '../index.js';
import {ThemeProvider, LightTheme} from '../../index.js';

export const name = 'table-rtl';

  <ThemeProvider theme={{...LightTheme, direction: 'rtl'}}>
    <div dir="rtl" style={{height: '400px', width: '800px'}}>
      <Table
        columns={[...new Array(3)].map(() => 'Column Name')}
        data={[...new Array(4)].map(() =>
          [...new Array(3)].map(() => 'Cell Data'),
        )}
      />
    </div>
  </ThemeProvider>
);
```

Once that is added you should get an automated PR against your branch containing new snapshot images. Once those are merged into your branch I can land this PR.

sijad:
I've added the test, but didn't received any PR, did I did something wrong?

PR #2544, opened 2019-12-23
chore(border): update Card Root to follow BorderT pattern

leyanlo:
Update Card Root to follow BorderT pattern so it can be overridden with theme values without Styletron warnings. Also fix Table Semantic StyledTableHeadCell, which could potentially break since JS Object order is not guaranteed.

### DIFF
--- a/src/card/styled-components.js
+++ b/src/card/styled-components.js
@@ -34,18 +34,9 @@ export const HeaderImage = styled<{}>('img', ({$theme}) => ({
 }));
 
 export const Root = styled<{}>('div', ({$theme}) => ({
-  borderTopWidth: '2px',
-  borderRightWidth: '2px',
-  borderBottomWidth: '2px',
-  borderLeftWidth: '2px',
-  borderTopStyle: 'solid',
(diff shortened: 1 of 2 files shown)


src/themes/shared/borders.js
```diff
@@ -9,34 +9,88 @@ import type {BordersT} from '../types.js';
 
 const borders: BordersT = {
   border100: {
-    borderColor: 'hsla(0, 0%, 0%, 0.04)',
-    borderStyle: 'solid',
-    borderWidth: '1px',
+    borderTopColor: 'hsla(0, 0%, 0%, 0.04)',
```
nadiia: I think its usage should be changed, not the theme object. Additionally, it would be a breaking change.

leyanlo: I wasn’t happy with this either. Updated to change only the styled components.

nadiia:
@leyanlo Reverting the 'longhand => shorthand' change from this PR due to mixing long- and shorthand props resulting in overrides inconsistency. Sorry, for missing it in the original PR.

PR #2545, opened 2019-12-23
docs(sidebar): fix link missing href

sandgraham:
Fix an error being thrown when `href` is not provided to `Link`.

Screen Shot 2019-12-23 at 3 38 45 PM

Solution: only use `Link` if an `href` is provided, otherwise use `StyledNavLink`.


leyanlo:
Ah good catch, thanks!

PR #2550, opened 2019-12-25
fix(flex-grid): re-implement FlexGridItem to avoid nth-child selectors

leyanlo:
#### Description

Re-implement FlexGridItem to avoid nth-child selectors. Original implementation relied on the order of rules in the stylesheet, which is not guaranteed with Styletron, sometimes causing layout bugs if rules are out of order.

By passing the index and count props to each FlexGridItem, we can explicitly set margin-right and margin-bottom without relying on nth-child selectors. This also cleans up the DOM and reduces the number of classes:

FlexGridItem classes before: 
flex-grid-item-classes-before
FlexGridItem classes after: 
flex-grid-item-classes-after

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/src/flex-grid/__tests__/flex-grid-item.test.js
+++ b/src/flex-grid/__tests__/flex-grid-item.test.js
@@ -19,16 +19,76 @@ import {LightTheme} from '../../themes/index.js';
 describe('FlexGridItem', () => {
   test('flexGridItemMediaQueryStyle', () => {
     const testProps = [
-      {flexGridColumnCount: 1, flexGridColumnGap: 0, flexGridRowGap: 0},
-      {flexGridColumnCount: 1, flexGridColumnGap: '1px', flexGridRowGap: '2px'},
-      {flexGridColumnCount: 2, flexGridColumnGap: '1px', flexGridRowGap: '2px'},
-      {flexGridColumnCount: 3, flexGridColumnGap: '1px', flexGridRowGap: '2px'},
+      {
+        flexGridColumnCount: 2,
+        flexGridColumnGap: 0,
+        flexGridRowGap: 0,
+        flexGridItemIndex: 0,
+        flexGridItemCount: 4,
+        snapshotName: '2 columns no gaps',
+      },
+      {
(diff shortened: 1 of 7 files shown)


src/flex-grid/__tests__/flex-grid-item.test.js
```diff
@@ -19,16 +19,76 @@ import {LightTheme} from '../../themes/index.js';
...
     const testProps = [
-      {flexGridColumnCount: 1, flexGridColumnGap: 0, flexGridRowGap: 0},
-      {flexGridColumnCount: 1, flexGridColumnGap: '1px', flexGridRowGap: '2px'},
-      {flexGridColumnCount: 2, flexGridColumnGap: '1px', flexGridRowGap: '2px'},
-      {flexGridColumnCount: 3, flexGridColumnGap: '1px', flexGridRowGap: '2px'},
+      {
+        flexGridColumnCount: 2,
+        flexGridColumnGap: 0,
+        flexGridRowGap: 0,
+        flexGridItemIndex: 0,
```
gergelyke: @leyanlo if you had to change the props fed into the component, does it mean that it's a breaking chnge?

leyanlo: Shouldn’t be a breaking change. These props are generated by FlexGrid and passed to FlexGridItems automatically, which is why none of the other usages in this codebase needed to be updated.

src/flex-grid/types.js
```diff
@@ -17,4 +17,9 @@ export type FlexGridPropsT = {
   flexGridRowGap?: ResponsiveT<ScaleT>,
 } & BlockPropsT;
 
-export type FlexGridItemPropsT = FlexGridPropsT;
+export type FlexGridItemPropsT = {
```
gergelyke: one last thing before merging - can you also update the typescript definitions?

leyanlo: Updated!

PR #2565, opened 2020-01-04
docs(theming): update guide

sandgraham:
Updates the theming guide.


documentation-site/pages/guides/theming.mdx
```diff
@@ -6,126 +6,668 @@ LICENSE file in the root directory of this source tree.
...
+};
+
+function App() {
+  return <BaseProvider theme={customTheme}>{/* ... */}</BaseProvider>;
+}
+```
+
+Note, you can use the custom type theme object with our provider so long as it _extends_ the default `ThemeT`. Base Web components will essentially ignore the additional properties, but the theme object will be available in component overrides.
+
+If you want to use that custom theme shape with our various styling utilties, you will want create new `styled`, `withStyle`, and `useStyletron` methods which use your custom theme type. Luckily, we make this pretty easy for you.
```
chasestarr: You may want to communicate that this step is only required when using flow. I'm not sure about typescript though, it may be `any`

documentation-site/pages/guides/theming.mdx
```diff
@@ -6,126 +6,668 @@ LICENSE file in the root directory of this source tree.
...
+
+Let's look at how you can use the theme object while extending our components and when building out your own interfaces.
+
+### Extension
+
+There are two supported methods for styling Base Web components: `withStyle` and `overrides`. Both have access to the nearest ancestor theme object.
+
+#### `withStyle`
+
+Whenever Base Web exports a Styletron styled component we prefix that component with `Styled`. You can use `withStyle` to extend a styled component with your custom styles:
```
chasestarr: may want to link to styletron docs for anyone unfamiliar

documentation-site/pages/guides/theming.mdx
```diff
@@ -6,126 +6,668 @@ LICENSE file in the root directory of this source tree.
...
+        color: $theme.colors.accent,
+      })}
+    >
+      Hello
+    </div>
+  );
+}
+```
+
+[`useStyletron`](/components/use-styletron) lets you generate classes to be passed directly to an element’s `classNames` prop. `useStyletron` is a wrapper around the built-in Styletron hook. Our wrapper makes sure you have access to your theme by including it as a second item in the hook’s returned array. This makes it easy to style arbitrary elements without having to create new components or replicate design tokens.
```
chasestarr:
```suggestion
[`useStyletron`](/components/use-styletron) lets you generate class names to be passed directly to an element’s `classNames` prop. `useStyletron` is a wrapper around the built-in Styletron hook. Our wrapper makes sure you have access to your theme by including it as a second item in the hook’s returned array. This makes it easy to style arbitrary elements without having to create new components or replicate design tokens.
```

PR #2568, opened 2020-01-06
fix(select): stop event propagation when option from select dropdown clicked

### LINKED ISSUE #2563: Calendar time selector does not call onChage
When using the calendar with the built-in time picker, the onChange prop is not called when a time is selected. 

Reproducible here: https://baseweb.design/components/datepicker/#composed-single-pickers

akshatnitd:
`Fixes #2563 

#### Description
Stop event propagation when option from select dropdown clicked.

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


gergelyke:
thank you @akshatnitd !

akshatnitd:
@gergelyke Not sure why the particular test case fails while e2e testing, whereas works as expected when checked using examples in documentation site. Any possible reason for this would be helpful.

gergelyke:
it fails because the zeit now deployments are failing - looking into that now

akshatnitd:
@gergelyke Fixed the errors. 

Some error was coming in different part of the codebase as well, fixed it in this PR itself ✌🏻

PR #2569, opened 2020-01-06
chore(project): update to the lts nodejs version

gergelyke:


package.json
```diff
@@ -216,7 +216,10 @@
     }
   },
   "volta": {
-    "node": "10.17.0",
+    "node": "12.14.0",
     "yarn": "1.19.1"
+  },
+  "engines": {
+    "node": ">=10.x"
```
gergelyke: the reason for the disconnect between the two versions is the Node.js release process: https://nodejs.org/en/about/releases/

10.x is still in maintenance, so we don't want to throw warnings for those consumers, but ideally we develop against the latest LTS

PR #2582, opened 2020-01-07
fix(phone-input): autocomplete to show phone numbers without dialcode

shreedharshetty:
Fixes XXX (removed link, Gergely)

#### Description
We shouldn't allow the user to select phone numbers with country code from the phone-input's autocomplete

This change will only show phone numbers without country code [Ref](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete)

This was also reported on [one](https://medium.com/lenskart-engineering/mobile-number-entry-done-right-aaba0b292cbe) of Lenskart's eng blogs
#### Scope

- [] Patch: Bug Fix


gergelyke:
@sandgraham with that said, should we close this PR, or is there anything we should add to our codebase / docs?

sandgraham:
@gergelyke We could add an explanation to the phone input page on the doc site. I'll make a ticket to document this decision, and perhaps include some more research into why the current default is `tel`. I wish I had kept a record of some of that initial research.

gergelyke:
thank you @sandgraham

sandgraham:
@gergelyke @shreedharshetty 

I did some testing again, comparing `tel` and `tel-national` on Desktop and Mobile versions of Chrome and Safari.

Somewhat surprisingly, Mobile Chrome, Mobile Safari, and Desktop Safari fill in your phone number without the country code when using both `autoComplete="tel"` and `autoComplete="tel-national"`.

Desktop Chrome omits the country code with `tel-national`, as desired, but includes the country code with just `tel`.

---

For some reason when I initially tested all of these out, I found `tel-national` did not prompt auto completion as often as `tel`. I do think some of this comes down to what the browser has collected on the user so I also tested when the user has the country code included and not included in their saved addresses. I found that `tel-national` still performs as intended in these cases.

Perhaps there were some flaws in my initial testing, or maybe browsers have progressed a bit on this front since last year, but regardless, after revisiting this, I do think we should change to `tel-national`.

shreedharshetty:
Just to add my findings,
With `tel`
Screenshot 2020-01-07 at 11 28 51 AM
With `tel-national`
Screenshot 2020-01-07 at 1 22 55 PM

More details here: [Ref](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#attr-fe-autocomplete-tel)

sandgraham:
The a11y test is failing for phone input because `autocomplete="tel-national"` is meant to be used with `type="text"`, not `type="tel"`. [Check the spec here](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#attr-fe-autocomplete-tel-national).

Interesting enough, according to the spec, it seems that the only valid `autocomplete` value for `type="tel"` is... `tel`. So do we even need `autocomplete`? Leaving off `autocomplete` makes desktop Chrome recommend any phone number you've entered, both full numbers with country codes and national numbers without country codes- so it gives you all the options, some of which will be correct, some incorrect.

Okay, one option is to just omit `autocomplete` entirely then. 

However, going back to the spec, `type="tel"` is really meant to collect a full phone number:
> ...Full telephone number, including country code

So the more "correct" approach, according to the spec, when breaking the number into multiple inputs, is to use `type="text"` with `autocomplete="tel-national"`.  We then also have to add `inputMode="tel"` to keep our mobile telephone keyboards.

In order to pass the a11y test, we either need to omit `autoComplete`, use `autoComplete="tel"`, or use the following attributes:

```
<input type="text" autoComplete="tel-national" inputMode="tel" />
```

I'm leaning towards the last option. @shreedharshetty @gergelyke

gergelyke:
@shreedharshetty can you help us update the PR per Graham's recommendation?

PR #2584, opened 2020-01-07
Release v9.36.0

nadiia:


gergelyke:
I think it should be 9.36.0, as there a few features on the master that was not released

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "baseui",
-  "version": "9.35.0",
+  "version": "9.35.1",
```
gergelyke:
```suggestion
  "version": "9.36.0",
```

gergelyke: we should also update the title of the PR

PR #2587, opened 2020-01-07
feat(card): add img props

### LINKED ISSUE #2549: [Card] Expose `srcset` attribute for <HeaderImage /> in <Card />
## Feature description
In addition to the [`src` attribute](https://github.com/uber/baseweb/blob/master/src/card/card.js#L67) on the `<Card />` component, it would be helpful to pass an optional `srcset` attribute ([mdn](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-srcset)). For example

```js
<Card
  srcset={'https://source.unsplash.com/user/erondu/700x400 w400, https://source.unsplash.com/user/erondu/700x1200 w1200'}
  headerImage={'https://source.unsplash.com/user/erondu/700x400'}
>
  Contents
</Card>
```

nbdn:
Fixes #2549 

#### Description

Add the ability to pass headerImage as an object instead of a string.

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change


src/card/types.js
```diff
@@ -27,6 +27,10 @@ export type CardsPropsT = {
   +hasThumbnail: ({+thumbnail?: string}) => boolean,
   /** Image src to be positioned at the top of the Card. */
   +headerImage?: string,
+  /** Possible header image text description alternative. */
+  +headerImageAlt?: string,
```
gergelyke: can you add this to the TS definitions too?

gergelyke:
cc @jonsadka

jonsadka:
Although a breaking change, I almost wonder if a better API for this is to turn `headerImage` into an object that defines the image instead of overloading the props? This would make the interface clearer and allow us to expend to more attributes over time
```js
<Card 
  ...
  headerImage={{
    altText
    src,
    srcSet
  }}
/>
```

gergelyke:
> Although a breaking change, I almost wonder if a better API for this is to turn `headerImage` into an object that defines the image instead of overloading the props? This would make the interface clearer and allow us to expend to more attributes over time
> 
> ```js
> <Card 
>   ...
>   headerImage={{
>     altText
>     src,
>     srcSet
>   }}
> />
> ```

That makes sense - also, we can get away without a breaking change: we could check if it's a string or an object, and do things accordingly.

In the next major, we can clean this up, or simply keep it as is

jonsadka:
> > Although a breaking change, I almost wonder if a better API for this is to turn `headerImage` into an object that defines the image instead of overloading the props? This would make the interface clearer and allow us to expend to more attributes over time
> > ```js
> > <Card 
> >   ...
> >   headerImage={{
> >     altText
> >     src,
> >     srcSet
> >   }}
> > />
> > ```
> 
> That makes sense - also, we can get away without a breaking change: we could check if it's a string or an object, and do things accordingly.
> 
> In the next major, we can clean this up, or simply keep it as is

I like that. 


> * Switch to getOverrides helper method for greenkeeping (in a separate commit)
> 
> I found that adding alt prop and switching to get overrides helper method was a good thing regarding the rest of the code, if you find that it is more appropriate in a separate PR, I can easily update this one.

I think it's best to keep that scope in a separate diff

nbdn:
I agree with you, it's better to switch to an object prop. 

Do you find relevant to make the same thing for the thumbnail of the card ? 

To be sure of making things in the right way, should I update this PR with the object prop handling without breaking change (check on string type) ? 

I understand too for the get overrides scope, I will remove this part of the code, do you want to keep a trace of this debt in an issue ?

gergelyke:
> To be sure of making things in the right way, should I update this PR with the object prop handling without breaking change (check on string type) ?

Yes, I'd do that

nbdn:
@gergelyke @jonsadka Just updated the PR in this way, `headerImage` can now be an object. However I noticed that we can't add multiple types in the yard config, so I added the object possibility in the description.

With React PropTypes, we can use the `PropTypes.oneOfType([...])` method in this usecase, maybe we can add something similar in the react-view library ?

src/card/index.d.ts
```diff
@@ -15,7 +15,7 @@ export interface CardProps {
   readonly action?: React.ReactNode;
   readonly children?: React.ReactNode;
   readonly hasThumbnail?: (props: {readonly thumbnail?: string}) => boolean;
-  readonly headerImage?: string;
+  readonly headerImage?: string | any;
```
gergelyke: can we create a type for the object here?

nbdn: of course, I will create it asap

nbdn:
> lgtm once we create a type for the headerImage object

Done 😄

src/card/index.d.ts
```diff
@@ -15,7 +16,7 @@ export interface CardProps {
   readonly action?: React.ReactNode;
   readonly children?: React.ReactNode;
   readonly hasThumbnail?: (props: {readonly thumbnail?: string}) => boolean;
-  readonly headerImage?: string;
+  readonly headerImage?: string | CardsPropsT;
```
jonsadka: This should be `string  | CardHeaderImagePropsT`, no?

src/card/types.js
```diff
@@ -18,15 +18,32 @@ export type CardComponentsT = {
   Title?: OverrideT<?{}>,
 };
 
+export type CardHeaderImagePropsT = {
```
jonsadka: I think `ImagePropsT` is more appropriate

gergelyke:
@nbdn thank you for the updates - happy to merge once Jon's comments are addressed

sandgraham:
Little late to this PR, but feel I should play devil's advocate here - what do we get from adding this as an explicit API vs using the `HeaderImage` override to pass these props down? I worry about explicitly proxying native APIs within our own API. It also adds another convention for collecting props where we already have one (overrides).

gergelyke:
> Little late to this PR, but feel I should play devil's advocate here - what do we get from adding this as an explicit API vs using the `HeaderImage` override to pass these props down? I worry about explicitly proxying native APIs within our own API. It also adds another convention for collecting props where we already have one (overrides).

hm, that makes sense too - would you mind providing a code example on how that would look?

nbdn:
@jonsadka @gergelyke done



> > Little late to this PR, but feel I should play devil's advocate here - what do we get from adding this as an explicit API vs using the `HeaderImage` override to pass these props down? I worry about explicitly proxying native APIs within our own API. It also adds another convention for collecting props where we already have one (overrides).
> 
> hm, that makes sense too - would you mind providing a code example on how that would look?

Using this lib, I thought this first too, but regarding the opened issue and the demand, I thought it was really needed.

For me we can do this with overrides like this:

```
<Card
      headerImage="https://source.unsplash.com/user/erondu/700x400"
      overrides={{
        HeaderImage: {
          props: {
            alt: "Alt"
          }
        }
      }}
    >
...
</Card
```

sandgraham:
It does seem odd to bury an important prop like `alt` in `overrides`- so I do think there is a legitimate case for extending the top-level API for `headerImage`.

I think this change looks useful - I just hadn't seen any mention of `overrides` so wanted to make sure we considered it before extending API.

jonsadka:
> It does seem odd to bury an important prop like `alt` in `overrides`- so I do think there is a legitimate case for extending the top-level API for `headerImage`.
> 
> I think this change looks useful - I just hadn't seen any mention of `overrides` so wanted to make sure we considered it before extending API.

I think that's a valid point. Are there any attributes defined in `ImagePropsT` that you think should move towards the overrides API? Personally I only find myself regularly using `alt`, `src`, `srcSet` and `sizes` but it is a bit arbitrary to expose some attributes and not others.

Wondering how this same decision was made for the `<Block>` component? 

```js
export type ImagePropsT = {
  alt?: string,
  complete?: boolean,
  crossOrigin?: string,
  currentSrc?: string,
  height?: number,
  decode?: Promise<void>,
  isMap?: boolean,
  naturalHeight?: number,
  naturalWidth?: number,
  sizes?: string,
  src?: string,
  srcset?: string,
  useMap?: string,
  width?: number,
};
```

nbdn:
Is you to see, I added more than the alt prop in the `ImagePropsT` type (all types defined in the flow `HTMLImageElement`) , and I can understand that you want a limit in the reproduction of native api, the question is: what is the real limit in the baseweb philosophy?

sandgraham:
@jonsadka The Block component is why I am cautious about proxying native APIs 😅.

With Block we added props on an as-needed basis. Someone would open up a PR with a new CSS property and we would just grow the API over time. It's a slippery slope, though in the case of this PR, the slope is much shorter than what we faced with Block. `img` has a smaller and more stable API than CSS in general.

@nbdn I say if we are going to have a proxy API, then just leave them all in, as you've done. You never know what folks will need and the duplication isn't that bad. It would be a little awkward to use both `headerImage` and an overrides object to pass props to the same `img`.

gergelyke:
@nbdn thank you for bearing with us :) the last thing for you to do is to merge the visual regression images

nbdn:
@gergelyke It's done! It's a pleasure to contribute to this project, I will continue my contributions

src/card/__tests__/card-image-object.scenario.js
```diff
@@ -0,0 +1,35 @@
...
+import {styled} from '../../styles/index.js';
+
+export const name = 'card-image-object';
+
+const Container = styled('div', {width: '328px'});
+
+const headerImage: ImagePropsT = {
+  src: 'https://source.unsplash.com/user/erondu',
+  srcSet:
+    'https://source.unsplash.com/user/erondu/500x200, https://source.unsplash.com/user/erondu/600x300 1.5x',
```
gergelyke: @nbdn we have to change this to a fix image - the VRT will keep changing with every run

nbdn: Allright, it's now done, to avoid adding a new base64, I reused the thumbnail's base64

gergelyke:
@nbdn one last time you have to update the PR with the VRT PR 🤞

gergelyke:
thank you @nbdn - it's released in v9.37.0

Issue #2586, opened 2020-01-07
[Select] Log warning if provided value is not an array


chasestarr:
The Select component fails with no information if the provided value is not an array. If the node env is dev, it should check and log a warning.

PR #2588, opened 2020-01-08
fix(select): add warning for value prop miss-use

gergelyke:
Closes https://github.com/uber/baseweb/issues/2586

### DIFF
--- a/src/select/select-component.js
+++ b/src/select/select-component.js
@@ -849,6 +849,15 @@ class Select extends React.Component<PropsT, SelectStateT> {
       value,
       filterOutSelected,
     } = this.props;
+
+    if (__DEV__) {
+      if (!Array.isArray(value)) {
+        console.warn(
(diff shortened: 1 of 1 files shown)


src/select/select-component.js
```diff
@@ -849,6 +849,13 @@ class Select extends React.Component<PropsT, SelectStateT> {
       value,
       filterOutSelected,
     } = this.props;
+
+    if (__DEV__ && !Array.isArray(value)) {
```
chasestarr: the dev check needs to be separate from other conditionals

gergelyke: okay, I'll make the changes - can you help me why they need to be separated?

chasestarr: the code won't be removed from prod builds

gergelyke: thank you Chase - just updated the PR

PR #2597, opened 2020-01-09
fix(tag): update tag types

sebastianpantin:
#### Description

The tag component was missing type for the title prop. This fixes that.

#### Scope

- [X] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


gergelyke:
thank you @sebastianpantin - would you mind signing the CLA so we can land this?

PR #2598, opened 2020-01-09
fix(typography): update label line heights to spec

sandgraham:
Update line heights for Labels:
```
LabelXSmall: 16px
LabelSmall:  16px
LabelMedium: 20px
LabelLarge:  24px
```
Also update padding on Buttons to keep sizing on spec.


sandgraham:
Looks like a ton of VRTs got updated - seems a bunch of things use the label types. I'll do a sweep to verify sizing stays on spec.

sandgraham:
@chasestarr @gergelyke Could you re-review? I added changes to `Button` as well to get sizing lined up with FIgma specs.

PR #2605, opened 2020-01-10
fix(datepicker): handle hyphens and em dashes

gergelyke:


src/datepicker/__tests__/datepicker.test.js
```diff
@@ -156,6 +156,19 @@ describe('Datepicker', () => {
     expect(renderedInput.props().value).toEqual('2019/01/01 – 2019/01/04');
   });
 
+  test('converts hyphen to en dashes', () => {
+    const onChange = jest.fn();
+    const date = new Date('2019 01 01');
+    const mask = '9999/99/99 - 9999/99/99';
+    const value = [date, addDays(date, 3)];
+    const component = mount(
+      <Datepicker mastk={mask} onChange={onChange} value={value} />,
```
tajo: `mastk`?

what is this test converting? the dash in mask value to en dash?

also should we test em dash conversion too?

gergelyke: @tajo ugh, good catch, thank you - updated

PR #2607, opened 2020-01-11
docs(tooltip): change default triggerType to hover

raunofreiberg:
#### Description

Currently the default `triggerType` for Tooltip `click`. Although, it is not possible to interact with the component in a clickable manner. This PR changes the `triggerType` to `hover`, which should be a more suitable default, and fix the issue.

Example:

image

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


chasestarr:
What do you mean by "_it is not possible to interact with the component in a clickable manner_"? Does that mean because it's not a button, the click trigger type should not be used from an a11y perspective?

PR #2608, opened 2020-01-12
fix(avatar): reset to error state when src is missing

dannysindra:
#### Description

When the avatar's `src` is updated from a valid source to a null-ish value (undefined, null), the `onError` callback will not be triggered and the `noImageAvailable` flag will still be false, thus the Avatar will display broken image.

Here is the [sandbox to reproduce the issue](https://codesandbox.io/s/avatar-xeqg4).

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/src/avatar/avatar.js
+++ b/src/avatar/avatar.js
@@ -38,7 +38,7 @@ export default class Avatar extends React.Component<PropsT, StateT> {
   }
 
   componentDidUpdate(prevProps: PropsT, prevState: StateT) {
-    if (prevProps.src != this.props.src) {
+    if (prevProps.src !== this.props.src) {
       this.setState({noImageAvailable: !this.props.src});
     }
   }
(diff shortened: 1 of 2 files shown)


src/avatar/avatar.js
```diff
@@ -38,8 +38,8 @@ export default class Avatar extends React.Component<PropsT, StateT> {
   }
 
   componentDidUpdate(prevProps: PropsT, prevState: StateT) {
-    if (prevProps.src !== this.props.src && prevState.noImageAvailable) {
-      this.setState({noImageAvailable: false});
+    if (prevProps.src != this.props.src) {
```
chasestarr: is strict equality possible?

dannysindra: yup it is possible. Initially i was thinking there is no need to setState if the `src` prop is updated from a null-ish to another null-ish value (e.g. null to undefined, and vice versa). But such scenario may never occur or not worth checking, so I can update this to strict equality

PR #2613, opened 2020-01-13
fix(semantic-table): warning for removing unstable exports

gergelyke:


gergelyke:
> On second thought, I think this should just not be included. Anyone opting into the change knows what they sign up for when using the unstable prefix

I agree, that makes perfect sense

PR #2616, opened 2020-01-13
feat(eslint): add eslint-plugin-baseui package

sandgraham:
This PR adds a new `eslint-plugin-baseui` package to the codebase.

### Rules

Currently there are three rules:

| Rule | Responsibility |
| --- | --- |
| `deprecated-theme-api` | Identify theme properties that are deprecated. |
| `deprecated-component-api` | Identify components and props that are deprecated. |
| `no-deep-imports` | Identify importing unsupported modules from `baseui` source code. |

Each rule is in a separate file in `src`. [Read here](https://eslint.org/docs/developer-guide/working-with-plugins) for more info on how rules are implemented.

### Versioning

The version should match `baseui`'s own version, so whenever `baseui` publishes a new version, `eslint-plugin-baseui` will have a matching version published.

### Tests

All the tests are in `__tests__`. They make use of eslint's [RuleTester](https://eslint.org/docs/developer-guide/nodejs-api#ruletester).

### Preview

vscode-fix-theme-property


packages/eslint-plugin-baseui/__tests__/no-deep-imports.js
```diff
@@ -0,0 +1,73 @@
+/*
+Copyright (c) 2018-2020 Uber Technologies, Inc.
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+
+'use strict';
+
+const RuleTester = require('eslint').RuleTester;
```
gergelyke: nit: I think we mostly use destructuring across the code base - may worth updating here too

packages/eslint-plugin-baseui/bump.js
```diff
@@ -0,0 +1,16 @@
...
+LICENSE file in the root directory of this source tree.
+*/
+
+// Sets local package.json to current baseui version.
+
+'use strict';
+
+const fs = require('fs');
+const baseuiVersion = require('../../package.json').version;
+const packageJSON = fs.readFileSync('package.json', 'utf-8');
```
gergelyke: probably it's safer to also parse, and update the version field - while being unlikely, it can happen that another string will match `0.0.0` first

tajo: should be safe since `replace` happens only for the first match

gergelyke: yeah, that makes sense - probably I was overthinking it (what if we change the order of props in the package.json file)

packages/eslint-plugin-baseui/src/no-deep-imports.js
```diff
@@ -0,0 +1,25 @@
...
+module.exports = {
+  create(context) {
+    return {
+      ImportDeclaration(node) {
+        const match = node.source.value.match(/baseui\/(\w|-)*\//);
+        if (match) {
+          context.report({
+            node,
+            message:
+              "Import only from baseui's public API. Source code may change in breaking ways without notice.",
```
chasestarr: Where is public api defined? May want to refer to one level deep or similar

packages/eslint-plugin-baseui/src/deprecated-theme-api.js
```diff
@@ -0,0 +1,434 @@
...
+              data,
+              fix,
+            });
+            return;
+          }
+
+          // Exit early if we are not in a function.
+          // We don't have access to the theme outside of functions.
+          // Ex: styled, withStyle, overrides, function component (useStyletron)
+          if (scope.type !== 'function') {
```
chasestarr: what about a class method?

packages/eslint-plugin-baseui/src/deprecated-theme-api.js
```diff
@@ -0,0 +1,434 @@
...
+
+          // Check if we are setting a property in createTheme:overrides
+          // Ex: createTheme({}, {datepickerBackground: "pink"});
+          if (
+            context
+              .getAncestors()
+              .some(
+                node =>
+                  node.type === 'CallExpression' &&
+                  node.callee.name === 'createTheme',
```
chasestarr: is it possible to check that `createTheme` was imported from `baseui`?

packages/eslint-plugin-baseui/src/deprecated-theme-api.js
```diff
@@ -0,0 +1,434 @@
...
+            node.parent.parent.type === 'ObjectPattern' &&
+            node === node.parent.value
+          ) {
+            return;
+          }
+
+          // Check if our identifier is accessing properties on a theme, $theme, or its theme concern.
+          // Ex: theme.colors.foreground
+          // Ex: props.$theme.colors.foreground
+          // Ex: colors.foreground
```
chasestarr: Is it possible to find all `styled`, `withStyle`, and `useStyletron` instances.. then identify if `($)theme` is being used as a function argument or is the second array value returned from `useStyletron`? I get a bit nervous about the 'name' heuristic and how this has the potential to raise false positives. Does this restrict any object named `theme` or `colors` unrelated to baseui? I would much prefer to be more passive in matching so that developers do not turn the feature off or become hostile towards it

packages/eslint-plugin-baseui/src/deprecated-theme-api.js
```diff
@@ -0,0 +1,434 @@
...
+                data,
+                fix,
+              });
+              return;
+            }
+          }
+
+          // Check if our identifier was fully destructured from a $theme object.
+          // Ex: styled('div', ({$theme: {colors: {background}}}) => {})
+          scope.block.params.forEach(param => {
```
chasestarr: can this be checked on the function name matching `styled` or `withStyle`?

packages/eslint-plugin-baseui/__tests__/deprecated-theme-api-test.js
```diff
@@ -0,0 +1,256 @@
...
+  parser: require.resolve('babel-eslint'),
+  parserOptions: {
+    ecmaVersion: 6,
+    sourceType: 'module',
+  },
+});
+
+const tests = {
+  valid: [
+    {
```
chasestarr: does the code below pass or fail?

```
const colors = {background: 'xyz'};
function MyComponent() {
  return <p style={{backgroundColor: colors.background}}>hello</p>
}

File: packages/eslint-plugin-baseui/__tests__/deprecated-theme-api-test.js [hunk 9]
```diff
@@ -0,0 +1,256 @@
...
+      errors: [{messageId: MESSAGES.replaceThemeProperty.id}],
+      output: `
+        const Foo = styled("div", ({ $theme }) => ({
+          ...$theme.typography.ParagraphMedium
+        }));
+      `,
+    },
+
+    // Invalid setting deprecated property in createTheme
+    {
```
chasestarr: does the code below pass or fail? (create theme not imported from baseui)

```
function createTheme() {}
createTheme({colors: {shadowFocus: 'pink'}});
```

packages/eslint-plugin-baseui/src/no-deep-imports.js
sandgraham: I rewrote this one a few times trying to find the right concise wording. How about:

> Please limit imports to "baseui" or one level deep modules such as "baseui/button". Importing more than one level deep is dangerous as these files are not considered part of our public API.

packages/eslint-plugin-baseui/src/deprecated-theme-api.js
sandgraham: How would you access the theme in a class method?

packages/eslint-plugin-baseui/__tests__/deprecated-theme-api-test.js
sandgraham: Currently `colors.background` will be flagged.

packages/eslint-plugin-baseui/__tests__/deprecated-theme-api-test.js
sandgraham: Currently `shadowFocus` will be flagged.

packages/eslint-plugin-baseui/src/deprecated-theme-api.js
sandgraham: Yeah, I started checking for imports in `deprecated-component-api` so I can use the same pattern here.

packages/eslint-plugin-baseui/__tests__/deprecated-theme-api-test.js
sandgraham: I'm going to add a couple test cases for this and add a check for `createTheme` being imported.

packages/eslint-plugin-baseui/src/deprecated-theme-api.js
sandgraham: I think one nice thing with this current approach is that it will catch functions that are defined before being passed to `styled|withStyle|overrides` because it keys on `({$theme})` in the parameters of the function. I figured this type of de-structuring signature is unique enough to our styling utilities that it didn't require further refining.

packages/eslint-plugin-baseui/src/deprecated-theme-api.js
chasestarr: render and a styled component created there, maybe not that common though

packages/eslint-plugin-baseui/src/no-deep-imports.js
chasestarr: Yeah I think it's better to be more verbose

packages/eslint-plugin-baseui/src/deprecated-theme-api.js
chasestarr: I personally feel that this should 'miss' sometimes rather than be overzealous in flagging errors

packages/eslint-plugin-baseui/src/deprecated-theme-api.js
sandgraham: I think the scope will always be a function because the node's parent scope is the body of the function passed to styled.

```js
class Foo extends React.Component {
  render() {
    const Boo = styled('div', ({$theme}) => {
      return {
        backgroundColor: $theme.colors.background // parent scope is arrow function
      }
    });
    return <Boo />;
  }
}
```

chasestarr: Ah good point 👍

kahwee:
Wow this is really cool if we can run `yarn lint --fix` and get code migrated to the latest `baseui`!

packages/eslint-plugin-baseui/src/deprecated-theme-api.js
sandgraham: Added a check for import here.

packages/eslint-plugin-baseui/__tests__/deprecated-theme-api-test.js
sandgraham: Fixed in refactor

packages/eslint-plugin-baseui/__tests__/deprecated-theme-api-test.js
sandgraham: Fixed in refactor

packages/eslint-plugin-baseui/src/deprecated-theme-api.js
sandgraham: In refactor only functions passed to style, withStyle, and overrides style properties will have a chance to flag an error.

packages/eslint-plugin-baseui/src/deprecated-theme-api.js
sandgraham: After refactor it works close to this - still keys off the deprecated property as an identifier but then checks scope to verify it is inside of a styled|withStyle|overrides style function or that useStyletron is used.

sandgraham:
I refactored `deprecated-theme-api` to be much stricter in what it will flag. It should be pretty hard to trigger a false positive now.

There are some edge cases unaccounted for- which I'd like to circle back to at some point (and have added tests/todos for), but I think this is pretty useful as is.

packages/eslint-plugin-baseui/src/deprecated-theme-api.js
chasestarr: sweeeet

packages/eslint-plugin-baseui/__tests__/deprecated-theme-api-test.js
```diff
@@ -0,0 +1,825 @@
...
+      `,
+    },
+    {
+      code: `
+        // TODO(should-fail)
+        // useStyletron - direct array access, who would do this though?
+        function Foo({ children }) {
+          const boo = useStyletron()
+          return (
+            <div className={css({ color: boo[0].colors.foreground })}>{children}</div>
```
chasestarr:
```suggestion
            <div className={boo[0]({ color: boo[1].colors.foreground })}>{children}</div>
```

PR #2621, opened 2020-01-14
fix(select): don't perform setState on unmounted component

AndrewUldin:
#### Description

1. Place select in overlay
2. Open menu
3. Close overlay by clicking outside of overlay
4. Select will try to update its state which causes the warning and memory leak

Fix contains changes in onBlur handler. Now there is a condition check whether component is still mounted in a dom before performing setState.

#2623 

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/src/select/select-component.js
+++ b/src/select/select-component.js
@@ -119,10 +119,13 @@ class Select extends React.Component<PropsT, SelectStateT> {
     isPseudoFocused: false,
   };
 
+  isMounted: boolean = false;
+
   componentDidMount() {
     if (this.props.autoFocus) {
       this.focus();
     }
+    this.isMounted = true;
   }
 
   componentDidUpdate(prevProps: PropsT, prevState: SelectStateT) {
@@ -147,6 +150,7 @@ class Select extends React.Component<PropsT, SelectStateT> {
     if (__BROWSER__) {
       document.removeEventListener('touchstart', this.handleTouchOutside);
     }
+    this.isMounted = false;
   }
 
   focus() {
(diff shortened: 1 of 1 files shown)


houmark:
This is SOOOO crazy. I am literally at this very moment circling back to an error I noticed a long time ago, where a modal with a `<Select>` is throwing this error in *some* situations when I close the modal. When clicking the X in the modal or escape it would not, but if I click a cancel button which calls a function that closes, then it would throw the error.

After one hour of debugging and unable to figure out what was happening no matter what I tried, I did a hack where I would `setTimeout` on the cancel button, which seems to circumvent the problem. And then I get the email with the notification on the PR.

Seems I can remove my hacks and hope this one fixes my situation as well. It makes sense.

Thanks for this!

AndrewUldin:
I'm happy that I could be helpful for you. Hopefully that PR will be accepted. Cheers!

houmark:
I'm sure it will, it's a pretty simple fix, but the hard part for me was to realize that the error was in fact in the select component and not by my incorrect use.

houmark:
@AndrewUldin Please sign the contributor license to get this PR approved faster. You can see the link in the review area or [click here](https://cla-assistant.io/uber/baseweb?pullRequest=2621).

AndrewUldin:
> I'm sure it will, it's a pretty simple fix, but the hard part for me was to realize that the error was in fact in the select component and not by my incorrect use.

You could see that in a stack trace shown in your browser console. For chrome it is not obvious that you should click on warning itself in order to see entire trace. Hope it will help you in a future.

Thank you for pointing out to agreement. Merge will anyway rely on accepting it by maintainers located in PST.

houmark:
Yes I’m aware and it pointed to the select. But I was not sure if the error happened due to wrong use in my code of said select.

Uber devs are all over the world :)

houmark:
Just confirming. This also fixed the same error with a Select in a Modal when you close the Modal in certain ways!

PR #2629, opened 2020-01-14
Release v9.39.1

nadiia:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "baseui",
-  "version": "9.39.0",
+  "version": "9.39.1",
```
gergelyke:
```suggestion
  "version": "9.40.0",
```

becasue of https://github.com/uber/baseweb/commit/0b8b4c3badbc9d5f452e55415294d2af71a02f00

gergelyke: also, can we hold off on this until zeit gets back to us on the deployment issues?

gergelyke: or at least wait till EOD

nadiia: hm, the change you linked should be a patch not minor

PR #2631, opened 2020-01-15
feat: add href support to menu item

### LINKED ISSUE #2555: [StatefulMenu] Add optional href for menu item
It would be useful to be able to supply an href alongside a label to navigate to that destination on click, instead of supplying an `onItemSelect` callback that pushes history. 

The benefit of the href over a callback is that you can "Open Link in New Tab", for example.

[A solution where you supply an anchor as the label](https://codesandbox.io/s/statefulmenu-zcso6) is insufficient because only a click on the text, and not anywhere on the entire menu item, results in navigating to the href.

## Feature description

- be able to optionally supply an `href` as a property on each of the objects in the `items` array supplied to `StatefulMenu`
- ensure the rendered anchor fills the entire height/width of the parent `ListItem`, such that clicking anywhere on the `ListItem` navigates to the href

rohitb4:
Fixes #2555 

#### Description

Add an optional href to menu item. 
When href is provided to a menu item, an anchor tag is rendered within the `li` with href as the link and target as `_blank`.

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/src/menu/option-list.js
+++ b/src/menu/option-list.js
@@ -12,7 +12,7 @@ import {getOverrides} from '../helpers/overrides.js';
 
 import {OPTION_LIST_SIZE} from './constants.js';
 import MaybeChildMenu from './maybe-child-menu.js';
-import {StyledListItem} from './styled-components.js';
+import {StyledListItem, StyledListItemAnchor} from './styled-components.js';
 import type {OptionListPropsT} from './types.js';
 
 function OptionList(props: OptionListPropsT, ref: React.ElementRef<*>) {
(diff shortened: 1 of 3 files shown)


src/menu/option-list.js
```diff
@@ -32,6 +32,18 @@ function OptionList(props: OptionListPropsT, ref: React.ElementRef<*>) {
     overrides.ListItem,
     StyledListItem,
   );
+  const getItem = item => {
+    return (
```
chasestarr: Rather than writing the conditional in jsx, can this be refactored into two separate return statements please?

src/menu/option-list.js
```diff
@@ -32,6 +32,18 @@ function OptionList(props: OptionListPropsT, ref: React.ElementRef<*>) {
...
+      <>
+        {!item.href && getItemLabel(item)}
+        {item.href && (
+          <StyledListItemAnchor href={item.href} target="_blank">
```
chasestarr: There will likely need to be an option to open in a new-tab/new-window/same-tab

nadiia: `StyledListItemAnchor ` has to be overrideable.

gergelyke: yes, that would make perfect sense

Issue #2609, opened 2020-01-13
Smaller size than compact


taviroquai:
## Smaller size than compact

Currently, there are 3 sizes in the Base Web:
https://baseweb.design/components/button/#sizes

This request is to add a smaller size than compact.
I would sugest in the order of a 20px height, and be called "mini".

Some user interfaces in some web applications are very complex that need a considerable number of buttons in a small area.

This would be applied to all components that inherit a _button.size_ property.

Thanks!

gergelyke:
@taviroquai for now, this is not supported by the design system - I'll follow up with design if there is a desire to support this. Closing for now, but will reopen if we want to move forward with this!

gergelyke:
@taviroquai - jus talked with design, and they'd be open for an even smaller size. We've agreed, that it will be engineering-led, and they'll incorporate the size we come up with. With that said, they recommend that we first try to use a one size smaller font type, and one size smaller paddings.

You can find the current compact padding values here: https://github.com/uber/baseweb/blob/ddd276b8240c124e3e0d01ca19e510f9b7953135/src/input/styled-components.js#L34-L39

And the fonts: https://github.com/uber/baseweb/blob/ddd276b8240c124e3e0d01ca19e510f9b7953135/src/input/styled-components.js#L57

any chance you can help us trying this out with the input component first?

taviroquai:
@gergelyke Thank you so much for the response!

Yes I can help, but I need some guidelines or a pratical example of how I can test this.

I've learned that I can create a custom theme... but how do I change these specific items (input padding and font size)?

Thanks!

gergelyke:
let me put up a quick PR with me ideas

gergelyke:
actually ended up doing the whole thing :D https://github.com/uber/baseweb/pull/2632

what do you think?

taviroquai:
@gergelyke Awesome work! Yes thats I meant.

I was thinking about an even smaller size:
```css
font-size: 11px;
padding: 3px 4px;
```
This ends up nearly with a 20px height maximum.

But I can't complaint for your fast and awesome work! Thanks! :)

gergelyke:
For now, this is the smallest we can do without more fundamental changes - the reason for this is the multiselect component:

<img width="594" alt="Screen Shot 2020-01-16 at 8 51 26 AM" src="https://user-images.githubusercontent.com/2174968/72545172-75e00e80-383d-11ea-80d0-474920705775.png">

This is the smallest input size that can still host the Tag component.

PR #2632, opened 2020-01-15
feat(components): add a mini size

gergelyke:
closes https://github.com/uber/baseweb/issues/2609


gergelyke:
> Should we update the visual snapshots for all the affected components?

yeah, for now it's just a WIP so we can start collaborating with design (one other known issue is the multi select height, because of the added tags)

gergelyke:
@sandgraham just updated the PR with with VRT tests

nadiia:
The new mini size looks good! 👏

gergelyke:
> LGTM.
> 
> The name `mini` still seems a bit funny to me, but I don't have any other ideas.
> 
> `wee`?
> `tiny`?
> `cramped`?

we could call it small? so we'd have small, compact, default and large

houmark:
`small` or `tiny` for me :)

gergelyke:
Just talked with design, they prefer `mini` - also, for the record, we are a bit diverged from the naming used in the design spec (small, default, large), so probably we'll have to change it in the next major

PR #2641, opened 2020-01-16
feat(data-table): added empty state

### LINKED ISSUE #2403: [DataTable] ability to change empty state component
## Feature description

currently it's not possible to customize [empty state](https://github.com/uber/baseweb/blob/db8c89e3284a63df9f28fa954d3430b8f5ac3ea4/src/data-table/data-table.js#L314-L323) component.

foodforarabbit:
Fixes #2403 

#### Description

Added EmptyState override for DataTable
Added emptyState entry into locale (locale.datatable.emptyState)
Added an example on the documentation page
Added simple test to ensure that component renders / that override works

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change


foodforarabbit:
> Thanks for the contribution @foodforarabbit. The example on the docs site looks like it's working. I'd like to go in a slightly different direction with the API. Sorry that this was not documented in the issue. Can this be refactored from overrides to a render-prop pattern? I'm trying to approach this component in a way such that most configurations won't need to use overrides. Eventually, we'll likely add it but not until some stability testing.
> 
> In that situation, there would be a new prop on the data table called `emptyMessage` or similar that accepts a string or a react node. If that prop is non-null, render the provided ui vs the default message.

Oh yea sure will do!

foodforarabbit:
updated to use render prop

Issue #2592, opened 2020-01-08
Full height DataTable


supersonicclay:
## Current Behavior

Specifying `height: 100vh` on parent container produces produces double scrollbars and sometimes flickering
https://codesandbox.io/s/full-height-datatable-z90rj

![doublescroll](https://user-images.githubusercontent.com/469582/72124302-79e1cd00-3321-11ea-816d-702732c4daa4.gif)

## Expected Behavior

Table fills parent container with no double scrollbars and no flickering.

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v9.36.0  |
| React   |         |
| browser |         |

- [x] I have searched the [issues](https://github.com/uber/baseweb/issues) of this repository and believe that this is not a duplicate.

supersonicclay:
If I also put in `overflow: hidden` for attempt 2, then it gets rid of the double scrollbars and flickering. But seems like this shouldn't be needed.

chasestarr:
The table grows to fill its parent element.  See [this example](https://codesandbox.io/s/full-height-datatable-j8nfv) illustrating how `height: 100%` is not sufficient. You may want to explore `vh` units to fill some portion of the viewport height

supersonicclay:
Ok, ignore the `height: 100%`. The problem was that `html` and `body` needed to be `height: 100%` in order for the outer `div` to be full height. Or, as you mentioned, using `height: 100vh`.

But the secondary issue of double scroll bars is still an issue. This happens even with `height: 100vh`. Seems like the table is just a pixel or two too big which is making an extra scrollbar.

I updated the description above.

chasestarr:
Thanks for the reproduction. I will look into fixing it. Feel free to continue using the component and it should be resolved in a future baseui version with no api changes

PR #2645, opened 2020-01-16
fix(data-table): does not introduce scroll bars if full height/width

chasestarr:
fixes https://github.com/uber/baseweb/issues/2592

#### Description

The height/width previously did not take into account the border widths

#### Scope

- [x] Patch: Bug Fix


src/data-table/data-table.js
```diff
@@ -616,10 +620,12 @@ export function Unstable_DataTable(props: DataTablePropsT) {
...
   const normalizedWidths = React.useMemo(() => {
     const sum = ns => ns.reduce((s, n) => s + n, 0);
-    const resizedWidths = measuredWidths.map((w, i) => w + resizeDeltas[i]);
+    const resizedWidths = measuredWidths.map(
+      (w, i) => Math.floor(w) + Math.floor(resizeDeltas[i]),
+    );
     if (gridRef.current) {
       // $FlowFixMe
-      const domWidth = gridRef.current.props.width;
+      const domWidth = gridRef.current.props.width - 2;
```
gergelyke: may worth adding a comment on why we need this?

chasestarr:
```suggestion
      // minus 2 to account for the border stroke width
      const domWidth = gridRef.current.props.width - 2;
```

PR #2646, opened 2020-01-16
chore(eslint-plugin-baseui): add some info to readme and package.json

sandgraham:


sandgraham:
Okay, I'll rename during squash.

PR #2650, opened 2020-01-17
feat(popover): add autoFocus and returnFocus

### LINKED ISSUE #2633: Click Popover should trap focus and return focus to trigger when closed
## General Summary

Popover with triggerType of "click" should follow the recommendations for dialogs and drawers in [WAI-ARIA Authoring Practices 1.1](https://www.w3.org/TR/wai-aria-practices/#dialog_drawer).

[Codesandbox](https://codesandbox.io/s/statefulpopover-jswmu)

## Current Behavior

Currently when popover opens the popover does not trap focus.  When closing the popover focus is reset to top of document.

## Expected Behavior

Popover's Trigger should trap focus inside popover and closing the popover should return focus back to trigger.

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | ^9.32.0  |
| React   |    ^16.8.6     |
| browser |   chrome   79  |

Not a duplicate but perhaps similar to https://github.com/uber/baseweb/issues/1807

foodforarabbit:
Fixes #2633 

#### Description

Added props to popover

focusLock?:  boolean = true - whether to render with FocusLock or not (required for components that use popover but don't expect to be focus locked)
autoFocus?:  boolean = true - whether to autofocus first focusable element within content
returnFocus?: boolean = true - whether to return focus to original element after closing popover

implemented using FocusLock component similar to Drawer

returnFocus needs to be set to `false` if focusing the original element opens the popover in the first place or the popover will open again on close. (i.e. DatePicker)

set `noFocusGuard={true}` prop on FocusLock since it violates a11y rule elements should not have tabindex greater than 0 (e.g. https://dequeuniversity.com/rules/axe/3.2/tabindex)

updated some snapshots to account for element changes
fixed some tests to work as it did before (mostly tests that grab values at specific indexes which are different now after inserting new stuff)

updated documentation with the three new props, and updated example element to include focusable components so that the autoFocus / returnFocus can be demonstrated

had to set `focusLock` to `false` on components that used popover since it was breaking e2e tests
- datepicker
- calendar-header
- country-select
- select-component
- maybe-child-menu (added new e2e test to test for functionality of navigating into child menu and navigating back out and to a different menu item)
#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/documentation-site/components/yard/config/popover.ts
+++ b/documentation-site/components/yard/config/popover.ts
@@ -5,6 +5,8 @@ import {
   PLACEMENT,
   TRIGGER_TYPE,
 } from 'baseui/popover';
+import {Button} from 'baseui/button';
+import {Input} from 'baseui/input';
 import {PropTypes} from 'react-view';
 import {TConfig} from '../types';
 
@@ -17,6 +19,8 @@ const PopoverConfig: TConfig = {
     'baseui/popover': {named: ['StatefulPopover']},
   },
   scope: {
+    Button,
+    Input,
     StatefulPopover,
     ACCESSIBILITY_TYPE,
     PLACEMENT,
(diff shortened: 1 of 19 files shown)


foodforarabbit:
wait, don't merge yet adding another fix for focusLock on childMenu preventing some menu functionality (e.g. left arrow return to original menu)

PR #2654, opened 2020-01-18
fix(datepicker): yard config autoFocusCalendar prop type

### LINKED ISSUE #2533: [Datepicker] Disable keyboard opening on mobile devices
Feature request from @jonsadka:

> Is it possible to disable the keyboard from opening on mobile devices when using the <Datepicker/> component? It feels like unwanted behavior.

foodforarabbit:
Fixes #2533 (kinda, fixes the documentation to expose this functionality)

#### Description

I was looking into #2533 and it turns out it was fixed already. But in the documentation the prop type was incorrect so I fixed that.

With autoFocusCalendar prop set to true, and after a single click:

image

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/documentation-site/components/yard/config/datepicker.ts
+++ b/documentation-site/components/yard/config/datepicker.ts
@@ -67,7 +67,7 @@ const DatepickerConfig: TConfig = {
     },
     autoFocusCalendar: {
       value: undefined,
-      type: PropTypes.Function,
+      type: PropTypes.Boolean,
       description:
         'Defines if the calendar is set to be focused on an initial render.',
       hidden: true,


documentation-site/components/yard/config/datepicker.ts
```diff
@@ -67,7 +67,7 @@ const DatepickerConfig: TConfig = {
     },
     autoFocusCalendar: {
       value: undefined,
```
tajo:
```suggestion
      value: false,
```

tajo:
```suggestion
       value: false,
```

tajo:
Closing this as a duplicate to https://github.com/uber/baseweb/pull/2656

PR #2655, opened 2020-01-18
chore(layer): add console warning for missing BaseProvider/LayersManager

### LINKED ISSUE #2513: [ThemeProvider][BaseProvider] Popover mounts new divs to end of body element
If an application does not use the `BaseProvider`, `Popover` portaling will leave behind an empty div. What are some ways that baseui can warn developers that `BaseProvider` is missing?

foodforarabbit:
Fixes #2513 

#### Description

Added a warning if no LayersManager / BaseProvider in the app (no host passed in from context consumer)

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


src/layer/layer.js
```diff
@@ -27,15 +27,21 @@ class LayerComponent extends React.Component<
...
     }
+
     // There was no LayersManager added if this.props.host === undefined.
     // Use document.body is the case no LayersManager is used.
-    const host =
-      this.props.host !== undefined ? this.props.host : document.body;
+    const hasLayersManager = layersManagerHost !== undefined;
+    if (!hasLayersManager) {
+      console.warn(
+        '`LayersManager` was not found. This occurs if you are attempting to use a component requiring `Layer` without using the `BaseProvider` at the root of your app.',
```
gergelyke: should we add a link to the relevant documentation site?

src/layer/layer.js
```diff
@@ -27,15 +27,21 @@ class LayerComponent extends React.Component<
...
       onMount && onMount();
       return;
     }
+
     // There was no LayersManager added if this.props.host === undefined.
     // Use document.body is the case no LayersManager is used.
-    const host =
-      this.props.host !== undefined ? this.props.host : document.body;
+    const hasLayersManager = layersManagerHost !== undefined;
+    if (!hasLayersManager) {
```
chasestarr: Can you please wrap this in a `__DEV__` conditional? There should be other instances if you search the codebase. Doing so will exclude this check from production builds - reducing bundle size and some excess operations at runtime

foodforarabbit: oops good catch, thanks!

Issue #1769, opened 2019-08-19
How to have two Toast's on the same page/component in two different locations?


houmark:
## Current Behavior

I have a UI where I would like to have a confirmation toast in the lower right side and an "information" toast which would be in the top center.

How can I add two toasts within the same component, and then programmatically address each one of them individually?

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v8.16.4  |
| React   |     16.8.6    |
| browser |    Chrome 76.x     |

- [x] I have searched the [issues](https://github.com/uber-web/baseui/issues) of this repository and believe that this is not a duplicate.

gergelyke:
hi @houmark - from a strictly UX point of view, probably notification should always pop up from the same spot, no matter if they are confirmation or information

with that said, I am not sure how to achieve that - Nadiia, can you help with this? is it possible with the current setup?

houmark:
I somewhat agree and disagree :) Notifications in relation to user interaction is always the same place, but notifications from third party changes (less important to notice) can go on a different location for me without being confusing.

houmark:
Awaiting response on this one, but I think this needs new development to be possible.

gergelyke:
> Awaiting response on this one, but I think this needs new development to be possible.

Hi @houmark - this is not something the core team can take on, as our internal design always use one Toast only. If you have the bandwidth to contribute, we can potentially land it :)

houmark:
Understood, but can we then set the proper labels and remove stale so it isn't forgotten?

gergelyke:
done :)

chasestarr:
I helped someone figure this out the other day, sharing here: https://codesandbox.io/s/toaster-basic-usage-4rqld

closing, let me know if this does not work for you

houmark:
Nice! That's great and simple. I had thoughts in that direction actually but didn't get around to try it out.

houmark:
Maybe this should be added to Toast docs for anyone else to know how this can be done.

gergelyke:
@houmark can you please open a PR adding this as an example?

PR #2657, opened 2020-01-19
fix(toast): deduplicate toasts with same key

### LINKED ISSUE #2333: Add ability to de-duplicate toaster notifications
The `toaster` export from `"baseui/toast"` is **very** handy. I was wondering if it might be possible to provide the ability to de-duplicate the toaster notifications such that I could prevent multiple "toast" notifications appearing with the same content and also allow me to update the properties of said notification. I imagine adding an optional `id` (or possibly use the existing `key` property) to the options object which if the `toaster` saw an `id` value which matched one of the currently visible "toast" notifications, it would not create an additional notification. Instead it would ideally update the existing notification with the new properties (e.g. replace the children, reset the `autoHideDuration` and timeout, etc.).

I understand that there is already `toaster.update(key, {options})`, but it is fairly tedious to keep track of which notifications are currently visible and use that info to switch between, say `toaster.update()` and `toaster.info()`.

foodforarabbit:
Fixes #2333 

#### Description
- fixed typing on some of the functions. There was a type ToastPropsShapeT (ToastPropsT minus children) being used for state.toasts and for the return value of getToastProps(), but actually the children prop is in both. children is separate in the toaster.show() function but then it is recombined in the toasterInstance.show() function.
- added helper function updateToastsMutationWithCount which performs the original toast update mutation but also returns the number of toasts updated
- added updateToastsMutationWithCount to toasterInstance.show() function - it will attempt to update toasts with matching key prop first, and then only push the new toast if nothing was updated
- clear existing autohide timeout in Toast.startTimeout if it already exists (i.e. reset the autohide timer)
- call Toast.startTimeout in componentDidUpdate ONLY if `autoHideDuration` property is changed - we can avoid a deep comparison to figure out if a toast has changed or not. Might want to make this more explicit somehow - suggestions welcome
- used `Layer` instead of creating a portal manually to the body
- added new scenario and e2e test for normal toast behaviour (no key provided, autogenerate key and pop up additional toasts) and same-key behaviour (if toast with same key found, update it instead of popping up a new one)
- added some text on the documentation page about `toaster` utility and the same-key update situation

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


foodforarabbit:
actually I was thinking we could implement a hidden counter in the toast config (e.g. `__updated`) that is increased every time a toast is updated. That way, we could know when to reset the autohide timer without any expensive comparisons while also making it automatic (right now I'm simply comparing autoHideDuration so to reset the autohide timer, you need to change this. Seems a bit hacky though). Not sure if this discussion belongs here or in the original issue though.

houmark:
Excited about this PR. I did not have the chance to test it out yet, but it seems like everything annoying about toasts fixed :) Nice work.

houmark:
About the `autoHideDuration` — Do I understand it correctly, that to extend an updated toast that had 10000 in its first show, to have another 10 seconds, I will need to submit the update/second show with 10001 for example? Basically a different value than the first time? If I do not submit a value or the same, it will hide 10 seconds after showing the first toast?

A different approach to both of those could be a new prop `extendAutoHideDuration` so I can decide when updating if I want it to extend the time or not (seems less hacky), but I don't have a strong opinion on either one of the solutions.

houmark:
> I agree that changing the `autoHideDuration` is not sufficient, and the update count approach seems more appropriate. Another prop called `resetAutoHideTimerOnReset` or similar can be added to control the behavior. If 'showing' a duplicate toast should it reset the timer by default, or the opposite? I personally am leaning towards reset.

Just a naming nitpick. `resetAutoHideTimerOnReset` should maybe be `resetAutoHideTimerOnUpdate`?

src/toast/toaster.js
```diff
@@ -193,12 +210,7 @@ export class ToasterContainer extends React.Component<
         if (__BROWSER__) {
           return (
             <>
-              {ReactDOM.createPortal(
-                root,
-                // $FlowFixMe
-                document.body,
-              )}
-              {this.props.children}
+              <Layer key={'toast-layer'}>{root}</Layer>,{this.props.children}
```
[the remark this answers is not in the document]
foodforarabbit: hm weird, not sure why it ended up like this. Fixed

documentation-site/components/yard/config/toast.ts
```diff
@@ -49,27 +50,53 @@ const toastConfig: TConfig = {
...
-            )}>Show more</Button>
-          </Block>);
-          toastKey = toaster.info((<>{msg}{showMore}</>), {
-            onClose: ()=>console.log('Toast closed.'),
-            overrides: {InnerContainer: {style: {width: '100%'}}}});
-        }}
-        >
-          Show notification
-        </Button>
+      value: `<>
```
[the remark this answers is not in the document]
foodforarabbit: I broke out the same-key button into its own example. I demonstrate both updating the toast (with different notification kinds) and resetting the timeout (timeout is set to 1000ms, clicking the button repeatedly will cause the toast to stay open)

src/toast/toaster.js
```diff
@@ -56,29 +56,47 @@ export class ToasterContainer extends React.Component<
...
+    toasts.forEach((t, index, arr) => {
+      if (t.key === key) {
+        arr[index] = {...t, ...this.getToastProps(props), key};
+        updated += 1;
+      }
+    });
+    return updated;
+  };
+
+  getToastProps = (props: ToastPropsT): ToastPropsT & {key: React.Key} => {
```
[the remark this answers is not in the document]
foodforarabbit: agreed, I prefer immutability as well just thought we were mutating for performance reasons. Rewrote it to be purely functional / immutable.

foodforarabbit:
> > I agree that changing the `autoHideDuration` is not sufficient, and the update count approach seems more appropriate. Another prop called `resetAutoHideTimerOnReset` or similar can be added to control the behavior. If 'showing' a duplicate toast should it reset the timer by default, or the opposite? I personally am leaning towards reset.
> 
> Just a naming nitpick. `resetAutoHideTimerOnReset` should maybe be `resetAutoHideTimerOnUpdate`?

caught this already :D

chasestarr:
> > > I agree that changing the `autoHideDuration` is not sufficient, and the update count approach seems more appropriate. Another prop called `resetAutoHideTimerOnReset` or similar can be added to control the behavior. If 'showing' a duplicate toast should it reset the timer by default, or the opposite? I personally am leaning towards reset.
> > 
> > 
> > Just a naming nitpick. `resetAutoHideTimerOnReset` should maybe be `resetAutoHideTimerOnUpdate`?
> 
> caught this already :D

🙄 ah good catch, typo

documentation-site/pages/components/toast.mdx
```diff
@@ -33,6 +34,11 @@ top/left, bottom/right, bottom/center, bottom/left. Toasts are intended to provi
 and self-dismiss after a short time with an option to self-dismiss. Each toast can be colored to fit
 it’s status type of neutral, positive, warning and alert.
 
+**Using the `toaster` utility:**
+
+Toasts can be created easily with some of the helper functions on `toaster`, as seen in the code sample above. If a key isn't provided, it will auto-generate a unique key for you which will be returned in the instantiation functions (`info, positive, warning, negative`).
+If a toast with an identical key already exists, that toast's contents will be updated instead of a new toast being created. The autoHide timeout will reset if the `resetAutoHideTimerOnUpdate` prop is set to true on the toaster.
```
nadiia: Could you specify that it's set to `true` by default? Maybe rephrase to something like `unless resetAutoHideTimerOnUpdate  prop is set to true on the toaster`.

src/toast/toaster.js
```diff
@@ -193,11 +204,7 @@ export class ToasterContainer extends React.Component<
         if (__BROWSER__) {
           return (
             <>
-              {ReactDOM.createPortal(
-                root,
-                // $FlowFixMe
-                document.body,
-              )}
+              <Layer key={'toast-layer'}>{root}</Layer>
```
nadiia: It would be nice to use `Layer` here, but in the current state, it will break applications with multiple layers since any later rendered layers will overlap toast notifications since the toasts container is added earlier. And that may not be the desired behavior.

src/toast/types.js
```diff
@@ -88,8 +93,10 @@ export type ToasterPropsT = {
   /** The number of milliseconds to wait before automatically dismissing a
    * notification. This behavior is disabled when the value is set to 0.*/
   autoHideDuration: number,
+  /** When updating a toast, should we reset the autohide timer */
```
nadiia:
```suggestion
  /** Defines if updating a toast resets the autohide timer */
```

chasestarr:
One potential bug on the docs site example:

repro steps:
1. click 'show notification'.
2. wait a few seconds, toast disappears.
3. click 'show notification'.
4. click 'show more'.
5. toast does not disappear.

I would expect the toast to disappear on step 5

nadiia:
> 4. click 'show more'.
> 5. toast does not disappear.
> 
> I would expect the toast to disappear on step 5

The `Show more` just updates the notification but doesn't dismiss it.

nadiia:
It seems that`resetAutoHideTimerOnUpdate` in Yard does not work.

src/toast/__tests__/toaster.scenario.js
```diff
@@ -13,18 +13,31 @@ import {toaster, ToasterContainer, PLACEMENT} from '../index.js';
...
+        autoHideDuration={500}
+      />
+      <Button
+        id="default"
+        onClick={() => {
+          toaster.info('hi');
+        }}
+      >
+        Info toast
+      </Button>
```
nadiia: nit: Could you add some spacing between buttons?

src/toast/toaster.js
foodforarabbit: gotcha

foodforarabbit:
fixing some code comments first, will check out the bugs next

foodforarabbit:
> One potential bug on the docs site example:
> 
> repro steps:
> 
> 1. click 'show notification'.
> 2. wait a few seconds, toast disappears.
> 3. click 'show notification'.
> 4. click 'show more'.
> 5. toast does not disappear.
> 
> I would expect the toast to disappear on step 5

Ah yes I set up the test incorrectly. 

1) I was passing in autoHideDuration as a toast prop instead of as a toaster prop
2) on update, we first spread the original props (e.g. autoHideDuration: 1000), then we see if there is an autoHideDuration override in the updated props (no, I wouldn't expect to have to set it again anyway) and finally we get the autoHideDuration from the ToasterContainer. **we might want to revisit this functionality since autoHideDuration is kind of a special prop - I will take it from the original toast props instead of letting it use the ToasterContainer value on update. I am thinking that it is a more common use-case for someone to change the autoHideDuration for a toast, and they would want to keep the changed autoHideDuration after updating that toast.
3) ToasterContainer and toaster uses a global ref... so basically only one ToasterContainer was being used as the props. I didn't realize this until I inspected it closer. I've removed the redundant ToasterContainers from the example component, and added some text to the button to make it more clear.

src/toast/toaster.js
```diff
@@ -41,6 +41,15 @@ export class ToasterContainer extends React.Component<
 
   constructor(props: ToasterPropsT) {
     super(props);
+
+    if (__DEV__) {
+      if (toasterRef !== undefined) {
+        // eslint-disable-next-line no-console
+        console.warn(
+          'A ToasterContainer already exists in your application. It is not recommended to have more than one ToasterContainer in an application.',
```
houmark: I have a bit of a problem with this one. We will be spammed with that message in development, as we plan to have multiple toasters in the near future. I know it's not recommended, but there are valid uses cases and technically there should be no issue if done with a `ref`. Maybe the check could be more sophisticated and check if each `ToasterContainer` has a `ref` before creating the warning.

See a working sandbox here: https://codesandbox.io/s/toaster-basic-usage-4rqld

Also discussed in #1769

foodforarabbit: good point, I didn't consider that it could be used that way. I took it out since there isn't an easy way to check for "ref" prop (could use React.forwardRef but... that will be different issue)

Also just thought if placement could be controlled by the toast itself (the ToasterContainer prop can jsut be default like autoHideTimeout is) that would cover your use case with a single ToasterContainer as well. It shouldn't be too hard to sort toasts based on placement and render them separately. But again, another time, another issue :)

houmark: > Also just thought if placement could be controlled by the toast itself (the ToasterContainer prop can jsut be default like autoHideTimeout is) that would cover your use case with a single ToasterContainer as well. It shouldn't be too hard to sort toasts based on placement and render them separately. But again, another time, another issue :)

I'd personally love to see that (in a different PR of course). It would make it simpler and then this warning could be reintroduced and the old behavior deprecated in the warning with a link to the documentation for that new feature. But as you may read in the issue I created to ask for this, it did not seem like this was a feature the Uber team wanted as it was not good practice, but they may be more open to it now (as they had a case where they helped someone implement it after I did my question)?

foodforarabbit:
hehe well I'll leave it up to them to decide. I'm still new around here,
just fixing bugs and filling in gaps. Don't know what the long term plans
are yet - was going to finally get started looking over the figma tomorrow!

but I believe that if they are going to allow it at all, props on a toast
is the lesser evil compared to two ToasterContainers controlled by directly
by ref puppet strings xD

chasestarr:
I think this is good to land once VRT changes are resolved. Thanks for looking into the auto-hide issue and raising the some of the complexities with multiple containers. I'm open to start addressing the placement-by-toast feature. It's not a common pattern, but there's at least one application here that needs it

foodforarabbit:
> I think this is good to land once VRT changes are resolved.

Not sure what I need to do with the VRT xD I saw a pull request into my branch and merged it in, saw 3 pngs but not sure what I was looking for.

chasestarr:
It's kind of a side-effect of how we combine e2e tests and visual regression tests. In this case, the new screenshots look fine

PR #2661, opened 2020-01-20
fix(progress-steps): refactor unncessary ContentTail :after into just plain div

### LINKED ISSUE #2562: ProgressSteps override to increase length of ContentTail doesn't work in IE11
We want to make the `ContentTail` longer than the default, so we used an override on `StyledContentTail`. This works in Chrome but not in IE11 - in IE11, the tail is still too short and doesn't connect all the way to the next dot.

Codesandbox:
https://codesandbox.io/s/progress-steps-ie-bug-nm238

The relevant override is:
```js
  Tail: {
    style: (props: {$theme: ThemeT}) => {
      const {
        $theme: {
          sizing: {scale100}
        }
      } = props;
      return {
        ':after': {
          height: `calc(100% - ${scale100})`
        }
      };
    }
  },
```

## Current Behavior
The `ContentTail` is too short, this is an IE11 screenshot:
image

## Expected Behavior
The `ContentTail` is just right, and leads nicely in to the next dot:
image

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v9.30.0  |
| React   |  v16.10.2       |
| browser |  Chrome 78.0.3904.108 (works), IE 11 (doesn't work)       |

foodforarabbit:
Fixes #2562 
#### Description

not a direct fix of issue but fix underlying cause which might help with issue
instead of having ContentTail comprised of an invisible div with visible portion implemented as an :after element, I've refactored it to only be a div that is visible which should make it easier to override.
One "breaking" change:
in the original implementation, the div acted as a kind of container for the tail, which was implemented by the :after element (e.g. the tail itself was moved down with padding, and extended in height to 100% + sizing)
in this new implementation, the div itself is the tail.

While it looks and works the same for our documentation / using the default implementation, any modifications that depended on calculations based on the div or the :after will probably be wrong.

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


gergelyke:
@foodforarabbit thank you for the contribution! can you help us verify that it works on IE?

foodforarabbit:
> @foodforarabbit thank you for the contribution! can you help us verify that it works on IE?



This is IE 11

src/progress-steps/styled-components.js
```diff
@@ -132,17 +132,12 @@ export const StyledContentTail = styled<StylePropsT>(
       position: 'absolute',
       [$theme.direction === 'rtl' ? 'right' : 'left']: '7px',
       top: 0,
-      height: '100%',
-      paddingBottom: 0,
+      height: `calc(100% + ${$theme.sizing.scale500})`,
```
nadiia: Could you add a comment on why `$theme.sizing.scale500` is used here, like where this specific value usage comes from?

foodforarabbit: There seems to be a gap in prod right now as well, I can fix it though.

As for the value, that was the height of the :after pseudo component that made up the original tail.

image

foodforarabbit: Hi, ok I fixed it to remove the gap. Used the calculations from the icon itself so that the spacing is more accurate now.

I also refactored the NumberContentTail to also get rid of the :after pseudo element in the same way, and adjusted the values to get rid of the gap.

foodforarabbit:
fixed gap

image

image

PR #2667, opened 2020-01-21
fix(form-control): add data attribute to form-control-caption

parkerdan:
#### Description

Adding data attribute to form-control caption as it is present on form-control label

#### Scope

- [ ] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


chasestarr:
We only add `data-baseweb` attributes to the 'root' element, used to signify a baseweb component is located on the page. Do you need this for testing? If so, I suggest using overrides to apply your own attribute like so https://baseweb.design/guides/understanding-overrides/#introducing-overrides

parkerdan:
> We only add `data-baseweb` attributes to the 'root' element, used to signify a baseweb component is located on the page. Do you need this for testing? If so, I suggest using overrides to apply your own attribute like so https://baseweb.design/guides/understanding-overrides/#introducing-overrides

In the current FormControl component, the data attribute is on the Label and Control Container.  The only missing part is the Caption

```
      <React.Fragment>
        {label && (
          <Label
            data-baseweb="form-control-label"
            htmlFor={onlyChildProps.id}
            {...sharedProps}
            {...getOverrideProps(LabelOverride)}
          >
            {typeof label === 'function' ? label(sharedProps) : label}
          </Label>
        )}
        <ControlContainer
          data-baseweb="form-control-container"
          {...sharedProps}
          {...getOverrideProps(ControlContainerOverride)}
        >
          {React.Children.map(children, (child, index) => {
            if (!child) return;

            const key = child.key || String(index);
            return React.cloneElement(child, {
              key,
              disabled:
                typeof onlyChildProps.disabled !== 'undefined'
                  ? onlyChildProps.disabled
                  : disabled,
              error:
                typeof onlyChildProps.error !== 'undefined'
                  ? onlyChildProps.error
                  : error,
              positive:
                typeof onlyChildProps.positive !== 'undefined'
                  ? onlyChildProps.positive
                  : positive,
            });
          })}
          {(caption || error || positive) && (
            <Caption {...sharedProps} {...getOverrideProps(CaptionOverride)}>
              {hint}
            </Caption>
          )}
        </ControlContainer>
      </React.Fragment>

```

chasestarr:
Oh interesting, I had not realized is was applied to all elements on the FormControl component. I'm fine with making this change for consistency's sake. Just need to get unit tests to pass

PR #2677, opened 2020-01-22
docs: monolithic engine for styletron

tajo:
Just testing the monolithic engine.


adenta:
What was the result here? Everything looked good?

I keep getting warnings for mixing shorthand properties

gergelyke:
Now you can opt-in to use the monolithic engine instead. You can find more info on it in the Styletron docs: https://www.styletron.org/getting-started#styletron-engine-monolithic

But essentially, it's a drop-in replacement for the atomic engine :)

PR #2681, opened 2020-01-23
Release v9.44.2

gergelyke:


gergelyke:
the vrt seems to be incorrect in this run, and previous runs on the master passed - going to merge it as is, and will keep an eye on it

Issue #2673, opened 2020-01-22
Make it possible to lock modal completely


houmark:
## Feature description

This may be possible already, but I did not find a way, so here we go.

I'd like to be able to lock a modal from being closed by clicking on the modal backdrop (and potentially even clicking or removing the X in such cases).

There are use cases where the user is doing very important work in a modal or you want them to fully acknowledge a change before they can move on, and for those cases, it feels a bit dangerous that they can just click the backdrop and have it close. I know, in a perfect world, you'd keep the state and have everything as they left it when they re-open the modal, but I still think a way to lock it up (even programmatically after opening it) would be nice. Clicking the right button would then unlock and close or whatever I'd like.

I'm still not fluent enough in TypeScript to contribute myself, but I see there's some vscode stuff coming in which may help me get started, but I wanted to vet the interest in this feature before doing more.

Or maybe it's already possible in some way I did not discover?

chasestarr:
Hmm I assumed that the `closeable` prop on `Modal` would do what you're looking for. https://codesandbox.io/s/modal-7khwy

This looks like a regression that needs to be fixed

houmark:
Yes, `closable` should do that but currently has no effect.

gergelyke:
I am working on this one

gergelyke:
@chasestarr @houmark referenced a PR above with the "fix" - it seems like none of us noticed the tiny typo in the codesandbox :)

houmark:
Reposting my comment here with additional thoughts:

That's so weird. I'm sure I tried this before and it didn't work. But I just tested and it works as intended with the right prop :)

Now that we are around, maybe deprecating this prop name and introduce either `isCloseable` or better `dismissOnClickOutside` which is the name used for Popover is a good idea to have more consistent naming?

After a second look at this, `closeable` will also remove the X in the upper right corner of the modal, which is not exactly my ideal expectation when using that prop. I would still like to have that close button/icon, so the user can click that fairly small X or a cancel button (if I include such a button) in the bottom of the modal. So with that in mind, the current prop should probably stay as is and then introduce `dismissOnClickOutside` which will keep the X but prevent closing on click on the backdrop? Best of both worlds and either can be used depending on the use case.

[...]

PR #2686, opened 2020-01-23
fix(modal): add warning for closeable prop

gergelyke:
Closes https://github.com/uber/baseweb/issues/2673

It was only a typo :) Added a warning for this, in case others run into it too. Also, a VRT test case

Screen Shot 2020-01-23 at 8 53 51 AM


houmark:
That's so weird. I'm sure I tried this before and it didn't work. But I just tested and it works as intended with the right prop :)

Now that we are around, maybe deprecating this prop name and introduce either `isCloseable` or better `dismissOnClickOutside` which is the name used for `Popover` is a good idea to have more consistent naming?

src/modal/modal.js
```diff
@@ -72,6 +72,12 @@ class Modal extends React.Component<ModalPropsT, ModalStateT> {
         Then pass backdrop overrides to DialogContainer instead. Tha will help you with
         the next major version upgrade.`);
       }
+      // $FlowFixMe: flow complains that this prop doesn't exist
+      if (this.props.closable) {
```
chasestarr: this should have a `__DEV__` guard

gergelyke: it's not visible in this diff, but it's part of a bigger __DEV__ condition: https://github.com/uber/baseweb/blob/4f21b3af2aa8bf793b496d515ede4be1fe36e2b1/src/modal/modal.js#L61-L80

PR #2711, opened 2020-01-24
fix(MaskedInput): pass style overrides, add datepicker example with readonly input attr

nadiia:


src/datepicker/datepicker.js
```diff
@@ -221,6 +223,12 @@ export default class Datepicker extends React.Component<
     }
   }
 
+  setAnyHover = (val: boolean): null => {
+    console.log('ANY HOVER IS SET to', val);
```
tajo:
```suggestion
```

src/datepicker/datepicker.js
```diff
@@ -284,6 +292,28 @@ export default class Datepicker extends React.Component<
                   mask={this.getMask()}
                   required={this.props.required}
                   clearable={this.props.clearable}
+                  overrides={{
+                    Input: {
+                      style: () => {
+                        console.log('OVERRIDE STYLE IS CALLED');
```
tajo:
```suggestion
```

tajo:
It seems strange to update the internal state as a part of CSS media query. The styles override could be called multiple times so it should not have side-effects?

nadiia:
This PR was for a test only, to test in "staging". Marked as WIP.

src/input/masked-input.js
```diff
@@ -71,13 +71,16 @@ export default function MaskedInput({
...
+        typeof inputOverride.style === 'function'
+          ? inputOverride.style
+          : {
+              ...(inputOverride.style || {}),
+            },
     },
     ...restOverrides,
   };
-
+  // $FlowFixMe
```
chasestarr: the code below will give no flow errors and also allows for component overrides, which was not supported before

```js

export default function MaskedInput({
  mask,
  maskChar,
  overrides: {Input: inputOverride, ...restOverrides} = {},
  ...restProps
}: MaskedInputPropsT) {
  let componentOverride = MaskOverride;
  let propsOverride = {};
  let styleOverride = {};

  if (typeof inputOverride === 'function') {
    componentOverride = inputOverride;
  } else if (typeof inputOverride === 'object') {
    componentOverride = inputOverride.component;
    propsOverride = inputOverride.props;
    styleOverride = inputOverride.style;
  }

  const nextOverrides = {
    Input: {
      component: componentOverride,
      props: {
        mask,
        maskChar,
        ...propsOverride,
      },
      style: styleOverride,
    },
    ...restOverrides,
  };

  return <Input {...restProps} overrides={nextOverrides} />;
}
```

PR #2714, opened 2020-01-25
bugfix(progress-bar): respect successValue in infinite progress mode

afzalsayed96:
#### Description

In `infinite` mode if the user passes `value` and `successValue` prop then we stop loading animation on 100% with this change

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/src/progress-bar/styled-components.js
+++ b/src/progress-bar/styled-components.js
@@ -41,27 +41,34 @@ export const BarProgress = styled<StylePropsT>('div', props => {
   const width = `${($value / $successValue) * 100}%`;
   const borderRadius = borders.useRoundedCorners ? sizing.scale0 : 0;
 
-  const animationStyles = $infinite
+  const animationStyles =
+    $value < $successValue
+      ? {
+          animationDuration: '2.1s',
+          animationIterationCount: 'infinite',
+          animationTimingFunction: 'cubic-bezier(0.65, 0.815, 0.735, 0.395)',
+          animationName: {
+            '0%': {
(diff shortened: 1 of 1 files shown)


gergelyke:
@afzalsayed96 I've tried reproducing this on the doc site, but I couldn't: https://bit.ly/2U4c4Uz

can you put together a codesandbox that reproduces the issue you are seeing?

afzalsayed96:
Sorry. What I mean is we **should** stop loading if user has provided `value` and `successValue` where `value >= successValue` IMO. This PR achieves said behavior.

Currently it's the reverse.

gergelyke:
Oh I see - while the current behavior may not be the most intuitive, I am hesitant to merge this, as it would be a breaking change in terms of the functionality. For now, I would hold off on merging this. What do you all think?

sandgraham:
I think I agree that we should keep this as is. I like how simple the current API is. If `infinite` is true you will see the infinite loading state. You can achieve the same effect with current implementation by toggling `infinite` off, either once you know `successValue` or once you know you are done.

afzalsayed96:
Ah! You're right.
Actually, there is a trade-off in this scenario. The longer the unintuitive behavior stays, more people get accustomed to it. Hence we affect more people by releasing later.
I'm interested to know what others think. In any case I would be fine with what you deem fit.

gergelyke:
Okay, closing for now, but we can re-open if anything feels strongly about this

PR #2715, opened 2020-01-25
chore(scripts): run scripts with node command

afzalsayed96:
#### Description

Running scripts with `./script/name` format doesn't work in windows as it doesn't support shebangs. This PR makes all scripts more explicit (and consistent) by calling them with `node` command.

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


gergelyke:
thank you @afzalsayed96

sandgraham:
VRTs should not have updated so no need to merge that PR in. Re-running now.

PR #2716, opened 2020-01-27
feat(tree-view): TreeLabel override, renderAll, singleExpanded

### LINKED ISSUE #2665: [Treeview] Ability to customize open/close overrides based on item
## Feature description
I'm looking to create a TreeView component where each node could be customized. Today this is not possible without some pretty ugly hacks.

I can see two API Designs to make this work.

1 - The current API renders everything via initial data. This approach can work; however, each node does not have access to each entry of the initial data as props.  Each entry could be used to override and create each variants from the overrides>component api.

[Codesandbox demo inflexible api](https://codesandbox.io/s/overrides-usage-2qkpo)

2 - Another API could follow Material design; the API design of Material TreeView is a more declarative api similar to Base Accordion:

[MaterialUI's Custom Tree View Example](https://codesandbox.io/s/ujp8m)

foodforarabbit:
Fixes #2665 

#### Description

Make the entire label (e.g. all the node content excluding children) overrideable while still allowing the existing overrides (TreeItemContent, IconContainer, ExpandIcon, CollapseIcon) to work with the default TreeLabel

added renderAll prop which allows all content to be rendered for seo purposes (regadless of isExpanded state)

added singleExpanded prop which will set all sibling nodes to isExpanded=false when setting one node to isExpanded=true
- this will ONLY trigger when a node is expanded via the StatefulContainer toggle
- if you want it to apply to the data before an expansion is triggered, it will have to be done manually

- added e2e tests, flow types, typescript types, etc...
#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change


gergelyke:
re the `renderAll ` - I am wondering, what would be the drawback of making this the default behavior here, and across all components? cc @chasestarr @nadiia @tajo @sandgraham

chasestarr:
The drawback would be wasted reconciliation ops if your application does not care about server-rendering that content. It would be misleading is a developer places an expensive operation in popover content expecting it to only be run when opened, but it ends up running more

gergelyke:
> The drawback would be wasted reconciliation ops if your application does not care about server-rendering that content. It would be misleading if a developer places an expensive operation in popover content expecting it to only be run when opened, but it ends up running more

thank you Chase, that makes perfect sense

documentation-site/examples/tree-view/label-overrides.js
```diff
@@ -0,0 +1,123 @@
...
+      case 1:
+        return 'h1';
+      case 2:
+        return 'h2';
+      case 3:
+      default:
+        return 'h3';
+    }
+  })(node.depth);
+  // $FlowFixMe
```
chasestarr: is it possible to handle the flow error in this documentation example

foodforarabbit: oops fixed!

foodforarabbit:
hmm dunno why tests keep failing, what does the error say? when I click onto the failing tests it says I need to sign in to something. 

I've pulled this and run the tests locally and they passed fine (had a few timeouts on the e2e tests but that was because my computer is slow haha, ran them again and they passed)

sandgraham:
@foodforarabbit Let's rerun and see how it goes - some of the e2e have been a bit flakey.

PR #2738, opened 2020-01-28
feat(data-table): numerical column accepts a format function

### LINKED ISSUE #2230: [DataTable] Numerical column accepts a format function
For example, suppose I want to render a number formatted to a particular currency that is not US dollar

joginder07:
Resolves #2230 

#### Description
Updated type of format option. 
Added test case.
#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change


src/data-table/column-numerical.js
```diff
@@ -33,7 +33,7 @@ type NumericalOperations =
 
 type OptionsT = {|
   filterable?: boolean,
-  format?: NumericalFormats,
+  format?: NumericalFormats | ((value: number) => string | number),
```
chasestarr: What's the use-case for the result to be a number?

```suggestion
  format?: NumericalFormats | ((value: number) => string),
```

src/data-table/column-numerical.js
```diff
@@ -428,7 +428,13 @@ function NumericalFilter(props) {
 
 const NumericalCell = React.forwardRef<_, HTMLDivElement>((props, ref) => {
   const [css, theme] = useStyletron();
-
+  const formattedValue =
```
chasestarr: This change can be encapsulated in the `format` function on L63 with an early return if `props.format` is a function

src/data-table/column-numerical.js
joginder07: The reason behind keeping it as a number was that a user can apply some mathematical transformation which is not supported by NUMERICAL_FORMATS, example Math.ceil.

Issue #2737, opened 2020-01-28
Button component uses shorthand css property for setting border width.


haeguri:
## Current Behavior

I using `Button` component in this project. And I set `border*Width` longhand css properties on my button component via `overrides` prop. But I saw this warning.

> Styles `{"borderWidth":"0px"}` and `{"borderTopWidth":"1px"}` in object yielding class "ae ak b7 fn d2 b9 dd du dv d4 bz b5 dw dm dx ex dz e0 e1 cu cv am e2 d7 d8 d5 d6 e7 e8 e9 ea bj b4 eb ec ed g9 gd gq" may result in unexpected behavior. Mixing shorthand and longhand properties within the same style object is unsupported with atomic rendering.

I'm already used longhand css property for setting border width.. So I read `Button` component source in this repository, I found out the suspect part.

https://github.com/uber/baseweb/blob/15886c506fa2f35ad5351e83968a8a67a5463110/src/button/styled-components.js#L18

This code is setting borderWidth using shorthand method. 

## Expected Behavior

I thought should use `border*Width: 0` longhand css properties, instead of `borderWidth: 0` shorthand css property. Then expects that warning is gone.

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v9.9.0  |
| React   | v16.12.0  |
| browser |  chrome 79.0.3945.130  |

- [x] I have searched the [issues](https://github.com/uber/baseweb/issues) of this repository and believe that this is not a duplicate.

sivaraj-v:
```
<Button
    className={css({
      ':hover': {
        backgroundColor: "#efefef",
        color: "#333"
      }
    })}
      overrides={{
        BaseButton: {
          style: ({ $theme }) => {
            return {
              backgroundColor: 'white',
              color: '#333',
              width: '100%',
              borderWidth: "1px",
              borderStyle: 'solid',
            };
          }
        }
      }}
      onClick={() => props.callback(props.provider)}>
      <img
        style={{
          padding: '0 14px',
          marginLeft: '-14px'
        }}
        src="./icons/google-icon.svg"
      />{' '}
      Continue with Google
    </Button>
```


Getting warning Messages as :
[...]

PR #2741, opened 2020-01-28
fix(button): longhand for border width

gergelyke:
Closes https://github.com/uber/baseweb/issues/2737


src/button/styled-components.js
```diff
@@ -74,7 +77,10 @@ export const LoadingSpinner = styled<SharedStylePropsT>(
       borderBottomRightRadius: '50%',
       borderBottomLeftRadius: '50%',
       borderStyle: 'solid',
```
nadiia: Would you mind changing `borderStyle` as well?

houmark:
Isn't this potentially (design) breaking for people having changed the shorthand property in overrides?

PR #2742, opened 2020-01-28
Modal prop types should be optional

akre54:
Causing flow errors if you don't pass an `animate` prop, e.g.

#### Description

Not sure if there's a better overall pattern for this... Maybe the Component in modal/modal.js can take `$Shape<ModalPropsT>` instead?

#### Scope

- [X] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/src/modal/types.js
+++ b/src/modal/types.js
@@ -32,21 +32,21 @@ export type ElementRefT = {current: React.ElementRef<any> | null};
 // Props shared by all flavors of modal
 export type ModalPropsT = {
   /** Sets whether the Modal should be displayed by easing in and out */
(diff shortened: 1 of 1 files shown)


gergelyke:
hi @akre54 - can you help us what flow version and base version are you using? 

Currently, all base components follow this pattern: https://flow.org/en/docs/react/components/#toc-using-default-props

gergelyke:
Also, can you create a code example that reproduces your issues?

gergelyke:
closing for now, but let's discuss it here if you still have the issue

Issue #2418, opened 2019-12-04
Datepicker: years before 2000 cause several issues


tconkling:
https://codesandbox.io/s/datepicker-tkn6r

## Current Behavior

If I set the current value or the minDate of a Datepicker to a date before Y2K, I see the following behaviors:

(See the [linked codesandbox](https://codesandbox.io/s/datepicker-tkn6r) for an example that behaves problematically)

- I can't use the left-arrow button to select months _before_ Jan 2000 (despite my minDate being earlier than that):

![disabled_left_arrow](https://user-images.githubusercontent.com/709022/70104094-e0245300-15f1-11ea-879d-b33f2c874d6e.jpg)

- I _can_ use the right-arrow button to select months _after_ my maxDate:
![right_arrow_should_disable](https://user-images.githubusercontent.com/709022/70104123-f3372300-15f1-11ea-8cb9-6176cd724f4f.jpg)

- The month dropdown doesn't contain any values:
![no_results](https://user-images.githubusercontent.com/709022/70104089-da2e7200-15f1-11ea-857b-0b447fab3a77.jpg)

## Expected Behavior

- Should be able to select dates prior to 2000, if they're within my date range
- Should not be able to select dates greater than my date range, even if my minDate is prior to 2000
- I should see values in the month dropdown, even if my minDate is prior to 2000

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v9.25.0  |
| React   |    16.8.6     |
| browser |    Chrome Version 78.0.3904.108 (Official Build) (64-bit)     |

- [x] I have searched the [issues](https://github.com/uber/baseweb/issues) of this repository and believe that this is not a duplicate.

gergelyke:
@tconkling can you help us look into this?

kantuni:
Hi @gergelyke,

I'll try send you the PR tomorrow.

kantuni:
Here you go: https://github.com/uber/baseweb/pull/2430

PR #2743, opened 2020-01-28
fix(datepicker): properly handle min and max values in calendar header

gergelyke:
Closes https://github.com/uber/baseweb/issues/2418


.eslintignore
```diff
@@ -11,3 +11,4 @@ documentation-site/posts.js
 documentation-site/cheat-sheet.js
 public
 src/codemods/__testfixtures__
+out
```
chasestarr: intentional?

```suggestion
```

src/datepicker/__tests__/stateful-calendar.scenario.js
```diff
@@ -13,5 +13,15 @@ import {StatefulCalendar} from '../index.js';
 export const name = 'stateful-calendar';
 
 export const component = () => (
-  <StatefulCalendar highlightedDate={new Date('2019-02-14T10:00:00Z')} />
+  <>
+    <StatefulCalendar highlightedDate={new Date('2019-02-14T10:00:00Z')} />
+    <br />
+    <StatefulCalendar
```
chasestarr: Based on the disabled logic change and the min/max date provided here, I would expect the left/right arrow buttons in these new VRTs to be disabled as well. They look enabled in the VRTs.

.eslintignore
gergelyke: yeah, it was intentional, as eslint fails on the out folder of the extension package

src/datepicker/__tests__/stateful-calendar.scenario.js
gergelyke: why is that? the `maxDate` is in 2001, the `minDate` is in 1999, and the initial value is in 2000 - so imho, the arrows should be enabled. Am I missing something?

On the master branch, the bug was that you couldn't paginate to 2000, because the prop was not taken into account. This was added to make sure it's enabled.

chasestarr: Oh sorry, I overlooked the years and only compared the month/day values

PR #2747, opened 2020-01-28
fix(vsce): add compile step before publishing

gergelyke:


packages/baseweb-vscode-extension/package.json
```diff
@@ -13,6 +13,26 @@
     "type": "git",
     "url": "https://github.com/uber/baseweb"
   },
+  "scripts": {
```
gergelyke: moved these up, so they are above the generated parts

PR #2765, opened 2020-01-29
feat(ci): migrate link checker script to node with some improvements

sandgraham:
Ports `link-checker.sh` to a node script.

Also resolves preview deployment URLs from ZEIT API since we can't always guess the URL. For instance, a long branch/author name will result in a truncated URL with a hash in it. Their documentation doesn't explain how that hash is generated. Regardless, using the API should be a bit safer going forward.


chasestarr:
may want to remove the shell script

scripts/link-checker.js
```diff
@@ -0,0 +1,120 @@
...
+      `yarn blc ${url} -ro --exclude components/avatar --exclude example.com --exclude github.com`,
+      {stdio: 'inherit'},
+    );
+  } else {
+    console.log('');
+    throw new Error('Maximum number of attempts reached.');
+  }
+}
+
+async function getDeploymentURL() {
```
gergelyke:
```suggestion
export async function getDeploymentURL() {
```

I'd like to reuse this for the documentation site vrt story

gergelyke: ops, it won't work this way - can you split this out to a module, so it can be reused?

gergelyke: or we can simply merge as is, and happy to do that in my pr

chasestarr: what is your idea for that?

gergelyke: I'd like to get the zeit deployment url to create a snapshot of the ProgressStep component (that seems to be one that doesn't change a lot), and one one a blogpost

chasestarr: I think running VRT tests across the network may be asking for bugs and heartache

gergelyke: That's a good call-out - I think another option we have is to build the doc site in the CI workflow. Previously we've removed that because it slowed down the pipeline considerably. 

I guess we have two options here

* try out and see if over-the-network puppeteer screenshots would be flaky
* try out building the doc site locally again

which one would you try first? did I miss a third one?

chasestarr: Is the requirement: VRT tests must be run on examples published to the docs site?

Could docs site builds be avoided by including the examples into the storybook/e2e server?

PR #2778, opened 2020-01-31
feat(scenario): refactor to remove name export

chasestarr:
Scenarios are annoying because they require inherent knowledge of the two named exports `name` and `component`. I've customized the scenario loader to no longer need a name. It now keys off of the file name. 
Now that it's a simple default export, scenario files align closer with docs site examples. There's likely a reasonable way to include examples in VRT tests if we felt strongly about doing so


sandgraham:
That VRT seems to have flaked again? Not sure how its doing that if the screen size doesn't change and there are no interactions.

chasestarr:
i just pushed a commit to turn off the flakey test for now

sandgraham:
Oh there is a `skip` option that config will read too if you want.

PR #2783, opened 2020-02-01
feat(theme): add borderAccentLight

sandgraham:
Updates `borderAccent` and adds a new semantic color token with the old value: `borderAccentLight`.


tajo:
Should be also added to the extension cheat sheet.

sandgraham:
@tajo Is adding to cheatsheet just adding it to the `App.tsx` file like I did?

tajo:
> @tajo Is adding to cheatsheet just adding it to the `App.tsx` file like I did?

Yes

PR #2790, opened 2020-02-04
docs(theming): always show dark theme & restyling

sandgraham:
A re-styling of the theme properties section.

- Include light & dark theme comparisons where appropriate.
- Show path to theme property.
- Some drying up and refactoring.

### DIFF
--- a/documentation-site/components/theming/borders.js
+++ b/documentation-site/components/theming/borders.js
@@ -15,7 +15,8 @@ export function Border({value}: {value: string}) {
   return (
     <ThemeComparison
       value={value}
-      renderSwatch={({mode, commonStyles}) => (
+      concern="borders"
+      renderBox={({mode, commonStyles}) => (
         <div
           className={css({
             ...commonStyles,
@@ -23,7 +24,7 @@ export function Border({value}: {value: string}) {
           })}
         ></div>
       )}
-      renderValues={({mode}) => (
+      renderValue={({mode}) => (
         <React.Fragment>
           <div>{mode.borders[value].borderStyle}</div>
           <div>{mode.borders[value].borderWidth}</div>
--- a/documentation-site/components/theming/borders.js
+++ b/documentation-site/components/theming/borders.js
@@ -8,27 +8,60 @@ LICENSE file in the root directory of this source tree.
 
 import * as React from 'react';
 import {useStyletron} from 'baseui';
-import {Property} from './common.js';
+import {Title, Value, Property} from './common.js';
+import {LightTheme, DarkTheme} from 'baseui/themes';
+
+function Swatch({value, mode, left = false}) {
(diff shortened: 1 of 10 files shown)


documentation-site/components/theming/borders.js
```diff
@@ -8,42 +8,50 @@ LICENSE file in the root directory of this source tree.
...
-      ></div>
-    </Property>
+    <PropertyCompareTheme
+      name={name}
+      concern="borders"
+      renderBox={({mode, commonStyles}) => (
+        <div
+          className={css({
+            ...commonStyles,
+            ...mode.borders[name],
```
nadiia: `mode` seems to be the `theme` obj here, correct? Why don't we keep the `theme` naming here to avoid any confusion?

documentation-site/components/theming/colors.js
```diff
@@ -8,42 +8,38 @@ LICENSE file in the root directory of this source tree.
...
+          className={css({
+            ...commonStyles,
+            ...mode.borders.border600,
+            backgroundColor: mode.colors[name],
+          })}
+        ></div>
+      )}
+      renderValue={({mode}) => (
+        <React.Fragment>
+          <div>{mode.colors[name]}</div>
```
nadiia: Same here regarding `mode` vs `theme`.

documentation-site/components/theming/common.js
```diff
@@ -13,52 +13,116 @@ LICENSE file in the root directory of this source tree.
...
+    fontFamily: monospaceFontFamily,
+    fontStyle: 'italic',
+    color: $theme.colors.contentSecondary,
+  };
+});
+
 export const Title = styled<{}>('div', ({$theme}) => {
   return {
-    ...$theme.typography.HeadingXSmall,
+    ...$theme.typography.ParagraphLarge,
```
nadiia: Since it's a title it makes sense to use one of the Heading components.

documentation-site/components/theming/borders.js
sandgraham: I need to distinguish between the theme for the doc site that is provided by `useStyletron` and `styled` and the fixed theme object that I import and pass around (`LightTheme`, `DarkTheme`). In some cases they are both in the same scope so I need two different names. Open to suggestions on what to call them though.

sandgraham: `docTheme` and `previewTheme` maybe?

nadiia: `previewTheme` sounds good 👍

sandgraham: cool - updated

documentation-site/components/theming/common.js
sandgraham: changed it back to a heading

PR #2794, opened 2020-02-04
fix(timepicker): handle null value prop

nadiia:


src/timepicker/timepicker.js
```diff
@@ -134,7 +134,7 @@ class TimePicker extends React.Component<TimePickerPropsT, TimePickerStateT> {
           ? undefined
           : {id: closestStep, label: secondsToLabel(closestStep)},
       });
-      this.handleChange(closestStep);
+      if (this.props.value) this.handleChange(closestStep);
```
chasestarr: what should happen if there is no value, but nullable is false?

nadiia: Good point! Added an additional condition for it.

src/timepicker/__tests__/time-picker.scenario.js
```diff
@@ -22,26 +22,28 @@ const overrides = {
...
   return (
     <React.Fragment>
       <TimePicker
-        format={format}
-        step={step}
         value={time}
-        onChange={setTime}
+        onChange={time => {
+          setTime(time);
+          onChange();
```
chasestarr: what does this do?

nadiia: Since it's a reusable component so I added it for testing with different console logs on separate instances. Just kept it there with a default noop.

Issue #2685, opened 2020-01-23
Popover memory leak


zsobral:
https://codesandbox.io/s/statefulpopover-gpkjx

## Current Behavior

the browser freeze after some clicks. I need to open chrome task manager and end the process. When checking the task manager the memory usage keeps increasing

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v9.43.0 |
| React   | 16.12.0 |
| browser | chrome |

- [x] I have searched the [issues](https://github.com/uber/baseweb/issues) of this repository and believe that this is not a duplicate.

gergelyke:
I suspect, this is an issue with codesandbox, as I couldn't reproduce here: https://baseweb.design/components/popover/

Closing for now, but please let us know if you can repro outside of codesandboc

zsobral:
sorry, forgot to tell that is caused when using `triggerType="hover"`

zsobral:
sorry, forgot to tell that is caused when using a StatefulPopover with content StatefulMenu and the  triggerType is hover

gergelyke:
@zsobral could you reproduce the issue outside of codesandbox too?

zsobral:
Yes, I'm using with create-react-app and chrome

Em qui, 23 de jan de 2020 18:28, Gergely Nemeth <notifications@github.com>
escreveu:

gergelyke:
thank you

PR #2797, opened 2020-02-05
fix(popover): avoid infinite focus loop

chasestarr:
Fixes: https://github.com/uber/baseweb/issues/2685

#### Description

infinite loop between focusing the content and the anchor element
on anchor hover, content focuses if it contains a tabbable element this is why the scenario includes tabIndex prop
on anchor click, the popover closes because of onClickOutside rules
anchor click focuses it, thus opening the popover and focusing popover content
content focus causes the anchor to blur, which causes the popover to close
because of how the focus lock was configured it refocused the anchor which restarts the loop

#### Scope

- [x] Patch: Bug Fix


houmark:
I was about to create a test case on this one. In Chrome we see the browser scrolling to the top when clicking a button that opens a popover since a few releases ago. Seems that Safari is handling it better.

chasestarr:
> I was about to create a test case on this one. In Chrome we see the browser scrolling to the top when clicking a button that opens a popover since a few releases ago. Seems that Safari is handling it better.

Focus was added in [this pr](https://github.com/uber/baseweb/pull/2650), which probably should have defaulted to off (going to consider making this change in a separate pr).  Scrolling to the top sounds its own issue, or are you noticing a similar loop?

houmark:
Not exactly sure, I just received a user report on the behavior and then I went to test in Safari and didn't see an issue.

Then I tested in Chrome and it does it pretty consistently in the use case we have (flex grid table, with an expand that opens a Grid). Anyways, my plan is to see if this PR does any changes to that behavior in Chrome and if not, then I will try to make a Code Sandbox with a simplified reproducible test case.

chasestarr:
this could be some relevant info to you if it's only reproducible in chrome https://github.com/theKashey/react-focus-lock#focusing-in-osx-safarifirefox-is-strange

houmark:
So `focusLock={false}` seems to correct the bad behavior in our use case. Still testing, but this may be an indicator of a bug with the feature or just incorrect use which happens due to being default to `true`.

PR #2802, opened 2020-02-06
feat(vscode-extension): add theme color samples in code

nadiia:


tajo:
It seems that there is some lag between updating settings and updating the editor's visuals. For example, when I am changing the `Editor: Font Size` it re-renders immediately.

I can pretty easily get into a state when the coloring spreads to surrounding characters: 

Screen Shot 2020-02-06 at 4 06 45 PM
Screen Shot 2020-02-06 at 4 01 20 PM

all I need to do is just keep typing.

`yarn compile` throws a lot of TS errors, that might break the deploy task.

packages/baseweb-vscode-extension/ext-src/coloring.ts
```diff
@@ -0,0 +1,100 @@
...
+    const workspaceConfig = vscode.workspace.getConfiguration('baseweb');
+    const isColoringOn = workspaceConfig.get('theme.coloring.enabled');
+    if (!isColoringOn) {
+      return;
+    }
+    const themeMode = workspaceConfig.get('theme.coloring.source');
+    const theme = themeMode === 'Light' ? LightTheme : DarkTheme;
+    const coloringStyle = workspaceConfig.get('theme.coloring.style');
+
+    const regEx = /colors\.\w+/g;
```
tajo: this regexp is probably too permissive? (problem with coloring neighbor characters)

packages/baseweb-vscode-extension/ext-src/coloring.ts
```diff
@@ -0,0 +1,100 @@
...
+      const startPos = activeEditor.document.positionAt(match.index + 7);
+      const endPos = activeEditor.document.positionAt(
+        match.index + match[0].length,
+      );
+      const colorVal = theme.colors[match[0].slice(7)] || 'transparent';
+      const decoration = {
+        range: new vscode.Range(startPos, endPos),
+        hoverMessage: `${colorVal} | ${themeMode}Theme`,
+      };
+      activeEditor.setDecorations(
```
tajo: should we rather skip coloring when there is no theme match instead of `"transparent"`, might be more efficient

tajo:
Also, `cheat sheet` stopped working?

nadiia:
> Also, `cheat sheet` stopped working?

The cheat sheet works actually ("on my machine"  🤣 ). Would you mind showing the repro? Do you see any errors?

packages/baseweb-vscode-extension/ext-src/coloring.ts
nadiia: Improved the regexp ^

packages/baseweb-vscode-extension/ext-src/coloring.ts
nadiia: Done, see a condition above ^.

nadiia:
Also reduced the timeout to `200`.

tajo:
It works great!  🎉

My only nit:

Screen Shot 2020-02-13 at 1 14 47 PM

It would be nice to detect when to use dark vs light text color based on the background. The default `underline` coloring is probably the best option but if somebody wants to use `background`, a lot of values make the code unreadable.

nadiia:
@tajo Would you be ok with merging it as is and looking into font colors as a follow-up PR?

nadiia:
@tajo added font color contrast calculation ^

PR #2806, opened 2020-02-06
fix(slider): improve focus-ring styles

tajo:
Going to piecemeal the focus-ring story a bit. I'm interested into your early feedback before I roll out this approach across other components. It relies on `:focus-visible` polyfill which is a piece of global JS that tracks user's interactions. The CSS selector is not supported outside of Firefox at this moment.

## Why do we need `:focus-visible` functionality?

We want to display focus ring only when an element is focused by keyboard. If mouse is used, we don't want to apply any new styles (there are some exceptions like inputs). This is good for both a11y and sleek design.

## Before

(applied when using mouse or keyboard)

Screen Shot 2020-02-06 at 3 29 28 PM

Screen Shot 2020-02-06 at 3 30 40 PM

## After 

(applied only with keyboard)

Screen Shot 2020-02-06 at 3 21 53 PM

Screen Shot 2020-02-06 at 3 21 37 PM

## Notes

- I moved the polyfill initialization into `BaseProvider` so we don't have to implement `componentDidMount` logic in every single component
- material-ui implementation is using React Hooks but we still use class components, so that's the only modification
- slider is a slightly special case since we also have to track which specific thumb is focused, for single element components the implementation will be even simpler


src/layer/layers-manager.js
```diff
@@ -56,7 +60,7 @@ export default class LayersManager extends React.Component<LayersManagerPropsT>
                 zIndex: this.props.zIndex,
               }}
             >
-              <AppContainer {...appContainerProps}>
+              <AppContainer {...appContainerProps} ref={this.containerRef}>
```
chasestarr: This will only apply to baseui elements that are not portaled right? Can we move it up a level or apply the polyfill to the layers container as well?

src/slider/styled-components.js
```diff
@@ -102,6 +102,9 @@ export const Thumb = styled<StylePropsT>('div', props => {
     borderBottomRightRadius: isLeft ? '1px' : '4px',
     backgroundColor: $theme.colors.mono100,
     display: 'flex',
+    outline: props.$isFocusVisible
```
chasestarr: does `$isFocusVisible` need to be added to `StylePropsT`?

src/slider/slider.js
```diff
@@ -135,6 +158,10 @@ class Slider extends React.Component<PropsT> {
                   {...props}
                   $thumbIndex={index}
                   $isDragged={isDragged}
+                  $isFocusVisible={
```
nadiia: Would it make sense to pass the `$isFocusVisible` as part of the `sharedProps`?

src/utils/focusVisible.js
```diff
@@ -0,0 +1,149 @@
+/*
+Copyright (c) 2018-2020 Uber Technologies, Inc.
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+
+// @flow
+/* global window */
```
nadiia: We should wrap `window` usages in the `if (__BROWSER__) {` instead. Those will be needed on a client-side only right?

tajo: functions using `window` are called only from componentDidMount or onFocus / onBlur. What would `if (__BROWSER__) {` achieve? If anything it's probably better to see server errors if we use those incorrectly.

src/layer/layers-manager.js
tajo: so this ref doesn't need to be a parent of component that uses `focusVisible` logic, we just need to get `node.ownerDocument` to attach some global event listeners like `mousedown`... so I was looking for a node that always gets mounted and this should be it?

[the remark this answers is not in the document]
tajo: AppContainer div is always rendered, right? If so this should be fine - added more details under the Chase's comment.

src/slider/slider.js
```diff
@@ -98,10 +100,31 @@ class Slider extends React.Component<PropsT> {
       StyledTickBar,
     );
     const sharedProps = this.getSharedProps();
+
+    const handleFocus = event => {
+      if (isFocusVisible(event)) {
+        this.setState({focusVisible: true});
+      }
+      const index = event.target.parentNode.firstChild === event.target ? 0 : 1;
+      this.setState({focusedThumbIndex: index});
```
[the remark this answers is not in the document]
tajo: You mean moving those into `stateful-slider-container`? We probably want to keep focus working even in the stateless slider. We are doing the same thing for `Input` I think.

src/layer/layers-manager.js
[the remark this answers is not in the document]
nadiia: I see, then shouldn't be a problem.

sandgraham:
Perhaps unorthodox sentiment here: I feel like the the focus ring color should be blue - its kind of hard to notice the black/white.

tajo:
> Perhaps unorthodox sentiment here: I feel like the the focus ring color should be blue - its kind of hard to notice the black/white.

I might agree. How would you feel about using `accent` and boxShadow (to preserve the border radius).

Screen Shot 2020-02-10 at 11 08 44 AM
Screen Shot 2020-02-10 at 11 08 34 AM

PR #2807, opened 2020-02-06
fix(popover): set bgColor to backgroundTertiary

nadiia:


sandgraham:
Yeah, looks nice in dark theme, but kind of weird in light theme.

nadiia:
I'd say it looks slightly weird on very small surfaces since it creates an illusion that it' just a shadow spot. Also, the `No results` phone input dropdown's background probably need to have it's own bg or the color contrast tuned 🤔 I'll reach out to designers regarding it.

nadiia:
Got 👍 from designers on this one. They may revisit tho.

houmark:
To avoid having to do triple overrides for the background if another background color like white is wanted for, this should have its own theme primitive IMHO.

We had to do this to get a white background on some popovers where that looks better in our UI:

```
{
  Arrow: { style: { backgroundColor: 'white' } },
  Inner: { style: { backgroundColor: 'white' } },
  Body: { style: { backgroundColor: 'white' } },
}
```

Alternatively `Arrow` and `Inner` could inherit the background color from Body unless overridden.

PR #2817, opened 2020-02-07
fix(popover): change popover autofocus false default

chasestarr:
Making the popover autofocus by default is kind of a dangerous change. Let's set it to false by default


nadiia:
There is also yard config that needs a change of the default value of the `focusLock` prop.

PR #2821, opened 2020-02-07
fix(button): properly position loading spinner

lessenadam:
#### Description

Makes the baseButton position relative. It is a fix for the issue as seen in this [codesandbox](https://codesandbox.io/s/button-1bkow), where the absolutely positioned spinner moves as the div scrolls.

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


chasestarr:
Can you please test the `table-grid-jobs` scenario? This exact property was edited somewhat recently: https://github.com/uber/baseweb/pull/1489/files#diff-796e653f3f8d3e6934873c730a56706c

chasestarr:
Below is an image that better illustrates the issue with this change. Notice how the `chevron-right` is above the header row
Screen Shot 2020-02-10 at 1 10 34 PM

lessenadam:
@chasestarr I see the repro of that issue in `table-grid-jobs`, however the CSS to build that table looks pretty obscure, and not sure how easy it will be to root cause why `position: relative` would cause that behavior. I think there are 2 potential options:
1) Ask the person who implemented the table to investigate the issue, as the CSS would be pretty hard to unpack
2) Revisit the absolutely positioned loading spinner inside the button to make it statically positioned as well

chasestarr:
`relative` creates a new [stacking context ](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context) so that's why we are getting the unpredictable layering. I wrote the table code which illustrates a standard [sticky position](https://developer.mozilla.org/en-US/docs/Web/CSS/position#Sticky_positioning) use-case. I personally find the need to use relative kind of problematic since it can break applications like the table. I suspect the fix is likely related to the absolute positioning on the spinner element

lessenadam:
Non-breaking table
image

Button width long content:
image

lessenadam:
Hey @chasestarr can you take a look at the proposed changes? I took 2 screenshots showing the UI working as expected. I'm not sure if we can get rid of `<LoadingSpinnerContainer {...loadingSpinnerContainerProps}>` altogether since I'm not sure what it's doing in all cases.

houmark:
I'm seeing the height of the button changing to be less tall when the loading spinner is running. This may be due to local overrides but I don't believe we have any such. I'll test more and create an issue if this continue happening.

Screen Recording 2020-02-11 at 05 09 PM

chasestarr:
Hmm I would have expected to see changes in our VRT tests, but I just looked at the react-view editor on the docs site and it does resize if `size != default`

lessenadam:
This did make me realize that there is an issue with this approach if the button 's child has a height greater than 20px. I tried to solve this with just CSS by decreasing the height of the child div by 20px using calc for example, but it's not really working. I'm not sure what the best way is to maintain the height of the button if its height is increased by the content of the button 🤔

chasestarr:
I'll fix it tomorrow, thanks for catching it @houmark

PR #2823, opened 2020-02-09
feat: rtl support for data table

### LINKED ISSUE #2589: [DataTable] support RTL
## Feature description

add RTL support for DataTable

rohitb4:
Fixes #2589 

#### Description

Support for RTL in data table. Many of the RTL features were supported by `react-window`
#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change


src/data-table/__tests__/data-table-rtl.test.js
```diff
@@ -0,0 +1,28 @@
...
+LICENSE file in the root directory of this source tree.
+*/
+// @flow
+import React from 'react';
+import {mount} from 'enzyme';
+import {columns, rows} from './data-table.scenario.js';
+import {Unstable_DataTable} from '../index.js';
+import {ThemeProvider, LightTheme} from '../../index.js';
+
+describe('rendering data-table in RTL direction', () => {
```
chasestarr: Can you please remove the snapshot and this unit test and convert it to a visual regression test by creating `data-table-rtl.scenario.js`? I would prefer to have a visual check rather than the jest snapshot. Those tend to just build up noise that's not meaningful

rohitb4: I'll add a scenario.

src/data-table/data-table.js
```diff
@@ -553,7 +553,11 @@ const InnerTableElement = React.forwardRef<
                     paddingLeft: theme.sizing.scale300,
                     paddingRight: theme.sizing.scale300,
                     position: 'absolute',
-                    right: 0 - ctx.scrollLeft,
+                    right:
+                      theme.direction !== 'rtl'
+                        ? 0 - ctx.scrollLeft
+                        : 'initial',
+                    left: theme.direction === 'rtl' ? 0 : 'initial',
```
nadiia: nit: could you make both above ternary expressions either `===` or `!==`?

chasestarr:
Is `3000-` in the numerical column correct? Numerical column may still prefer to be right-aligned since we can better scan values, but I'm not sure what the standard convention is in rtl languages

rohitb4:
> Is `3000-` in the numerical column correct? Numerical column may still prefer to be right-aligned since we can better scan values, but I'm not sure what the standard convention is in rtl languages

Not sure about that, In base web Table, I see similar behaviour. But in Google sheets, I see different behaviour.

Base web screenshots:
Screenshot 2020-02-14 at 6 54 25 AM
Screenshot 2020-02-14 at 6 54 08 AM


Google sheets screenshots:
Screenshot 2020-02-14 at 6 58 03 AM
Screenshot 2020-02-14 at 6 58 14 AM

chasestarr:
I think we should maintain right-align on the numerical column so that the values can be optimally viewed. Here's an [example](https://he.wikipedia.org/wiki/%D7%92%D7%95%D7%92%D7%9C_%D7%9B%D7%A8%D7%95%D7%9D) of wikipedia rendering Hebrew numbers right-aligned

Issue #2630, opened 2020-01-14
[DataTable] Expose functionality to pass initial filters to `StatefulDataTable`, and call back on addition/removal of filters


lianga888:
## Feature description

Provide the ability to provide optional props on `StatefulDataTable` to pass in initial filter state, and callbacks to call when we change these filters. It looks like these props are exposed on `DataTable` but not on `StatefulDataTable`.

The use-case in mind is enabling the ability to persist filter state, such that a user could define some default filters and have those same filters initialized whenever reloading the page.

To do so, we would need a mechanism to provide these initial filters, and well have execute some callbacks when these filters change

Issue #2402, opened 2019-12-01
[DataTable] loading state


sijad:
## Feature description

show loading inside DataTable to let user know data are fetching from external resources.

chasestarr:
Linked PR contributes a loading state to the flexbox-based `table` component. The loading state on this `data-table` will simply be text similar to how the empty state is rendered
<img width="620" alt="Screen Shot 2020-01-07 at 11 04 56 AM" src="https://user-images.githubusercontent.com/5317799/71921362-902f3380-313d-11ea-965b-06a9781f8df8.png">

PR #2840, opened 2020-02-12
feat(data-table): various small features/bug fixes

chasestarr:
closes: https://github.com/uber/baseweb/issues/2630
closes: https://github.com/uber/baseweb/issues/2402

#### Description

Screen Shot 2020-02-12 at 9 44 17 AM

This PR includes a couple 'touch up' features that I think have been blocking from stabilizing the component.

- now supports dark theme
- resizable columns off by default
- loading state
- conditionally search/filter the stateful-data-table
- add/remove filter callbacks on stateful-data-table
- initial filters on stateful-data-table

#### Scope

- [x] Minor: New Feature


src/data-table/types.js
```diff
@@ -106,14 +116,15 @@ export type StatefulContainerPropsT = {|
   ...StatefulDataTablePropsT,
   children: ({|
     filters: Map<string, {description: string}>,
-    onFilterAdd: (filterParams: {description: string}, title: string) => void,
+    onFilterAdd: (title: string, filterParams: {description: string}) => void,
```
nadiia: This would be a breaking change, right? Can we keep it as is?

chasestarr: It's unstable and this particular prop is not used anywhere at uber so I think it's safe. The argument ordering was rather inconsistent so I took the opportunity to standardize it

Issue #2837, opened 2020-02-11
[Button] height changes between loading/not loading states


chasestarr:
Latest baseui release. Related to https://github.com/uber/baseweb/pull/2821#issuecomment-584880396. It can be reproduced by changing the size prop on Button doc page and then toggling the loading state.

chasestarr:
@houmark Thanks for raising the issue. A fix has been published in v9.52.2

houmark:
I'm sorry @chasestarr — but I'm still seeing the same height change with v9.52.2 for both a large button and a default-sized button. This may again be due to overrides or a different font-size or something. I'll test more tonight and provide more feedback.

Having said that, I'm not sure the fix with different `sizing.scale` is the safest solution, but I may not understand it well enough. Another option would be to measure the height of the button on the normal state, save it in memory, and then set that height when in the loading state?

chasestarr:
I can't reproduce on our docs site editor, but good point about overrides. If button font size has been overridden, it will continue to change height. Is that what you have going on in your application?

houmark:
No, we don't have any font sizes changes, our theme tweaks are only colors and media breakpoints.

I've narrowed it down to a global style we have:

```
div {
  box-sizing: border-box;
}
```

This should not really break anything and while this one may be obtrusive, it's also "normal" practice to have this style in modern applications. If we drop that style, then most of our design breaks. Maybe this is simply missing in the div's for the spinner / button?

houmark:
I think this change is gonna create a lot of problems. Maybe this should be reverted until a proper solution that as sensible can be thought out.

This is a Code Sandbox, forked from the documentation site, and no other modifications has been done than a quick and dirty loading state change:
https://codesandbox.io/s/button-zvifg

Note how the button sizes down when loading. This may be due to missing Baseweb components surrounding the button, but shouldn't each component be stable on its own?

houmark:
I've now added the border-box CSS style in index.html and the button is resizing its height when the loading state changes.

houmark:
I did a "hot patch" on our CSS to be like this, which fixed the height changes on buttons when `isLoading` is true:

```
div {
  box-sizing: border-box;
}
button div {
  box-sizing: unset;
}
```

Baseweb should maybe handle that?

I think a better long term solution is a ref on the button to measure the height and keeping that exact height while loading, though I am not sure if different box-sizing will still affect that, though it should be possible to reset that and make a combination of the two.

[...]

PR #2855, opened 2020-02-13
fix(button): Change spinner to use span instead of div

chasestarr:
closes: https://github.com/uber/baseweb/issues/2837

#### Scope

- [x] Patch: Bug Fix


houmark:
@chasestarr I think this one is good to go? It should maybe be renamed to "Change spinner to use span instead of div" though.

chasestarr:
> @chasestarr I think this one is good to go? It should maybe be renamed to "Change spinner to use span instead of div" though.

Sounds good, thank you

Issue #2870, opened 2020-02-17
[TimePicker] Selecting 12:00 AM returns current date


VladimirMilenko:
## Current Behavior
When selecting 12:00 AM value is ignored as seconds equals 0. And therefore time picker selects current date.

https://github.com/uber/baseweb/blob/master/src/timepicker/timepicker.js#L163
```js
this.props.value && seconds ? new Date(this.props.value) : new Date();
```

## Expected Behavior

Set 12:00 AM of current date.

## Your Environment

https://codesandbox.io/s/timepicker-n5968

| Tech    | Version |
| ------- | ------- |
| Base UI | v9.54.0  |

- [x] I have searched the [issues](https://github.com/uber/baseweb/issues) of this repository and believe that this is not a duplicate.

PR #2873, opened 2020-02-18
fix(timepicker): handle midnight properly

gergelyke:
Closes https://github.com/uber/baseweb/issues/2870

### DIFF
--- a/src/timepicker/timepicker.js
+++ b/src/timepicker/timepicker.js
@@ -159,8 +159,7 @@ class TimePicker extends React.Component<TimePickerPropsT, TimePickerStateT> {
   };
 
   handleChange = (seconds: number) => {
-    const date =
-      this.props.value && seconds ? new Date(this.props.value) : new Date();
+    const date = this.props.value ? new Date(this.props.value) : new Date();
     const [hours, minutes] = secondsToHourMinute(seconds);
     date.setHours(hours, minutes, 0);
     this.props.onChange && this.props.onChange(date);


src/timepicker/timepicker.js
```diff
@@ -160,7 +160,9 @@ class TimePicker extends React.Component<TimePickerPropsT, TimePickerStateT> {
 
   handleChange = (seconds: number) => {
     const date =
-      this.props.value && seconds ? new Date(this.props.value) : new Date();
+      this.props.value && typeof seconds === 'number'
```
chasestarr: Based on the type definition, it needs to be a number or flow will error. In what situation will seconds not be defined?

gergelyke: seconds can be zero, making it a falsy value - this change tries to cover that case

chasestarr: can `&& seconds` be removed?

gergelyke: that's a very good point - yes

gergelyke: @chasestarr fixed - @nadiia does this make sense to you too?

gergelyke:
> a test would better ensure the reported bug is fixed, but logically this should work

Would you write a unit test for this?

chasestarr:
I'd probably write an e2e test

gergelyke:
will follow up with a test pr later this week

PR #2887, opened 2020-02-19
feat(pagination): add size prop

gergelyke:


gergelyke:
> missing yard config update, but otherwise lgtm

great catch, adding now

PR #2888, opened 2020-02-19
fix(flow): transform cup globals before dist copy

chasestarr:
#### Description

I put together a quick fusion js project with flow and baseui to test various flow versions but was presented with errors related to `__DEV__` and similar globals. The published `.js.flow` files don't need this information so i've reconfigured the build process to transform cup globals before copying the flow files to `dist`. Doing so allows applications to run baseui + flow without `create-universal-package`.

```
yarn run v1.10.1
$ flow
Error --------------------------------------------------------------- node_modules/baseui/checkbox/checkbox.js.flow:57:9

Cannot resolve name `__DEV__`.

   57|     if (__DEV__ && this.props.checkmarkType === STYLE_TYPE.toggle) {
               ^^^^^^^
```

#### Scope

- [x] Patch: Bug Fix


chasestarr:
sigh.. component sizes should build based on separate import requirements `.flow` files should never end up in any final bundles

PR #2890, opened 2020-02-19
fix(button): improve a11y of the loading state

tajo:
Hm, tested a few solutions with the screenreader, this seems to produce the best result 
 
aria-label = `loading ${String(this.props.children)}`
aria-busy = "true"
 
Reader says "loading Foo Button" when the button is focused.


tajo:
> have you considered first checking the typeof children? not sure how `[object Object]` sounds on a screen reader

updated, yea probably better to not say anything

PR #2900, opened 2020-02-21
fix(ci): expose docker-compose env var to e2e server

chasestarr:


docker-compose.yml
```diff
@@ -17,6 +17,8 @@ services:
       interval: 5s
       timeout: 10s
       retries: 5
+    environment:
+      - COMPOSE_HTTP_TIMEOUT
```
gergelyke: is this coming from buildkite? if so, we may have to add here too https://github.com/uber/baseweb/blob/master/.buildkite/pipeline.yml

chasestarr: i added the value in the buildkite UI

gergelyke: oh I see, you are right, I mixed up the two files 🤦‍♂

PR #2912, opened 2020-02-25
feat(menu): improvements to nested menus

sandgraham:
- Fix issue with null refs remaining in menu list. This was partially `setState` timing (between add/remove) and partially us not properly filtering null refs.
- Reset parent menu on escape (by passing reset handler to `onEsc`).
- Make doc site version selector keyboard accessible (original objective).

Maybe this PR should be broken up into a `fix`, a `feat`, and a `docs` PR.

<details><summary>Old Behavior</summary>
<p>

</p>
</details>

<details><summary>New Behavior</summary>
<p>

</p>
</details>


src/menu/maybe-child-menu.js
```diff
@@ -44,6 +44,7 @@ export default function MaybeChildMenu(props: PropsT) {
             mountNode={ctx.mountRef.current ? ctx.mountRef.current : undefined}
             onMouseEnterDelay={30}
             onMouseLeaveDelay={30}
+            onEsc={ctx.onEsc ? ctx.onEsc : undefined}
```
chasestarr: Thanks for adding this, but at the moment I believe it's unmounting a bit too much on escape. This should call `resetParentMenu` from props. Calling that should close the top-most child menu and move selection back to the parent.

documentation-site/components/version-selector.js
```diff
@@ -81,8 +81,11 @@ const VersionSelector = () => {
     <StatefulPopover
       placement={PopoverPlacement.bottomLeft}
       dismissOnClickOutside={false}
+      focusLock
+      // eslint-disable-next-line jsx-a11y/no-autofocus
+      autoFocus
       content={({close}) => (
-        <NestedMenus>
+        <NestedMenus onEsc={close}>
```
chasestarr: I think `onEsc` on `NestedMenus` may not be required. I believe escape handling is not working here because of the changed focus element. You may want to reconfigure this to be a stateless popover if you need more control

sandgraham: Ah, didn't notice that method- seems to do the trick. I'm wondering if its okay to have it be the default (esc resets parent menu). I pushed up changes to review.

src/menu/maybe-child-menu.js
sandgraham: Ah, didn't notice that method- seems to do the trick. I'm wondering if its okay to have it be the default (esc resets parent menu). I pushed up changes to review.

documentation-site/components/version-selector.js
sandgraham: Yeah the new pop over created by maybe-child-menu doesn't get any `onEsc/Escape` handlers. But I think this works fine with reseting parent now. (Two escapes to close but makes sense).

PR #2915, opened 2020-02-25
feat(phone-input): add standalone phone picker component

nadiia:
The alternative phone input has the exact same API as the non-split phone input and shares the implementation with additional overrides to the reused select component. 
The change includes addition of `clearable` top level prop for the both phone input components.
Screen Shot 2020-02-25 at 7 59 28 AM


nadiia:
> Looks pretty good to me. Just have a few questions:
> 
> * What is the relationship between the country picker and country selector? Which one would I import if I wanted to use on its own? Maybe we should add some description or an example to doc site for this.

Since the `CountrySelect` includes the dial code presentation (it's more like the whole `Before` subcomponent for the input) and it's not really usable as a standalone component. It's a part of the public API and I couldn't generalize it on the level of `CountrySelect` without breaking change and with a clean API. 
For the future major we should get rid of the `CountrySelect` in favor of `CountryPicker`, and the dial code rendering inside the input should be just a part of the composed phone input component. I can't think of a case where `CountrySelect` could be used on its own so I don't think we should publicize the `CountrySelect` vs `CountryPicker` to avoid further confusion. Maybe a comment in the codebase for us to remember the note about? I'll add it.

> * In the figma designs, the positive and negative states do not affect the country dropdown. I think that makes sense because the dropdown itself shouldn't ever be in an error state where as the input might be. Do you prefer it one way or another?

The component reflects the current design for positive and error state in Figma:
Screen Shot 2020-02-25 at 10 46 36 AM
It's inconsistent though across the designs of different size phone input. I'll clarify it with the designers, shouldn't be a blocker.

> * On preview site, focus is not transferring to the input after selecting a country for the new alternate design.

Now that it's a two separate components navigation is similar to the one in a form, so will involve tabbing into the input vs switching the focus programmatically.

> * On preview site, when selecting an autocompletion option on phone input, the react-live instance gets filled but not the example input you were interacting with. Not sure what the deal with this is - it doesn't happen on production doc site. Maybe this is just a preview deployment issue.

The same happens on [the current release (production doc site)](https://baseweb.design/components/phone-input/#basic-usage). I believe it's the missing form controls name attribute is the cause.

src/phone-input/__tests__/phone-input-alternative-dropdown.scenario.js
```diff
@@ -0,0 +1,30 @@
...
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+// @flow
+
+import React from 'react';
+
+import {StatefulPhoneInputAlternative} from '../index.js';
+// import {Input as DefaultInput} from '../../input/index.js';
```
chasestarr:
```suggestion
```

src/phone-input/__tests__/phone-input-alternative-dropdown.scenario.js
```diff
@@ -0,0 +1,30 @@
...
+
```
chasestarr:
```suggestion
```

src/phone-input/__tests__/phone-input.scenario.js
```diff
@@ -9,6 +9,7 @@ LICENSE file in the root directory of this source tree.
 import React from 'react';
 
 import {StatefulPhoneInput, PhoneInputLite} from '../index.js';
+// import {Input as DefaultInput} from '../../input/index.js';
```
chasestarr:
```suggestion
```

src/phone-input/phone-input-alternative.js
```diff
@@ -0,0 +1,105 @@
...
+import {getOverrides, mergeOverrides} from '../helpers/overrides.js';
+import type {LitePropsT} from './types.js';
+
+PhoneInputAlternative.defaultProps = {
+  ...defaultProps,
+  countries: COUNTRIES,
+  clearable: true,
+};
+
+export default function PhoneInputAlternative(props: LitePropsT) {
```
chasestarr: Could the name be more descriptive of the result? Does design what this to eventually be the default? If so, maybe add `Next` suffix and add a warning for default usage.

nadiia: I named it with the intent of coexisting components. And since the API is the same at this point it could potentially be silently replaced. Unless we want to remove the usage of the non-split version completely before the v10 I think the warning might be a little too noisy. Happy to change the naming.

chasestarr: If it's safe to replace with the new version? I assume the two have different overrides. My main concern is that `Alternate` does not communicate how it's different from the default. Maybe `Separated`?

nadiia: There is a wrapper root element added for the split version to contain the two separate controls.

src/phone-input/country-picker.js
```diff
@@ -0,0 +1,121 @@
...
+          [SIZE.mini]: sizing.scale400,
+          [SIZE.compact]: sizing.scale500,
+          [SIZE.default]: sizing.scale600,
+          [SIZE.large]: sizing.scale700,
+        };
+        const styleOverride = {
+          paddingLeft: sizeToLeftPadding[props.$size || SIZE.default],
+          paddingRight: sizeToRightPadding[props.$size || SIZE.default],
+        };
+        // do not add positive and error color borders wjen not focused
```
gergelyke:
```suggestion
        // do not add positive and error color borders when not focused
```

documentation-site/pages/components/phone-input.mdx
```diff
@@ -40,6 +41,12 @@ It contains a dropdown which allows the user to choose from a list of countries
   <PhoneInputBasic />
 </Example>
 
+<Example title="Alternative design" path="phone-input/alternative.js">
```
gergelyke:
```suggestion
<Example title="Phone Input Next" path="phone-input/alternative.js">
```

documentation-site/pages/components/phone-input.mdx
```diff
@@ -40,6 +41,12 @@ It contains a dropdown which allows the user to choose from a list of countries
...
+  <PhoneInputNext />
+</Example>
+
+Use `PhoneInputNext` for an alternative look where the country picker is split from the input element.
```
gergelyke:
```suggestion
Starting in version 10,  `PhoneInputNext` will replace the current `PhoneInput`, where the country picker is split from the input element.
```

PR #2918, opened 2020-02-25
fix(checkbox): applies screen-reader improvements

chasestarr:
#### Description

- adds the `aria-checked` value `mixed` when the checkbox is `indeterminate`.
- applies a `bui` id to the caption element on `form-control`
- if `form-control` has an error, it will send the caption id as `aria-errormessage`
- if `form-control` has a caption, it will send the caption id as `aria-describedby`
- adds `aria-errormessage` prop to `input`

#### Scope

- [x] Patch: Bug Fix

### DIFF
--- a/src/checkbox/types.js
+++ b/src/checkbox/types.js
@@ -41,6 +41,10 @@ export type DefaultPropsT = {
 };
 
 export type PropsT = {
+  /** Id of element which contains a related caption */
+  'aria-describedby'?: string,
+  /** Id of element which contains a related error message */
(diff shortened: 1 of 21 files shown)


src/checkbox/types.js
```diff
@@ -41,6 +41,10 @@ export type DefaultPropsT = {
 };
 
 export type PropsT = {
+  /** Id of element which contains a related caption */
+  'aria-describedby'?: string,
```
gergelyke: should we add it to the TS definitions too?

PR #2928, opened 2020-02-26
feat(toast): add autoFocus prop

sandgraham:
This PR adds `autoFocus` to `Toast` and `Toaster`.

Previously if you wanted to focus a Toast you would have to tab through the natural tab order of the page to get to the dismiss button. Now, you can add the `autoFocus` prop, probably when using `Toaster`, in order to transfer focus to the Toast on mount. It will also try to restore focus to the previously focused element on dismissal.

All this said - people should prefer to set an auto close duration and not require user interaction to close. Focusing alerts is bad for a11y because the alert content is read at least twice on screenreaders. Read more about accessibility and toasts [here](https://getbootstrap.com/docs/4.2/components/toasts/#accessibility), [here](https://medium.com/@sheribyrnehaber/designing-toast-messages-for-accessibility-fb610ac364be), and [here](https://www.scottohara.me/blog/2019/07/08/a-toast-to-a11y-toasts.html).


src/toast/styled-components.js
```diff
@@ -139,6 +139,10 @@ export const Body = styled<SharedStylePropsArgT>('div', props => {
     transitionTimingFunction: $theme.animation.easeInOutCurve,
     display: 'flex',
     justifyContent: 'space-between',
+    ':focus': {
```
tajo: This will be visible even when not using keyboard right? It might be worth to use the same approach as in other components - detecting `isFocusVisible`

sandgraham:
I'm thinking now maybe we should focus the dismiss button and not the toast body. Maybe rename this to `focusDismissOnMount` or something.

vrt/config.js
```diff
@@ -340,6 +340,9 @@ const config = {
       },
     ],
   },
+  'toaster-focus': {
+    skip: true,
```
gergelyke: that's the reason for skipping it?

sandgraham: Since I added an e2e test for functionality I didn't think we needed visual coverage.

PR #2929, opened 2020-02-27
fix(select): communicate highlighted option to screen reader

chasestarr:
#### Description

Before this change the select component was not accessible and did not follow aria guidelines. The largest change was that the input value should change along with changing the highlighted list item. See [reach-ui](https://reacttraining.com/reach-ui/combobox/) and [google maps](https://maps.google.com) as examples. Notice how the UX is more like, 'I want to autocomplete text-input' rather than 'I want to select from a range of choices'

This pr adds a `transientInputState` so that we can update the input value based on highlight, and also maintain the typed-in value. Separating this into another variable also avoids 'filtering' the list based on the transient value.

In addition to the select-specific changes, some features/bug-fixes were made to the menu component. First, a callback was added to the `StatefulMenu` component to communicate when the highlight index changes. Second, `aria-activedescendant` was broken and now it applies a generated id, or user-provided.

I'd like to add some tests, but this is close to complete

#### Scope

- [x] Patch: Bug Fix


tajo:
I was trying to test this. The input's value changes correctly; however, the voiceover is reading only parts of the highlighted value or reads the previous value. Do you also observe that?

chasestarr:
> I was trying to test this. The input's value changes correctly; however, the voiceover is reading only parts of the highlighted value or reads the previous value. Do you also observe that?

Yeah I had seen it when reaching the last item in the list, but on trying again it seems to be bugging out even on the first item

src/menu/stateful-container.js
```diff
@@ -219,15 +228,13 @@ export default class MenuStatefulContainer extends React.Component<
       this.props.onItemSelect({item, event});
       this.internalSetState(STATE_CHANGE_TYPES.click, {
         highlightedIndex: index,
-        activedescendantId,
```
nadiia: It seems like getting rid of `activedescendantId` will be a breaking change.

src/select/select-component.js
```diff
@@ -515,6 +558,14 @@ class Select extends React.Component<PropsT, SelectStateT> {
     }
   };
 
+  handleHighlightChange = (option?: OptionT) => {
+    if (option) {
+      this.setState({transientInputValue: option[this.props.labelKey]});
```
nadiia: The selected value rendering can be defined by `getValueLabel` and the final displayed value can be different from the value of the `labelKey` prop. This will create inconsistency in the display value while navigating through options and when one of the options selected.

nadiia: [Example](https://baseweb-git-select-screen-reader.uber-ui-platform.now.sh/components/select/?code=import%20*%20as%20React%20from%20%22react%22%3B%0Aimport%20%7B%20Select%20%7D%20from%20%22baseui%2Fselect%22%3B%0A%0Aexport%20default%20()%20%3D%3E%20%7B%0A%20%20const%20%5Bvalue%2C%20setValue%5D%20%3D%20React.useState(%5B%0A%20%20%20%20%7B%20label%3A%20%22Aquamarine%22%2C%20id%3A%20%22%237FFFD4%22%20%7D%0A%20%20%5D)%3B%0A%20%20return%20(%0A%20%20%20%20%3CSelect%0A%20%20%20%20%20%20options%3D%7B%5B%0A%20%20%20%20%20%20%20%20%7B%20label%3A%20%22AliceBlue%22%2C%20id%3A%20%22%23F0F8FF%22%20%7D%2C%0A%20%20%20%20%20%20%20%20%7B%20label%3A%20%22AntiqueWhite%22%2C%20id%3A%20%22%23FAEBD7%22%20%7D%2C%0A%20%20%20%20%20%20%20%20%7B%20label%3A%20%22Aqua%22%2C%20id%3A%20%22%2300FFFF%22%20%7D%2C%0A%20%20%20%20%20%20%20%20%7B%20label%3A%20%22Aquamarine%22%2C%20id%3A%20%22%237FFFD4%22%20%7D%2C%0A%20%20%20%20%20%20%20%20%7B%20label%3A%20%22Azure%22%2C%20id%3A%20%22%23F0FFFF%22%20%7D%2C%0A%20%20%20%20%20%20%20%20%7B%20label%3A%20%22Beige%22%2C%20id%3A%20%22%23F5F5DC%22%20%7D%0A%20%20%20%20%20%20%5D%7D%0A%20%20%20%20%20%20value%3D%7Bvalue%7D%0A%20%20%20%20%20%20getValueLabel%3D%7B(%7B%20option%20%7D)%20%3D%3E%0A%20%20%20%20%20%20%20%20%60%24%7Boption.id%7D%20-%20%24%7Boption.label%7D%60%0A%20%20%20%20%20%20%7D%0A%20%20%20%20%20%20placeholder%3D%22Select%20color%22%0A%20%20%20%20%20%20onChange%3D%7Bparams%20%3D%3E%20setValue(params.value)%7D%0A%20%20%20%20%2F%3E%0A%20%20)%3B%0A%7D)

chasestarr: hmm, it doesn't really work because `getValueLabel` can return a react node rather than simply a string

PR #2932, opened 2020-02-27
fix(select): fix for an error in an empty clearable select on esc

nadiia:


src/select/select-component.js
```diff
@@ -546,10 +546,12 @@ class Select extends React.Component<PropsT, SelectStateT> {
   clearValue = (event: KeyboardEvent | MouseEvent | TouchEvent) => {
     if (isClick(event) && !isLeftClick(event)) return;
 
-    const resetValue = this.props.value.filter(
-      item => item.clearableValue === false,
-    );
-    this.setValue(resetValue, null, STATE_CHANGE_TYPE.clear);
+    if (this.props.value) {
```
chasestarr: `this.props.value` is not optional in types, how do we expect to trigger the error? It can happen if applications are not type-checking, but I thought we had agreed early on to not support that?

nadiia: That's a good question. You can see the repro in our current docs [in this example](https://baseweb.design/components/datepicker/#datepicker-with-quick-select-actions).

chasestarr: I'm fine with merging it because it's net-positive for the project, but am suspicious why it's not caught by flow. It may be worth tracking down why that example raising the bug

PR #2933, opened 2020-02-27
feat(layers): handle doc click event on the layers level

nadiia:
#### Description

Handle document click event on the LayersManager level and run the document click handler of the top-most layer only. 
It allows layers triggered from another layer to be rendered outside of the triggering layer container.

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change


nadiia:
Working on e2er test for it ^

tajo:
What is it fixing?

nadiia:
> What is it fixing?

The layering issue we can see [on the docs site](https://baseweb-git-focus-pickers.uber-ui-platform.now.sh/components/datepicker/#calendar-with-multiple-months).

src/popover/popover.js
```diff
@@ -442,6 +425,7 @@ class Popover extends React.Component<PopoverPropsT, PopoverPrivateStateT> {
             key="new-layer"
             mountNode={this.props.mountNode}
             onEscape={this.props.onEsc}
+            onDocumentClick={this.onDocumentClick}
```
chasestarr: Cool! Now clicks outside will close only the top-most popover. You may want to make this same change for modal and drawer as well. See https://baseweb-git-wpt-5568.uber-ui-platform.now.sh/components/select/#select-in-a-modal-dialog or maybe save to another pr

nadiia: Oh, somehow I missed the modal while searching for usage. I'll add this change to the modal in a follow-up PR.

PR #2942, opened 2020-02-28
feat(modal): handle modal closure through the Layer's onDocumentClick handler

nadiia:
Handling modal closure on a backdrop click through the Layer's `onDocumentClick` handler solves the issue when multiple layers rendered on top of a modal layer and a backdrop click removes all those layers instead of handling only the top-most layer's handlers.
An example here is a modal that renders a select component. When the select's dropdown is open a click on the backdrop (that is also outside of the select dropdown click) will only close the select's dropdown and the modal will remain open.

### DIFF
--- a/src/modal/__tests__/modal.test.js
+++ b/src/modal/__tests__/modal.test.js
@@ -20,7 +20,6 @@ import {
   StyledDialog,
   CLOSE_SOURCE,
 } from '../index.js';
-
 import {styled} from '../../styles/index.js';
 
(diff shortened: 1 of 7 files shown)


src/modal/__tests__/modal.test.js
```diff
@@ -105,21 +104,6 @@ describe('Modal', () => {
     });
   });
 
-  test('backdrop triggers close', () => {
```
nadiia: replaced this one with e2e

PR #2931, opened 2020-02-27
fix(tree-view): update default styling to chevron

chrisrauh:
#### Description

Update the default icons to be chevrons. 
Update the override docs accordingly so they still demonstrate overrides. 

The reference design is below though it had to be slightly adapted from the 14px designed to the 16px standard font size of the component.

image

#### Scope

- [X] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change

PR #2944, opened 2020-02-28
fix(tree-view): update default styling to chevron (#2931)

gergelyke:


gergelyke:
landing as is, and will send a follow up pr

PR #2947, opened 2020-02-28
fix(select): communicate component state to screen reader

chasestarr:
#### Description

Improves the select component so that screen readers can successfully operate it. See https://github.com/uber/baseweb/pull/2929 for an approach that would make `Select` a semantic combobox, but may lead to unintended changes in behavior. This Diff seems to be a good middle ground between making it operational, but not changing functionality.

#### Scope

- [x] Patch: Bug Fix


src/select/types.js
```diff
@@ -179,6 +179,7 @@ export type PropsT = {
 };
 
 export type SelectStateT = {
+  activeDescendant: ?string,
```
gergelyke: should we add these to the TS type definitions too?

tajo:
This works pretty great overall. However; it seems that backspace doesn't work anymore.

chasestarr:
> This works pretty great overall. However; it seems that backspace doesn't work anymore.

Can you please provide some reproduction steps and also validate against master? I didn't make any changes related to backspace in this PR. It may be a separate regression

nadiia:
> This works pretty great overall. However; it seems that backspace doesn't work anymore.

Do you mean to remove the selected value? It only works for `multi` I think.

tajo:
Ah, it's not a regression. When you select a value you are not able to delete part of it to search for other options. That's a bit annoying.

PR #2948, opened 2020-02-28
build(vrt): sanitize branch names

sandgraham:


vrt/ci.js
```diff
@@ -346,6 +346,12 @@ function getRepositoryOwnerAndNameFromURL(url) {
   return [owner, name];
 }
 
+function sanitizeBranchName(branchName) {
+  // Colons are not permitted in git branch names and for some reason
+  // Buidlkite will report the branch name as "user:branch".
```
sandgraham:
```suggestion
  // Buildkite will report the branch name as "user:branch".
```

vrt/ci.js
```diff
@@ -346,6 +346,12 @@ function getRepositoryOwnerAndNameFromURL(url) {
   return [owner, name];
 }
 
+function sanitizeBranchName(branchName) {
+  // Colons are not permitted in git branch names and for some reason
+  // Buildkite will report the branch name as "user:branch".
+  return branchName.replace(/:/g, '-');
```
tajo: Should we sanitize even more characters and use some 3rd party lib?

sandgraham: This should be fine since we are only protecting against the colon that buildkite might add to the actual branch name. The user's actual branch name can't be created in the first place if it has bad characters. Not sure why buildkite does this, but I think if the branch is called `master` - buildkite will append the repo owner (ex: `sandgraham:master`).

PR #2951, opened 2020-02-29
fix(flow): fix flow warnings that failed yarn:test

### LINKED ISSUE #2943: [build] yarn test fails on fresh clone
On a fresh clone, `yarn test` fails with the error below

```
$ yarn flow check --max-warnings 0
yarn run v1.21.1
$ /Users/crauh/uDev/baseweb/node_modules/.bin/flow check --max-warnings 0
Warning ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/datepicker/styled-components.js:592:5
Unused suppression comment.
     589│       borderBottomRightRadius: '100%',
     590│       ...(getDayStyles(code, props.$theme)[':after'] || {}),
     591│     },
     592│     // $FlowFixMe fails flow check in 0.111+
     593│     ...($range
     594│       ? {
     595│           // :before pseudo element defines a grey background style that extends
Warning ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/timepicker/__tests__/timepicker.test.js:38:5
Unused suppression comment.
     35│     const [hours, minutes] = secondsToHourMinute(option.id);
     36│     date.setHours(hours, minutes, 0);
     37│     expect(onChange).toHaveBeenCalled();
     38│     // $FlowFixMe
     39│     expect(onChange.mock.calls[0][0].getMinutes()).toEqual(date.getMinutes());
     40│     // $FlowFixMe
     41│     expect(onChange.mock.calls[0][0].getHours()).toEqual(date.getHours());
Warning ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/timepicker/__tests__/timepicker.test.js:40:5
Unused suppression comment.
     37│     expect(onChange).toHaveBeenCalled();
     38│     // $FlowFixMe
     39│     expect(onChange.mock.calls[0][0].getMinutes()).toEqual(date.getMinutes());
     40│     // $FlowFixMe
     41│     expect(onChange.mock.calls[0][0].getHours()).toEqual(date.getHours());
     42│   });
     43│ });
Found 3 warnings
error Command failed with exit code 2.
```

chrisrauh:
Fixes #2943 

#### Description

Fixes the flow warnings that caused yarn:test to fail from a clean clone.

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/src/datepicker/styled-components.js
+++ b/src/datepicker/styled-components.js
@@ -589,7 +589,6 @@ export const StyledDay = styled<SharedStylePropsT>('div', props => {
       borderBottomRightRadius: '100%',
       ...(getDayStyles(code, props.$theme)[':after'] || {}),
     },
-    // $FlowFixMe fails flow check in 0.111+
     ...($range
       ? {
           // :before pseudo element defines a grey background style that extends
--- a/src/timepicker/__tests__/timepicker.test.js
+++ b/src/timepicker/__tests__/timepicker.test.js
@@ -35,9 +35,7 @@ describe('Timepicker', () => {
     const [hours, minutes] = secondsToHourMinute(option.id);
     date.setHours(hours, minutes, 0);
     expect(onChange).toHaveBeenCalled();
-    // $FlowFixMe
     expect(onChange.mock.calls[0][0].getMinutes()).toEqual(date.getMinutes());
-    // $FlowFixMe
     expect(onChange.mock.calls[0][0].getHours()).toEqual(date.getHours());
   });
 });


tajo:
> $FlowFixMe fails flow check in 0.111+

This might solve it for the current 0.110 but 0.111 will break? Hm I wonder if the only solution here is to start ignoring flow warnings? @chasestarr

chasestarr:
It's strange that this fails locally but not in CI master. I first want to root-cause that. Ideally we should fix the flow errors since we may get into a messy spot where flow suppression comments begin to be 'append only' in the source code. If no other options then yes, we will need to remove [these tests](https://github.com/uber/baseweb/blob/master/src/styles/__tests__/styled.test.js#L112) and maybe place them into the `flow-check-build` package?

tajo:
I'm getting 76 errors with `flow-bin@0.111` and 340 with `flow-bin@0.119`.  I feel it's a doomed effort trying to compatible across multiple flow versions. Should we just focus on a single version? We are likely the only consumer of flow types anyway.

chasestarr:
How do you arrive at the 76 errors on v0.111?

tajo:
> How do you arrive at the 76 errors on v0.111?

*75 errors

1. master branch
2. `yarn add -D flow-bin@0.111`
3. `yarn flow`

chasestarr:
But we don't need to worry about that right? Just errors in what gets published. We could stay at our current version for as long as we wanted. React for example is still on an [old version](https://github.com/facebook/react/blob/master/package.json#L61) 

Another way to handle the problem would be to only publish type definition files similar to how we handle typescript. This could give us flexibility to adopt typescript in the source code if we find that valuable, but continue to maintain type defs. I haven't evaluated what benefits we would get from doing so, but may be worth decoupling implementation type-checking from publish

chrisrauh:
@chasestarr @tajo - not clear what exactly is the next step here, can you clarify?

tajo:
> not clear what exactly is the next step here, can you clarify?

I'm working on this. We actually don't run the flow check that's part of `yarn test` - it errors out when there are warnings. I'm now trying to fix the root cause (usage of `$FlowFixMe` because of v0.111).

PR #2953, opened 2020-02-29
fix(tree-view): fix leaf item indent and add indent guides

chrisrauh:
#### Description

- Fixes the leaf item indent by adding a blank icon. Added capability to override this icon because why not.

- Adds indent guides to increase readability of long lists of items on complex trees.

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


src/tree-view/index.d.ts
```diff
@@ -14,6 +14,8 @@ export interface TreeLabelOverrides {
   IconContainer?: Override<{}>;
   ExpandIcon?: Override<{}>;
   CollapseIcon?: Override<{}>;
+  LeafIconContainer?: Override<{}>;
+  LeafIcon?: Override<{}>;
```
tajo: `indentGuides` should be added into this file too

tajo:
Nice changes!

> Fixes the leaf item indent by adding a blank icon. Added capability to override this icon because why not.

What if I want to make this icon dynamic? Imagine this is a filesystem and I want to have a different icon for `.jpg` vs `.txt`.  This override API can't really support that use-case.

Maybe we could add something useful into `sharedProps` so the override can make that decision. Or even more robust solution - having the icon component as a part of the `data`:

```js
const data = [
  {
    id: 1,
    label: 'image1.jpg,
    icon: <IconJpg />,
  },
  {
    id: 2,
    label: 'file.txt,
    icon: <IconTxt />,
  },
]
```

chrisrauh:
> Nice changes!
> 
> > Fixes the leaf item indent by adding a blank icon. Added capability to override this icon because why not.
> 
> What if I want to make this icon dynamic? Imagine this is a filesystem and I want to have a different icon for `.jpg` vs `.txt`. This override API can't really support that use-case.
> 
> Maybe we could add something useful into `sharedProps` so the override can make that decision. Or even more robust solution - having the icon component as a part of the `data`:
> 
> ```js
> const data = [
>   {
>     id: 1,
>     label: 'image1.jpg,
>     icon: <IconJpg />,
>   },
>   {
>     id: 2,
>     label: 'file.txt,
>     icon: <IconTxt />,
>   },
> ]
> ```

I thought about that as well. The simple override in place was sort of a side effect of fixing the indentation. But agree that different icons for different file types is a likely scenario. Not against doing something like this. At this point, one could override one of the containers to add that capability though - similarly to the variable heading size example.

tajo:
> In the meantime, one could override one of the containers to add that capability though - similarly to the variable heading size example.

I guess it's fine for now.

src/tree-view/index.d.ts
chrisrauh: Added.

Issue #2825, opened 2020-02-10
FormControl doesn't propagate error, positive, disabled props


Mater:
## Current Behavior

FormControl has logic to propagate error, positive and disabled props for child control.
It doesn't work for controls with `defaultProps`.

Example: FormControl with disabled and error. Input has default values for this props and propagation doesn't work. For positive it works, because no positive default in Input.
https://codesandbox.io/s/formcontrol-94fvj

Logic in FormControl (props propagation) is redundant, because it doesn't work as expected with baseweb controls, and need pass these props explicitly for proper work.

## Expected Behavior

It should work, if no props passed explicitly to control.

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v9.52.0  |
| React   |  16.8.6 |
| browser |  Google Chrome Version 79.0.3945.130 (Official Build) (64-bit) |

- [x] I have searched the [issues](https://github.com/uber/baseweb/issues) of this repository and believe that this is not a duplicate.

gwuah:
I'm picking this up. cc. @chasestarr

PR #2954, opened 2020-02-29
fix(form-control): propagate 'disabled' prop to children

gwuah:
Fixes https://github.com/uber/baseweb/issues/2825

#### Description

This pull request fixes the bug that was causing the disabled property from being properly propagated to children components.

#### Scope

- [x] Patch: Bug Fix


gergelyke:
thank you @gwuah

PR #2957, opened 2020-03-02
build(vrt): fix logic for generating snapshot branches from forks

sandgraham:
Update logic for determining the snapshot branch name.


sandgraham:
@chasestarr yeah AFAIK that env variable did not used to prefix the author from forked repos- otherwise the github pr logic would never have worked.

PR #2968, opened 2020-03-03
feat(timepicker): Swappable date libraries

### LINKED ISSUE #2921: [Datepicker / Timepicker] Swappable date processing in the using date.io
## Feature description

Would you guys be open to me submitting a PR to introduce swappable date processing in Datepicker and Time picker? 

The idea is to move all the date processing calls into a single utils objects that would be passed down through context, similar to how it's implemented [here](https://material-ui-pickers.dev/getting-started/installation) in material ui. This would allow us to rely on luxor for the date processing instead of date-fns.  I was also planning on using the same [date.io](https://github.com/dmtrKovalenko/date-io#readme) interface that material ui is using. I'd probably start on Timepicker because its simpler as a proof of concept and then move to Datepicker. 

## Background

We need the ability to have a picker run correctly for a different timezone than the browser, and we're running into issues with date-fns and the reliance on the creation of native date objects in the components. Right now, our only options are to fork and update the date processing, or just not use baseweb at all, we love the rest of the library, so we're hoping to find a way to get past this problem. 

## Questions

My plan was to move the calls one by one, rerunning the entire test suite every time to confirm that nothing broke.  Is the test coverage of the Date and Timepicker complete enough to ensure that nothing will break as I make the swaps? 

If the coverage isn't sufficient to make the swap, do you have any recommendations for how I might increase it to make that possible?

cbranch101:
Fixes #2921 

#### Description

This adds the ability to swap out date-fns with other date processing libraries like moment or luxor in the TimePicker.  It achieves this by adding a new adapter prop that contains the date functions that can be swapped out.   If `adapter` isn't provided Timepicker should be behave exactly the same way. 

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change


src/timepicker/timepicker.js
```diff
@@ -224,10 +237,9 @@ class TimePicker extends React.Component<TimePickerPropsT, TimePickerStateT> {
     // $FlowFixMe
     selectProps.overrides = selectOverrides;
 
-    const value =
-      this.props.value instanceof Date
-        ? this.buildSelectedOption(this.props.value, this.props.format)
-        : this.state.value;
+    const value = utils.isValid(this.props.value)
```
cbranch101: This is the only change in here that doesn't map 1 to 1 to old date-fns api.  The problem is that now that the `value` could be any number of different objects, used a `instanceof` check here is going to be too strict.  `isValid` is the generic api `date.io` uses to achieve the same goal.

chasestarr: makes sense 👍

src/datepicker/utils/date-utils-provider.js
```diff
@@ -0,0 +1,27 @@
...
+
+const baseInstance = new DateFnsUtils();
+
+export const DateUtilsContext: React.Context<ThemeT> = React.createContext(
+  baseInstance,
+);
+
+const DateUtilsProvider = (props: {utils: any, children: ?React.Node}) => {
+  const {utils: Utils, children} = props;
+  const utilsInstance = React.useMemo(() => new Utils(), [Utils]);
```
chasestarr: Initializing the `utils` class here feels a bit strange, can we guarantee the `date-io` adapters take no arguments? What's the downside to the provider's parent calling the constructor?

src/datepicker/utils/date-utils-provider.js
```diff
@@ -0,0 +1,27 @@
...
+LICENSE file in the root directory of this source tree.
+*/
+// @flow
+import * as React from 'react';
+import DateFnsUtils from '@date-io/date-fns';
+
+const baseInstance = new DateFnsUtils();
+
+export const DateUtilsContext: React.Context<ThemeT> = React.createContext(
+  baseInstance,
```
chasestarr: Can this move to a property on the context object? Maybe like `React.createContext({adapter: new DateFnsUtils()});`. Doing so would signal a wrapped api in the code more clearly.

cbranch101: Yep, I was going back and forth on this.  It definitely makes it more clear.

src/datepicker/utils/date-utils-provider.js
cbranch101: Different types of the utils class from date-io can accept different arguments.  In the Material UI implementation, all of those arguments are just props on the DateUtilsProvider.  

I was pretty much just following the Material UI implementation across the board, but now that you mention it, just providing the instance as a prop is probably simpler.  I guess the only real benefit would be similarity to the API of Material UI's date swapping.

cbranch101: I made this change, and it ended up simplifying the types as well

src/datepicker/utils/types.js
```diff
@@ -0,0 +1,133 @@
...
+
+export type DateT = moment$Moment | Date;
+
+type DateValues = DateT | string | number;
+type DateInput = ?DateValues;
+
+type Comparison = (value: DateT, comparing: DateT) => boolean;
+type DateInDateOut = (value: DateT) => DateT;
+
+type DateFuncT = (DateInput | void) => DateT;
```
cbranch101: This ended up being a little bit of a compromise.  As you can see [here](https://github.com/dmtrKovalenko/date-io/blob/master/packages/core/IUtils.d.ts#L60), `date` can actually return null.  The idea is that, if you pass `null`, it will just always return `null`. 

The problem was, this was introducing all kinds of type refinement issues when functions that were previously expecting dates, were receiving this type that could be null.   I experimented with using [predicate functions](https://flow.org/en/docs/types/functions/#toc-predicate-functions) to let the flow know that if you pass anything but null into date, it will always be defined, but the fact that it was an externally defined function made that much more complicated. 

My compromise solution was just to act as if this null path didn't exist.  It seems like a little bit of an edge case situation anyway, so this seemed like the best option.

src/timepicker/__tests__/time-picker.e2e.js
```diff
@@ -220,4 +221,36 @@ describe('TimePicker', () => {
       expect(option1).toBe('00:15');
     });
   });
+  describe('when using moment', () => {
```
cbranch101: This first test was just to establish that the provider was working, and get the ball rolling on the PR.  Long term, it seems like the ideal solution would be to create some kind of setup where all of the tests defined in this file could be rerun for every date-io function type.  This would have required a massive refactoring of this file and would have made it much harder to follow, so I decided to hold off on that, but I do think it would be the most complete way to ensure that the timepicker is fully working with the other support date libraries.

cbranch101:
You can see that I deleted the single `TimePicker` unit test.  The problem was that the new context api [isn't working](https://github.com/enzymejs/enzyme/issues/2117) with shallow render in enzyme yet. 

Changing the `DateUtilsProvider` to use the old context api seemed like a bad idea, switching to use `mount` instead of shallow render would have totally changed the test, and so that left just deleting the test altogether.  As it was only a single test anyway, and the end to end seemed to be covering everything the unit tests were covering, that seemed like the best bet.

src/timepicker/timepicker.js
```diff
@@ -112,16 +119,20 @@ class TimePicker extends React.Component<TimePickerPropsT, TimePickerStateT> {
...
   state = {steps: [], value: null};
+  static contextType: React.Context<DateUtilsContextT> = DateUtilsContext;
+  context: DateUtilsContextT;
 
   componentDidMount() {
+    const {utils} = this.context;
     const steps = this.buildSteps();
-    if (this.props.value instanceof Date) {
+
+    if (utils.isValid(this.props.value) && this.props.value) {
```
cbranch101: Using `instanceof` was specific enough that the type could be refined to Date inside of the if statement, this was the only way I could find to keep it working the same way

src/data-table/column-datetime.js
```diff
@@ -397,7 +400,7 @@ function DatetimeFilter(props) {
                     value={rangeDates[0]}
                     onChange={time =>
                       setRangeDates([
-                        applyTimeToDate(rangeDates[0], time),
+                        applyTimeToDate(rangeDates[0], utils.toJsDate(time)),
```
cbranch101: This was the only place where the separation between the TimePicker and the Datepicker changes leaked out. 

The idea was to update everything in the TimePicker first, but because `applyTimeToDate` is used in this component and DatePicker, it broke down that separation. 

Once DatePicker has been fully moved over `applyTimeToDate`, I'll use utils and correctly handle all of the potential date objects that could be passed into it, but for now the easiest solution is to just convert it back into a JS Date.  And, luckily, in the default `date-fns` [date-io implementation](https://github.com/dmtrKovalenko/date-io/blob/master/packages/date-fns/src/date-fns-utils.ts#L190-L192) `toJsDate` just returns the value passed into it, and so this call won't add any extra conversions for normal users.

cbranch101:
I rebased master into this branch before I pushed it, and it's showing no changes available, but I'm still getting the `This branch is out-of-date` message on the PR.  Do I still just need to merge master in as well?

src/datepicker/utils/types.js
chasestarr: This does add some nuance to consuming date values from props like [here](https://github.com/uber/baseweb/pull/2968/files#diff-a551f5d9fe815b88dbfdfccf4f31a6afR178) for example. Does flow raise an error if `undefined` is omitted from that line?

Without running the code, my guess is that because `DateInput` is a [maybe type](https://github.com/uber/baseweb/pull/2968/files#diff-73c5142ac939310b4e94f825bad665dbR12), the argument type provided to `DateFuncT` would expand out to `moment$Moment | Date | string | number | null | void`.  Does that seem to be the case?

My worry is that we will miss handling the null case and cause a fairly common pattern, initializing controlled state variables to `null`, to break amongst different date-io adapters

src/timepicker/__tests__/time-picker.e2e.js
chasestarr: Agreed, this looks like a good first step

src/timepicker/__tests__/time-picker.scenario.js
```diff
@@ -42,8 +50,8 @@ const Controlled = ({
         size={size}
         {...restProps}
       />
-      <p data-e2e="hours">hour: {time ? time.getHours() : 'null'}</p>
-      <p data-e2e="minutes">minute: {time ? time.getMinutes() : 'null'}</p>
+      <p data-e2e="hours">hour: {time ? getHours(time) : 'null'}</p>
```
chasestarr: Will methods on the `Date` class continue to work as normal if no adapter is used? Just confirming, but was this change  made to avoid code duplication? It may be worth separating to assert existing code patterns continue to function as-expected

src/timepicker/timepicker.js
```diff
@@ -159,10 +173,14 @@ class TimePicker extends React.Component<TimePickerPropsT, TimePickerStateT> {
...
-    const [hours, minutes] = secondsToHourMinute(seconds);
-    date.setHours(hours, minutes, 0);
-    this.props.onChange && this.props.onChange(date);
+    const {utils} = this.context;
+    const {setHours, setMinutes, setSeconds} = utils;
+    const date = utils.date(this.props.value || undefined);
+    const [hours, minutes] = secondsToHourMinute(seconds, utils);
+    const hourDate = setHours(date, hours);
+    const minuteDate = setMinutes(hourDate, minutes);
+    const updatedDate = setSeconds(minuteDate, 0);
```
chasestarr:
```suggestion
    const date = utils.date(this.props.value || undefined);
    const [hours, minutes] = secondsToHourMinute(seconds, utils);
    const hourDate = utils.setHours(date, hours);
    const minuteDate = utils.setMinutes(hourDate, minutes);
    const updatedDate = utils.setSeconds(minuteDate, 0);
```

nit: I find it a bit tough to follow the multiple destructures (was unsure for a moment where the functions were coming from). No need to commit, just a suggestion

src/timepicker/__tests__/time-picker.scenario.js
cbranch101: This change was made so that this function would work correctly for moment version in the test. I could split it out, but it was kind of nice to be able to just re-run the original test fully. 

And yah, it behaves the exact same way for the standard version.  It's making the same calls inside of `utils`

cbranch101: So, I spent a little bit more time digging into this null type issue, and I'm a little stuck on it. 

`const seconds = dateToSeconds(utils.date(), utils);`

That call is a prime example of the problem I'm hitting. 

If the argument passed into `date` is null, it will just return null, otherwise it will return a date.  This means the return type has to be `Date | null` Handling for this gave me two options, but of which seemed bad. 

**1. Always refine `utils.date`**

```
const seconds = dateToSeconds(utils.date() || new Date(), utils);
```

This solves the problem of refining the date down to the right type I needed for `dateToSeconds` but to do it I have to introduce this super weird condition that can't ever be reached.  The second part of that expression only exists because flow's type system isn't currently smart enough to know that the returned object will always be there. 

**2. Always handle null in date processing functions**

I could change it to be able to handle that null value, and would fix the type system, but this also seems not great.  I know that `dateToSeconds` is expecting an explicit date(because date() is being passed), but the only way to make it work is to handle a null value that can never happen. 

I also explored somehow wrapping `date` with `%checks` functions that the type system could figure out that `date` was returning a date object if you passed it undefined, but all of those solutions still require some kind of if statement for the refining to happen, and so we're back in the same situation as the first bad option above. 

I think your concern about null types not being handled correctly is totally valid, I just think that this solution as it is might cover it.  

Passing null to `date` will throw a flow error in a case when it's actually okay to pass null.  The problem here is more that we're preventing people from doing something that would potentially be convenient, rather than allowing them to do something potentially harmful.  Given that, I think this is maybe an acceptable tradeoff. 

If you disagree, the `dateToSeconds(utils.date() || new Date(), utils)` style call everywhere we're calling `date` seems like next best to me. 

Thanks for the super prompt and clear feedback on this PR

cbranch101: @chasestarr Going back and reading my comment and this comment it was referring to 
> My worry is that we will miss handling the null case and cause a fairly common pattern, initializing controlled state variables to null, to break amongst different date-io adapters

I assumed you were talking about my decision to just act like the ability to pass null and get back null in date just didn't exist, but I'm not sure that's the case. The two links you include in your comment don't seem to have linked to the lines in question correctly, they just go to the whole file.  Could you update those so I can follow better what you mean?

cbranch101:
@chasestarr I reworked this PR to use props instead of context like we discussed.  Not much in here should be surprising, aside from maybe the fact that I moved some of the handlers around inside of Timepicker.  I wanted a clear division between what should go in `DateHelpers` and what should be in the component.  Any reusable functionality that uses the adapter goes in date helpers, and anything that calls a date helper got moved to the class.

src/timepicker/timepicker.js
```diff
@@ -132,15 +62,18 @@ class TimePicker extends React.Component<TimePickerPropsT, TimePickerStateT> {
         steps: steps,
         value: this.props.nullable
           ? undefined
-          : {id: closestStep, label: secondsToLabel(closestStep)},
+          : {
+              id: closestStep,
+              label: this.secondsToLabel(closestStep, undefined),
```
chasestarr:
```suggestion
              label: this.secondsToLabel(closestStep, this.props.format),
```

src/timepicker/timepicker.js
```diff
@@ -11,117 +11,47 @@ import * as React from 'react';
...
   };
+  dateHelpers: DateHelpers<T>;
 
   state = {steps: [], value: null};
 
   componentDidMount() {
     const steps = this.buildSteps();
-    if (this.props.value instanceof Date) {
+    const {adapter} = this.props;
+    this.dateHelpers = new DateHelpers(adapter);
```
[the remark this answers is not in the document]
chasestarr: The downside of assigning this to a class property is that it will not update in response to props changes. While that is likely a contrived example that should not come up a simple trade-off to an adapter-getter method resolves the issue.

```js
Class Component {
  getDateIOAdapter = () => {
    return this.props.adapter || dateFnsAdapter;
  }
}
```

src/timepicker/timepicker.js
```diff
@@ -158,11 +91,86 @@ class TimePicker extends React.Component<TimePickerPropsT, TimePickerStateT> {
     this.handleChange(seconds);
   };
 
+  secondsToLabel = (seconds: number, format: mixed) => {
```
chasestarr:
```suggestion
  secondsToLabel = (seconds: number, format: '12' | '24') => {
```

src/timepicker/timepicker.js
```diff
@@ -194,23 +202,27 @@ class TimePicker extends React.Component<TimePickerPropsT, TimePickerStateT> {
...
     }
     return filterOptions(options, filterValue, excludeOptions, newProps);
   };
 
-  buildSelectedOption(value: Date, format: '12' | '24' = '12') {
-    const secs = dateToSeconds(value);
+  buildSelectedOption: (T, ?'12' | '24') => OptionT = (
+    value,
+    format = '12',
+  ) => {
```
chasestarr: flow is thankfully working on inferring generics from outside of the function scope (skolemization) https://medium.com/flow-type/what-were-building-in-2020-bcb92f620c75 until then, we need to provide an 'input' to the function like below

```suggestion
  buildSelectedOption = <T>(
    value: T,
    format: '12' | '24' = '12',
  ): OptionT => {
```

src/datepicker/utils/types.js
```diff
@@ -0,0 +1,127 @@
...
+*/
+// @flow
+
+type DateValues<T> = T | string | number;
+type DateInput<T> = ?DateValues<T>;
+
+type Comparison<T> = (value: T, comparing: T) => boolean;
+type DateInDateOut<T> = (value: T) => T;
+
+type DateFuncT<T> = (DateInput<T> | void) => T;
```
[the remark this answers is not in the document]
cbranch101: It can, but only if we you pass null, and since we're using the types to prevent people from passing null, this seems like a reasonable compromise.  If we don't do this, every single time we call `adapter.date()` we have to do `adapter.date() || new Date` even though the second part of that expression can never happen.  This was that whole predicate function issue I was talking about before.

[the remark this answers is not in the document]
chasestarr: Does this line error if it does not fallback to `undefined`?
https://github.com/uber/baseweb/pull/2968/files#diff-a551f5d9fe815b88dbfdfccf4f31a6afR168

src/timepicker/timepicker.js
cbranch101: @chasestarr mentioned we should keep this to maintain the previous behavior

src/timepicker/timepicker.js
cbranch101: We decided to punt on this change, because it would require manually passing the type all the call site like `buildSelectedOption<Date>()`

src/datepicker/utils/types.js
```diff
@@ -0,0 +1,127 @@
...
+
+  getWeekdays(): string[],
+  getWeekArray(date: T): T[][],
+  getYearRange(start: T, end: T): T[],
+
+  /** Allow to customize displaying "am/pm" strings */
+  getMeridiemText(ampm: 'am' | 'pm'): string,
+};
+
+export type DateFormatsT = {
```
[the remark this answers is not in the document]
cbranch101: It's just used when you configure the adapter, we may need that for the Datepicker PR, but I'll just take it out for now

src/datepicker/utils/date-helpers.js
```diff
@@ -0,0 +1,30 @@
...
+
+const MINUTE = 60;
+const HOUR = MINUTE * 60;
+
+class DateHelpers<T> {
+  adapter: DateIOAdapter<T>;
+  constructor(adapter: DateIOAdapter<T>) {
+    this.adapter = adapter;
+  }
+  dateToSeconds: T => number = date => {
```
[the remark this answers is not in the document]
cbranch101: Same as below, we're going to punt on this change

src/timepicker/__tests__/time-picker.scenario.js
```diff
@@ -42,8 +48,37 @@ const Controlled = ({
...
```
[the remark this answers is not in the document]
cbranch101: This should be reverted now

src/timepicker/__tests__/time-picker.scenario.js
```diff
@@ -12,8 +12,13 @@ import startOfDay from 'date-fns/startOfDay';
 
 import {TimePicker} from '../index.js';
 import {SIZE} from '../../input/index.js';
+import dateFnsAdapter from '../../datepicker/utils/date-fns-adapter.js';
+import MomentUtils from '@date-io/moment';
+import moment from 'moment';
```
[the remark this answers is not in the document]
cbranch101: Yep, that makes more sense, I did that

src/index.d.ts
```diff
@@ -8,6 +8,8 @@ import {
 import {Override, Overrides} from './overrides';
 import {Locale} from './locale';
 import {Theme, ThemePrimitives} from './theme';
+import {IUtils} from '@date-io/core/IUtils';
+import {Moment} from 'moment';
```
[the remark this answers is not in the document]
cbranch101: Good call

src/timepicker/timepicker.js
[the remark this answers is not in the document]
cbranch101: Based on what we discussed on slack, I initially set `DateHelpers` in the constructor, and then reset it in `shouldComponentUpdate`

src/datepicker/utils/types.js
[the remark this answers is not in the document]
cbranch101: This required making the types a little bit more specific, but now it does show a helpful message

cbranch101:
Because we're switching from using `date-fns` directly to using an adapter that includes every possible function we would need from `date-fns` this loss in the ability to tree shake has an effect on build size.  Using a demo app that only imports `Timepicker` as an example, it would increase by 48.75 KB from 465.64 to 514.39.  

One thing that might be relevant to this increase is the fact that we're currently using both the date-io adapter and standard import of `date-fns` at the same time.  Once all of Datepicker is migrated to use the adapter, removing `date-fns` as a dependency might help this somewhat, but that I'm not sure about.

src/datepicker/utils/types.js
```diff
@@ -0,0 +1,76 @@
...
+*/
+// @flow
+
+type DateValues<T> = T | string | number;
+type DateInput<T> = DateValues<T>;
+
+type Comparison<T> = (value: T, comparing: T) => boolean;
+type DateInDateOut<T> = (value: T) => T;
+
+type DateFuncT<T> = (DateInput<T> | void) => T;
```
sandgraham: nit: only type with trailing `T`

src/timepicker/timepicker.js
```diff
@@ -11,117 +11,51 @@ import * as React from 'react';
...
 
+  constructor(props: TimePickerPropsT<T>) {
+    super(props);
+    this.dateHelpers = new DateHelpers(props.adapter);
+  }
+
   componentDidMount() {
     const steps = this.buildSteps();
-    if (this.props.value instanceof Date) {
+    const {adapter} = this.props;
```
sandgraham: nit: the other props are accessed directly on `this.props`

src/timepicker/timepicker.js
```diff
@@ -132,21 +66,28 @@ class TimePicker extends React.Component<TimePickerPropsT, TimePickerStateT> {
...
     const formatChanged = prevProps.format !== this.props.format;
     const stepChanged = prevProps.step !== this.props.step;
+    const adapterChanged = prevProps.adapter !== this.props.adapter;
     if (formatChanged || stepChanged) {
       const steps = this.buildSteps();
       this.setState({steps});
     }
+    if (adapterChanged) {
+      this.dateHelpers = new DateHelpers(this.props.adapter);
+    }
```
sandgraham: Maybe this should go before the potential `buildSteps` and `setState` calls? My concern is that the new list would use the old adapter.

src/timepicker/timepicker.js
```diff
@@ -158,11 +99,86 @@ class TimePicker extends React.Component<TimePickerPropsT, TimePickerStateT> {
...
+      return {id: secs, label: this.secondsToLabel(secs, format)};
+    });
+  };
+
   handleChange = (seconds: number) => {
-    const date = this.props.value ? new Date(this.props.value) : new Date();
-    const [hours, minutes] = secondsToHourMinute(seconds);
-    date.setHours(hours, minutes, 0);
-    this.props.onChange && this.props.onChange(date);
+    const {adapter} = this.props;
```
sandgraham: nit: as above

src/timepicker/timepicker.js
```diff
@@ -194,23 +210,26 @@ class TimePicker extends React.Component<TimePickerPropsT, TimePickerStateT> {
...
-    const secs = dateToSeconds(value);
+  buildSelectedOption: (T, ?'12' | '24') => OptionT = (
+    value,
+    format = '12',
+  ) => {
+    const secs = this.dateHelpers.dateToSeconds(value);
     return {
       id: secs,
-      label: secondsToLabel(secs, format),
+      label: this.secondsToLabel(secs, format || '12'),
```
sandgraham: Is the `|| '12'` required when it has a default value already?

src/timepicker/types.js
```diff
@@ -40,3 +42,10 @@ export type TimePickerStateT = {
   /** Internal value of the selected time as an integer since midnight (0) */
   value: ?OptionT,
 };
+
+export type TimePickerDefaultPropsT = {
+  format: '12',
+  step: 900,
+  creatable: false,
+  adapter: DateIOAdapter<Date>,
+};
```
sandgraham: Why do we need default props to be typed? Also is it valid Flow to include values like so? I haven't seen that before.

src/timepicker/timepicker.js
cbranch101: It is yah, unfortunately.  The long term solution to this is probably to not pass it around at all and just use the format prop directly in these functions, but there was some old behavior that that could have potentially broken, so I decided to wait on that.

src/timepicker/types.js
cbranch101: This is the only way the default adapter prop can be typed correctly. The tricky part here is that we have a more specific date io adapter type in default props then we have in prop types, so this was the only way I could find to allow for passing in the generic type, but then also have the specific type work by default.

cbranch101:
@sandgraham Regarding your question about swapping date libraries, the solution for that would be to create a new wrapper component for `Timepicker` that passes the adapter prop and then use that throughout your app. 

This is actually something @chasestarr and I discussed quite a bit working through this PR.  The original implementation used context, which made the swapping of date libraries simpler.  You could just pass the adapter instance to a provider once and you'd be good to go .  

The problem was, this made the types complicated to a point of pretty much being unworkable.  I was able to cobble together something that worked, but it was extremely hard to follow.  Switching to a prop was the compromise we came up with, with the thinking being that most end users will be creating a single modified version of `Timepicker` to use in their apps anyway, so adding that new adapter prop to that modified version wasn't too much of an ask.  

If we see a proliferation of components that consume this adapter prop, that could become an unworkable solution, but for now it's just `Timepicker`, and eventually `Datepicker` and `DatetimeColumn`, so it seemed like a good way forward.

src/timepicker/timepicker.js
cbranch101: Yep, that would 100% lead to problems, good catch.

src/timepicker/types.js
sandgraham: Interesting, thanks for explanation. Haven't seen this technique before.

chasestarr:
I did a bit more digging on bundle-size measurements before and after this change. It will grow by about `~7-8%` if just importing the `TimePicker` component, but when paired with `Datepicker` in the same app, growth over baseline will be reduced to about `~3%`. Both deltas use compressed builds

Issue #2908, opened 2020-02-24
[LayoutGrid] Conditionally render Cell in Grid


houmark:
## Current Behavior

Trying to conditionally render a `<Cell>` and returning `null` crashes the grid:

```javascript
Uncaught TypeError: Cannot read property 'props' of null
    at grid.js:28
    at mapSingleChildIntoContext (react.development.js:1283)
    at traverseAllChildrenImpl (react.development.js:1145)
    at traverseAllChildrenImpl (react.development.js:1161)
    at traverseAllChildren (react.development.js:1226)
    at mapIntoWithKeyPrefixInternal (react.development.js:1308)
    at Object.mapChildren [as map] (react.development.js:1332)
    at Grid (grid.js:19)
    at renderWithHooks (react-dom.development.js:16260)
    at mountIndeterminateComponent (react-dom.development.js:18794)
    at beginWork$1 (react-dom.development.js:20162)
    at HTMLUnknownElement.callCallback (react-dom.development.js:336)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:385)
    at invokeGuardedCallback (react-dom.development.js:440)
    at beginWork$$1 (react-dom.development.js:25780)
    at performUnitOfWork (react-dom.development.js:24695)
    at workLoopSync (react-dom.development.js:24671)
    at performSyncWorkOnRoot (react-dom.development.js:24270)
    at react-dom.development.js:12199
    at unstable_runWithPriority (scheduler.development.js:697)
    at runWithPriority$2 (react-dom.development.js:12149)
    at flushSyncCallbackQueueImpl (react-dom.development.js:12194)
    at flushSyncCallbackQueue (react-dom.development.js:12182)
    at discreteUpdates$1 (react-dom.development.js:24423)
    at discreteUpdates (react-dom.development.js:1438)
    at dispatchDiscreteEvent (react-dom.development.js:5881)
```

Condition example:

```javascript
{condition ? <Cell span={2}>Content</Cell> : null}
```

## Expected Behavior

It would be great if the grid would do a null check and simply move on if null/undefined. A workaround has been to use `span={0}` but it doesn't feel quite right.

| Tech    | Version |
| ------- | ------- |
| Base UI | v9.56.0  |
| React   |    Lastest     |
| browser |    Chrome 79.x     |

- [x] I have searched the [issues](https://github.com/uber/baseweb/issues) of this repository and believe that this is not a duplicate.

gwuah:
[...]

PR #2971, opened 2020-03-03
fix(layout-grid): enable conditional rendering

gwuah:
Fixes https://github.com/uber/baseweb/issues/2908

#### Description

This pr contains a fix that enables users to conditionally render a grid.

#### Scope

- [ x] Patch: Bug Fix


chasestarr:
Thanks @gwuah - you can run `yarn flow` locally to see how static type checking fails

gwuah:
Hi @chasestarr,
So I ran the command locally and I saw the issue.
I tried adding null to possible child types(see image below) but that broke a lot of things.
I can confirm the fix works though. I just want to know if there is a better to test my fix? 

Screen Shot 2020-03-04 at 10 22 38 AM

gwuah:
Okay, so turns out I placed the `null` in the wrong position. 
Sorry for the notification. I've updated the pull request.

chasestarr:
Thank you

PR #2977, opened 2020-03-04
fix(ci): exclude zeit.co in link checker

chasestarr:


chasestarr:
I had previously assumed it would exclude it.. not sure why

chasestarr:
hmm looks like making this change excludes all links

gergelyke:
do we still want to land this?

PR #2978, opened 2020-03-04
fix(side-nav): pass item to StyledNavItemElement

gergelyke:
Before version 9.31, one could do something like this to override the side-nav:

```
overrides={{
        NavItem: {
          style: ({item: {isInvalid}}) => ({
            color: isInvalid ? 'red' : 'black',
          }),
        },
      }}
```

after 9.31, that's no longer possibe

### DIFF
--- a/src/side-navigation/__tests__/nav-override.scenario.js
+++ b/src/side-navigation/__tests__/nav-override.scenario.js
@@ -1,0 +1,40 @@
+/*
+Copyright (c) 2018-2020 Uber Technologies, Inc.
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+// @flow
+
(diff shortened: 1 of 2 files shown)


sandgraham:
Note, you can achieve this by passing through `item` manually:

```js
<Navigation
  items={[
    {
      title: "111",
      itemId: "111",
      isInvalid: true
    },
    {
      title: "222",
      itemId: "222",
      isInvalid: false
    }
  ]}
  activeItemId={activeItemId}
  onChange={({ item }) => setActiveItemId(item.itemId)}
  overrides={{
    NavItem: {
      props: props => ({
        ...props,
        $item: props.item
      }),
      style: ({ $item }) => ({
        color: $item.isInvalid ? "red" : "black"
      })
    }
  }}
/>
```

PR #2989, opened 2020-03-05
fix(table-grid): provides suggestion to improve accessibility

chasestarr:
This PR includes a react-hook called `useCellNavigation` which applies key handlers to cells in a table. Not only does it handle even matrices, it also operates fine when 'holes' exist (column/row spans). Cell navigation is applied to all scenarios and a new example was created. The hook is not published at the moment.. since I think it may be too opinionated in functionality but can serve as a good starting point.
Screen Shot 2020-03-05 at 3 37 12 PM

#### Scope

- [x] Patch: Bug Fix


chasestarr:
> Works well! I think we should export this hook with the component if it can't be built-in.

Thanks - I worry about this being too opinionated and inflexible to officially publish. The code is available to copy from the example if wanted.

PR #2992, opened 2020-03-06
feat(input): accept number type for value

gergelyke:


src/payment-card/utils.js
```diff
@@ -8,7 +8,7 @@ LICENSE file in the root directory of this source tree.
 
 import valid from 'card-validator';
 
-export const addGaps = (gaps: number[], value: string) =>
+export const addGaps = (gaps: number[], value: string | number) =>
```
chasestarr: you may want to coerce the argument to a string rather than change the function signature

gergelyke: that is a very good point, thank you

PR #2997, opened 2020-03-07
fix(dropdown): defer initial autoFocus and add focus lock

tajo:
The problem was that popover initially sets position to 0,0 (or even negative), does some measurements and then sets the correct coordinates. If you combine that with the focusLock and autoFocus the page jumps to the top.

The solution is to delay `autoFocus` until the popover gets some non-zero/non-negative position. A bit hacky but couldn't find a better solution.


src/popover/types.js
```diff
@@ -174,6 +174,7 @@ export type PopoverPrivateStateT = {
   placement: TetherPlacementT,
   isLayerMounted: boolean,
   isMounted: boolean,
+  doAutoFocus: boolean,
```
chasestarr: nit: to me, `doAutoFocus` does not really communicate the correct meaning. maybe consider `autoFocusAfterPositioning` or similar

houmark:
Could this also affect `Modal`'s in some similar situations? I've seen the modal scroll to the top sometimes, especially in Safari, and we have autoFocus on most forms in those modals.

PR #3007, opened 2020-03-09
feat(styles): add withWrapper to base module

sandgraham:
Adds the `withWrapper` styling util to the base level import path.


chasestarr:
We probably want good type support before this is published

PR #3010, opened 2020-03-10
fix(datepicker): improve a11y

nadiia:
Keyboard navigation:
- fix tabbing in the calendar when a display month changed
- fix the calendar focus on the arrow key down from the input when the focus previously changed by a click into the input element


src/datepicker/__tests__/datepicker-unreliable.e2e.js
```diff
@@ -136,8 +135,7 @@ describe('Datepicker', () => {
     await page.click(selectors.rightArrow);
     const value = await page.$(selectors.monthYearSelectButton);
     const text = await page.evaluate(element => element.textContent, value);
-    // (Month YearTriangle Down) because it renders an icon within the element
-    expect(text).toBe('December 2030Triangle Down');
+    expect(text).toBe('December 2030');
```
gergelyke: 👏

src/datepicker/calendar-header.js
```diff
@@ -405,7 +407,7 @@ export default class CalendarHeader extends React.Component<
         >
           {monthYearTitle}
           <MonthYearSelectIconContainer {...monthYearSelectIconContainerProps}>
-            <TriangleDown />
+            <TriangleDown title="" />
```
gergelyke: shouldn't we use `aria-hidden` instead?

tajo: and `role="presentation"`

tajo:
works nicely!

src/datepicker/calendar-header.js
nadiia: yeah, I can add the 'presentation' role but the title still seems not useful here

PR #3016, opened 2020-03-11
docs(theming): respect media prop

konradkalemba:
#### Description

Documentation site's 'Theming' page provides misleading information regarding `mediaQuery` property usage. This pull request makes `Breakpoint` component to respect `media` prop when passing concern prop to `Property` component, therefore fixes the bug.

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/documentation-site/components/theming/breakpoints.js
+++ b/documentation-site/components/theming/breakpoints.js
@@ -20,7 +20,7 @@ export function Breakpoint({
   return (
     <Property
       name={name}
-      concern="breakpoints"
+      concern={media ? 'mediaQuery' : 'breakpoints'}
       renderValue={() =>
         media
           ? LightTheme.mediaQuery[name]


konradkalemba:
OK, so while we are waiting for @sandgraham's review, let me explain more in detail what, in my opinion, the bug is:
Currently in [Theme Properties](https://baseweb.design/guides/theming/#theme-properties) there are two subsections (listed below) that refer to the same properties - `theme.breakpoints.[size]`, but both return different values.

https://baseweb.design/guides/theming/#breakpoints
image

https://baseweb.design/guides/theming/#media-query
image

In [Media Query](https://baseweb.design/guides/theming/#media-query) subsection wrong property names are used. `theme.mediaQuery.[size]` should be written there.

I actually wasted some time figuring out why media queries don't have any effect. Then I realized that instead of CSS media query, `theme.breakpoints.[size]` returns a width - which is completely expected, but reading the documentation made me believe that I should use it to get media query.

chasestarr:
Oh right - thank you! Sorry about misunderstanding

PR #3017, opened 2020-03-11
feature(timepicker): applies parameterized generics to date-io adapter

chasestarr:
#### Description

Based on https://github.com/uber/baseweb/pull/2968 where @cbranch101 did the bulk of implementation work.

@cbranch101 Check out this diff for where my head is at on the date-io feature, and please let me know what you think. Relevant changes from your PR are in the last commit. The significant changes include applying generics to the adapter type and the timepicker props and shifting the adapter injection from context to props.

An earlier issue is how the timepicker's `onChange` [handler type](https://github.com/uber/baseweb/pull/2968/files#diff-0c14dacf7afab82607bb673499a78e3dR30) was changed to include either a moment class or js date. Given the code example below, application code would not be able to statically infer the argument type therefore requiring type refinement that previously was not required. There's also no way to infer a component's position within a react tree to ensure that it sees a moment adapter ancestor.

```jsx
function MyComponent() {
  function handleChange(time: moment$Moment | Date) {
    // line below will fail flow check because
    // moment class does not have getDate method
    console.log(time.getDate());
  }
  return (
    <Timepicker onChange={handleChange} />
  );
}
```

The 'fix' in this case was to pull the adapter out of context and into statically analyzable props. Also notice [this](https://github.com/uber/baseweb/compare/pr/2968?expand=1#diff-5dbd815ddb875b3630974a915b5b8140R25) type annotation to propagate moment types through to event handlers. Please this diff into consideration to pull in some changes into your PR. No need to merge this one. I just wanted to share some actual code with you. Let me know if this approach can work for your use-case


cbranch101:
I can definitely see where this makes the typing more specific eliminates the need to refine types in onChange, but it also makes swapping out the date processing quite a bit more complicated(one provider vs a unique prop for every component that needs to use date processing).
			
For our use case, it seems like that trade-off might not be worth it, but just to make sure I understand the problem correctly, I was hoping I could clarify a couple of things. 

As far the example you provided, there's one thing that's confusing me.  I was under the impression that the flow types were only for use within the `baseweb` repo(they're not really exported, etc), while Typescript types are used for the external api that can actually be consumed in the apps.  That assumption was what made me think this change was acceptable.  While it would require new type refinements for all Timepicker calls within `baseweb`(of which there's only the one anyway), that didn't seem like too big of an annoyance.  For app users, it seems we could just make the type script prop types generic to solve this issue. 

Also, for the `Timepicker` calls within `baseweb` I think treating it as a generic type that can be either a normal date or a moment is actually probably easier to follow, and closer to the reality that the library is facing.  It really could be either type, and we need to handle for it accordingly, but then once we're in app land(and it's TS types) we can make it more specific. 

Let me know if this makes sense or not, and thanks again for all of the time you've put in to get this merged in.

chasestarr:
Thanks for reviewing the proposal. We do publish flow types and it's used almost exclusively at uber. See the [accordion component's npm published build for example](https://unpkg.com/browse/baseui@9.63.0/accordion/). Please keep in mind that the flow types are being relied on by ~500 or so applications. Introducing type checking errors will result in CI failures and would slow down teams from upgrading the package to take advantage of new features/bug fixes.

I agree that this approach is more cumbersome that the context alternative, and I would be very happy to merge in a solution that addresses the downsides of both. I think investigating generic date type along with context may prove fruitful, but I ran into challenges in coordinating the type defined at context instantiation and consumption. The problem with context is that we can't link the position in the react tree to a component's call-site, which results in flow not able to infer what `DateT` is at runtime. 

To be honest I'm not very familiar with TS. Will it ask for refinement in the `onChange` example above? How would you handle that in your application? Do you need to explicitly cast the `time` type to `moment$Moment`? Flow in that situation would not allow simply that, but you could probably cast through `any` then to moment type.

cbranch101:
Yep, I had no idea you guys were using these flow types in your internal apps, this helps a lot in clarifying the need for not changing the external type api at all.  I'll mess with it and see if can find a way using context. 

As far as the TS questions, I'm pretty shaky there myself.  I was pretty much just following the typing from the theme provider, it might be something we need to go a little deeper on once there's a solution for the flow typing.

PR #3019, opened 2020-03-11
fix(data-table): screen reader improvements

tajo:
- "add filter" popover gets correct focusLock and styling
- filter menu items are now announced
- added resize-observer-polyfill since the data-table is currently crashing in Safari
- replaced some custom buttons with `baseui/buttons`

So overall major improvements to the filter menu UIs.

For the table cell navigation itself there is not much we can do right now since react-window keeps all cells in a single HTML element (they have some perf reasons). So the navigation works but the screen-reader use can't tell what the shape of table is. Not ideal.


src/data-table/column-categorical.js
```diff
@@ -57,43 +59,11 @@ function FilterQuickControls(props: {
...
-          cursor: 'pointer',
-        })}
-        onClick={props.onClearSelection}
-      >
-        Clear
-      </button>
-    </React.Fragment>
+    <ButtonGroup size={SIZE.mini} kind={KIND.minimal}>
+      <Button onClick={props.onSelectAll}>Select All</Button>
+      <Button onClick={props.onClearSelection}>Clear</Button>
```
chasestarr: These buttons should have `type="button"` I believe. Clicking on either submits the form.

src/data-table/filter-menu.js
```diff
@@ -127,6 +146,7 @@ function Options(props: OptionsPropsT) {
             // handled on the wrapper element
             // eslint-disable-next-line jsx-a11y/click-events-have-key-events
             <li
+              id={`bui-${index}`}
```
chasestarr: Maybe not that important, but probably will want to use the `getBuiId` utility because there could be id collisions

chasestarr:
I'm also running into the 'new dependency' fail on the flow check build script on another pr. Looking into how to best resolve that

PR #3026, opened 2020-03-12
fix(input): adds inputMode to props in typescript

grrowl:
#### Description

Adds `inputMode` prop to `Input` component — exists in `types.js` but is missing in `index.d.ts`

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


gergelyke:
thanks Tom!

PR #3041, opened 2020-03-14
fix(tree-view): use node-id as key instead of index

### LINKED ISSUE #2999: [Tree-view] Replace index with id as key
Index of tree node is set as key which needs to be replaced with getId(node).

## Current Behavior

On tree node mount, node id is stored in ref. When a node is replaced which has new id componentDidMount is not invoked because index is same and therefore new id is not stored in ref.

## Expected Behavior

Node id should be set as key. Secondly, we need to remove ref in componentWillUnmount. 

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v9.62.0  |
| React   |         |
| browser |         |

gwuah:
Fixes #2999

#### Description

This pull request fixes the issue in #2999 

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


tajo:
Thanks!

PR #3061, opened 2020-03-17
fix(input): fix color of inputTextDisabled for Input

loganyu:
#### Description

When the `Input` component is disabled, the color should equal the `inputTextDisabled` color. Currently, `inputTextDisabled` color is only used for the placeholder and `contentSecondary` is used for the disabled color.

The `Select` component uses `inputTextDisabled` for both disabled input text and disabled placeholder.

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/src/input/styled-components.js
+++ b/src/input/styled-components.js
@@ -318,7 +318,7 @@ export const InputContainer = styled<SharedPropsT>(
 function getInputColors($disabled, $isFocused, $error, colors) {
   if ($disabled) {
     return {
-      color: colors.contentSecondary,
+      color: colors.inputTextDisabled,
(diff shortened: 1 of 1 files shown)


nadiia:
Hi Logan, thank you for bringing up this potential difference with the latest Figma design and component. I created a JIRA story to revisit the colors and align it with the latest design and closing this issue here. Also the change to use the semantic `contentSecondary` color prop was intentianal, but we'll revisit it to make sure there is alignment with the input design.

loganyu:
Thanks @nadiia

PR #3069, opened 2020-03-18
feat(input): add inputPlaceholder and inputPlaceholderDisabled to themes

loganyu:
#### Description
This allows changing the color of the placeholder text for input.

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change


src/themes/dark-theme/color-component-tokens.js
```diff
@@ -357,6 +357,8 @@ export default (
   inputEnhancerFill: themePrimitives.mono500,
   inputEnhancerFillDisabled: themePrimitives.mono700,
   inputEnhancerTextDisabled: themePrimitives.mono500,
+  inputPlaceholder: themePrimitives.primary400,
+  inputPlaceholderDisabled: themePrimitives.primary400,
```
sandgraham: I think these should be `mono` colors since some folks will have a `primary` color that is blue or green or whatever.

src/themes/light-theme/color-component-tokens.js
```diff
@@ -181,6 +181,8 @@ export default (
   inputEnhancerFill: themePrimitives.mono300,
   inputEnhancerFillDisabled: themePrimitives.mono300,
   inputEnhancerTextDisabled: themePrimitives.mono600,
+  inputPlaceholder: themePrimitives.primary600,
+  inputPlaceholderDisabled: themePrimitives.primary600,
```
sandgraham: Same with these two.

src/themes/light-theme/color-component-tokens.js
```diff
@@ -181,6 +181,8 @@ export default (
   inputEnhancerFill: themePrimitives.mono300,
   inputEnhancerFillDisabled: themePrimitives.mono300,
   inputEnhancerTextDisabled: themePrimitives.mono600,
+  inputPlaceholder: themePrimitives.mono800,
+  inputPlaceholderDisabled: themePrimitives.mono800,
```
nadiia: I think these two values should nor be the same. 
Placeholder in a disabled state has `#AFAFAF` color, and  `#757575` when an input is enabled.

loganyu: I changed the disabled state colors to the same values as the `inputTextDisabled` for the light and dark themes.
Also, it looks like for the light theme, the placeholder color is `#545454` not `#757575`

src/themes/dark-theme/color-component-tokens.js
loganyu: I changed these to use the `mono` colors

src/themes/light-theme/color-component-tokens.js
nadiia: The values in my previous comment are from our latest and the current design ^

src/themes/dark-theme/color-component-tokens.js
```diff
@@ -357,6 +357,8 @@ export default (
   inputEnhancerFill: themePrimitives.mono500,
   inputEnhancerFillDisabled: themePrimitives.mono700,
   inputEnhancerTextDisabled: themePrimitives.mono500,
+  inputPlaceholder: themePrimitives.mono200,
```
nadiia:
```suggestion
  inputPlaceholder: themePrimitives.mono300,
```

PR #3074, opened 2020-03-19
fix(codesandbox-focus): make sure all focusVisible flags are global

tajo:
Somehow (can't explain), when in the CodeSandbox (only), these top level flags are different inside the `isFocusVisible` function compare to the module level. This ensures it's always same.

There must be some simple explanation but I am not seeing it.

I tested it here: https://codesandbox.io/s/basic-usage-n7e2p


src/utils/focusVisible.js
```diff
@@ -67,7 +74,7 @@ function handleKeyDown(event) {
   if (event.metaKey || event.altKey || event.ctrlKey) {
     return;
   }
-  hadKeyboardEvent = true;
+  window[hadKeyboardEventKey] = true;
```
sandgraham: Do we have to check for `window` for ssr?

sandgraham: nevermind lol

PR #3075, opened 2020-03-19
fix(select): support backspace/delete for single select

chrisirhc:
#### Description

Previously, backspace key would do nothing. This change fixes that by simulating a backspace so it renders the input's label without the last character. This aligns the behavior with what's typically expected in an input.

This change also fixes the behavior for deleteRemoves on single select, so that it actually removes.

We discovered this issue after receiving customer feedback that users weren't able to edit the input after a selection was made. We found this behavior closer to user expectations when leveraging `select` as a search with autocomplete functionality.

#### Scope

- [X] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


gergelyke:
thank you @chrisirhc - can you help us fix the eslint issue, and merge in the screenshots?

src/select/__tests__/select-backspace-behavior.e2e.js
```diff
@@ -0,0 +1,31 @@
...
+  it('backspace one character', async () => {
+    await mount(page, 'select-backspace-behavior');
+    const selector = `#backspace-behavior ${SELECT_INPUT}`;
+    await page.waitFor(selector);
+    const input = await page.$(selector);
+
+    await input.type('a');
+    await page.keyboard.press('Enter');
+    const select = await page.$eval(selector, i => i.value);
+    // expect(select).toBe('');
```
chasestarr:
```suggestion
```

src/select/__tests__/select-backspace-behavior.scenario.js
```diff
@@ -0,0 +1,40 @@
...
+            {id: 'Aqua', color: '#00FFFF'},
+            {id: 'Aquamarine', color: '#7FFFD4'},
+            {id: 'Azure', color: '#F0FFFF'},
+            {id: 'Beige', color: '#F5F5DC'},
+          ]}
+          labelKey="id"
+          valueKey="color"
+          backspaceRemoves={true}
+          onChange={({value}) => {
+            debugger;
```
chasestarr:
```suggestion
```

src/select/select-component.js
```diff
@@ -531,17 +531,33 @@ class Select extends React.Component<PropsT, SelectStateT> {
...
+    const item = this.popValue();
+    if (!item) {
+      return;
+    }
+    const valueLength = this.props.value.length;
+    const renderLabel = this.props.getValueLabel || this.getValueLabel;
+    const labelForInput = renderLabel({option: item, index: valueLength - 1});
+    // label might not be a string, it might be a Node of another kind.
+    if (typeof labelForInput === 'string') {
+      const remainingInput = labelForInput.slice(0, -1);
```
chasestarr: Can the `remainingInput` value be configurable based on the `backspaceRemoves` prop?

If `backspaceRemoves === true` `remainingInput` should be an empty string, completely reseting the input value
otherwise, pop the last character like you have here.

chrisirhc: I looked into this a bit more as I was implementing the fix and found that `backspaceRemoves` actually toggles whether the backspace key can remove items or not. The change in behavior you proposed changes the meaning of `backspaceRemoves` and would be a breaking change in behavior, since it would mean that true/false values would always allow backspace key to remove the item.

chasestarr: Good point. It does not make sense to repurpose the existing flag. I was concerned with how multi-select currently works with `backspaceRemoves`, which does not repopulate the input value. This PR does change default behavior in that way. To make for a completely non-breaking change we will need to include a new prop to control this interaction and default it to clearing the value.

src/select/__tests__/select-backspace-behavior.e2e.js
chrisirhc: Doh.

src/select/select-component.js
chrisirhc: Yep, this goes to whether the behavior is expected or if it's unexpected to decide that.
If we're not sure, a new option might be good (if that's not considered too heavy-handed). `backspaceRemovesCharacter`?

chasestarr: I tend to have a higher tolerance for making small behavior changes than others so I'm personally fine with landing, especially given that this seems like a preferable alternative. If it's requested, I can make a follow up to add the config. It's important to note that this change has come as a result of customer feedback to improve component UX and we believe this default behavior will make for better applications.

chrisirhc: Great point here. I'll update the description.

chrisirhc:
> The backspace behavior is inconsistent. In the case of a customized label, it removes the item and leaves input value empty, while with the default label it reuses previously selected value's label in the search input. See the screencast below. It's probably fine but needs to be documented in the `backspaceRemoves` property description.
Great point. This was intentional behavior since if the renderLabel returns something that's not a string, it would bail by removing the whole item. This is mainly because I'm not sure how to convert a label containing DOM into a string. Perhaps I should only be using `labelKey`?

> Also, a new search is not triggered when the dropdown is not opened on the first character removal.
Urgh, this is a miss and affects usability.

PR #3078, opened 2020-03-19
Release v9.64.4

chasestarr:


chasestarr:
> with `feat(vscode): show more theme prop values in tooltip hints ` merged, should it be a minor?

I'm not sure that we should dictate semantic versioning of the main baseui package based on the extension

gergelyke:
In that case, should we update https://github.com/uber/baseweb/commit/5241b04a6e0cbb14a001ce68bebfc7b4791bd485 to say, that we version based on the main package only?

Also, if we do that, how would you handle releases when we let's say, add things to the extension only?

PR #3080, opened 2020-03-19
fix(select): backspaceRemoves fixes

gergelyke:
fixes the issues Nadiia brought up in  #3075


chrisirhc:
Wow, you're so fast 🙏 . Thank you!

PR #3082, opened 2020-03-19
fix(build): only run deploy script on tags with semantic pattern

chasestarr:
testing the regex: https://regex101.com/r/sbLzfr/1/


gergelyke:
should we add it to the deploy-docs too?

PR #3084, opened 2020-03-20
fix(input): remove Button dependency

sandgraham:
Refactor the input password mask button so that it uses a simple styled component rather than our full fledged `Button` component. This keeps the `Button` out of your bundle if only importing `Input`.

Here are some pictures:

_Before (with Button in bundle)_
Screen Shot 2020-03-19 at 4 39 20 PM

_After (without Button in bundle)_
Screen Shot 2020-03-19 at 4 39 04 PM

Technically, this change could nullify some functionality passed through override props or styles to the old `Button` component but this edge case seems pretty minor to me.


gergelyke:
in dark mode,  shouldn't the eye stay the same color as before?

src/input/base-input.js
```diff
@@ -184,13 +184,15 @@ class BaseInput<T: EventTarget> extends React.Component<
     }[this.props.size];
     return (
       <MaskToggleButton
-        aria-label={label}
```
chasestarr: should `aria-label` be maintained?

src/input/base-input.js
```diff
@@ -184,13 +184,15 @@ class BaseInput<T: EventTarget> extends React.Component<
...
-        kind={KIND.minimal}
-        onClick={() => this.setState({isMasked: !this.state.isMasked})}
-        title={label}
```
chasestarr: as well as the title?

src/input/base-input.js
```diff
@@ -184,13 +184,15 @@ class BaseInput<T: EventTarget> extends React.Component<
...
-        type="button"
```
chasestarr: button type defaults to `submit` so removing this may cause regressions in application forms

src/input/base-input.js
sandgraham: oops - i forgot to add these props back in after trying something

gergelyke:
the build passed in buildkite, but it did not report status - merging it as is

PR #3091, opened 2020-03-20
fix(docs): adds another disabled version

chasestarr:


tajo:
Should we just not display this version since it was not published?

chasestarr:
> Should we just not display this version since it was not published?

That sounds fine to me. I'll make a note to update it this afternoon

PR #3100, opened 2020-03-21
fix: properly override clearicon styles and props

### LINKED ISSUE #3066: Select component's ClearIcon override doesn't work
Styling for ClearIcon from Select component cannot be overridden.

Repro: https://codesandbox.io/s/select-2igju

## Current Behavior

Following the docs: 
image

shows that styles cannot be overridden.

## Expected Behavior

Styles can be overridden.

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v9.64.2  |
| React   |     16.8.6     |
| browser |    chrome     |

- [ yes ] I have searched the [issues](https://github.com/uber/baseweb/issues) of this repository and believe that this is not a duplicate.

sendilkumarn:
Fixes #3066

#### Description

Styling for ClearIcon from Select component can be overridden.

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/src/select/select-component.js
+++ b/src/select/select-component.js
@@ -763,12 +763,12 @@ class Select extends React.Component<PropsT, SelectStateT> {
           Svg: {
             component: StyledClearIcon,
             props:
-              overrides.StyledClearIcon && overrides.StyledClearIcon.props
-                ? overrides.StyledClearIcon.props
+              overrides.ClearIcon && overrides.ClearIcon.props
+                ? overrides.ClearIcon.props
                 : {},
             style:
-              overrides.StyledClearIcon && overrides.StyledClearIcon.style
-                ? overrides.StyledClearIcon.style
+              overrides.ClearIcon && overrides.ClearIcon.style
+                ? overrides.ClearIcon.style
                 : {},
           },
         }}


src/select/select-component.js
```diff
@@ -763,12 +763,12 @@ class Select extends React.Component<PropsT, SelectStateT> {
           Svg: {
             component: StyledClearIcon,
             props:
-              overrides.StyledClearIcon && overrides.StyledClearIcon.props
-                ? overrides.StyledClearIcon.props
+              overrides.ClearIcon && overrides.ClearIcon.props
```
nadiia: Since `ClearIcon` is a react component and not a styled element it should not use props or styles that are passed on `ClearIcon` directly but the props and styles from `ClearIcon`'s overrides for the `Svg` element.
We need to extract overrides from `clearIconProps` and merge it with our default overrides:
```
import {getOverrides, mergeOverrides} from '../helpers/overrides.js';
...
const clearIconOverrides = mergeOverrides(
      {Svg: {component: StyledClearIcon}},
      clearIconProps && clearIconProps.overrides,
    );
clearIconProps.overrides = clearIconOverrides
```
and remove the `overrides` prop at the line 762.

sendilkumarn: Thanks for the comment. It uses the `ClearIcon` which will be (defaulted to) a component from `baseUI/icon`.  Also the `svg` override uses the `styled` element and we apply the props and styles to that right?

p.s. Also isn't this similar to `renderArrow()` below?  

Also it doesn't work when we do the `mergeOverrides`.

nadiia: It is similar to the `renderArrow` and it has the same issue. If you could fix both of them with this PR that would be really helpful.
Could you give more details on what doesn't work when you use `mergeOverrides`?
To clarify what the overrides for the `svg` element will look like:
```
<Select overrides={{
  ClearIcon: {
    props: {
      // top level Base Web's [icon component](https://baseweb.design/components/icon/) props (`DeleteAlt` in our case)
      size={24}, 
      overrides: {  
        Svg: {
          props: {'data-attr': 'data-attr-val'}, 
          style: {cursor: 'default'}},
      }
    }
  }
}}>
```

nadiia: To summarize our zoom meeting notes: 
- we decided to ship this fix as is since the overrides implementation for the `ClearIcon` and `SelectArrow` was introduced incorrectly before the latest bug where `ClearIcon` was replaced with `StyledClearIcon`,
- we'll be shipping the overrides fix as discussed above as a follow-up PR.

PR #3107, opened 2020-03-23
feat(form control): add input to themes

loganyu:
#### Description
This allows changing the color of the caption text instead of having to use the same color as `contentSecondary`

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/documentation-site/components/yard/config/form-control.ts
+++ b/documentation-site/components/yard/config/form-control.ts
@@ -15,7 +15,7 @@ const TextareaConfig: TConfig = {
     FormControl,
     Input,
   },
-  theme: [],
+  theme: ['caption'],
   props: {
     children: {
       value: '<Input />',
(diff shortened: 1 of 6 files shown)


loganyu:
@sandgraham I updated the color. I think I needed to rebase to get the updated color.

sandgraham:
@loganyu Just unblocked the CI run. It should generate new vrt images for you.

nadiia:
I actually don't think we should replace this semantic color with another new component specific prop. I can see how more granular overrides it would provide but also how it can create inconsistency across the components.

sandgraham:
@loganyu I'm going to close this PR because adding more of the component specific theme props conflicts with our plans for the theme. This wasn't something we documented anywhere explicitly (that I know of) so I'm sorry to have to revoke your work.

For now I recommend setting your `contentSecondary` or using overrides to customize your form control. Like I wrote earlier, we have plans to tweak how theming works so we can make cases like this easier but are not ready to share anything yet.

I hope we will see more of your contributions in the future!

loganyu:
thanks @sandgraham for the explanation! I'll use one of the workarounds.

PR #3111, opened 2020-03-24
docs(spinner): add example showcasing $as prop

sandgraham:
Adds a new example to the `Spinner` docs to show how you can manipulate the underlying `StyledSpinnerNext` element.

### DIFF
--- a/documentation-site/examples/spinner/next-span.js
+++ b/documentation-site/examples/spinner/next-span.js
@@ -1,0 +1,10 @@
+// @flow
+
+import * as React from 'react';
+import {StyledSpinnerNext} from 'baseui/spinner';
+
+export default () => (
+  <p>
+    <StyledSpinnerNext $as="span" $style={{display: 'block'}} />
+  </p>
+);
--- a/documentation-site/examples/spinner/next-span.js
+++ b/documentation-site/examples/spinner/next-span.js
@@ -5,6 +5,6 @@ import {StyledSpinnerNext} from 'baseui/spinner';
 
 export default () => (
   <p>
-    <StyledSpinnerNext $as="span" $style={{display: 'block'}} />
+    <StyledSpinnerNext $as="span" />
   </p>
 );
--- a/documentation-site/pages/components/spinner.mdx
+++ b/documentation-site/pages/components/spinner.mdx
@@ -81,7 +81,7 @@ Spinners provide a visual cue that an action is processing awaiting a course of 
 
 ## Spinner Next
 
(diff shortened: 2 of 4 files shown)


documentation-site/examples/spinner/next-span.js
```diff
@@ -0,0 +1,10 @@
+// @flow
+
+import * as React from 'react';
+import {StyledSpinnerNext} from 'baseui/spinner';
+
+export default () => (
+  <p>
+    <StyledSpinnerNext $as="span" $style={{display: 'block'}} />
```
nadiia: I'd incline not to use `$style` in the examples. Can we replace it with `withStyle`?

documentation-site/pages/components/spinner.mdx
```diff
@@ -111,3 +112,9 @@ If you want to customize the default styling of the spinner you can extend it us
 </Example>
 
 In this example we modify the size of the spinner (via `width`, `height`, and `borderWidth`) as well as the highlight color (via `borderTopColor`). You could also use `withStyle` to add margins to the spinner or align it within another element.
+
+<Example title="$as" path="spinner/next-span.js">
+  <SpinnerNextSpan />
+</Example>
+
+By default the spinner styles a single `div` element. You can change the underlying element type by using the `$as` prop. In this example we change it to a `span` and add `display: block`. This allows us to use the `Spinner` inside of a `p` element. Note, we could also set `display: inline-block` to prevent the spinner from wrapping to a new line.
```
nadiia: While helping someone who brought up the issue mentioned here ^ I also notices that the API for this styled element is different from the current `Spinner` component but we didn't document it anywhere. Could you add it here as well?

documentation-site/examples/spinner/next-span.js
gergelyke: just out of curiosity, what's wrong with `$style`? we document it on the styletron website (https://www.styletron.org/react/#style-prop), so I think one would assume it's part of the public API?

sandgraham:
@nadiia I changed this up a bit. I added `display: block` to the default styling, which shouldn't be breaking since it is already `div` default, but lets you change to `span` without having to set and styles. Also updated some of the text in this section to make it easier to read (imo).

documentation-site/examples/spinner/next-span.js
nadiia: I know that we document it in Styletron docs, but it's not a part of the base web API we've been encouraging people to use. I also think it may create more confusion around how and when one would use it. It's more of an internal implementation imho.

PR #3118, opened 2020-03-25
fix(docs): fix incorrect breakpoint documentation

colinricardo:
Fixes incorrect large breakpoint value in documentation.

#### Description

As per `src/themes/shared/breakpoints.js`, the large breakpoint value is actually `1136` not `1280`.

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


colinricardo:
I can't add a label to this, but it's `bugfix`.

PR #3128, opened 2020-03-27
fix(codesandbox): fix DarkTheme, focus-ring, add CI integration

tajo:
- fixes the CodeSandbox setup, `baseui/esm` needs to be used when importing `BaseProvider` so everything is importing from the same source, otherwise things like React context doesn't work, this is a problem only in CodeSandbox since its bundler doesn't respect our esm mapping (Webpack does)
- reverts changes made to not displaying focus-ring,  it was a workaround for this issue... the mystery is solved!
- adds CodeSandbox CI, so every PR should come up with sandbox that will have PR's `baseui` built and loaded

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "build:copy-files": "shx cp README.md dist/README.md",
     "build:copy-flow-files": "node ./scripts/flow-copy-src.js",
     "build:copy-ts-files": "node ./scripts/ts-copy.js",
-    "build:package-json": "shx cp package.json dist/package.json & node ./scripts/build-package-json.js",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -26,7 +26,7 @@
     "build:copy-files": "shx cp README.md dist/README.md",
     "build:copy-flow-files": "node ./scripts/flow-copy-src.js",
     "build:copy-ts-files": "node ./scripts/ts-copy.js",
-    "build:package-json": "shx cp package.json dist/package.json & node ./scripts/build-package-json.js",
+    "build:package-json": "shx cp package.json dist/package.json && node ./scripts/build-package-json.js",
```
gergelyke: good catch!

Issue #3099, opened 2020-03-20
[DataTable] Application should know what rows are rendered


chasestarr:
Suppose I need to export the table to csv after filter and sorting. This is not possible at the moment.

It can be achieved by adding a callback function (maybe named `onIncludedRowsChange` or similar) to the component called when `rows` [is recalculated](https://github.com/uber/baseweb/blob/master/src/data-table/data-table.js#L783).

Deliaz:
Hi Chase, 
It sounds as a missing feature that we also need in our project. 
We plan to download the table data and to re-calculate some numbers based on what's left after filters have been applied. Just wondering, is there any ETA?
Thank you :)

chasestarr:
I plan to start looking into it this week

PR #3130, opened 2020-03-27
feat(data-table): add onIncludedRowsChange callback

chasestarr:
closes: https://github.com/uber/baseweb/issues/3099

#### Description

Adds a new prop to DataTable that is called when the included rows in the table change. This is useful if the application needs to do something with the filtered/sorted 'view' that the user has configured. Please note, that it's important to have a stable function provided to this prop. See [here](https://github.com/uber/baseweb/blob/893c273df08a95a55a03a2eb3821c319249f0170/src/data-table/__tests__/data-table-included-rows-change.scenario.js#L38) for an example. If the function is in-lined, it causes an update loop.

#### Scope

- [x] Minor: New Feature


chasestarr:
Typescript defs on `data-table` are way out of date. I'll update it once it's 'stable'

PR #3132, opened 2020-03-28
fix(deps): remove main field

tajo:
This should make both CodeSandbox and Webpack happy. We don't need this field since our main file is `./index.js` anyway.


tajo:
ok, this makes sense to me: https://codesandbox.io/s/romantic-star-1fdqk

PR #3137, opened 2020-03-29
Update ListItem to be a ForwardRefExoticComponent [typescript]

grrowl:
#### Description

Updates `ListItem`'s typescript definition to match implementation, which is that `ListItem` forwards refs to its inner `HTMLLIElement`. Now we can do:

```tsx
<ListItem ref={ref} {...props}>
```

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


grrowl:
@chasestarr done :+1: cheers

PR #3140, opened 2020-03-29
feat: added a check before overriding isActive

### LINKED ISSUE #3139: [ProgressStep] We should let Step component override isActive.
ProgressStep is overriding the Step's prop isActive. So the property on the Step is useless as it is never used. I suggest we let the Step component override this prop.

## Feature description
Screenshot 2020-03-29 at 11 16 15 PM

YashGadle:
Fixes #3139

#### Description

Step's isActive was always being overriden by ProgressStep

#### Scope

- [ ] Patch: Bug Fix
- [X] Minor: New Feature
- [ ] Major: Breaking Change


src/progress-steps/progress-steps.js
```diff
@@ -23,12 +23,17 @@ function ProgressSteps({
     if (!child) return;
 
     const childOverrides = child.props.overrides || {};
+
+    const isActive = child.props.isActive
```
gergelyke: we may want to check using "typeof", so one can input false too

YashGadle: yes, makes sense. fixed it in latest commit

sandgraham:
@YashGadle Could you also add a visual regression test that covers why you are adding this functionality? You can do so by adding a new storybook scenario to the `src/progress-steps/__tests__` directory.

Issue #2884, opened 2020-02-19
[Select] option with href property renders as anchor in dropdown


chasestarr:
This is an unintended side-effect of https://github.com/uber/baseweb/pull/2631 which adds anchor rendering to the `Menu` component. I can't think of a use-case for rendering the anchor in the select dropdown. We should probably disable this feature when rendered from the select component since it's somewhat of a regression on existing applications

## Current Behavior

options with the `href` property are rendered as an anchor in the dropdown
https://codesandbox.io/s/select-xr8t9

## Expected Behavior

options should not be rendered as anchors

| Tech    | Version |
| ------- | ------- |
| Base UI | v9.55.x  |

rohitb4:
Do existing applications use href within select, I thought #2631 introduced href in menu and select.

chasestarr:
I edited the original message to be more clear. It doesn't make much sense to render an anchor element in a select dropdown. The select component is supposed to act similarly to html select element, which does not cause a page navigation when an option is clicked

PR #3142, opened 2020-03-30
fix(select): disable select href render as anchor

chasestarr:
Fixes: https://github.com/uber/baseweb/issues/2884

#### Scope

- [x] Patch: Bug Fix


chasestarr:
The `Menu` component is heavily dependent on the overrides interface for configuration. If you look at that docs page now, you'll find that there is very little props for such a complicated component

Issue #2836, opened 2020-02-11
[Select] in loading state shows spinner deprecation warning


chasestarr:
Developers can't take any action on spinners nested in other components. We need to either silence the error where we want to or move to the new spinner in select. Main downside of moving to the new spinner is that it does not expose the same overrides

PR #3143, opened 2020-03-30
fix(spinner): only warn on usage outside of baseui

chasestarr:
fixes: https://github.com/uber/baseweb/issues/2836

### DIFF
--- a/src/file-uploader/file-uploader.js
+++ b/src/file-uploader/file-uploader.js
@@ -157,7 +157,7 @@ function FileUploader(props: PropsT) {
                         />
                       ) : (
                         <Block marginBottom="scale300">
-                          <Spinner size={40} />
+                          <Spinner silenceV10DeprecationWarning size={40} />
                         </Block>
                       )}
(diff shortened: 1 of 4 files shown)


AndrewUldin:
@chasestarr could you please also take a look at this warning? I'm getting it after updating to 9.67.2

`Warning: React does not recognize the `silenceV10DeprecationWarning` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `silencev10deprecationwarning` instead. If you accidentally passed it from a parent component, remove it from the DOM element.`

skbogner:
Me too.

gergelyke:
@AndrewUldin would you mind creating a codesandbox for this? tried to repro it on the doc site, but couldn't (I assume, somewhere it's passed down to a Styletron component)

AndrewUldin:
@gergelyke yes, sure: https://stackblitz.com/edit/react-crzz7k?file=Hello.js
`isLoading={true}` does that.

Issue #3067, opened 2020-03-18
[datepicker] After setting date by clicking [Day], and then changing the value of the Calendar component, leaves the startDate selected by click highlighted. This feels like a glitch.


minseoksuh:
(Our team use version 9.0.0 datepicker from base ui, 
but I think the same issue is there in the most recent version)

As described in the title,
If you set Date range by clicking the days (Event), 
and then if you change the date range through "value" props of Calender component 
(without Event: like quickSelect)
then you see a highlight remaining in the calendar for the startDate of the "Clicked Range."

Users felt like that is a glitch and a bug. 
And it does seem weird.

However, as I looked into your code base, I just thought you guys are not resetting the highlight value when the "value" props update.

When, I just resetted the highlightedDate state in componendDidUpdate,
I felt everything work generally fine without the problem people found strange.
<img width="918" alt="스크린샷 2020-03-18 오후 6 53 03" src="https://user-images.githubusercontent.com/61446966/76947975-c2cd8780-6949-11ea-86e9-9048613ef63b.png">

## Feature description

Is it possible that you guys allow another option for the datepicker
like "resetHighLightOnChange" for the Calendar component
when "true"

that does something like this?
<img width="620" alt="스크린샷 2020-03-18 오후 6 56 59" src="https://user-images.githubusercontent.com/61446966/76948294-45eedd80-694a-11ea-94c0-9ba4ac8d07c2.png">

This is my first time writing anything on other repositories, so please excuse me if my format was not proper.

I will fix and reissue if you guys want me to.

Hope you guys look into the issue.

Thanks!!

gergelyke:
@minseoksuh would you mind creating a code sandbox for this? you can fork this one https://codesandbox.io/s/7t0oh?module=/example.js

minseoksuh:
https://codesandbox.io/s/datepicker-y7xvr

I forked it and set up the setting in which the issue occurs.

If you select a range by clicking, and then if you select option in quickSelect,
the highlight for the startDate of the "Clicked range" remains. (Although the actual range has changed to select option range i.e. last week).
(Note: when I quickSelect in codeSandbox the whole picker sometimes disappear. This is not the scenario I have in mind. In our project the the calendar always stays there when we select "last week" )

Is this what you wanted me to do?

I would like to apply the changes I made to the basueui dependency, 
but I do not know how to do that in code sandbox.

Let me know what

chasestarr:
[...]

PR #3147, opened 2020-03-31
fix(datepicker): reset highlight on value change

minseoksuh:
https://github.com/uber/baseweb/issues/3067

#### Description

- added resetHighlightOnValueChange props for calender in datepicker/types.js
- added logic to reset highlight in componentDidUpdate in calendar.js
- wrote unit test for resetHighlihgtOnValue in calendar.test.js
- added description of the feature in documentation

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


gergelyke:
@chasestarr @minseoksuh I wonder if this should be the default behavior? that case, we wouldn't even have to introduce a new prop

chasestarr:
Yes, I agree that this can be the default

gergelyke:
@minseoksuh would you mind updating the PR?

nadiia:
A few things here:
- I couldn't reproduce the issue you described in the ticket on the latest version
- The codesandbox example in [#3067](https://github.com/uber/baseweb/issues/3067#issuecomment-601001375) doesn't demonstrate the issue described, also, it doesn't use the latest baseui version. Could you provide the repro of the issue with the latest version? 
- We should not add a new component property for this issue

minseoksuh:
> A few things here:
> 
> * The codesandbox example in [#3067](https://github.com/uber/baseweb/issues/3067#issuecomment-601001375) doesn't demonstrate the issue described, also, it doesn't use the latest baseui version. But I could repro the issue on the masters storybook
> * We should not add a new component property for this issue
> * We should set the highlighted value in the state to the first selected value if any, otherwise to the highlighted value that comes from props, if none is provided and no selected value then we set it to today's date.

1. I just added the quickSelect option to the calendar, and I updated the sandbox to use the latest version. When I did this, the issue action was not reproduced. However, in this case, today was highlighted even though the selected range have changed. (Different response from the original issue, still looks glitchy)

2. Ok I can get rid of the props.

3. I think it makes sense to have the first date of the currently selectedDate highlighted. Should I implement @nadiia 's suggestion and re-commit to this PR?

Thanks

src/datepicker/calendar.js
```diff
@@ -112,8 +111,16 @@ export default class Calendar extends React.Component<
...
 
-    if (prevProps.value !== this.props.value) {
+    if (prevProps.value !== this.props.value && !this.state.focused) {
+      if (this.props.value instanceof Date) {
+        this.setState({highlightedDate: this.props.value});
+      } else if (this.props.value instanceof Array) {
+        this.setState({highlightedDate: this.props.value[0]});
+      } else {
+        this.setState({highlightedDate: new Date()});
+      }
```
minseoksuh, line 121: Changed this part to incorporate Nadia's suggestion. Added !this.state.focused to the conditional because without it highlight on click (specifically clicking to select range) is bugged. The second click to select an end day of range will render the endDate as blank (without the number inside rendered). !this.state.focused condition makes sure the reset occurs only when value is changed through external measures (like quickselect)

src/datepicker/calendar.js
```diff
@@ -372,15 +379,8 @@ export default class Calendar extends React.Component<
...
-    if (selected && isSameMonth(selected, date) && isSameYear(selected, date)) {
-      nextState = {highlightedDate: selected};
-    } else {
-      nextState = {
-        highlightedDate: date,
-      };
-    }
-    this.setState(nextState);
+    if (selected) this.setState({highlightedDate: selected});
+    else this.setState({highlightedDate: new Date()});
```
minseoksuh, line 383: explained this part in the commit message.

nadiia:
Works well, and good to go after addressing the last couple comments 👍

minseoksuh:
> Works well, and good to go after addressing the last couple comments 👍

Yup.

Added the check logic and added comments for the update section.

Thanks for the feedback.

Please let me know if anything else needs to be done.

nadiia:
The issue is when keyboard navigating to a next/prev month there is no focusable date in the calendar in a new month rendered and therefore no way to tab into the calendar.
datepicker-keyboard-nav

minseoksuh:
> The issue is when keyboard navigating to a next/prev month there is no focusable date in the calendar in a new month rendered and therefore no way to tab into the calendar.

Allowed the user to tab into Sunday's of the calendar (if that is ok) even if it is not highlighted.
Also, when tabbing out of the calender's day section, the highlight remained, so added a logic to reset the highlight to today's date in that situation.

nadiia:
2020-04-14 at 3 33 42

minseoksuh:
> It adds a regression in the navigation, seel below:
> 2020-04-14 at 3 33 42

Uh don't understand the problem.

Can you please describe the proper behavior in more detail?

nadiia:
Sorry about it. When a calendar day is focused and the tab key is pressed it should focus on the next focusable element outside of the calendar month. So it would be the quick select input if any or the first focusable element (previous month button) in the calendar's header. In the current change tho it tabs through the days in the calendar before reaching anything outside of the calendar. I added an e2e test for it in #3207

minseoksuh:
> Sorry about it. When a calendar day is focused and the tab key is pressed it should focus on the next focusable element outside of the calendar month. So it would be the quick select input if any or the first focusable element (previous month button) in the calendar's header. In the current change tho it tabs through the days in the calendar before reaching anything outside of the calendar. I added an e2e test for it in #3207
> 
> Basically, only a single day at a time in a calendar month should have a tabindex=0. If there is no user-highlighted date it should be the 1st of the month.

Fixed the behavior. Thanks for the detailed explanation.

minseoksuh:
@nadiia Hi just wondering how the PR is going. Let me know if anything else need to be done.
Thanks!

nadiia:
Hi @minseoksuh, since this change affects too many details in the calendar navigation and, unfortunately, we didn't have enough tests to cover all those interesting and tricky behaviors, I assumed it would make sense to take care of it on our end. So I opened a PR (#3256) with a fix for the brought up issue. But I would like to thank you for starting working on the fix.
Please check out the built preview of the fix and I'll be closing this PR.

PR #3151, opened 2020-04-01
Release v9.67.1

gergelyke:


chasestarr:
FYI the last release was v9.66.0 https://github.com/uber/baseweb/releases/tag/v9.66.0 so this should be v9.66.1

gergelyke:
we've looked into this, and the problem was that our probot did not create a release for this

Issue #3103, opened 2020-03-23
Functional props override with ListItemAnchor breaks link


sandgraham:
https://codesandbox.io/s/menu-anchor-overrides-props-function-bug-7rswb

When you use the functional props override with Menu's `ListItemAnchor` the link becomes un-clickable. If you change to the object form of props override things work as expected.

[Current work around](https://codesandbox.io/s/temp-fix-for-props-override-on-menu-anchor-0vlq6) if you really need dynamic props is to use the component level override.

chasestarr:
The problem is that when we have a prop function, we dynamically create an HOC component in the `getOverrides` function. Whenever the component re-renders, it unmounts/mounts the component leading to the click event no longer having an associated dom element. Fixing this will require a significant refactor of how overrides work internally. It all comes down to how we instantiate override components in render, rather than statically creating the HOC. I've prototyped a solution below, but have not ironed out the edge cases

```js
// overrides.js
import {isValidElementType} from 'react-is';
function Override(Base) {
  const Result = React.forwardRef((allProps, ref) => {
    const {override, ...props} = allProps;
    if (!override) {
      return <Base ref={ref} {...props} />;
    }

    if (isValidElementType(override)) {
      const ComponentOverride = override;
      return <ComponentOverride ref={ref} {...props} />;
    }

    if (typeof override === 'object') {
      let Component = Base;
      if (isValidElementType(override.component)) {
        Component = override.component;
      }

      let overrideProps = {};
      if (typeof override.props === 'object') {
        overrideProps = override.props;
      } else if (typeof override.props === 'function') {
        overrideProps = override.props(props);
      }

      if (override.style) {
        overrideProps.$style = override.style;
      }

      return <Component ref={ref} {...overrideProps} />;
    }

    return <Base ref={ref} {...props} />;
  });

  return Result;
}
```

```js
// component.js
import {Override} from 'overrides.js';
const BaseComponent = Override(StyledBaseComponent);
function Parent(props) {
  return <BaseComponent override={props.override.BaseComponent} />
}
```

sandgraham:
Oh I really like this.

chasestarr:
We deprecated the ability to pass dynamic props function to overrides

PR #3153, opened 2020-04-01
fix(overrides): hoc pattern implementation

chasestarr:
Adresses: https://github.com/uber/baseweb/issues/3103

#### Description

This PR includes the HOC implementation and unit-tests. It has not been added to any components yet

#### Scope

- [x] Patch: Bug Fix


src/styles/__mocks__/styled.js
```diff
@@ -84,6 +84,7 @@ export function styled(
           ref={this.props.forwardedRef}
           styled-component="true"
           test-style={this.state.styles}
+          test-style-stringified={JSON.stringify(this.state.styles)}
```
tajo, line 87: Hm, do we really want to add this?  It creates a lot of noise. At least it should be formatted into multiple lines so future diffs are reviewable.

tajo:
Lgtm but the new style attribute should not end up in snapshots.

src/styles/__mocks__/styled.js
chasestarr: I can format it, but it with react-testing-lib getting the attr `test-style` will only say `[object: Object]`. Do you have any ideas for how to test?

chasestarr: I'm going to replace our mock with https://github.com/styletron/styletron/tree/master/packages/styletron-engine-snapshot

chasestarr:
removed the dup styles

PR #3157, opened 2020-04-02
fix(datepicker): single-picker value parsing with formatString

VladimirMilenko:
#### Description

Typing from the first character will change not the month, but the day.
This PR addresses that and parses the string according to formatString prop passed to it.
Example: https://codesandbox.io/s/datepicker-qqrck

Due to the fact, that dates like `21.12.2`[1] and `21.12.0002`[2] are from Date object perspective, formatting them straight away makes component unusable [1] immediately transforms to [2] in the input – formatting is only applied when the length of input date is equal to formatString. This is kind of "strict" formatting.

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


src/datepicker/datepicker.js
```diff
@@ -219,7 +219,16 @@ export default class Datepicker extends React.Component<
         }
       }
     } else {
-      const date = new Date(inputValue);
+      let date = new Date(inputValue);
```
gergelyke, line 222: shouldn't we use datefn's `parse` here instead?

VladimirMilenko: That's a good question. parse requires string format. In theory that should be `dd/MM/YYYY` as it's default for datepicker.

gergelyke: we could do something like `this.props.formatString || 'yyyy/MM/dd'` (I think this is the default, based on https://github.com/uber/baseweb/blob/master/src/datepicker/datepicker.js#L115)

VladimirMilenko:
> Visual changes were detected on this branch. Please review the following PR containing updated snapshots: [VladimirMilenko#2](https://github.com/VladimirMilenko/baseui/pull/2)

I've moved old `datepicker-int` test to `datepicker-int-range`

src/datepicker/datepicker.js
VladimirMilenko: Also, I think, that it doesn't really make sense. We're handling the parsing of date further, and keep it as default. 
As I can see, in range it's done same way.

gergelyke: for the range, we do this:

```
if (formatString) {
  startDate = parse(left, this.normalizeDashes(formatString), new Date());
  endDate = parse(right, this.normalizeDashes(formatString), new Date());
}
```

do you think it would make sense for the single select to be something like this

```js
const dateString = this.normalizeDashes(inputValue);
let date = new Date(datestring)

if (formatString) {
  date = parse(dateString, this.normalizeDashes(formatString), new Date());
}

      isValid(date) &&
        this.props.onChange &&
        this.props.onChange({
          date,
        });
```

VladimirMilenko: normalizeDashes does makes sense. But the if condition is a missing the case of early-parsing.
If you parse the date `11.12.2` it would be parsed as `11.12.0002` and will be replaced in input. So, the strict lenght check is needed.

gergelyke: understood, that makes sense - thanks Vladimir

src/datepicker/__tests__/datepicker-int.scenario.js
```diff
@@ -9,18 +9,14 @@ LICENSE file in the root directory of this source tree.
 import * as React from 'react';
 
 import {StatefulDatepicker} from '../index.js';
-import {hu} from 'date-fns/locale/index.js';
```
nadiia, line 12: Could you keep the locale usage in here?

src/datepicker/__tests__/datepicker-int.scenario.js
```diff
@@ -9,18 +9,14 @@ LICENSE file in the root directory of this source tree.
...
   return (
     <StatefulDatepicker
       aria-label="Select a date"
-      formatString="yyyy EEEE d"
-      placeholder="dd.mm.yyyy – dd.mm.yyyy"
-      locale={hu}
-      highlightedDate={new Date('March 10, 2019')}
-      mask={null}
-      range
+      formatString="dd.MM.yyyy"
```
nadiia, line 17: How about we create a separate scenario for a formatting test case and keep the localization one as is?

src/datepicker/datepicker.js
```diff
@@ -219,7 +219,19 @@ export default class Datepicker extends React.Component<
...
       }
     } else {
-      const date = new Date(inputValue);
+      const dateString = this.normalizeDashes(inputValue);
+      let date = new Date(dateString);
+      const formatString = this.props.formatString;
+      if (formatString) {
+        // Prevent early parsing of value.
+        // Eg 25.12.2 will be transformed to 25.12.0002 formatted from date to string
+        if (dateString.replace(/ /g, '').length < formatString.length) {
```
nadiia, line 228: What is the regex supposed to look for? Also, wouldn't whitespace be valid in some formatting cases?

VladimirMilenko: This regex is erasing whitespaces generated by masked input. But there is a good catch. I should actually compare replaced whitespaces for both of them.
Here is `mask="99 99 9999"` and `formatString="dd MM YYYY"`, if your input value is `"1"` then masked input will transform it to `"1         "` . 
So, we should compare `"1".length` to `"ddMMYYYY".length`

src/datepicker/__tests__/datepicker-int.scenario.js
VladimirMilenko: I was more leaning towards the point, that formatting is a part of internationalization. 
But if you think that would be preffered, then for sure i can move it to separate test.

src/datepicker/datepicker.js
nadiia: Correct, you'd need to trim the `dateString` or  get rid of spaces in both strings that are being compared. 
Regarding the space regex, I'm thinking we should look for any whitespace `/s`.

src/datepicker/__tests__/datepicker-int.scenario.js
nadiia: The locale is used in the formatting fn.

src/datepicker/__tests__/datepicker-int.scenario.js
VladimirMilenko: Done

src/datepicker/__tests__/datepicker-int.scenario.js
VladimirMilenko: Done

src/datepicker/datepicker.js
VladimirMilenko: Done + added extra test

src/datepicker/datepicker.js
```diff
@@ -174,7 +174,7 @@ export default class Datepicker extends React.Component<
     const inputValue = event.currentTarget.value;
 
     if (
-      inputValue === this.getMask().replace(/9/g, ' ') ||
+      inputValue === this.getMask().replace(/ /g, ' ') ||
```
nadiia, line 177: Is this change intentional? I think this condition supposed to check if the input is empty. With a mask `9999/99/99` the input

PR #3160, opened 2020-04-03
feat(themes): add createDarkTheme & createLightTheme

sandgraham:
- Adds two new exported functions: `createLightTheme` (aliased to `createTheme`) and `createDarkTheme`.
- Loosens some types (Flow/TS) which were overly restrictive.
- Renames a few things for consistency
- Updates docs.


nadiia:
> To allow users to create dark themes, I've added a third argument to createTheme, an options object which currently has only one property, isDark: boolean.

This doesn't seem to be scalable or have much value for OSS side. It ties it to our current light/dark themes set, but will have the same issue once we want to add/create a medium light, midnight dark, or pastel theme.

> I've changed light-theme and dark-theme to use createTheme.

I don't agree that our default themes implementation needs to use theme creator but we should keep it simple. Ideally, we should retire `createTheme` with its current interface (in v10) and it should provide a generic solution for creating a custom theme that is not considered an extension of one of our ready to use themes.

documentation-site/pages/guides/theming.mdx
```diff
@@ -157,20 +157,19 @@ As you might expect, `createTheme` is a factory function for Base Web compliant
...
-const theme = createTheme(/* primitives */, /* overrides */);
+const theme = createTheme(/* primitives */, /* overrides */, /* options */);
 ```
 
 #### `primitives`
 
 Our theme object has a lot of properties. Most of these properties are assigned to a subset of reoccurring values. We call these reoccurring values, theme building `primitives`. Primitives are made up mostly of colors, (`primary`, `accent`, etc), gradients of those colors (`primary100`, `primary200`, etc) as well as a `primaryFontFamily`. To see the exact shape for `primitives`, reference the [Flow](https://github.com/uber/baseweb/blob/master/src/themes/types.js#L534) or [Typescript](https://github.com/uber/baseweb/blob/master/src/theme.ts#L668) definitions.
 
-These are passed as the first (and only required) argument for `createTheme`.
+These are passed as the first argument to `createTheme`.
```
nadiia, line 167: I think the `required` part was just a misleading wording since it could be just an empty object.

src/index.d.ts
```diff
@@ -14,8 +14,9 @@ export function createThemedUseStyletron<Theme>(): UseStyletronFn<Theme>;
 export const useStyletron: UseStyletronFn<Theme>;
 
 export function createTheme<P extends object>(
-  primitives: ThemePrimitives,
+  primitives: Partial<ThemePrimitives>,
```
nadiia, line 17: 👍

src/themes/create-theme.js
```diff
@@ -0,0 +1,53 @@
...
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+// @flow
+
+import animation from './shared/animation.js';
+import borders from './shared/borders.js';
+import breakpoints from './shared/breakpoints.js';
+import darkBorders from './dark-theme/borders.js';
```
nadiia, line 12: As I mentioned in my previous comment we should only use as a base a single theme, whichever we consider as a default or base (currently it's LightTheme and it's also used as a default theme in the provider).
And with growing the theme base options the bundle size will also grow.

src/themes/create-theme.js
```diff
@@ -0,0 +1,53 @@
...
+
+export default function createTheme(
+  // Used to derive various theme properties.
+  primitives?: $Shape<PrimitivesT> = {},
+  // Used to override default theme property values derived from primitives.
+  overrides?: {},
+  // Options for deriving the final theme object.
+  options?: {isDark?: boolean} = {},
+): ThemeT {
+  const {primaryFontFamily, ...customColorTokens} = primitives;
```
nadiia, line 32: Could you keep the comments from the deleted file or add new ones on why we do all the destructuring and overrides?

src/themes/move-theme/typography.js
```diff
@@ -5,12 +5,9 @@ This source code is licensed under the MIT license found in the
 LICENSE file in the root directory of this source tree.
 */
 // @flow
-import type {FontTokensT} from '../types.js';
 
-export const fontTokens: FontTokensT = {
```
nadiia, line 10: I'd keep it as `...tokens` or anything extendable, so we can potentially include the secondary font in it.

src/themes/shared/colors.js
```diff
@@ -0,0 +1,49 @@
...
+*/
+// @flow
+import lightColorTokens from '../light-theme/color-tokens.js';
+import getLightSemanticColorTokens from '../light-theme/color-semantic-tokens.js';
+import getLightComponentColorTokens from '../light-theme/color-component-tokens.js';
+import getLightDeprecatedSemanticColorTokens from '../light-theme/color-deprecated-semantic-tokens.js';
+import darkColorTokens from '../dark-theme/color-tokens.js';
+import getDarkSemanticColorTokens from '../dark-theme/color-semantic-tokens.js';
+import getDarkComponentColorTokens from '../dark-theme/color-component-tokens.js';
+import getDarkDeprecatedSemanticColorTokens from '../dark-theme/color-deprecated-semantic-tokens.js';
```
nadiia, line 15: We should not combine dark and light themes. First, it's not scalable, second, it will increase the bundle size with a default theme.

sandgraham:
I checked how this impacts bundle size and it adds **3.25KB** (gzip) to the overall bundle size of `baseui`. Note this is a flat cost, meaning it affects all apps using `baseui`, even ones that do not use `createTheme`.

If we don't use `createTheme` to construct `LightTheme` and `DarkTheme` internally then that 3.25KB would only impact folks importing `createTheme` themselves. The actual 3.25KB comes from the additional import of all the dark theme color assignment functions (via `getColors`). 

With that in mind, it makes sense to revert using `createTheme` for our own themes.

The next question is how (and if) we provide folks with a way to construct dark themes (using our already existing `dark-theme` getters). Here are some options:

We could just not provide this utility at all. I think that would be a bit sad. I would like to allow folks to use our dark "getters" since they already exist and help a lot when creating a dark theme.

We could leave the current changes (`options.isDark`) in `createTheme`. The 3.25KB seems like a decent tradeoff for the added utility and it would only affect folks who use `createTheme` themselves. My thinking is that if you really care about bundle size enough to notice a few KB then you probably aren't using `createTheme` at all.

Another option is to unify the light and dark getter functions and accept an `isDark` option. This would end up adding to the overall bundle but probably a lot less than if they were separate (and you used both).

We could also create an entirely new function, say `createDarkTheme`, to completely remove any impact from existing apps. This is a pretty safe option.

I think I'll pursue this last option for now since it rocks the boat the least but still accomplishes the goal of letting folks create a dark theme as easily as a light theme. I might also add a section to the docs about optimizing custom themes and recommending `deepMerge` or even static builds for those who are concerned with bundle size.

documentation-site/pages/guides/theming.mdx
sandgraham: It would error if `primitives` and specifically `primitives.primaryFontFamily` were not provided. Updated types and got rid of the potential error so its all optional now.

src/themes/create-theme.js
sandgraham: I'll create a new function just for dark themes.

src/themes/create-theme.js
sandgraham: Added some back in.

documentation-site/pages/guides/theming.mdx
```diff
@@ -224,22 +224,47 @@ const overrides = {
...
-const theme = createTheme(lightThemePrimitives, overrides);
+const theme = createTheme(primitives, overrides);
 ```
 
 In this example, we first reassign the accent color on our theme primitives to hot pink. Then we use `overrides` to re-map the secondary button colors to use our hot pink accent colors rather than the default primary colors.
 
-Between `parameters` and `overrides`, you should have a fair amount of control over theme object creation. Ideally, you can use `createTheme` to quickly construct a Base Web compliant theme object.
+### `createDarkTheme`
+
+One of the main reasons to have a dynamic theme is to toggle your app between light and dark modes. `createTheme` will by default map `primitives` to theme colors in a way suited to light themes. If you want to create a dark theme you can instead use the `createDarkTheme` function.
```
nadiia, line 234:
```suggestion
One of the main reasons to have a dynamic theme is to toggle your app between light and dark modes. `createTheme` will by default map `primitives` to theme colors in a way suited to light themes. If you want to create a custom theme on top of our dark theme you can instead use the `createDarkTheme` function.
```

documentation-site/pages/guides/theming.mdx
```diff
@@ -224,22 +224,47 @@ const overrides = {
...
+  accent100: '#FCD3F9',
+  accent200: '#F89FF3',
+  accent300: '#F45AEA',
+  accent400: '#F127E4',
+  accent500: '#B71DAD',
+  accent600: '#901788',
+  accent700: '#600F5B',
+};
+
+const darkTheme = createDarkTheme(primitives);
```
nadiia, line 251: Should we include the second `overrides` arg here as an example?

documentation-site/pages/guides/theming.mdx
```diff
@@ -224,22 +224,47 @@ const overrides = {
...
+```
+
+This works exactly the same way as `createTheme`, but `createDarkTheme` will map all of your `primitives` to theme properties in a way better suited to dark themes.
+
+If you aren't sure why this mapping changes, try creating a dark theme with `createTheme`. You might notice that the colors look somewhat off. This is because, in a typical light theme, things start light and then get darker as they require more attention, whereas, in a dark theme, things start dark and become lighter. We need to know which direction (light/dark) to go when assigning your primitive colors to theme properties. By separating the assignment logic for light and dark theme creation we can better estimate which values to assign to various properties.
+
+Between `parameters` and `overrides` you should have a fair amount of control over theme object creation. Ideally, you can use `createTheme` to quickly construct a Base Web compliant theme object.
```
nadiia, line 258:
```suggestion
Between `primitives` and `overrides` you should have a fair amount of control over theme object creation. Ideally, you can use `createTheme` to quickly construct a Base Web compliant theme object.
```

documentation-site/pages/guides/theming.mdx
```diff
@@ -224,22 +224,47 @@ const overrides = {
...
 ### Customizing Typography
 
 The `primitives` object has an optional property, `primaryFontFamily`, which allows you to set a custom font family string for typography components.
 
 ```js
-import {createTheme, lightThemePrimitives} from 'baseui';
+import {createTheme} from 'baseui';
 
 const primitives = {
-  ...lightThemePrimitives,
```
nadiia, line 242: 👍

src/themes/light-theme/create-light-theme.js
```diff
@@ -0,0 +1,62 @@
...
+      ...getComponentColorTokens(colorTokens),
+      ...getDeprecatedSemanticColorTokens(colorTokens),
+      ...getSemanticColorTokens(colorTokens),
+    },
+    direction: 'auto',
+    grid,
+    lighting,
+    mediaQuery,
+    sizing,
+    typography: getTypography(fontTokens),
```
nadiia, line 53: It looks like it will end up in an `undefined`  font-family since the `fontTokens` arg will always be defined (on line 35) even when a custom `primaryFontFamily` is not passed in `primitives`  and will result in the default arg value not applied inside `getTypography` fn.

src/themes/shared/typography.js
```diff
@@ -5,119 +5,123 @@ This source code is licensed under the MIT license found in the
...
     'system-ui, "Helvetica Neue", Helvetica, Arial, sans-serif',
 };
 
-// color constants
-export default (themePrimitives: FontTokensT = fontTokens): TypographyT => {
+export default (
+  customFontTokens?: $Shape<FontTokensT> = defaultFontTokens,
+): TypographyT => {
+  const {
+    primaryFontFamily = defaultFontTokens.primaryFontFamily,
```
nadiia, line 19: Could we avoid assigning the default in two places (16 and 19)?

documentation-site/pages/guides/theming.mdx
sandgraham: Added a second snippet to show `overrides` too

src/themes/shared/typography.js
sandgraham: I was trying to update the function so it could accept nothing, a partial font tokens object, or a full font tokens object. Right now it only accepts nothing or a full font tokens object, which is why we need to call `getTypography` differently depending on whether or not `primaryFontFamily` is defined. I thought it might be nice to allow the font tokens to also be partially provided going forward. but i'll take it out and if we add new font tokens we can worry about it then.

PR #3163, opened 2020-04-06
fix(overrides): migrates accordion to override hoc

chasestarr:
#### Description

- migrates the accordion component to use the new override hoc
- fixes related issue where concern of icon overrides from theme was leaking outside of the icon implementation. encapsulates it there and move icon to new override hoc
- migrates other components using icon to override hoc (spinner, file-uploader)

vrt changes:
- adds new accordion overrides scenario to protect against regressions
- edits icon overrides scenario to the correct state (previously it documented a bug)
- toast component autofocuses to ensure that feature does not regress

#### Scope

- [x] Patch: Bug Fix


src/accordion/panel.js
```diff
@@ -90,62 +95,32 @@ class Panel extends React.Component<PanelPropsT, {isFocusVisible: boolean}> {
...
               onClick={this.onClick}
               onKeyDown={this.onKeyDown}
-              onFocus={forkFocus(headerProps, this.handleFocus)}
-              onBlur={forkBlur(headerProps, this.handleBlur)}
+              onFocus={
+                overrides.Header && overrides.Header.props
+                  ? forkFocus(overrides.Header.props, this.handleFocus)
+                  : this.handleFocus
+              }
+              onBlur={
```
nadiia, line 118: Is it guaranteed that this 'onBlur' handler won't get overwritten by `overrides.Header.props.onBlur`? Same for other event handlers.

chasestarr: Good point, the blur fork will get clobbered if `onBlur` exists in override props. I will look into how to resolve. Other than these two handlers, we will want overrides to take precedent right? I had noticed that `headerProps` were provided before the event handlers on this component, but thought it was a bug.

nadiia: It looks like all the handlers/props that go after `{...headerProps}` in the old code are not supposed to be overwritten.

chasestarr: I personally think we get into a very messy situation to selectively apply overrides. This change should be equivalent for `onClick` and `onKeyDown` since those prop overrides were never applied before. Now, that function props overrides are better supported, we can do something like below and get the best both situations.

```js
const overrides = {
  Header: {
    props: (props) => {
      return {
        ...props,
        onClick: (event) => {
          if (props.onClick) props.onClick(event);
          // add anything I want
        },
        onKeyDown: (event) => {
          if (props.onKeyDown) props.onKeyDown(event);
          // add anything I want
        },
        onFocus: (event) => {
          // maintains the fork
          if (props.onFocus) props.onFocus(event);
          // add anything I want
        },
        onBlur: (event) => {
          // maintains the fork
          if (props.onBlur) props.onBlur(event);
          // add anything I want
        },
      };
    },
  },
};
```

nadiia: I like it, but that would be a breaking change now. I think for the v10 tho this would be the way to move to. 
`onKeyDown` and `onClick` currently accepted as top-level props so they are applied, just no need to pass them through overrides.

src/toast/styled-components.js
```diff
@@ -147,7 +147,7 @@ export const CloseIconSvg = styled<SharedStylePropsArgT>('svg', props => {
   return {
     ...getSvgStyles(props),
     cursor: 'pointer',
-    width: '16px',
+    width: props.$size || '16px',
```
sandgraham, line 150: what's this for?

chasestarr: lol I spent maybe 2 hours investigating why when I overridded the `$size` property on `CloseIcon` in `Toast` it was always clamped to `16px`. Turned out that it was hard-coded here

src/accordion/panel.js
chasestarr: resolved and added a test: https://github.com/uber/baseweb/pull/3163/files#diff-71e348c99a67e7b6ca25cb442bd39f1dR111

src/icon/types.js
```diff
@@ -27,6 +27,8 @@ export type IconPropsT = {
   size?: SizeT,
   /** Color of icon, will be used as svg fill */
   color?: ColorT,
+  /** A ref to access an svg element. */
+  svgRef?: React.ElementRef<*>,
```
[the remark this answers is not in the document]
chasestarr: Adding a `forwardRef` was a big problem related to types. The problem was, we use `React.Node` all across the source code to allow for slot injection, like in `Button` enhancers. `forwardRef` is _not_ a `React.Node` type! Going forward, we should probably use `React.AbstractComponent` in similar situations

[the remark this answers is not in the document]
chasestarr: popovers will continue to not work by default since it's not forwarded from the ref prop, but through this other name

src/accordion/panel.js
```diff
@@ -90,62 +113,32 @@ class Panel extends React.Component<PanelPropsT, {isFocusVisible: boolean}> {
...
               aria-disabled={disabled || null}
               {...sharedProps}
-              {...headerProps}
               {...(ariaControls ? {'aria-controls': ariaControls} : {})}
               onClick={this.onClick}
               onKeyDown={this.onKeyDown}
-              onFocus={forkFocus(headerProps, this.handleFocus)}
-              onBlur={forkBlur(headerProps, this.handleBlur)}
+              onFocus={this.handleFocus}
+              onBlur={this.handleBlur}
```
nadiia, line 144: Since the above props are merged in the `HeaderOverrides` do we still need it here?

src/accordion/panel.js
```diff
@@ -90,62 +113,32 @@ class Panel extends React.Component<PanelPropsT, {isFocusVisible: boolean}> {
...
-    const ToggleIconComponent = isIconOverriden
-      ? ToggleIcon
-      : expanded
-      ? CheckIndeterminateIcon
-      : PlusIcon;
+    const HeaderOverrides = mergeOverrideResources(overrides.Header || {}, {
+      props: {
+        // constrain focus handlers, overrides handled in component methods
+        onFocus: this.handleFocus,
+        onBlur: this.handleBlur,
```
nadiia, line 121: I think `onClick={this.onClick}` and `onKeyDown={this.onKeyDown}` should be included in here also.

src/accordion/panel.js
chasestarr: This is so that if `override.props` is a function, consuming applications could read the initial value

PR #3166, opened 2020-04-07
feat(app-nav-bar): initial implementation of the navigation bar

nadiia:


src/app-nav-bar/styled-components.js
```diff
@@ -0,0 +1,168 @@
...
+export const StyledSpacing = styled<{}>('div', props => {
+  const {$theme} = props;
+  return {
+    boxSizing: 'border-box',
+    height: '100%',
+    paddingTop: $theme.sizing.scale700,
+    paddingBottom: $theme.sizing.scale700,
+    display: 'flex',
+    alignItems: 'center',
+    [`@media screen and (max-width: ${$theme.breakpoints.medium - 1}px)`]: {
```
sandgraham, line 65: Is there a reason to arrange media queries like this rather than using `theme.mediaQuery.medium` or `min-width`?

src/app-nav-bar/styled-components.js
```diff
@@ -0,0 +1,168 @@
...
+    display: 'flex',
+    alignItems: 'center',
+    color: $active ? colors.contentPrimary : colors.contentTertiary,
+    marginLeft: sizing.scale700,
+    marginRight: sizing.scale700,
+    outline: $isFocusVisible ? `3px solid ${colors.accent}` : 'none',
+    outlineOffset: '-3px',
+    borderBottom:
+      $active && !$isFocusVisible
+        ? `2px solid ${colors.primary}`
```
sandgraham, line 125: Should we use the long hand form for border styles?

src/app-nav-bar/app-nav-bar.js
```diff
@@ -0,0 +1,95 @@
...
+    appDisplayName,
+    appDisplayNameLink,
+  } = props;
+
+  return (
+    <StyledRoot>
+      {/* Mobile Nav Experience */}
+      <div
+        className={css({
+          [`@media screen and (min-width: ${breakpoints.large}px)`]: {
```
sandgraham, line 38: Can we use `theme.mediaQuery.large` here?

src/app-nav-bar/logo.js
```diff
@@ -0,0 +1,22 @@
...
+import * as React from 'react';
+import {StyledLink} from '../link/index.js';
+import {StyledAppName} from './styled-components.js';
+
+export default function Logo(props: {
+  appDisplayName: ?React.Node,
+  link: ?string,
+}) {
+  return (
+    <StyledLink $style={{textDecoration: 'none'}} href={props.link || ''}>
```
chasestarr, line 18: Accepting an href with no overrides will be problematic when there will inevitably be a request to make this work with react router or similar. My first thought is to reformat this to accept either a string or abstract component in top-level props. If a string, render this else render the provided component? Downside being we may see some non-standard display names in the top-left

src/list/types.js
```diff
@@ -44,10 +44,10 @@ export type LabelPropsT = {|
 export type MenuAdapterPropsT = {
   ...PropsT,
   // eslint-disable-next-line flowtype/no-weak-types
-  item: any,
-  onMouseEnter: MouseEvent => mixed,
-  onClick: MouseEvent => mixed,
-  $size: string,
-  $isHighlighted: boolean,
-  $disabled: boolean,
+  item?: any,
```
chasestarr, line 47: why relax this type?

src/app-nav-bar/types.js
```diff
@@ -0,0 +1,31 @@
...
+Copyright (c) 2018-2020 Uber Technologies, Inc.
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+// @flow
+
+import * as React from 'react';
+
+export type AppNavBarPropsT = {
```
chasestarr, line 11: Below is a suggestion to refine the types and avoid `any` definitions. I probably missed some nuance, so please let me know where my exploration of the diff went awry.

```js
// eslint-disable-next-line flowtype/no-weak-types
type ItemT = any;

type MainNavItemT = {|
  active?: boolean,
  icon?: React.AbstractComponent<{||}>,
  item: ItemT,
  // maybe include this to customize the menu adapter children?
  mapItemToNode?: React.AbstractComponent<{|item: ItemT|}>,
  mapItemToString: ItemT => string,
  subnav?: MainNavItemT[],
  subnavExitIcon?: React.AbstractComponent<{||}>,
|};

type UserNavItemT = {|
  item: ItemT,
  // maybe include this to customize the menu adapter children?
  mapItemToNode?: React.AbstractComponent<{|item: ItemT|}>,
  mapItemToString: ItemT => string,
|};

export type AppNavBarPropsT = {|
  displayName: string | React.AbstractComponent<{||}>,
  mainNavItems: MainNavItemT[],
  onMainNavItemSelect?: (params: {|item: MainNavItemT|}) => mixed,
  onUserNavItemSelect?: (params: {|item: UserNavItemT|}) => mixed,
  userNavItems?: UserNavItemT[],
  username?: string | React.AbstractComponent<{||}>,
  usernameSubtitle?: string | React.AbstractComponent<{||}>,
  userImgUrl?: string,
|};

export type UserMenuPropsT = {|
  username?: string | React.AbstractComponent<{||}>,
  usernameSubtitle?: string | React.AbstractComponent<{||}>,
  userImgUrl?: string,
|};
```

src/app-nav-bar/user-menu/user-menu-dropdown.js
```diff
@@ -0,0 +1,52 @@
...
+      onItemSelect={(...args) => {
+        props.onNavItemSelect(...args);
+        props.close();
+      }}
+      overrides={{
+        List: {
+          component: React.forwardRef(({children, ...restProps}, ref) => (
+            <StyledList {...restProps} ref={ref}>
+              <StyledUserMenuListItem>
+                {/* Replace with a renderer: renderUserProfileTile() */}
```
chasestarr, line 34: What does this mean?

src/app-nav-bar/app-nav-bar.js
nadiia: I want to make it customizable in a follow-up change so will leave it as is for now.

src/app-nav-bar/styled-components.js
```diff
@@ -0,0 +1,168 @@
...
+export const StyledAppName = styled<{}>('span', ({$theme}) => ({
+  ...$theme.typography.font650,
+  color: $theme.colors.primary,
+  textDecoration: 'none',
+  [`@media screen and (max-width: ${$theme.breakpoints.medium - 1}px)`]: {
+    ...$theme.typography.font550,
+  },
+}));
+
+// $FlowFixMe
```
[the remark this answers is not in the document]
nadiia: resolved

src/list/types.js
nadiia: Don't want to restrict the menu item type since with custom renderers they may want to include various additional data to be reused on render.

src/app-nav-bar/types.js
nadiia: Added the items types per your suggestion, that seem to work pretty well. But kept `React.Node` for those that accept a string or a node as is since the AbstractComponent type seems to be not the right fit for it.

src/app-nav-bar/user-menu/user-menu-dropdown.js
nadiia: Comments for myself where to allow rendering customizations that will be added as a follow-up change.

src/app-nav-bar/logo.js
nadiia: In this current PR there are not many overrides allowed, but it will be coming in as follow-up changes due to the size of this initial PR.

src/app-nav-bar/user-menu/user-menu-dropdown.js
chasestarr: ah ok a render-prop

src/app-nav-bar/types.js
```diff
@@ -0,0 +1,48 @@
...
+  item: ItemT,
+  mapItemToNode: ItemT => React.Node,
+  mapItemToString: ItemT => string,
+|};
+export type AppNavBarPropsT = {
+  appDisplayName?: React.Node,
+  appDisplayNameLink?: string,
+  // eslint-disable-next-line flowtype/no-weak-types
+  mainNav?: MainNavItemT[],
+  onNavItemSelect: (params: {item: MainNavItemT | UserNavItemT}) => mixed,
```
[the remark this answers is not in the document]
nadiia: I can't think of a case where navigating to one of user menu items would be handled differently. But the handler also gets the item selected to navigate to and it potentially can be handled differently based on the received item data.

src/app-nav-bar/types.js
```diff
@@ -0,0 +1,48 @@
...
+import * as React from 'react';
+
+// eslint-disable-next-line flowtype/no-weak-types
+type ItemT = any;
+
+export type MainNavItemT = {|
+  active?: boolean,
+  icon?: React.AbstractComponent<{||}>,
+  item: ItemT,
+  mapItemToNode: ItemT => React.Node,
```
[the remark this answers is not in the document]
nadiia: My thinking is that `mapItemToNode` should be always used to render an item so it has to be required otherwise we'll need to make assumptions about `item` object structure, or use `mapItemToNode` as a backup. The `mapItemToString` is really needed for a11y aria values and a case like `Avatar` rendering where the `name` prop only accepts a string.

src/app-nav-bar/types.js
```diff
@@ -0,0 +1,48 @@
...
+|};
+
+export type UserNavItemT = {|
+  active?: boolean,
+  icon?: React.AbstractComponent<{||}>,
+  item: ItemT,
+  mapItemToNode: ItemT => React.Node,
+  mapItemToString: ItemT => string,
+|};
+export type AppNavBarPropsT = {
```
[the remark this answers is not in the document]
nadiia: Yeah, I think we can make it exact.

src/app-nav-bar/types.js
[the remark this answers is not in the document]
chasestarr: Can this component not have a reasonable default? I think if this is required we will be seeing lots of non-standard looking menu items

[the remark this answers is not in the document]
nadiia: It's a good point. I'll add usage of the `mapItemToString` as a backup then.

PR #3171, opened 2020-04-07
feat(themes): add animation and easing values

andersonoliveira:
#### Description

- Adds new timing for animations
- Adds new easing for animations

P.S.: Our Base Motion use beats in **ms**

More info:
http://t.uber.com/motion


sandgraham:
@andersonba will we eventually need `timing250` at all? I wonder if we should update anything currently using `250ms` to be `200ms` or `300ms` instead and avoiding adding 250 entirely. My only concern is that once it is added we can't really get rid of it (until a major).

gergelyke:
I believe we also need to add these to the type definitions: https://github.com/uber/baseweb/blob/master/src/themes/types.js#L654-L662

andersonoliveira:
> @andersonoliveira will we eventually need `timing250` at all? I wonder if we should update anything currently using `250ms` to be `200ms` or `300ms` instead and avoid adding 250 entirely. My only concern is that once it is added we can't really get rid of it (until a major).

Yes we can remove entirely. Just wasn't sure about the PR size with all the changes. Should I add the documentation at the same PR with all changes?

andersonoliveira:
> I believe we also need to add these to the type definitions: https://github.com/uber/baseweb/blob/master/src/themes/types.js#L654-L662

Yes.

src/themes/shared/animation.js
```diff
@@ -8,13 +8,22 @@ LICENSE file in the root directory of this source tree.
...
+  timing200: '200ms',
+  timing300: '300ms',
+  timing400: '400ms',
+  timing500: '500ms',
+  timing600: '600ms',
+  timing700: '700ms',
+  timing800: '800ms',
+  timing900: '900ms',
+  timing1000: '1000ms',
   easeInCurve: 'cubic-bezier(.8, .2, .6, 1)',
```
chrisrauh, line 21: @sandgraham @andersonba - should we also update these default ease values to match quintic since that is our default in the guidelines? 

Making this update could require adjusting some of the timing where they are being used. It could be something done in a separate PR to keep them more atomic.

sandgraham: I'm for updating them - no preference on if a new pr is warranted.

andersonoliveira:
> @andersonoliveira If we update `timing100` we should probably also change instances of it to `timing200` or `timing300` in the same PR. I don't think it's a problem if PR includes all that since it's just a find/replace.
> 
> Oh, there are also Typescript types in `src/theme.ts` that should be updated in tandem with Flow types. 😅

@sandgraham I'll update that today.

src/themes/shared/animation.js
andersonoliveira: > @sandgraham @andersonoliveira - should we also update these default ease values to match quintic since that is our default in the guidelines?
> 
> Making this update could require adjusting some of the timing where they are being used. It could be something done in a separate PR to keep them more atomic.

@chrisrauh I'll update the timing and documentation first. After that will look into the easing and get a new PR. Thanks

andersonoliveira: Branch updated.

PR #3175, opened 2020-04-08
greenkeeping(ts): fix and update types for tree-view

jayantbh:
#### Description

Updated a couple of types related to TreeView. That's all.
The changes are backward compatible.

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


jayantbh:
Hmm, not sure what's wrong with the PR Label here.

gergelyke:
just added the label, thanks @jayantbh !

gergelyke:
can you help us why do we need to change the flow typdes?

jayantbh:
Hi @gergelyke!
Don't we want the TS and Flow types to be _similar_?

Like, if the TS type is a generic, the flow type should be too? At least that's what I was thinking.

jayantbh:
Ah, I see the flow tests failed. Lemme fix.

jayantbh:
Hmm, generics with defaults are apparently not supported in flow. Is that correct?

src/tree-view/types.js
```diff
@@ -27,13 +27,13 @@ export type TreeViewOverridesT = {
 
 export type TreeNodeIdT = number | string;
 
-export type TreeNodeT = {
+// eslint-disable-next-line flowtype/no-weak-types
+export type TreeNodeT<T: any> = {
```
chasestarr, line 31: Default generics in flow use <T = any> maybe that helps? Colon only constrains to a subtype specified. I'm not running the change locally, so not sure if this resolves the flow check

jayantbh:
So, I changed how I was declaring the flow type a generic, and refactored all usages of that type.

PR #3180, opened 2020-04-08
feat(accordion): add stateless version

sandgraham:
Adds a stateless accordion.


tajo:
I think it would be nice if you could control the state from the top level with `Accordion` and maybe adding `onlyOnePanelCanBeExpanded` prop (there must be a better name) so you can expand multiple panels. This would be actually built on top of the API in this PR.

```jsx
import * as React from "react";
import { Accordion, Panel } from "baseui/accordion";

export default () => {
  // first panel is expanded
  const [expanded, setExpanded] = React.useState(["0"])
  return (
    <Accordion
      onChange={({ expanded }) => setExpanded(expanded)}
      expandedPanels={expanded}
      onlyOnePanelCanBeExpanded={false}
    >
      <Panel title="Panel 1">Content 1</Panel>
      <Panel title="Panel 2">Content 2</Panel>
      <Panel title="Panel 3">Content 3</Panel>
    </Accordion>
  );
}
```

sandgraham:
@tajo, yeah I was debating having the state be at the top level vs distributed to each panel. I'll see how that feels.

Also, the `onlyOnePanelCanBeExpanded` prop is basically the existing `accordion` prop. If you set it to false, you can open more than one panel. I wasn't sure what it did until today.

nadiia:
We already have that prop that makes Accordion behave like one: [`accordion`](https://github.com/uber/baseweb/blob/master/src/accordion/types.js#L58).

nadiia:
I don't think it makes sense for the stateless accordion to handle what panels are closed or opened, including managing the `accordion` prop, since it would lose the `stateless`/`controlled` meaning.

tajo:
> Question though: why is the state a "0" for the first panel?

It's a bit strange but it's what we currently use for the stateful state afaik.

> We already have that prop that makes Accordion behave like one: accordion.

Oh, I thought so. So I guess we just don't document it today.

> I don't think it makes sense for the stateless accordion to handle what panels are closed or opened, including managing the accordion prop, since it would lose the stateless/controlled meaning.

"stateless meaning" - to me stateless doesn't mean that component has no internal state. Some of our other stateless components like `Select` also have internal state (is dropdown open? is focus visible?...).  Imo stateless should provide consumer a full control over the most essential behavior while making the usage practical.

My use-case for "stateless" accordion was for yard to open style override panels that have non-default value (even after the initial mount so `initialState` doesn't cut it). I could use `Panel`s directly and implement my own `StatelessAccordion` component (a decent chunk of work). It seems like something that we should provide out of the box?

nadiia:
I don't argue about "stateless" meaning and agree on the "stateless doesn't mean that component has no internal state". I was saying that handling "accordion" functionality (how many panels can be opened at the same time) doesn't fall under a "controlled" definition.

sandgraham:
Updated based on feedback.
- Moved `expanded` to `StatelessAccordion`, which accepts `React.Key[]`.
- Added `onChange` to `StatelessAccordion`, which is called with `{key: React.Key, expanded: React.Key[]`.

I played around with different APIs for `onChange` and ultimately it did feel like providing the next `expanded` list was a nice feature. You can ignore it and use the `key` returned, or just pass down your `expanded` state however you like.

PR #3190, opened 2020-04-09
feat(icon): fowards ref

chasestarr:
### Description

Changes Icon to forward refs to the underlying svg element. This can merge safely once https://github.com/uber/baseweb/pull/3196 is landed


documentation-site/examples/tree-view/overrides.js
```diff
@@ -83,10 +83,10 @@ export default function TreeViewOverrides() {
           },
         },
         CollapseIcon: {
-          component: CheckIndeterminateIcon,
+          component: props => <CheckIndeterminateIcon {...props} />,
```
chasestarr, line 86: This signals a huge problem with the diff. If an icon is passed to overrides it will fail flow check.

documentation-site/examples/notification/overrides.js
```diff
@@ -12,7 +12,7 @@ export default () => (
         }),
       },
       CloseIcon: {
-        component: DeleteAlt,
+        component: props => <DeleteAlt {...props} />,
```
chasestarr, line 15: Here's another instance of overrides needing to change because of flow check fail

documentation-site/examples/tree-view/overrides.js
tajo: wait, how is this not a breaking change?

chasestarr: Yeah that's my point. Although, the previous code should work the same it will just not pass flow-check

documentation-site/examples/notification/overrides.js
nadiia: Why do we need to wrap it tho? Does it mean that types are incorrect?

documentation-site/examples/tree-view/overrides.js
chasestarr: sigh.. you're right these two things are not equivalent

src/button/button-internals.js
```diff
@@ -15,6 +16,22 @@ import {getOverrides} from '../helpers/overrides.js';
 
 import type {ButtonPropsT} from './types.js';
 
+function RenderEnhancer(props) {
+  const {Enhancer, ...restProps} = props;
+  if (
```
tajo, line 21: Why not just to check for `isValidElementType` and delete this first branch?

chasestarr: Fixed it up so it now only handles strings early. This is required because `'div'` is a valid element type and can be called like `<'div' {...props />`. This provides a good example of why we should never allow `React.Node` as a render prop type and default to `React.AbstractComponent`

PR #3196, opened 2020-04-10
fix(overrides): removes prop generic. now set to any

chasestarr:
### Description

Typing the props value on overrides ultimately blocked us more that it was helping. Maybe a quarter of components were already setting it to `any`.

#### Scope

- [x] Patch: Bug Fix


chasestarr:
Some additional information regarding this change: We don't consider the `OverridesT` type to be part of baseui's public API and therefore have categorized this diff as a non-breaking change. The type is only available by importing from a specific deep file path, and is not included in the documentation site.

If this change does negatively affect your application, use this [comby replacement script](https://bit.ly/3aJoRk6) (click 'Run in Terminal' to view). See [comby documentation](https://comby.dev/) for more info.

PR #3200, opened 2020-04-12
fix(codesandbox): module resolution

sandgraham:
Fix codesandbox module resolution issue. It seems this stems from codesandbox resolving some modules as commonjs and some as esm. This was leading to strange bugs because there were multiple instances of providers (which should be singletons).

At this time, codesandbox prefers `browser` over `main`/`module` so I've added it to our `package.json`. `browser` only affects bundling targeting the client, but since it points at the same file as `module` there should not be any changes.

- [x] Tested with CRA
- [x] Tested with Fusion

---
For more context:
https://github.com/codesandbox/codesandbox-client/issues/3305
https://github.com/codesandbox/codesandbox-client/pull/3319


sandgraham:
Adding `browser` to `package.json` seems to sort things out 🤔

PR #3212, opened 2020-04-15
fix(input): adds missing props (pattern and aria-errormessage) to the TS defs

grrowl:
#### Description

Adds `pattern`  to `BaseInput`'s TS types. Also `aria-errormessage` was missing (it's in the flow types) so I added that too.

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


gergelyke:
thank you @grrowl

PR #3216, opened 2020-04-15
fix(button): fix button background timing function

andersonoliveira:
#### Description

Fix the button background timing function to be smooth on hover.

#### Scope

- [X] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


src/themes/types.js
```diff
@@ -668,6 +668,7 @@ export type AnimationT = {
   easeInQuintic: string,
```
chrisrauh, line 668: For consistency, should we add "Curve" at the end of the easeXXQuintic or remove it from the other ones? (cc @sandgraham)

andersonoliveira: Yes. I'll update that.

PR #3220, opened 2020-04-16
chore(project): mergify for renovate

gergelyke:


gergelyke:
I believe it can be - this one was simply from my machine, for some reason, the repo was not picked up in their UI. They may have needed the initial PR to be opened manually? Not sure

chasestarr:
Ok, I wanted to confirm that it would not auto-merge without 2 approvals

gergelyke:
Yeah, it won't - this yml file is the source of truth for how it runs

PR #3227, opened 2020-04-16
fix(docs): don't preload move fonts and move the link to the body end so it's not blocking

nadiia:
**Background story** 
We prefetched `fonts.css` but that css with font-faces was never applied and so the fonts were never downloaded.

**Solution**
In order to reduce/eliminate the flash of the system fonts, first, we need to link `fonts.css` in a regular way and not just `prefetch` (also it doesn't make sense to prefetch that small css file), second, we need to preload font files that are used on the page. There are trade-offs in this approach like potential increase in the page load time, also preloaded resources get prioritizes that may not be what one wants for font files. Despite the above mentioned, in our specific case, I did not notice any significant load time difference while testing it on a slower network.

### DIFF
--- a/documentation-site/pages/_document.js
+++ b/documentation-site/pages/_document.js
@@ -70,7 +70,19 @@ export default class MyDocument extends Document {
             rel="stylesheet"
             href="https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.css"
           />
-          <link rel="prefetch" as="style" href="/static/fonts.css" />
+          <link
+            rel="preload"
+            href="https://d1a3f4spazzrp4.cloudfront.net/dotcom-assets/fonts/UberMoveText-Regular.woff2"
+            as="font"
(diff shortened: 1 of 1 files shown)


sandgraham:
I think the preload was to avoid the font flickering from `system-ui` to `UberMove`. Curious what the issue is with loading the font that way?

nadiia:
The thing is that we never prefetched fonts, only the css file. But the issue here is that the actual fonts are never downloaded after css's prefetch despite we use those fonts on a page. Not prefetching css resolves the fonts downloading issue and the css link doesn't actually need to be at the end. 
Even tho I don't think a flash is a big deal for the move fonts but I'm working on a little different solution here. Just want to test perf for it to see if it even makes sense to do any prefetch. Stay tuned for an update here.

sandgraham (approved):
Looks like it works pretty well now.

gergelyke (approved):

PR #3236, opened 2020-04-20
fix(accordion): replace includes with indexOf for crossbrowser support, docs edits

nadiia:


gergelyke:
maybe I am missing something, but shouldn't babel take care of cross-browser support? I believe browsers which do not support includes, won't support es modules either, so without compilation, they can't use baseui - is that correct?

src/accordion/accordion.js
```diff
@@ -122,12 +114,16 @@ export default class Accordion extends React.Component<
...
-    const sharedProps = this.getSharedProps();
     const {overrides = {}} = this.props;
     const {Root: RootOverride} = overrides;
     const [Root, rootProps] = getOverrides(RootOverride, StyledRoot);
     return (
-      <Root data-baseweb="accordion" {...sharedProps} {...rootProps}>
+      <Root
+        data-baseweb="accordion"
+        $disabled={this.props.disabled}
+        $isFocusVisible={false}
```
sandgraham, line 124: Always false?

nadiia:
> maybe I am missing something, but shouldn't babel take care of cross-browser support? I believe browsers which do not support `includes`, won't support es modules either, so without compilation, they can't use baseui - is that correct?

You're right it does take care of `includes` for the non-es build and IE doesn't support es modules. You think it's better to leave as is (`include`)?

src/accordion/accordion.js
nadiia: see above, nothing changed there

gergelyke:
Yeah, I'd leave it as is - I feel like with this, we'd go down on a [rabbit hole](https://github.com/uber/baseweb/search?q=includes&unscoped_q=includes). Also, I am sure we'd like to start using new JS features as they become available - so I'd just let babel take care of this.

PR #3238, opened 2020-04-20
fix(datepicker): Changes in the year Selection

### LINKED ISSUE #3230: [Datepicker] Inputting a date into input field results in wrong date
## Current Behavior
If you go [here](https://baseweb.design/components/datepicker/), clear the field and start typing "1993/01/01" you end up with "2993/01/01"
I haven't really explored why this is happening, but it makes inputting the field in the textfield fairly useless. You can paste 1993/01/01 and it works fine, which makes me think it's somehow turning the incomplete year into a completely different one.

## Expected Behavior
It should only display what you are typing. Perhaps one way to do this is to ignore onChange events until the requisite number of digits are in the field.

6d8eb892-1b32-4792-88ec-9ad238ad688d

jayanthchandra:
Fixes #3230 

#### Description
The current date is calculated by passing the dateString to the Date constructor, in doing this the default date is always shown as 2001 when the first input is entered and with this the user entered date is discarded.
In the commit, i am creating a date object by considering the first value as the year so the user entered value is shown in the datepicker. 

#### Scope

- [X] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/src/datepicker/datepicker.js
+++ b/src/datepicker/datepicker.js
@@ -220,7 +220,16 @@ export default class Datepicker extends React.Component<
       }
     } else {
       const dateString = this.normalizeDashes(inputValue);
-      let date = new Date(dateString);
+      let yearString = dateString.split('/')[0].trim();
+      let date;
+
+      if (yearString != '' && yearString.length < 4) {
+        yearString = yearString + '0'.repeat(4 - yearString.length);
+        date = new Date(yearString, 0, 1);
+      } else {
+        date = new Date(dateString);
+      }
+
       const formatString = this.props.formatString;
       if (formatString) {
         // Prevent early parsing of value.


src/datepicker/datepicker.js
```diff
@@ -220,7 +220,16 @@ export default class Datepicker extends React.Component<
       }
     } else {
       const dateString = this.normalizeDashes(inputValue);
-      let date = new Date(dateString);
+      let yearString = dateString.split('/')[0].trim();
```
nadiia, line 223: How does it handle i18n cases or custom formatting?

jayanthchandra: My premise here is that datepicker works well with format with "/" delimiter. 
This change is applied to any datestring where the length of the first subset is less than 4 
so for all the date formats with YYYY as the first format of the date, you could see the the change in the popup calendar UI. 
For the custom formats which starts with MM or DD, this change would be applied but the formatCheck would make the date as null so you wouldn't see any change there. 

If there is a cleaner way to do this, do let me know.

jayanthchandra: For i18n, 
I noticed that localisation works with the calender input not the manual entry.Thought this as bug so didnt want to include in this PR.  The inputted date doesn't reflected in the calendar UI. 
ezgif-6-89d40bdcc8c5

If i am missing any cases or context here let me know.

jayanthchandra:
The reason i said that datepicker works well with "/" delimeter is that if we change the formatString as DD-MM-YYYY then the date chosen from the Calender UI would have the correct prop but when we do manual date entry with that format the new date is not saved in the prop. Could you confirm if this is a bug or not. 
ezgif-6-09034117c17f

src/datepicker/datepicker.js
VladimirMilenko: @nadiia  formatString property is actually not triggering this behavior.

VladimirMilenko:
> The reason i said that datepicker works well with "/" delimeter is that if we change the formatString as DD-MM-YYYY then the date chosen from the Calender UI would have the correct prop but when we do manual date entry with that format the new date is not saved in the prop. Could you confirm if this is a bug or not.
> ezgif-6-09034117c17f

This behavior was introduced in https://github.com/uber/baseweb/pull/3157 `formatString` wouldn't trigger date parsing if the date is not completely entered

nadiia:
> The reason i said that datepicker works well with "/" delimeter is that if we change the formatString as DD-MM-YYYY then the date chosen from the Calender UI would have the correct prop but when we do manual date entry with that format the new date is not saved in the prop. Could you confirm if this is a bug or not.

Sounds like a bug, could you create a repro in a codesandbox?
Regarding the "/" as an individual case: we should look into adding a solution that considers and able to handle or allows handling any date formats.

nadiia:
The PR #3275 fixing the issue in #3230. Closing this one.

PR #3239, opened 2020-04-20
feat(datepicker) swappable date libraries

cbranch101:
I'm working on applying the same changes I made to [Timepicker](https://github.com/uber/baseweb/pull/2968) to Datepicker.  

Because Datepicker is so much more complex, I needed a more involved strategy to make sure I didn't break anything.  Because the test coverage isn't quite in the place where I could change individual date processing calls one by one and re-run the tests to ensure that nothing broke, I settled on a strategy of testing the helpers in isolation first.  

One by one, I added tests to every helper that was exported from`utils/index`, then re-implemented them using `date-io`, removed all of the old helpers from `utils/index` and re-exported them from an instantiated `Date-Helpers` class. Finally, I added a check to the tests confirm that the newly created exported version, and the DateHelpers versions all behave the same way.  This should give us pretty close to 100% confidence on the handler code. 

Pretty much everything in here follows the same patterns established in the Timepicker PR, anything that's more interesting, I'll highlight in comments.


src/datepicker/utils/date-helpers.js
```diff
@@ -15,6 +25,19 @@ class DateHelpers<T> {
   constructor(adapter: DateIOAdapter<T>) {
     this.adapter = adapter;
   }
+  getAdapterWithNewLocale: mixed => DateIOAdapter<T> = locale => {
```
cbranch101, line 28: This was the biggest blocker in this PR.  The problem is that the `date-io` instance that we're passing in as a prop accepts a `locale` when its created, but there are helpers that are reliant on passing locale as an argument. 

This means that we either need a way to change the locale for the instance, or we need to recreate the instance with a new locale.  As a short term solution, I settled on recreating the instance.  Doing that requires the `adapterOptionMap` above, which makes it possible to us to get the options passed into the `Utils` class when it was instantiated, so we can create it over again with a new locale. 

The thing is, I really don't like this solution, so I [submitted a PR](https://github.com/dmtrKovalenko/date-io/issues/347) to see if they make some changes on their end to make this process a little bit easier to manage.  I just didn't want to have to stop making progress until that problem is solved.

src/datepicker/utils/date-helpers.js
```diff
@@ -25,6 +48,263 @@ class DateHelpers<T> {
...
+    return dates.reduce((minDate, date) => {
+      return this.adapter.isBefore(date, minDate) ? date : minDate;
+    });
+  };
+  max: (Array<T>) => T = dates => {
+    return dates.reduce((maxDate, date) => {
+      return this.adapter.isAfter(date, maxDate) ? date : maxDate;
+    });
+  };
+  getEffectiveMinDate: ({minDate?: T, includeDates?: Array<T>}) => T = ({
```
cbranch101, line 153: `minDate` and `includeDates` used to not be optional, but then they include all of this code for checking if they're defined.  I decide to make them optional just so I could cover all of that code and not have it trigger flow errors, the other option would be to actually take that code out if it's really not optional.

src/datepicker/utils/date-helpers.js
```diff
@@ -25,6 +48,263 @@ class DateHelpers<T> {
...
+    } else if (includeDates && includeDates.length) {
+      return this.min(includeDates);
+    } else if (!(includeDates && includeDates.length) && minDate) {
+      return minDate;
+    }
+    // this condition can't ever be reached
+    // but flow isn't smart enough to see that all of the conditions are covered
+    return this.adapter.date();
+  };
+  getEffectiveMaxDate: ({maxDate?: T, includeDates?: Array<T>}) => T = ({
```
cbranch101, line 171: `maxDate` and `includeDates` were made optional

cbranch101: Basically all of these functions with the second argument that comes from props

src/datepicker/utils/date-helpers.js
```diff
@@ -25,6 +48,263 @@ class DateHelpers<T> {
...
+  isOutOfBounds: (T, {minDate: ?T, maxDate: ?T}) => boolean = (
+    day,
+    {minDate, maxDate} = {},
+  ) => {
+    return (
+      (!!minDate && this.differenceInCalendarDays(day, minDate) < 0) ||
+      (!!maxDate && this.differenceInCalendarDays(day, maxDate) > 0)
+    );
+  };
+  setSeconds: (T, number) => T = (date, seconds) =>
```
cbranch101, line 285: I don't love having to define all of these functions manually even though they're just the same calls with the exact same arguments in `date-io` the problem was, this was the only way I could find where those functions would be correctly typed.  Every thing else dynamic function creation and that caused flow to break down.

src/datepicker/utils/index.js
```diff
@@ -6,256 +6,52 @@ LICENSE file in the root directory of this source tree.
...
-    month: getMonth(date),
-    date: getDate(date),
-  });
-}
+import defaultAdapter from './date-fns-adapter';
+import DateHelpers from './date-helpers';
+
+const defaultDateHelpers = new DateHelpers<Date>(defaultAdapter);
+
+export const formatDate = defaultDateHelpers.formatDate;
```
cbranch101, line 15: I couldn't think of a way to have all of these be exported as named exports without explicitly defining them like this

src/datepicker/utils/types.js
```diff
@@ -23,7 +23,7 @@ export type DateIOAdapter<T> = {
 
   isNull(value?: T): boolean,
   isValid(value: DateInput<T>): boolean,
-  getDiff: Comparison<T>,
+  getDiff: (from: T, to: T) => number,
```
cbranch101, line 26: A few of these types where just set up incorrectly in the first PR

src/datepicker/utils/date-helpers.js
```diff
@@ -7,6 +7,16 @@ LICENSE file in the root directory of this source tree.
 // @flow
 import type {DateIOAdapter} from './types.js';
 
+const adapterOptionMap = {
```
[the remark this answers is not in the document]
cbranch101: Yep, I feel the same way

src/datepicker/utils/date-helpers.js
cbranch101: That makes a ton of sense.  What I'll do is push forward on making the `Datepicker` changes(which is probably going to take me a while) and if I can get a solution from `date-io` in the meantime, I'll apply that, otherwise, I'll add the deprecation warnings.

src/datepicker/utils/date-helpers.js
cbranch101: Just kidding, I just misread this type signature, ignore this comment

src/datepicker/utils/date-helpers.js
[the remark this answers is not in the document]
cbranch101: I renamed the function, but this should be moved now

src/datepicker/calendar-header.js
```diff
@@ -66,45 +53,68 @@ function idToYearMonth(id) {
...
+    super(props);
+    this.dateHelpers = new DateHelpers(props.adapter);
+  }
+
+  state = {
+    isMonthYearDropdownOpen: false,
+    isFocusVisible: false,
+  };
+
+  getDateProp: () => T = () => {
```
cbranch101, line 81: The old pattern of creating a javascript date as a default prop isn't going to work with the generic date type. 

The simplest solution would be to assign `dateHelpers.date()` as the default instead, but because `dateHelpers` isn't available until after the constructor is called, moving the default handling into a function seemed like the best bet.

src/datepicker/datepicker.js
```diff
@@ -96,21 +103,25 @@ export default class Datepicker extends React.Component<
     this.props.onChange && this.props.onChange(data);
   };
 
-  formatDate(date: ?Date | Array<Date>, formatString: string) {
+  formatDate(date: ?T | Array<T>, formatString: string) {
+    const format = date => {
+      if (formatString === DEFAULT_DATE_FORMAT) {
```
cbranch101, line 108: `date-io` has two different methods for formatting dates.  You can call `formatDate` which lets you provide a specific format string like `yyyy/mm/dddd`, or you can call `format` and pass in a named format string like `slashDate` or `fullDate` and then have that named format resolve to the right format string for the library being used in the date adapter. 

The tricky part here, is making those two scenarios work together correctly.  If `formatString` is provided as prop, I'm using formatDate, because its the responsibility of the user to make sure they pass the right date format for the library they're using, but if we're using the default, we need to make sure it's the correct default for the library.  

This is the case for both parsing and formatting

src/datepicker/datepicker.js
```diff
@@ -196,32 +207,52 @@ export default class Datepicker extends React.Component<
       isInputUsed: true,
     });
 
+    const parseDateString = dateString => {
+      const formatString = this.normalizeDashes(this.props.formatString);
+      if (formatString === DEFAULT_DATE_FORMAT) {
```
cbranch101, line 212: This follows the same pattern as `formatDate`

src/datepicker/stateful-calendar.js
```diff
@@ -10,18 +10,21 @@ import StatefulContainer from './stateful-container.js';
...
-  );
-}
+type PropsT<T> = StatefulDatepickerPropsT<CalendarPropsT<T>>;
 
-StatefulComponent.defaultProps = {
-  initialState: {},
-  stateReducer: (type, nextState) => nextState,
-  onSelect: () => {},
-};
+class StatefulComponent<T = Date> extends React.Component<PropsT<T>> {
```
cbranch101, line 15: There seems to be a bug in prettier where passing a default in a generic type(ie. `<T = Date>`) breaks its parser.  There happened to only be two simple components that used to be functional components that needed to have default generics passed to them, and so the easiest short term solution was to swap them over to classes.  Ideally that bug gets resolved and we can switch back.

src/datepicker/stateful-container.js
```diff
@@ -16,31 +16,35 @@ import type {
...
 
-const defaultStateReducer: StateReducerT = (type, nextState) => nextState;
+type InputProps<T> = CalendarPropsT<T> | DatepickerPropsT<T>;
 
-type PropsT = StatefulContainerPropsT<CalendarPropsT | DatepickerPropsT>;
-class StatefulContainer extends React.Component<PropsT, ContainerStateT> {
-  static defaultProps = {
+type PropsT<T> = StatefulContainerPropsT<InputProps<T>, T>;
+
+class StatefulContainer<T = Date> extends React.Component<
```
cbranch101, line 23: This was the other component that needed to be switched over

src/datepicker/stateful-container.js
```diff
@@ -16,31 +16,35 @@ import type {
...
-  constructor(props: PropsT) {
+  constructor(props: PropsT<T>) {
     super(props);
-    const value = props.range ? [] : (null: ?Date);
+    const value = props.range ? [] : (null: ?T);
     this.state = {value, ...props.initialState};
   }
 
-  onChange = (data: {date: ?Date | Array<Date>}) => {
+  onChange: ({date: ?T | Array<T>}) => mixed = data => {
```
cbranch101, line 39: There seems to be some buggy behavior in flow(noted [here](https://github.com/facebook/flow/issues/5193) and [here](https://github.com/facebook/flow/issues/4743)) related to the way that generic types flow through class methods. 

The only consistent solution is to switch to the kind of type declaration shown above.  The problem is, because the behavior is buggy, there doesn't actually seem to be a clear pattern as to which types of functions are a problem and which type of functions aren't.  I just ended up having to swap over every function that flow was complaining about, even if it didn't really make sense why.  It works as solution, it can just be a little confusing to figure out why one type signature had to change while another similar one didn't.

src/datepicker/types.js
```diff
@@ -201,7 +210,8 @@ export type DatepickerPropsT = CalendarPropsT & {
 };
 
 export type SharedStylePropsT = {
-  $date: Date,
+  // eslint-disable-next-line flowtype/no-weak-types
+  $date: any,
```
cbranch101, line 214: I tried to swap this over to `T` but discovered that all style properties were getting `automatically` typed as `any` anyway, so I just made this any to make it consistent.

src/datepicker/types.js
```diff
@@ -53,109 +58,116 @@ export type DatepickerOverridesT = {
   Popover?: OverrideT,
 };
 
-export type DayPropsT = {
+export type DayPropsT<T = Date> = {
```
cbranch101, line 61: Which there are a lot of changes to these types, almost all of them(aside from the addition of the adapter prop) should just be switching from `Date` to `T`

src/datepicker/types.js
```diff
@@ -223,101 +233,78 @@ export type SharedStylePropsT = {
...
-  nextState: NavigationContainerStateT,
-  currentState: NavigationContainerStateT,
-) => NavigationContainerStateT;
+  nextState: NavigationContainerStateT<T>,
+  currentState: NavigationContainerStateT<T>,
+) => NavigationContainerStateT<T>;
 
-export type StatefulContainerPropsT<T> = {
-  children: T => React.Node,
+export type StatefulContainerPropsT<PropsT, T = Date> = {
```
cbranch101, line 265: I had to rename the generic here so it didn't conflict with the new generic date type

chasestarr:
The change looks overall good to me, but there are some test failures. Can you please look into those?

chasestarr:
I've been looking into the vercel deploy failure, but haven't found a solution just yet

chasestarr:
Ran a canary test of this change on uber projects and it looks ready to land! Nice work @cbranch101

PR #3242, opened 2020-04-20
docs: add tooltip example for complex content

sandgraham:
Tooltips default to a high contrast background color. This means that components (such as links or buttons) may not have enough contrast inside of a tooltip. I don't think there is a good non-breaking fix for this outside of asking devs to use a theme or customize the content they put into the tooltip. IMO, if you pass complex content you are taking responsibility for making sure it looks good (and we ensure default use case of just text looks good).

Considered:
- Changing Link to current color. 👎 Specs specify black/white
- Passing adjusted versions of Link and Button. 👎Where does it end?
- Wrapping tooltip in dark theme by default. 👎 How would this work with custom themes?

CC @lucasterra


lucasterra:
Looks good to me. Thanks for such detailed examples.
In my code I ended up manually styling the content. I will probably wrap it with another ThemeProvider, like in your examples.

Thanks!

documentation-site/examples/tooltip/stateful-complex-content.js
```diff
@@ -1,7 +1,18 @@
 // @flow
 import * as React from 'react';
 import {StatefulTooltip} from 'baseui/tooltip';
-import {useStyletron} from 'baseui';
+import {Button} from 'baseui/button';
```
nadiia, line 4: Button seems to be not in use here.

PR #3245, opened 2020-04-21
feat(app-nav-bar): add horizontal subnav

nadiia:


documentation-site/examples/app-nav-bar/basic.js
```diff
@@ -9,7 +9,10 @@ import {
   Overflow as UserIcon,
   Upload as Icon,
 } from 'baseui/icon';
-import {Unstable_AppNavBar as AppNavBar} from 'baseui/app-nav-bar';
+import {
+  Unstable_AppNavBar as AppNavBar,
+  NAV_POSITION,
```
gergelyke, line 14: should we just call this `POSITION`? historically, we never prefixed constants (like in the Button component: https://github.com/uber/baseweb/blob/master/src/button/constants.js)

documentation-site/examples/app-nav-bar/basic.js
```diff
@@ -34,6 +37,7 @@ const MAIN_NAV = [
     mapItemToNode: renderItem,
     mapItemToString: renderItem,
     navExitIcon: Delete,
+    navPosition: {desktop: NAV_POSITION.horizontal},
```
gergelyke, line 40: following my previous comment,  should it be just `position`?

documentation-site/examples/app-nav-bar/basic.js
```diff
@@ -114,11 +118,37 @@ const USER_NAV = [
   },
 ];
 
+function findAndSetActiveChain(item, arr): [boolean, Array<any>] {
```
gergelyke, line 121: I wonder, can it be somehow part of the nav internals, so folks do not ave to copy-paste every time they use the nav?

src/app-nav-bar/constants.js
```diff
@@ -0,0 +1,17 @@
...
+LICENSE file in the root directory of this source tree.
+*/
+// @flow
+
+export const NAV_POSITION = {
+  horizontal: 'horizontal',
+  vertical: 'vertical',
+};
+
+export const NAV_ITEM_KIND = {
```
gergelyke, line 14: following my first comment, how about we just name this `ITEM_KIND`?

documentation-site/examples/app-nav-bar/basic.js
nadiia: 👍 I think in consts it can be `POSITION`.

documentation-site/examples/app-nav-bar/basic.js
nadiia: Since it relates to the sub-navigation positioning and not the item positioning itself I'd keep it as `navPosition` in this case.

gergelyke: oh, I see, at first, it was not clear for me - in that case, should name it `subNavPosition`?

documentation-site/examples/app-nav-bar/basic.js
nadiia: I wouldn't make it uncontrolled or let the nav component handle what items are currently active, but let the user set it explicitly. This part of the code could potentially be shipped as a helper function but not sure how valuable it would be.

documentation-site/examples/app-nav-bar/basic.js
nadiia: Since I use `nav`|`navExitIcon` thi one would be `navPosition` to match. I'm open to renaming all of the mentioned to `subnav`|`subnavExitIcon`|`subnavPosition` if you think it would reflect it's purpose better, but I personally like `nav` prefix so it's like "navigation under current item regardless how deep it's nested".

gergelyke: my thought process was that if it applies to the subnav, probably we should clarify, but I don't feel strongly either way

maybe a fresh pair of eyes could help? cc @chasestarr @tajo @sandgraham

documentation-site/examples/app-nav-bar/basic.js
chasestarr: I believe there's a helper function for this kind of thing in the tree-view and drag-drop components. An `isItemActive` function may be more useful than storing this information in the array

documentation-site/examples/app-nav-bar/basic.js
chasestarr: my comment from the last pr that suggested `navPosition` https://github.com/uber/baseweb/pull/3166#discussion_r406465811

src/app-nav-bar/app-nav-bar.js
```diff
@@ -29,7 +31,17 @@ export default function AppNavBar(props: AppNavBarPropsT) {
     appDisplayName,
     appDisplayNameLink,
   } = props;
-
+  const activeMainNavItem = mainNav.find(item => item.active);
+  const activeSubNav =
```
chasestarr, line 35: Maybe refactor this to a `let` + `if statement`? Ultimately resolving to a value due to subtle JS interactions leads to less quickly understandable code

documentation-site/examples/app-nav-bar/basic.js
nadiia: The tree helper function is very specific to the tree component. Additionally, in a real-world application, the active state would probably be defined based on the current url.

PR #3256, opened 2020-04-23
fix(datepicker): prioritize selected date highlight over preset highlight date

### LINKED ISSUE #3067: [datepicker] After setting date by clicking [Day], and then changing the value of the Calendar component, leaves the startDate selected by click highlighted. This feels like a glitch.

nadiia:
Fixes #3067

Credit to @minseoksuh for bringing up the issue and starting working on the fix in #3147 .

The added e2e test looks pretty long but it covers another case that would be affected by the change in #3147 and result in a regression.

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change


src/datepicker/__tests__/datepicker-keyboard-nav.e2e.js
```diff
@@ -107,4 +114,165 @@ describe('Datepicker - keyboard navigation', () => {
     isElActive = await isActiveEl(page, selectors.apr2);
     expect(isElActive).toBe(true);
   });
+
+  it('calendar sets highlighted date appropriately when selecting a new date or navigating around', async () => {
```
gergelyke, line 118: can we break this test case into a few smaller ones?

nadiia: I want to test the whole chain vs separate actions to make sure that, for example, selecting a date doesn't affect the rest of functionality.

PR #3264, opened 2020-04-24
feat(layout grid): use Context in grid so React.Fragments can be used

### LINKED ISSUE #3263: [grid-layout]: Be able to put Cells in a subcomponent
Currently, Grid and Cells need to be in the same component like this
```
function Example() {
    return <Grid><Cell></Cell></Grid>
}
```
But if you have a grid with a lot of cells, you probably want to split it up into subcomponents (especially if the number of Cells is variable) like this
```
function ExampleCells() {
    return (
        <>
           <Cell>Stuff</Cell>
           <Cell>Stuff2</Cell>
           <Cell>Stuff3</Cell>
           <Cell>Stuff4</Cell>
           <Cell>Stuff5</Cell>
        </>
    )
}

function Example() {
    return <Grid><ExampleCells /></Grid>
}
```

Currently the grid component doesn't allow this due to the way it injects props into the cell components

https://github.com/uber/baseweb/blob/master/src/layout-grid/grid.js#L42
```
      {React.Children.map(children, child => {
        if (child) {
          return (
            <Cell
              gridColumns={gridColumns}
              gridGaps={gridGaps}
              gridGutters={gridGutters}
              gridUnit={gridUnit}
              {...child.props}
            />
          );
        }
        return null;
      })}
```

I think perhaps it might make the grid layout more flexible instead of injecting props like this, it could use the Context API to inject props into the Cell components, so you can wrap cell components in components that render them like HoCs and React.Fragment.

jasongi-actu:
Fixes #3263 

#### Description

Change grid to use context so that non-cell elements can be used as children to the grid. This will allow you to split up cells into subcomponents that have a react fragment wrapped around them, or be wrapped in HoCs/other wrapped components that still render Cells.

This change should be entirely backwards compatible with any existing code that uses Baseweb.

#### Scope

- [ ] Patch: Bug Fix
- [x] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/src/layout-grid/cell.js
+++ b/src/layout-grid/cell.js
@@ -9,7 +9,7 @@ LICENSE file in the root directory of this source tree.
 import React from 'react';
 import {getOverrides} from '../helpers/overrides.js';
 import {StyledCell as DefaultStyledCell} from './styled-components.js';
-
+import {GridContext} from './grid.js';
 import type {CellPropsT} from './types.js';
 
 export default function Cell({
@@ -28,13 +28,14 @@ export default function Cell({
     overrides.Cell,
     DefaultStyledCell,
   );
+  const gridContext = React.useContext(GridContext);
   return (
     <StyledCell
       $align={align}
-      $gridColumns={gridColumns}
-      $gridGaps={gridGaps}
-      $gridGutters={gridGutters}
-      $gridUnit={gridUnit}
+      $gridColumns={gridColumns || gridContext.gridColumns}
+      $gridGaps={gridGaps || gridContext.gridGaps}
+      $gridGutters={gridGutters || gridContext.gridGutters}
+      $gridUnit={gridUnit || gridContext.gridUnit}
       $order={order}
       $skip={skip}
       $span={span}
--- a/src/layout-grid/cell.js
+++ b/src/layout-grid/cell.js
@@ -32,6 +32,7 @@ export default function Cell({
   return (
     <StyledCell
       $align={align}
+      // TODO(v10): Remove the four grid props, get them solely from GridContext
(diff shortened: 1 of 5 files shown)


src/layout-grid/cell.js
```diff
@@ -28,13 +28,14 @@ export default function Cell({
...
     <StyledCell
       $align={align}
-      $gridColumns={gridColumns}
-      $gridGaps={gridGaps}
-      $gridGutters={gridGutters}
-      $gridUnit={gridUnit}
+      $gridColumns={gridColumns || gridContext.gridColumns}
+      $gridGaps={gridGaps || gridContext.gridGaps}
+      $gridGutters={gridGutters || gridContext.gridGutters}
+      $gridUnit={gridUnit || gridContext.gridUnit}
```
sandgraham, line 38: Do we need the `||` here? There isn't any reason to pass values that differ from the grid context. In that case, we can remove these `gridXYZ` props from `Cell` altogether.

jasongi-actu: I did this so that it would be entirely backwards compatible with the current Cell component. I don't really see why someone would want to pass the grid props directly into a Cell component, but if we remove the props it turns this PR into a breaking change. Am happy to do it if you think it's worth making a breaking change?

sandgraham: Good point, can you just add a comment here to remind us to clean it up in v10? Something like this:
```
// TODO(v10): Remove gridXYZ props
```

jasongi-actu: OK, added in the latest commit
https://github.com/uber/baseweb/pull/3264/commits/d7e77300056b7da07ca428911848ed508850102d

PR #3269, opened 2020-04-24
fix(build): add buildkite job to create github release

chasestarr:


scripts/github-release.js
```diff
@@ -0,0 +1,102 @@
...
+  const releaseResponse = await fetch(`${root}/releases`, options);
+  if (!releaseResponse.ok) {
+    throw new Error(`Failed to create release. ${releaseResponse.statusText}`);
+  }
+
+  const release = await releaseResponse.json();
+  console.log(`Successfully created release at: ${release.html_url}`);
+}
+
+main();
```
gergelyke, line 102:
```suggestion
main().catch(console.error);
```

not sure if our node.js version would hide the errors 🤔

scripts/github-release.js
```diff
@@ -0,0 +1,102 @@
...
+  const commitsSinceTag = [];
+  for (let commit of commits) {
+    if (commit.sha === tags[0].commit.sha) {
+      break;
+    }
+    commitsSinceTag.push(commit);
+  }
+  const prs = await Promise.all(commitsSinceTag.map(c => prFromCommit(c.sha)));
+
+  const changelogEntries = [];
```
gergelyke, line 68: do I understand correctly, that this would not group commits based on the label the PR was assigned? not a big deal, if we want it later, we can do that as a followup

chasestarr: Correct, I found it not very helpful to organize since we release often. There tends to be maybe 10 prs at most

PR #3275, opened 2020-04-28
fix(datepicker): inputting a date into input field results in right date

### LINKED ISSUE #3230: [Datepicker] Inputting a date into input field results in wrong date

devanshj:
Fixes #3230

#### Description

The scenario in the issue is already taken care of, the author just forgot you'll always have a `formatString` if not the passed then the default

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change

<sup>Btw, I love baseweb! Especially the overrides pattern, it's stellar, like it embraces the fact that you'll always have to override something at some point so why not provide a way to do that in the first place. And not to mention it's not just trivial style overrides but [actual overrides in behavior and logic too](https://baseweb.design/guides/understanding-overrides/#override-the-entire-subcomponent) So thank you :)</sup>

### DIFF
--- a/src/datepicker/__tests__/datepicker.test.js
+++ b/src/datepicker/__tests__/datepicker.test.js
@@ -229,7 +229,7 @@ describe('Datepicker', () => {
     });
   });
 
-  test('fires the onchange only when date length is same as formatString with single date', () => {
+  test('does not fire the onchange when date length is less than passed formatString with single date', () => {
     const onChange = jest.fn();
     const format = 'dd.MM.yyyy';
     const date = new Date('2019 01 01');
@@ -258,6 +258,21 @@ describe('Datepicker', () => {
       date: parse(newDate, format, new Date()),
     });
     expect(onChange.mock.calls).toHaveLength(1);
+  });
+
+  // Issue #3230
+  test('does not fire the onchange when date length is less than default formatDate with single date when no formatString is passed', () => {
+    const onChange = jest.fn();
+    const component = mount(<Datepicker onChange={onChange} value={null} />);
+
+    // $FlowFixMe
+    component.instance().handleInputChange({
+      currentTarget: {
+        value: '1',
+      },
+    });
+
+    expect(onChange.mock.calls).toHaveLength(0);
   });
 
   test('returns an array of date objects on input change', () => {
--- a/src/datepicker/datepicker.js
+++ b/src/datepicker/datepicker.js
@@ -35,6 +35,7 @@ export default class Datepicker extends React.Component<
   static defaultProps = {
     'aria-describedby': 'datepicker--screenreader--message--input',
     value: null,
+    formatString: 'yyyy/MM/dd',
   };
 
   calendar: ?HTMLElement;
@@ -112,7 +113,7 @@ export default class Datepicker extends React.Component<
   formatDisplayValue(date: ?Date | Array<Date>) {
     let formatDisplayValue = this.props.formatDisplayValue || this.formatDate;
     formatDisplayValue = formatDisplayValue.bind(this);
-    return formatDisplayValue(date, this.props.formatString || 'yyyy/MM/dd');
+    return formatDisplayValue(date, this.props.formatString);
   }
 
   open = () => {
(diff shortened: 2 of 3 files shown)


src/datepicker/__tests__/datepicker.test.js
```diff
@@ -229,7 +229,7 @@ describe('Datepicker', () => {
     });
   });
 
-  test('fires the onchange only when date length is same as formatString with single date', () => {
+  test('does not fire the onchange when date length is less than passed formatString with single date', () => {
```
devanshj, line 232: I hope it's okay I changed this a bit to match the implementation because the guard is (a < b) not (a !== b)

src/datepicker/__tests__/datepicker.test.js
```diff
@@ -260,6 +260,21 @@ describe('Datepicker', () => {
     expect(onChange.mock.calls).toHaveLength(1);
   });
 
+  // Issue #3230
+  test('does not fire the onchange when date length is less than default formatDate with single date when no formatString is passed', () => {
+    const onChange = jest.fn();
+    const component = mount(<Datepicker onChange={onChange} value={null} />);
+
+    // $FlowFixMe
```
devanshj, line 268: I don't have flow extension installed so I guess if there's an error in above test it also would be in this one so I kept $FlowFixMe as it is

devanshj:
> [CLA assistant check](https://cla-assistant.io/uber/baseweb?pullRequest=3275)
> Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/uber/baseweb?pullRequest=3275) before we can accept your contribution.
> You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.io/check/uber/baseweb?pullRequest=3275) it.

Eh do y'all really need my phone number for the Contributor License Agreement? That's too personal I guess, I can provide it but is there a way to bail out?

src/datepicker/datepicker.js
```diff
@@ -221,18 +221,17 @@ export default class Datepicker extends React.Component<
...
-        // Eg 25.12.2 will be transformed to 25.12.0002 formatted from date to string
-        if (
-          dateString.replace(/(\s)*/g, '').length <
-          formatString.replace(/(\s)*/g, '').length
-        ) {
-          date = null;
-        } else {
-          date = parse(dateString, formatString, new Date());
-        }
+      const formatString = this.props.formatString || 'yyyy/MM/dd';
```
devanshj, line 224: I could have extracted `'yyyy/MM/dd'` to something like `DEFAULT_FORMAT_STRING` but it's hardcoded everywhere so I just followed that for consistency

nadiia: Could you add this string as a default value for `formatString`? Then we won't need a conditional assignment.

nadiia:
> Eh do y'all really need my phone number for the Contributor License Agreement? That's too personal I guess, I can provide it but is there a way to bail out?
> 
> Edit: I'm just asking to confirm, the answer can be a "No" and I'll sign it anyway xD

Unless it's a non-required field otherwise the asked data needs to be provided. 
Thank you for the fix.

devanshj:
> Unless it's a non-required field otherwise the asked data needs to be provided.

Cool, signed

> Thank you for the fix.

You're welcome

src/datepicker/datepicker.js
devanshj: You mean in the `static defaultProps`?

nadiia: correct

devanshj:
Umm looks like flow doesn't support inference from `defaultProps`... Haven't worked with flow before how do you fix it?

PR #3282, opened 2020-04-29
fix(tag): forwards ref to root element

chasestarr:
#### Description

This change enables tags to be simply wrapped with `Popover` and `Tooltip` components like in [this example](https://codesandbox.io/s/tooltip-stateful-example-with-complex-content-pyyj4?file=/example.js)

#### Scope

- [x] Patch: Bug Fix


src/tag/__tests__/tag.test.js
```diff
@@ -6,76 +6,57 @@ LICENSE file in the root directory of this source tree.
 */
 // @flow
 import * as React from 'react';
-import {mount} from 'enzyme';
+import {render, fireEvent, getByTestId} from '@testing-library/react';
```
gergelyke, line 9: 👏

Issue #3290, opened 2020-05-01
Import declaration conflicts with local declaration of 'SIZE'.


derekmisler:
There's a naming conflict in `button-group/index.d.ts` and it's causing TS errors for us:

```
import {SHAPE, SIZE, KIND} from '../button';

...

export interface SIZE {
  compact: 'compact';
  default: 'default';
  large: 'large';
  mini: 'mini';
}
```

## Current Behavior

Typecheck fails with:
```
> tsc --build tsconfig.json

node_modules/baseui/button-group/index.d.ts:3:16 - error TS2440: Import declaration conflicts with local declaration of 'SIZE'.

3 import {SHAPE, SIZE, KIND} from '../button';
                 ~~~~
```

## Expected Behavior

There should be no conflict

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v9.73.2  |
| React   | 16.13.1 |
| browser | Firefox 76.0b8 (64-bit) |
| npm | 6.13.4 |
| Node | v12.16.1 |

- [x] I have searched the [issues](https://github.com/uber/baseweb/issues) of this repository and believe that this is not a duplicate.

PR #3291, opened 2020-05-01
fix(button-group): remove duplicate definition

gergelyke:
Closes https://github.com/uber/baseweb/issues/3290

### DIFF
--- a/src/button-group/index.d.ts
+++ b/src/button-group/index.d.ts
@@ -11,13 +11,6 @@ export interface STATE_CHANGE_TYPE {
   change: 'change';
 }
 
-export interface SIZE {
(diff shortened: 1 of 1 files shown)


src/button-group/index.d.ts
```diff
@@ -11,13 +11,6 @@ export interface STATE_CHANGE_TYPE {
   change: 'change';
 }
 
-export interface SIZE {
```
nadiia, line 14: Should we re-export the imported `SIZE` above?

gergelyke: yeah, it's re-exported at the bottom of the file, just not visible in this changeset

PR #3292, opened 2020-05-01
fix(toast): respect the css `transition-duration` for unmount

### LINKED ISSUE #3194: [Toast] Zombie toast not being removed from DOM
After pressing the "x" to close the toast, mouse out and back over the toast before it disappears. Then try to click the "Show Notification" button again.
zombie-toast

https://codesandbox.io/s/toastercontainer-4d1gc?file=/example.js

## Current Behavior
Unable to click "Show Notification" button again as the toast hasn't been removed from the DOM.

## Expected Behavior
Able to click "Show Notification" button again.

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v9.68.0  |
| React   | v16.8.6 |
| browser | Chrome 81.0.4044.92 |

devanshj:
Fixes #3194

#### Description

Problem was the `setTimeout` was set for arbitrary 600ms (maybe it was the default for previous releases idk) and now the default transition duration is 200ms hence if you mouseover in those 400ms gap it doesn't unmount (as the gif in the issue shows).

Now using 200 instead of 600 is also not good enough because we could have overrides via prop or even directly (like outside of styletron and baseweb which we shouldn't bother about but okay) so we're using `getComputedStyle` to get the transition duration.

#### Scope

- [x] Patch: Bug Fix
- [ ] Minor: New Feature
- [ ] Major: Breaking Change

### DIFF
--- a/src/toast/toast.js
+++ b/src/toast/toast.js
@@ -48,6 +48,7 @@ class Toast extends React.Component<ToastPropsT, ToastPrivateStateT> {
   animateOutCompleteTimer: ?TimeoutID;
   closeRef: ?{current: ?mixed};
   previouslyFocusedElement: ?HTMLElement;
+  bodyRef: ?{current: ?mixed};
 
   state = {
     isVisible: false,
@@ -58,6 +59,7 @@ class Toast extends React.Component<ToastPropsT, ToastPrivateStateT> {
   constructor(props: ToastPropsT) {
     super(props);
     this.closeRef = React.createRef();
+    this.bodyRef = React.createRef();
     this.previouslyFocusedElement = null;
   }
 
@@ -137,11 +139,23 @@ class Toast extends React.Component<ToastPropsT, ToastPrivateStateT> {
 
   animateOut = (callback: () => mixed = () => {}) => {
     this.setState({isVisible: false});
+
+    let animationDuration =
+      (__BROWSER__ &&
+        this.bodyRef &&
+        this.bodyRef.current &&
+        Number(
+          getComputedStyle(this.bodyRef.current)
+            .getPropertyValue('transition-duration')
+            .replace('s', ''),
+        ) * 1000) ||
+      200; // default backup just in case
+
     // Remove the toast from the DOM after animation finishes
     this.animateOutCompleteTimer = setTimeout(() => {
       this.setState({isRendered: false});
       callback();
-    }, 600);
+    }, animationDuration);
   };
 
   dismiss = () => {
@@ -226,6 +240,7 @@ class Toast extends React.Component<ToastPropsT, ToastPrivateStateT> {
             {...sharedProps}
             {...bodyProps}
             // the properties below have to go after overrides
+            ref={this.bodyRef}
             onBlur={this.onBlur}
             onFocus={this.onFocus}
             onMouseEnter={this.onMouseEnter}
--- a/src/toast/toast.js
+++ b/src/toast/toast.js
@@ -5,7 +5,7 @@ This source code is licensed under the MIT license found in the
 LICENSE file in the root directory of this source tree.
 */
 // @flow
-/* global document */
+/* global document, window */
 import * as React from 'react';
 import {getOverrides, mergeOverrides} from '../helpers/overrides.js';
 import DeleteIcon from '../icon/delete.js';
@@ -141,17 +141,15 @@ class Toast extends React.Component<ToastPropsT, ToastPrivateStateT> {
     this.setState({isVisible: false});
 
     let animationDuration =
-      (__BROWSER__ &&
-        this.bodyRef &&
-        this.bodyRef.current &&
-        Number(
-          // $FlowFixMe: Body is typed as `mixed` but actually is Element so this isn't a problem
-          window
-            .getComputedStyle(this.bodyRef.current)
-            .getPropertyValue('transition-duration')
-            .replace('s', ''),
-        ) * 1000) ||
-      200; // default backup just in case
+      __BROWSER__ && this.bodyRef && this.bodyRef.current
+        ? Number(
+            // $FlowFixMe: Body is typed as `mixed` but actually is Element so this isn't a problem
+            window
+              .getComputedStyle(this.bodyRef.current)
+              .getPropertyValue('transition-duration')
+              .replace('s', ''),
+          ) * 1000
+        : 200; // default backup just in case
 
     // Remove the toast from the DOM after animation finishes
     this.animateOutCompleteTimer = setTimeout(() => {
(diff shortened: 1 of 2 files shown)


src/toast/toast.js
```diff
@@ -137,11 +139,25 @@ class Toast extends React.Component<ToastPropsT, ToastPrivateStateT> {
...
+        this.bodyRef &&
+        this.bodyRef.current &&
+        Number(
+          // $FlowFixMe: Body is typed as `mixed` but actually is Element so this isn't a problem
+          window
+            .getComputedStyle(this.bodyRef.current)
+            .getPropertyValue('transition-duration')
+            .replace('s', ''),
+        ) * 1000) ||
+      200; // default backup just in case
```
devanshj, line 154: Should I remove the `|| 200` altogether? Also if not I just realised need to handle the zero duration case coz that would also be falsy, will send patch (if we're keeping it)

src/toast/toast.js
```diff
@@ -137,11 +139,23 @@ class Toast extends React.Component<ToastPropsT, ToastPrivateStateT> {
 
   animateOut = (callback: () => mixed = () => {}) => {
     this.setState({isVisible: false});
+
+    let animationDuration =
```
chasestarr, line 143: please separate the environment checks from runtime checks like below

```js
let animationDuration = 200;
if (__BROWSER__) {
  if (this.bodyRef && this.bodyRef.current) {
    // set animation duration
  }
}
```

devanshj: So should I also change it [here](https://github.com/uber/baseweb/blob/dae593cc719adc2ad42839df0a601b361853dede/src/toast/toast.js#L68)?
Also may I know why does it matter? Is the compiler/minifier not smart enough to truncate ternaries but can truncate if branches?

devanshj:
> This is a great catch, but I don't think we should go so far in the implementation. We shouldn't be expected to handle every customization possible via overrides. If we want to support a specific customization like this we should do so with a prop or using a theme property (`timing200`) to sync the timing in all the relevant places.

I guess you mean we should take a `transitionDuration` kinda prop? Like [blueprintjs does](https://i.imgur.com/BBztGRF.png)?
I get your point that it feels as if we're going too far but with something like that we make users responsible for syncing `transition-duration` css value and `transitionDuration` prop which is kinda awkward.

One improvement to this is we could not respect `transition-duration` css at all, we always overwrite it with whatever the `transitionDuration` prop says but that still is not good enough because someone could write `transition-duration: something` expecting it to work but we behind the scenes overwrite it with default value of `transitionDuration` prop. We would have to make sure we only overwrite opacity transition duration.

Also just realised even this PR is actually buggy because we actually care about the `opacity` and computed value of transition duration can be `0.2s, 0.3s`. Which can be fixed of course.

Let me know how should I go forward with this. My preference would be to get transition duration of opacity and let the user have the ultimate source of truth via css / overrides.

devanshj:
> I think we need to solve this issue by something like marking the toast as closed when the `dismiss` is called that will prevent from resetting timers on mouse events. Checking for the transition duration value here seems to be an overkill.

Yes we could do that but that seems we're removing the feature that you could make the toast not disappear when it's mouseovered when it has let's say 50% opacity. For longer transition duration like let's say 1 second this would degrade the ux as the user would expect hovering would stop the disappearance.

src/toast/toast.js
chasestarr: To me it follows convention set by [C directives](https://www.techonthenet.com/c_language/directives/if.php) and my general aversion to multi-line ternary expressions, but no blocking need to change this or other code

devanshj: Ahh gotcha but I'm quite an [expressions-ov](https://fsharpforfunandprofit.com/posts/expressions-vs-statements/)[er-statements](http://blog.agiledeveloper.com/2015/08/functional-programming-favors.html) guy we could have a nice middle-ground if javascript had an expression version of if-else haha. Anyways I'll make the changes once I know where this PR is going.

nadiia:
Got a quick fix for it in #3296. Thanks for looking into the issue! Closing this PR.

devanshj:
You're welcome! But just to make it clear: your fix removes the feature that I mentioned (and has other problematic things like arbitrary 600 delay for onClose, redundant timeouts, etc) If we actually want to compromise on that feature then that's okay if not then let's continue the discussion on how to move forward.
So the question is are we removing that feature/behavior?

nadiia:
Not sure what you mean by `your fix removes the feature that I mentioned`. The change does not remove any of the currently available features.

devanshj:
> Yes we could do that but that seems we're removing the feature that you could make the toast not disappear when it's mouseovered when it has let's say 50% opacity. For longer transition duration like let's say 1 second this would degrade the ux as the user would expect hovering would stop the disappearance.

I was referring to this, I was suspicious didn't make myself clear. Follow these steps to see what feature I'm talking about
1. Go to [this codesandbox](https://codesandbox.io/s/pensive-napier-vtudv?file=/src/index.tsx)
2. Click on close button of "With feature" toast
3. Make the mouse leave the toast
4. Before it disappears make the mouse enter the toast again
5. Observe not it stops disappearing
6. Repeat step 2 to 4 for "Without feature" toast and observe it doesn't stop disappearing

The difference is one OPENED & CLOSED states and other has OPENED, CLOSING & CLOSED states. The CLOSING state and the behavior when the component is in that state is the feature.

We need that CLOSING state hence if have all those timeouts, clearTimeouts, etc in the first place. The following line is responsible for "cancelling" the process of disappearance
https://github.com/uber/baseweb/blob/dae593cc719adc2ad42839df0a601b361853dede/src/toast/toast.js#L162

I'm sorry if I dumbed it out too much I just want to make sure I make myself clear :)

PS:
I haven't tried it but you can reproduce this behavior for baseweb by having `transitionDuration` as `"3s"` [here](https://github.com/uber/baseweb/blob/dae593cc719adc2ad42839df0a601b361853dede/src/toast/styled-components.js#L138) and having timeout as `3000` instead of `600` [here](https://github.com/uber/baseweb/blob/dae593cc719adc2ad42839df0a601b361853dede/src/toast/toast.js#L144)

devanshj:
Away from keyboard but one solution we haven't explored is using the `transitionEnd` event that would not require to compute transition duration... Hopefully if it doesn't fire when transitions are cancelled we can use it, I'll check it out tommo...

nadiia:
The toast should not reset its explicitly called dismiss event on `mouseenter`.

devanshj:
> The toast should not reset its explicitly called dismiss event on mouseenter.

I doubt that's the expected behaviour, if it was we wouldn't have the `clearTimeout` for cancelling mount and `onClose` and all that logic in the first place.

PR #3293, opened 2020-05-01
chore(.github): PR template

devanshj:
Makes the template not use lists with checkboxes as github thinks those are tasks as it adds description like "1 of 3 Tasks completed" in the linked issues, PRs, etc

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -5,7 +5,8 @@
 <!-- Describe your changes below in as much detail as possible -->
 
 #### Scope
-
-- [ ] Patch: Bug Fix
-- [ ] Minor: New Feature
-- [ ] Major: Breaking Change
+<!-- Pick one:
+Patch: Bug Fix
+Minor: New Feature
+Major: Breaking Change
+-->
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -5,7 +5,8 @@
...
 
 #### Scope
-
-- [ ] Patch: Bug Fix
-- [ ] Minor: New Feature
-- [ ] Major: Breaking Change
+<!-- Pick one:
+Patch: Bug Fix
+Minor: New Feature
+Major: Breaking Change
```
gergelyke, line 11:
```suggestion
```

gergelyke: we might as well remove this, as we do not except breaking changes from contributors

devanshj: I have committed the change although I think contributors should still be allowed to send PRs that have a breaking change (like a bug fix could be a breaking change), you then take a call whether to merge it when you release a major or reject it altogether.

gergelyke: hm, that's a good point 🤔 I'd be happy to merge either way, let me know what you perefer

devanshj: I guess we should keep as it was, I'll revert in that case

PR #3294, opened 2020-05-03
fix(button): make .d.ts reflect that it's uses forwardRef

### LINKED ISSUE #2652: Missing addEventListener and removeEventListener for Button type, Button should extend HTMLButtonElement
The Button element does not include types for addEventListener or removeEventListener, so Typescript will complain when using those functions.

## Current Behavior

Typescript shows an error when using a HTMLButtonElement property that is not part of Button:

"Property 'addEventListener' does not exist on type 'Button'.ts(2339)"

The consumer cannot create a new type which extends both Button and HTMLButtonElement because the focus() function on Button doesn't match the focus function on HTMLButtonElement. If Button extended HTMLButtonElement then they would have to match.

## Expected Behavior

Button should extend HTMLButtonElement

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v9.42.0  |

devanshj:
Fixed #2652

#### Description

`Button` is already forwarding the ref to html button element so there won't be any problems in runtime in code on the issue page, only the types weren't reflecting that. [Here's the test](https://codesandbox.io/s/summer-sun-ejnct?file=/example.tsx) for this PR. (there's no type testing strategy in the repo hence testing this way)

Also there are a lot of components that forward the ref but the types don't reflect that so can send a PR that fixes all those types?

#### Scope
Patch: Bug Fix

### DIFF
--- a/src/button/index.d.ts
+++ b/src/button/index.d.ts
@@ -60,7 +60,7 @@ export const StyledLoadingSpinner: StyletronComponent<any>;
 export const StyledLoadingSpinnerContainer: StyletronComponent<any>;
 
 export const Button: React.ForwardRefExoticComponent<
-  React.PropsWithoutRef<ButtonProps> & React.RefObject<HTMLButtonElement>
+  React.PropsWithoutRef<ButtonProps> & React.RefAttributes<HTMLButtonElement>
 >;
 
 export const KIND: KIND;
--- a/src/button/index.d.ts
+++ b/src/button/index.d.ts
@@ -59,10 +59,9 @@ export const StyledEndEnhancer: StyletronComponent<any>;
 export const StyledLoadingSpinner: StyletronComponent<any>;
 export const StyledLoadingSpinnerContainer: StyletronComponent<any>;
 
-export class Button extends React.Component<ButtonProps> {
-  internalOnClick(...args: any): void;
-  focus(): void;
-}
+export const Button: React.ForwardRefExoticComponent<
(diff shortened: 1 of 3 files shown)


src/button/index.d.ts
```diff
@@ -59,10 +59,9 @@ export const StyledEndEnhancer: StyletronComponent<any>;
 export const StyledLoadingSpinner: StyletronComponent<any>;
 export const StyledLoadingSpinnerContainer: StyletronComponent<any>;
 
-export class Button extends React.Component<ButtonProps> {
-  internalOnClick(...args: any): void;
```
devanshj, line 63: There's no way you can access `internalOnClick` because the ref is being forwarded to the html button element not the class

src/button/index.d.ts
```diff
@@ -59,10 +59,9 @@ export const StyledEndEnhancer: StyletronComponent<any>;
...
-  focus(): void;
-}
+export const Button: React.ForwardRefExoticComponent<
```
devanshj, line 62: [Here's the reference](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/6de1126c1b450880518bdb58f85fe84d27879943/types/react/index.d.ts#L805) where this type came from

devanshj:
You're welcome!

> Also there are a lot of components that forward the ref but the types don't reflect that so can send a PR that fixes all those types?

Would that be good?

PR #3295, opened 2020-05-03
fix(popover): don't clone anchor with ref if it's not using forwardRef

### LINKED ISSUE #3265: Popover is not appearing if anchor component does not forward ref
## Feature description
Popover should not be dependent on attributes of child component.

Check 2 examples in the codesandbox for better understanding
https://codesandbox.io/s/statefulpopover-56pd9?file=/example.js

devanshj:
Closes #3265

#### Description
We're already logging a warning if the anchor ref is not being set, but if it's a functional component without `forwardRef` React logs an error that you can't set ref and bails out thus our warning doesn't log. So the fix doesn't set the ref while cloning hence out warning would get logged.

#### Scope
Minor: New Feature
(going for a minor because it's also kinda a feature)

### DIFF
--- a/src/layer/tether.js
+++ b/src/layer/tether.js
@@ -48,9 +48,9 @@ class Tether extends React.Component<TetherPropsT, TetherStateT> {
           if (__DEV__) {
             // eslint-disable-next-line no-console
             console.warn(
-              `[baseui][TetherBehavior] ref has not been passed to the Popper's anchor element.
-              See how to pass the ref to an anchor element in the Popover example
-              http://baseui.design/components/popover#anchor-ref-handling-example`,
+              `[baseui][TetherBehavior] ref has not been passed to the Popper's anchor element.\n` +
+                `See how to pass the ref to an anchor element in the Popover example\n` +
+                `http://baseui.design/components/popover#anchor-ref-handling-example`,
             );
           }
         } else {
--- a/src/layer/tether.js
+++ b/src/layer/tether.js
@@ -48,9 +48,9 @@ class Tether extends React.Component<TetherPropsT, TetherStateT> {
           if (__DEV__) {
             // eslint-disable-next-line no-console
             console.warn(
-              `[baseui][TetherBehavior] ref has not been passed to the Popper's anchor element.\n` +
-                `See how to pass the ref to an anchor element in the Popover example\n` +
-                `http://baseui.design/components/popover#anchor-ref-handling-example`,
+              '[baseui][TetherBehavior] ref has not been passed to the Popper's anchor element.\n' +
+                'See how to pass the ref to an anchor element in the Popover example\n' +
+                'http://baseui.design/components/popover#anchor-ref-handling-example',
             );
           }
         } else {
--- a/src/popover/popover.js
+++ b/src/popover/popover.js
@@ -361,10 +361,20 @@ class Popover extends React.Component<PopoverPropsT, PopoverPrivateStateT> {
       return null;
     }
 
-    const isValidElement = React.isValidElement(anchor);
-    const anchorProps = this.getAnchorProps();
+    let anchorProps = this.getAnchorProps();
 
-    if (typeof anchor === 'object' && isValidElement) {
+    if (typeof anchor === 'object' && React.isValidElement(anchor)) {
+      if (typeof anchor.type === 'function') {
+        /* if is non-forwardRef functional component
+         * (if it's wrapped by forwardRef then the type is no longer a function)
+         * then remove ref because rendering non-forwardRef functional component throws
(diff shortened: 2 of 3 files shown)


devanshj:
Hmm the snapshots are apparently passing locally 🤔 let me try undoing them to see if it passes in the CI

src/popover/popover.js
```diff
@@ -361,10 +361,20 @@ class Popover extends React.Component<PopoverPropsT, PopoverPrivateStateT> {
       return null;
     }
 
-    const isValidElement = React.isValidElement(anchor);
-    const anchorProps = this.getAnchorProps();
+    let anchorProps = this.getAnchorProps();
+
+    if (typeof anchor === 'object' && React.isValidElement(anchor)) {
+      if (typeof anchor.type === 'function') {
```
chasestarr, line 367: Reading undocumented values on the react element is putting this component at risk of regressions on react minor version changes. Silently not forwarding the ref is going to lead to more unexpected behavior

src/popover/popover.js
```diff
@@ -361,10 +361,20 @@ class Popover extends React.Component<PopoverPropsT, PopoverPrivateStateT> {
...
+        /* if is non-forwardRef functional component
+         * (if it's wrapped by forwardRef then the type is no longer a function)
+         * then remove ref because rendering non-forwardRef functional component throws
+         * an error and which makes tether unable to show the ref not set warning
```
chasestarr, line 371: Please see the related issue, there is still a warning provided by react that seems actionable-enough to me

src/popover/popover.js
devanshj: Yeah fair enough let me see if I can come up with something

src/popover/popover.js
devanshj: Yeah I replied there

src/popover/popover.js
chasestarr: 👍 Thanks for looking into the problem. I'm happy to land a solution, but hopefully in a more maintainable way

chasestarr:
> Also how does buildkite work like what does "passed & block" mean? Coz it said that before and now that I see it had failures... Like does a maintainer has to do something and unblock it and then it actually runs?

It doesn't run CI until we have a chance to read through the diff and verify nothing suspicious is happening. Someone with commit access will need to 'unblock' CI

src/popover/popover.js
devanshj: Yeah no problems 👍

devanshj:
> It doesn't run CI until we have a chance to read through the diff and verify nothing suspicious is happening. Someone with commit access will need to 'unblock' CI

Ahhh exactly what I guessed, I used to think the "Passed & Blocked" part meant it ran and passed and when I come here after sometime boom it fails xD

src/layer/tether.js
```diff
@@ -48,9 +48,9 @@ class Tether extends React.Component<TetherPropsT, TetherStateT> {
           if (__DEV__) {
             // eslint-disable-next-line no-console
             console.warn(
-              `[baseui][TetherBehavior] ref has not been passed to the Popper's anchor element.
-              See how to pass the ref to an anchor element in the Popover example
-              http://baseui.design/components/popover#anchor-ref-handling-example`,
+              `[baseui][TetherBehavior] ref has not been passed to the Popper's anchor element.\n` +
```
nadiia, line 51:
```suggestion
              '[baseui][TetherBehavior] ref has not been passed to the Popper's anchor element.\n' +
```

src/layer/tether.js
```diff
@@ -48,9 +48,9 @@ class Tether extends React.Component<TetherPropsT, TetherStateT> {
...
+                `See how to pass the ref to an anchor element in the Popover example\n` +
```
nadiia, line 52:
```suggestion
                'See how to pass the ref to an anchor element in the Popover example\n' +
```

src/layer/tether.js
```diff
@@ -48,9 +48,9 @@ class Tether extends React.Component<TetherPropsT, TetherStateT> {
...
+                `http://baseui.design/components/popover#anchor-ref-handling-example`,
```
nadiia, line 53:
```suggestion
                'http://baseui.design/components/popover#anchor-ref-handling-example',
```

chasestarr:
Closing for now, but feel free to reopen

devanshj:
Yeah cool! I was actually gonna send a proposal on react repo to have `isRefableElement` gaurd in `react-is` package (it already has `isForwardRef` but that's not enough coz the element could be a dom element too) so we could use that but haven't got time to send it. In case they accept the proposal I'll open a PR to use it here.

PR #3297, opened 2020-05-04
feat(combobox): adds component

chasestarr:
#### Description

[relevant aria-practices](https://www.w3.org/TR/wai-aria-practices/#combobox)

Screen Shot 2020-05-11 at 3 55 14 PM

#### Scope
Minor: New Feature


chasestarr:
Latest commit here adds documentation

documentation-site/pages/components/combobox.mdx
```diff
@@ -0,0 +1,63 @@
...
+
+<Example title="Filtered options" path="combobox/filtered-options.js">
+  <FilteredOptions />
+</Example>
+
+<Example title="Custom option node" path="combobox/replacement-node.js">
+  <ReplacementNode />
+</Example>
+
+<Example title="Available sizes" path="combobox/sizes.js">
```
gergelyke, line 53: As these are available from `react-view`, should we still have them? Less code to maintain 😬

kahwee:
Can't wait to test this out in one of our apps! Good job!

kahwee:
Can you share an example where the options are fetched remotely (with settimeout) asynchronously as the user types?

tajo:
When looking into the example, I am not completely sure what the relation between `val`, `label` and `id` is. Also there is `mapOptionToString`. What string?

1. Should `mapOptionToString` be rather called `mapOptionToLabel`?
2. I would expect `onChange` to return the full option object or having mandatory `mapOptionToValue` prop.

Also the layering doesn't work but that's probably WIP?

Screen Shot 2020-05-11 at 11 28 56 AM

chasestarr:
> Can you share an example where the options are fetched remotely (with settimeout) asynchronously as the user types?

@kahwee here is an example of async options https://baseweb-olkgexc1l.now.sh/components/combobox#async-options

chasestarr:
> When looking into the example, I am not completely sure what the relation between `val`, `label` and `id` is. What is used as a `key` / needs to be unique?

Can you clarify what you mean here? I updated the `val` keyword to `nextValue` to be more precise. Is there more that should be improved?
 
> Also there is `mapOptionToString`. Should `mapOptionToString` be rather called `mapOptionToLabel` or even `mapOptionToValue`?

`mapOptionToString` serves two purposes, the first is to render option text and it instructs how to render intermediate values in the input element. Given that added context, what do you think it should be called?

> Also the layering doesn't work but that's probably WIP?

Yeah I've been saving this bug hunting for the last step

tajo:
> > When looking into the example, I am not completely sure what the relation between `val`, `label` and `id` is. What is used as a `key` / needs to be unique?
> 
> Can you clarify what you mean here? I updated the `val` keyword to `nextValue` to be more precise. Is there more that should be improved?
> 
> > Also there is `mapOptionToString`. Should `mapOptionToString` be rather called `mapOptionToLabel` or even `mapOptionToValue`?
> 
> `mapOptionToString` serves two purposes, the first is to render option text and it instructs how to render intermediate values in the input element. Given that added context, what do you think it should be called?
> 
> > Also the layering doesn't work but that's probably WIP?
> 
> ~Yeah I've been saving this bug hunting for the last step~ fixed in the most recent commit

Ah ok, it makes all sense to me now.

chasestarr:
> Looks pretty nice. Do you think it would be useful to add a couple "behavior" type VRTs?

I'll make a follow-up pr

PR #3305, opened 2020-05-07
fix(tab): spread overrides for tabs too

gergelyke:


src/tabs/tabs.js
```diff
@@ -44,6 +50,7 @@ export default class Tabs extends React.Component<TabsPropsT> {
         $orientation: orientation,
         onSelect: () => this.onChange({activeKey: key}),
         children: child.props.title,
+        overrides,
```
nadiia, line 53: wonder if we should merge it with `child.props.overrides` with latest takes precedence

gergelyke: that's a good point - I have to do that, thanks Nadiia

src/tabs/tabs.js
```diff
@@ -44,6 +50,7 @@ export default class Tabs extends React.Component<TabsPropsT> {
         $orientation: orientation,
         onSelect: () => this.onChange({activeKey: key}),
         children: child.props.title,
+        overrides: child.props.overrides || overrides,
```
nadiia, line 53:
```suggestion
        overrides: mergeOverrides(overrides, child.props.overrides || {})
```

nadiia: `mergeOverrides` from `import {getOverrides, mergeOverrides} from '../helpers/overrides.js';`

PR #3309, opened 2020-05-11
feat(side-nav): add disabled prop to items

gergelyke:


src/side-navigation/nav-item.js
```diff
@@ -43,15 +43,19 @@ class NavItem extends React.Component<NavItemPropsT> {
...
+    const tabIndex = {
+      tabIndex: item.disabled ? -1 : undefined,
+    };
     return (
       <NavLink
         href={item.itemId}
+        {...tabIndex}
         {...sharedProps}
         {...linkProps}
         {...(item.itemId
```
nadiia, line 55:
```suggestion
        {...(item.itemId && !item.disabled
```

src/side-navigation/nav-item.js
```diff
@@ -43,15 +43,19 @@ class NavItem extends React.Component<NavItemPropsT> {
...
           ? {
-              onClick: this.handleClick,
-              onKeyDown: this.handleKeyDown,
+              onClick: item.disabled ? () => {} : this.handleClick,
```
nadiia, line 57:
```suggestion
              onClick: this.handleClick,
```

src/side-navigation/nav-item.js
```diff
@@ -43,15 +43,19 @@ class NavItem extends React.Component<NavItemPropsT> {
...
+              onKeyDown: item.disabled ? () => {} : this.handleKeyDown,
```
nadiia, line 58:
```suggestion
              onKeyDown: this.handleKeyDown,
```

src/side-navigation/styled-components.js
```diff
@@ -69,6 +70,16 @@ export const StyledNavItemElement = styled<SharedPropsT>('div', props => {
   const paddingSuffixDir: string =
     $theme.direction === 'rtl' ? 'paddingLeft' : 'paddingRight';
 
+  let cursor = $selectable ? 'pointer' : 'default';
+  let color = $active ? colors.primary : null;
+  let hoverColor = $selectable ? colors.primary : null;
+
+  if ($disabled) {
+    cursor = 'not-allowed';
```
tajo, line 78: Personally I would keep it as `default`. `not-allowed` seems too aggressive and we don't use it for disabled elements?

src/side-navigation/styled-components.js
```diff
@@ -78,15 +89,15 @@ export const StyledNavItemElement = styled<SharedPropsT>('div', props => {
     [borderWidthDir]: '4px',
     [borderStyleDir]: 'solid',
     [borderColorDir]: $active ? colors.primary : 'transparent',
-    color: $active ? colors.primary : null,
-    cursor: $selectable ? 'pointer' : 'default',
+    color: color,
```
tajo, line 92:
```suggestion
    color,
```

nit

src/side-navigation/styled-components.js
```diff
@@ -78,15 +89,15 @@ export const StyledNavItemElement = styled<SharedPropsT>('div', props => {
...
+    cursor: cursor,
```
tajo, line 93:
```suggestion
    cursor,
```

src/side-navigation/styled-components.js
gergelyke: we've discussed this offline, but for prosperity's sake: turns out ,we use it in quite a few places, so we've decided to keep things as is https://github.com/uber/baseweb/search?q=not-allowed&unscoped_q=not-allowed

PR #3310, opened 2020-05-11
docs: figma

sandgraham:
Adds a POC for exporting Figma documentation to a web page. This is still a bit rough. I'm sure things could be faster and more efficient in some places.

You can find the root index at https://baseweb-git-design-docs-poc-figma.uber-ui-platform.now.sh/guidelines.

Figma returns a tree of nodes when you query a file. This translates to the DOM, and in turn, React components pretty cleanly.

Some known issues:

- Everything is absolutely positioned. If we restructure our mockups a bit we could probably get responsive layouts working.
- There is no dark mode. Originally I pulled from the theme to convert values and respect dark mode, but there are too many instances of one off values to fully rely on theme tokens right now.
- Still need to implement navigation and search.
- Types would be useful for the large amount of Figma structures. Most flow types I've found are outdated or buggy.
- Need a convention for SEO/meta-data - currently not much happening there.


gergelyke:
what URL should we check to see the PoC?

sandgraham:
@gergelyke - it will be at `/guidelines`. Right now there is an issue building the page on Now that I am debugging. I'll remove draft status when it is fixed.

sandgraham:
@chasestarr - another possibility is leveraging Figma's [auto-layout](https://www.figma.com/blog/announcing-auto-layout/). If people place content in auto-layout containers it would be pretty similar to how we use layout components.

I need to work with doc writers to see what sorts of conventions we can establish in Figma so that we can translate things to a better web experience.

PR #3324, opened 2020-05-13
fix: call focus on treeItemRefs if ref is not null

jayantbh:
#### Description
Fixes the error where a ref may be null, but still be fired.
image

It could be that I'm using the component wrong, maybe I'm not forwarding a ref, but this check should still be in place.

#### Scope
Patch: Bug Fix

### DIFF
--- a/src/tree-view/tree-view.js
+++ b/src/tree-view/tree-view.js
@@ -48,7 +48,9 @@ export default function TreeView(props: TreeViewPropsT) {
   const focusTreeItem = (id: TreeNodeIdT | null) => {
     if (!id) return;
     setSelectedNodeId(id);
-    treeItemRefs.current[id].current.focus();
+
+    const node = treeItemRefs.current[id].current;
+    if (node) node.focus();
   };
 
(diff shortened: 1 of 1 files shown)


src/tree-view/tree-view.js
```diff
@@ -48,7 +48,7 @@ export default function TreeView(props: TreeViewPropsT) {
   const focusTreeItem = (id: TreeNodeIdT | null) => {
     if (!id) return;
     setSelectedNodeId(id);
-    treeItemRefs.current[id].current.focus();
+    treeItemRefs.current[id].current?.focus();
```
chasestarr, line 51:
```suggestion
    const item = treeItemRefs.current[id].current;
    if (item) {
      item.focus();
    }
```

jayantbh:
Aw sucks. Is optional chaining work in progress for this repo?

Also, updated.

chasestarr:
> Aw sucks. Is optional chaining work in progress for this repo?
> 
> Also, updated.

Thank you. We haven't looked into it yet, but I think babel needs to be updated.

PR #3327, opened 2020-05-13
docs(list): fix List example

chrisrauh:
#### Description

On codesandbox the List example was giving an error and the end enhacer was unstyled.

#### Scope
Patch: Bug Fix


gergelyke:
thank you @chrisrauh 🙏

PR #3329, opened 2020-05-13
fix(side-navigation): avoid setting href if disabled

chasestarr:
#### Description

If a side-nav item was disabled, a user could still right-click to navigate to the link. Now we avoid setting the href in this case

#### Scope
Patch: Bug Fix


src/side-navigation/nav-item.js
```diff
@@ -48,7 +48,7 @@ class NavItem extends React.Component<NavItemPropsT> {
     };
     return (
       <NavLink
-        href={item.itemId}
+        href={item.disabled ? null : item.itemId}
```
nadiia, line 51: I think it shouldn't be a link element if disabled.

chasestarr: Thank you. Updated to span if the item is disabled

PR #3331, opened 2020-05-14
docs: use stackblitz for code examples

sandgraham:
We've had some issues with the CodeSandbox bundler. Stackblitz is an alternate online IDE sandbox type thing. It doesn't have as many features, but it doesn't have the same bundling issues that CodeSandbox does. It is also a lot faster, just from some basic eye-testing. We can use it for now and if CodeSandbox fixes stuff on their end or releases some killer feature we can reevaluate.

Note, this PR doesn't remove _every_ trace of Codesandbox. Old links and embedded demos will stay as they are. Also the CodeSandbox CI is still set up to run on PRs - since it uses webpack it bundles correctly and might still be useful to keep around. If folks think otherwise I can remove it though.


tajo:
Nice!

https://baseweb-git-stackblitz.uber-ui-platform.now.sh/getting-started/learn/#base-web-with-data-fetching still uses CodeSandbox.

Also mentioned here: https://baseweb-git-stackblitz.uber-ui-platform.now.sh/components/layout-grid/#examples

nadiia:
Examples load so much faster! 🚀

PR #3340, opened 2020-05-14
fix(list): update list-item styles to spec

chasestarr:
### Description

Screen Shot 2020-05-14 at 3 48 00 PM

- For sizes up to 36 the artwork is centered on a 64 container
- If there is no artwork, then the label has a 16 left padding
- For artwork larger than 36 the artwork will have 16 padding on both sides

#### Scope
Patch: Bug Fix


chrisrauh:
The implementation looks right to spec. 👍

PR #3342, opened 2020-05-15
docs: update examples with prefilled arrays

sandgraham:
Noticed that StackBlitz compiles `new Array` a bit strangely and so subsequent `map` calls don't work correctly. https://github.com/stackblitz/core/issues/546

Seems that if you construct an array any other way it works fine. In this PR I've switched to `Array.from` as recommended in the linked issue.


chasestarr:
> can we also add a linter rule to make sure we won't run into this issue next time?

I think all of these cases are mine 😅 so I'll make a mental linter rule

PR #3345, opened 2020-05-18
docs: fix grammatical error in Learning Base Web

jkhaui:
`Fixes #1, Fixes #2`


sandgraham:
Looks like there is an unrelated issue with VRTs. Should be fine to merge in.

PR #3346, opened 2020-05-18
docs(textarea): fix shared props crash

sandgraham:


sandgraham:
I'm just going to include the VRTs - I think there was an update to Puppeteer/Chrome/Ubuntu or something.

PR #3350, opened 2020-05-19
feat(tree-view): export TreeLabelInteractable component for wrapping interactable tree node labels

nadiia:
There was an additional issue with a UI reflection of a false focus state, when keyboard navigating and switching focus to a focusable element in the tree node label the node would stay marked as focused but further tree keyboard navigation wouldn't work as the node is not actually focused.
This one I fixed with a check for a focus event on the node element.

The click issue wasn't that straightforward to work around with event targets due to multiple implementations of the components like different types of inputs. So I created a separate wrapper container that handles events' propagation.

#### Scope
Minor: New Feature


gergelyke:
can you also add an example to the doc site?

PR #3355, opened 2020-05-21
chore(project): update to next@9.4

gergelyke:
The static folder was deprecated, that's why the renaming: https://github.com/zeit/next.js/blob/master/errors/static-dir-deprecated.md


chasestarr:
Screen Shot 2020-05-21 at 9 21 01 AM
 images seem to be broken in the preview

gergelyke:
yeah, originally, I forgot to exclude the public dir in gitignore, so I think you see a previous build - will double check before merging

gergelyke:
my bad - I actually had to update the path of these static resources

PR #3359, opened 2020-05-23
docs(bedrock): remove references to bedrock from public documentation

angus-c:
It can be viewed on the public site and links to a google doc with comments from Uber employees. This feels inappropriate.


gergelyke:
@angus-c it only resolves for Uber employees - without an Uber login, nothing is visible, not even the google doc's url

what's the downside of keeping it as is?

angus-c:
> @angus-c it only resolves for Uber employees - without an Uber login, nothing is visible, not even the google doc's url
> 
> what's the downside of keeping it as is?

But they still see the text/link? Not a big deal I guess, I just don't see many OS products which have internal references in their README (also Bedrock is deprecated). I'm ok if you want to close it though.

gergelyke:
yeah, that sounds good, especially because it's already deprecated - for the record, we have t.uber.com links for a few more use-cases, but I'd rather keep those, as they are still actively used (like a link for the icons package). would that be ok?

angus-c:
sounds good, thanks!

PR #3365, opened 2020-05-27
fix(app-nav-bar): pass a str to Avatar, no drawer menu if no menu items

### LINKED ISSUE #3348: App Nav Bar - mainNav should hide if false or null array
## Current Behavior
In the AppNavBar Component, the mainNav if the value is not give or empty array, the list icon appears and on performing click event on icon it fails and asks for avatar and other components.

**Icon Screenshot** - list icon
image

**OnClick Event Screenshot**
image

## Expected Behavior
In the AppNavBar Component, the mainNav if the value is empty array, the list icon should appear but event should not happen.

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v9.72.0 |
| React   |  16.13.1   |
| Browser |  Chrome/ Firefox/ Safari |

nadiia:
Fixes #3348 

#### Description

It errored out when no username was provided.

#### Scope
Patch: Bug Fix

### DIFF
--- a/src/app-nav-bar/user-menu/user-menu.js
+++ b/src/app-nav-bar/user-menu/user-menu.js
@@ -35,7 +35,7 @@ export default function UserMenu(props: AppNavBarPropsT) {
       triggerType={TRIGGER_TYPE.click}
     >
       <Button overrides={{BaseButton: {component: StyledUserMenuButton}}}>
-        <Avatar name={username} src={userImgUrl} size={'32px'} />
+        <Avatar name={username || ''} src={userImgUrl} size={'32px'} />
         {isOpen ? (
           <ChevronUpSmallFilled
             size={28}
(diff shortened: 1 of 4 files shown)


src/app-nav-bar/user-menu/user-menu.js
```diff
@@ -35,7 +35,7 @@ export default function UserMenu(props: AppNavBarPropsT) {
       triggerType={TRIGGER_TYPE.click}
     >
       <Button overrides={{BaseButton: {component: StyledUserMenuButton}}}>
-        <Avatar name={username} src={userImgUrl} size={'32px'} />
+        <Avatar name={username || ''} src={userImgUrl} size={'32px'} />
```
gergelyke, line 38: in the case where the username is `undefined`, do we want to have an empty string as the prop value? or should we remove the prop entirely?

nadiia: `name` is a required prop in `Avatar` (could be replaced with custom `initials`) and it doesn't provide a default value, therefore, errors out.

PR #3373, opened 2020-05-28
docs: fix ie errors

### LINKED ISSUE #3354: Popover broken in IE11
ie11-base-web-popover 2020-05-20 02_02_18

It looks like the Popover component does not work on IE11. I looked at older versions on the documentation site, and the last working version was in Base Web v7.8.0.

ie11-base-web-popover-v7 2020-05-20 02_09_22

sandgraham:
Fixes #3354 

Basically there is a bunch of dependency code not being transpiled for some reason. I've added a list of modules which needs to be transpiled (via [next-transpile-modules](https://www.npmjs.com/package/next-transpile-modules)) to get things working in IE11.

I've also added a bail out for the `Yard` component when IE11 is detected - I believe the runtime would need to transpile the `react-view` code which doesn't seem worthwhile. If we don't bail out or transpile the page just shows some random Yard error.

There are also a couple style fixes just cause.

---

I'm wondering how maintainable this approach is- it was quite painful tracking down each module breaking IE11 and adding it to the transpile-list, but I couldn't find a nicer catch-all way to transpile dependencies using the next.js webpack config. Even the library I used, `next-transpile-modules`, is supposed to accept package scopes which would help with all the `@babel` packages, but alas, it didn't work out. Maybe there is a way to avoid the explicit listing.

Also, Vercel/Next works on IE11 out of the box, so not sure where we are going wrong in our configuration.

---

TODO

- [ ] check bundle size impact
- [ ] check deploy time impact


documentation-site/components/search.js
```diff
@@ -79,15 +79,15 @@ const IconWrapper = themedStyled<{$inputVisible: boolean}>(
   'div',
   ({$inputVisible, $theme}) => ({
     position: 'absolute',
+    top: '9.5px',
+    height: '22px',
```
nadiia, line 83:
```suggestion
```

PR #3377, opened 2020-05-28
fix(datepicker): handle selected time preservation when date is changed

nadiia:
#### Description

There are a few changes in here:
- time select now set to be nullable when no date selected
- the time selected is saved in the calendar's state and gets applied to appropriate dates when later changed

#### Scope
Patch: Bug Fix


src/datepicker/calendar.js
```diff
@@ -88,6 +88,11 @@ export default class Calendar extends React.Component<
       date: dateInView,
       quickSelectId: null,
       rootElement: null,
+      time: Array.isArray(this.props.value)
```
chasestarr, line 91: it may be useful for readability to pull the multiple ternaries into a mutable variable

chasestarr:
```js
let time = [];
if (Array.isArray(this.props.value) {
  time = this.props.value;
} else if (this.props.value) {
  time = [this.props.value];
}
```

PR #3378, opened 2020-05-28
docs(progress-bar): remove infinite example

gergelyke:
Yard provides the same functionality, and this has a bug in it, so 🔥


chasestarr:
strange.. the earliest instance of it I could find looks broken https://v9-47-0.baseweb.design/components/progress-bar

gergelyke:
yeah, I have no idea what's going on here :/

PR #3380, opened 2020-05-28
fix(accordion): update typescript accordion overrides type

### LINKED ISSUE #3376: AccordionOverrides typescript type incomplete.
## Current Behavior
Only `Root` is defined in the typescript type for AccordionOverrides. There are several other valid overrides as [described in the documentation](https://baseweb.design/components/accordion/). This causes typescript warnings and the typescript production build fails due to it thinking that an incorrect override is provided.

## Expected Behavior
There are no warnings with using accordion overrides other than `Root`.

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v9.78.0  |
| React   | v16.13.1        |
| Typescript | v3.9.3        |

balaclark:
Fixes #3376

#### Description

Adds in the missing accordion overrides to the typescript type.

#### Scope
Patch: Bug Fix

### DIFF
--- a/src/accordion/types.js
+++ b/src/accordion/types.js
@@ -33,7 +33,11 @@ export type PanelStateReducerT = (
 ) => PanelStateT;
 
 export type AccordionOverridesT = {
+  Content?: OverrideT,
+  Header?: OverrideT,
+  PanelContainer?: OverrideT,
   Root?: OverrideT,
+  ToggleIcon?: OverrideT,
 };
 
 export type PanelOverridesT = {


balaclark:
@gergelyke Thanks for approving this. This is my first contribution so I'm not sure of the process, do you know roughly when this should land?

gergelyke:
I was about to merge, but just double checked again - are you sure it would work? Looking at https://github.com/uber/baseweb/blob/master/src/accordion/accordion.js#L116-L130, it seems we do not pass the top level overrides down to the panels

balaclark:
The code you linked is just for the Root component, the child components (e.g. Panel) are rendered in a different method and receive the non-root overrides: https://github.com/uber/baseweb/blob/master/src/accordion/accordion.js#L69-L105

Specifically all other overrides come from props here: https://github.com/uber/baseweb/blob/master/src/accordion/accordion.js#L77 and are passed to the child components here: https://github.com/uber/baseweb/blob/master/src/accordion/accordion.js#L97 

The list of overrides I added comes from the documentation at: https://baseweb.design/components/accordion/. The documentation shows the other overrides working, you can also see this stackblitz with `Header` and `Content` overrides: https://stackblitz.com/edit/accordion-override?file=src/example.js

balaclark:
Thanks!

PR #3385, opened 2020-06-01
feat(combobox): include option in change handler params if suggestion is selected

chasestarr:
#### Description

In some situations, it's useful to know whether a combobox change came from user-keyboard-input vs user-selection. If a suggestion is selected, the option value will be provided to the change handler, otherwise null.

#### Scope
Minor: New Feature

### DIFF
--- a/src/combobox/combobox.js
+++ b/src/combobox/combobox.js
@@ -124,9 +124,12 @@ function Combobox<OptionT>(props: PropsT<OptionT>) {
       });
     }
     if (event.keyCode === ENTER) {
-      setIsOpen(false);
-      setSelectionIndex(-1);
-      onChange(tempValue);
+      let clickedOption = options[selectionIndex];
+      if (clickedOption) {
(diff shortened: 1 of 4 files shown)


src/combobox/combobox.js
```diff
@@ -170,7 +173,7 @@ function Combobox<OptionT>(props: PropsT<OptionT>) {
       const stringified = mapOptionToString(clickedOption);
       setIsOpen(false);
       setSelectionIndex(index);
-      onChange(stringified);
+      onChange(stringified, clickedOption);
```
nadiia, line 176: Would you mind adding a comment for v10 to change the `onChange` handler to be aligned with the rest of the components and use a type like
```
onChange: (params: {
  value: ValueT,
  type: ChangeSourceT,
}) => mixed,
```

chasestarr: 👍 added a comment in the `types.js` file

PR #3391, opened 2020-06-02
feat(combobox): add flag to disable autocomplete

chasestarr:
#### Description

If `autocomplete = false`, keyboard navigation in the listbox will not temporarily change the input value.

#### Scope
Minor: New Feature


chasestarr:
Not sure why there's no CI report on this PR

PR #3397, opened 2020-06-03
feat(link): add underline animation option

andersonoliveira:
#### Description
Add underline animation option for the link component
@chrisrauh 

#### Scope
Minor: New Feature


chasestarr:
Thanks @andersonba should this be the default behavior? If not, what situations call for this treatment? Do you mind adding some guidelines to follow?

andersonoliveira:
@chasestarr the default behavior is a standard link with just the underline and not animated. 
A good example of using movement is to ensure that a state change is noticeable other than just a color change as a standard link.

gergelyke:
as Chase pointed out too, I think we should add some description on when to use the default and when to use the animated version before landing this

andersonoliveira:
Yes. I'm adding some description and will push an update. Thanks guys.

documentation-site/pages/components/link.mdx
```diff
@@ -5,6 +5,15 @@ This source code is licensed under the MIT license found in the
 LICENSE file in the root directory of this source tree.
 -->
 
+# When to use the default version:
```
chasestarr, line 8: Could you please position this description above the `API` heading below on line 30? 

If you can, please write the description as if the reader is not aware of the option. For example, this could be described as: "Your application may need to draw more attention to some links on the page. Apply the `animateUnderline` prop to the component if the link is a call-to-action or initiates a UX flow." 

Also, please consider that many visitors to this website will not be uber employees. Expressing brand personality is not something that audience is interested in and is often a detractor towards using baseui in their personal projects. It's best to leave those points out of public docs

chasestarr:
According to [these visual regression tests](https://github.com/andersonoliveira/baseweb/pull/2) (click doc icon next to 'viewed' to display the before/after images), it looks like the links are now underlined by default. Can you please investigate and ensure there are no unintended changes to defaults?

andersonoliveira:
> nderlined by d

They were always underlined. Only difference now is the animated option.

chasestarr:
According to [these visual regression test results](https://github.com/andersonoliveira/baseweb/pull/2/files?short_path=e112092#diff-e112092b8a5e2d48bb387a3b4334482b), there are 31 changes from previous. Some scenarios have underlines where there were none previously, and others where the anchor wraps to two lines only have the underline applied to the last line. Run `yarn storybook` in the repository to view those examples locally

andersonoliveira:
> According to [these visual regression test results](https://github.com/andersonoliveira/baseweb/pull/2/files#diff-0467bef5bb2550a3072d787fb39f9db7), there are 31 changes from previous. Some scenarios have underlines where there were none previously, and others where the anchor wraps to two lines only have the underline applied to the last line. Run `yarn storybook` in the repository to view those examples locally

Thanks @chasestarr 
I made some changes. Hope this works.
One thing I notice is that in some snapshots, some links have a underline a little lower. We adjusted that so is not touching the copy.

gergelyke:
is it on purpose, that the distance between the underline and the text is different, based on the animation prop?

andersonoliveira:
> d the text is different, based on the animation prop?



> rline and the text is different, based on the animation prop?

Probably because the underline is a little over the background used for the animation.

jgillich:
Erm guys [this](https://github.com/uber/baseweb/pull/3397/files#diff-f273b0038d25492a8e7fb60c79726e8dR30) adds an underline to ALL links. If this is intentional, this is a pretty big change and should at least get a mention in the changelog.

chasestarr:
@jgillich can you please share an example of StyledLink code that was not underlined before? I can go back to a version before this change and still see underlines by default https://v9-80-0.baseweb.design/components/link

jgillich:
I'm sorry, you're totally right. I had `className={css({textDecoration: 'none'})}` on my link component, although that did stop working. Fix is to add an !important. I found that you can't use overrides on StyledLink, is that correct?

chasestarr:
> I'm sorry, you're totally right. I had `className={css({textDecoration: 'none'})}` on my link component, although that did stop working. Fix is to add an !important. I found that you can't use overrides on StyledLink, is that correct?

`StyledLink` is itself a styletron-component that can be customized with `withStyle`. See the docs here: https://www.styletron.org/react#reusing-styled-components

jgillich:
Oh, thanks. I think I got a little confused there here because [the docs mention an override property](https://baseweb.design/components/link/#api).

PR #3402, opened 2020-06-05
Add a controlRef prop to <Select/>

### LINKED ISSUE #3398: Expose ref to lowest level components
## Feature description
Would it be possible to cascade a ref of the lowest level component up the chain for form components (specifically select)?  I need to be able to focus the select component, but cannot as I can't access the ref for the underlying component.  Cascading the ref up using forwardRef should fix this

https://reactjs.org/docs/forwarding-refs.html

airhorns:
#### Description

Similar to `<Input/>`, it's useful to be able to get a ref to the inner element so developers can imperatively do stuff to it, like focus it. This adds a `controlRef` prop to `<Select/>` so you can do just that!

It's called `controlRef` and not `inputRef`, which is different than `<Input/>`, but on purpose, because if the Select isn't searchable, then there isn't actually an `<input>` created, just a `<div>` as a container. The Flow and TypeScript types reflect this, but I thought it might be good to make the name and docs clear that it won't always be an input.

Added a test and an example in the docs.

Fixes #3398

#### Scope

Minor: New Feature

### DIFF
--- a/documentation-site/examples/select/focus.js
+++ b/documentation-site/examples/select/focus.js
@@ -1,0 +1,45 @@
+// @flow
+import * as React from 'react';
+import {useStyletron} from 'baseui';
+import {Select} from 'baseui/select';
+import {Button} from 'baseui/button';
+
+export default () => {
+  const [css, theme] = useStyletron();
+  const [value, setValue] = React.useState([]);
+  const controlRef = React.useRef(null);
+  return (
+    <div className={css({display: 'flex'})}>
+      <Select
+        controlRef={controlRef}
+        creatable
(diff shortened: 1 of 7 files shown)


chasestarr:
Thanks for contribution @airhorns. Please review this test failure [here](https://buildkite.com/uberopensource/baseweb/builds/16458#1fa002ff-8800-45be-902a-b1ee07e5c0b7/181-1226)

airhorns:
Oops, the test was working when run by itself but not in the group, fixed a silly mistake for the scope of the `wrapper` variable. Should be good to go!

PR #3409, opened 2020-06-10
fix(countries): update Macedonia to North Macedonia

yrezgui:
#### Description

Macedonia officially changed its name to North Macedonia on the 12th of February 2019. This should be reflected on the `PhoneInput` list.

Sources:
* https://en.wikipedia.org/wiki/North_Macedonia
* https://www.theguardian.com/world/2019/feb/12/nato-flag-raised-ahead-of-north-macedonias-prospective-accession

#### Scope
Patch: Bug Fix


chasestarr:
Thank you @yrezgui

PR #3415, opened 2020-06-11
fix(datepicker): split range datepickers update from text input

chasestarr:
#### Description

This fixes a regression added to datepicker that avoids calling `onChange` if the text string is not fully populated. Split range datepickers operated by hacking the display value, but this pr introduces a top-level prop called `displayValueAtRangeIndex` to only operate on start/end date in the range.

#### Scope
Patch: Bug Fix


src/datepicker/__tests__/datepickers-composed-range.e2e.js
```diff
@@ -0,0 +1,123 @@
...
+const END_DATE_INPUT = '[id="end-date"] input';
+
+const DISPLAY_START_DATE = '[id="display-start-date"]';
+const DISPLAY_END_DATE = '[id="display-end-date"]';
+
+const SET_UNDEFINED_BTN = '[id="set-undefined"]';
+
+describe('datepicker-composed-range', () => {
+  jest.setTimeout(20 * 1000);
+
```
chasestarr, line 23:
```suggestion

```

src/datepicker/datepicker.js
```diff
@@ -235,11 +265,49 @@ export default class Datepicker extends React.Component<
...
+            } else {
+              if (isAfter(right, left) || isEqual(left, right)) {
+                onChange({date: [left, right]});
+              } else {
+                // See comment above about resetting dates on invalid range
+                onChange({date: [...value]});
+              }
+            }
+          }
+          // onChange({date: [left, right]});
```
chasestarr, line 306:
```suggestion
```

chasestarr:
@sandgraham are you familiar with the figma api errors in the zeit build?

PR #3422, opened 2020-06-15
fix(tree-view): don't refer to non-existent typescript types

airhorns:
#### Description

The TreeView TypeScript types were accidentally referring to a Flow type variable `OverrideT` which should just be `Override` like the rest of the TypeScript types, as well as using an older and unnecessary style of expressing a `children` prop. The TypeScript `React.ReactNode` type includes arrays of `React.ReactNode` as well, so the extra stuff isn't necessary.

#### Scope

Patch: Bug Fix


gergelyke:
thank you @airhorns

PR #3423, opened 2020-06-15
fix(toast): add tabIndex back to Toast body so it can be focused

jcready:
This was previously fixed in https://github.com/uber/baseweb/commit/34748e94615a962de7a6e83083b3a36d61ae532c
but then subsequently broken in https://github.com/uber/baseweb/commit/1aab1e6372be78f0a5c4b733ba9bde277ba87dfa

#### Description

Adds a `tabIndex` attribute back to the `<Toast/>` body so it can be focused and allow the the `onFocus`/`onBlur` event handlers to work correctly.

#### Scope
Patch: Bug Fix

### DIFF
--- a/src/toast/toast.js
+++ b/src/toast/toast.js
@@ -225,6 +225,7 @@ class Toast extends React.Component<ToastPropsT, ToastPrivateStateT> {
           <Body
             role="alert"
             data-baseweb={this.props['data-baseweb'] || 'toast'}
+            tabIndex={-1}
             {...sharedProps}
             {...bodyProps}
             // the properties below have to go after overrides


chasestarr:
@tajo was this change intentional?

tajo:
I think the idea was that the body should not be focusable since it's not interactive, only the close button is. Is there a reason to make the body focusable?

jcready:
@tajo there are `onFocus` and `onBlur` props for the Toast component. We used this to update the toast with additional information on focus.

nadiia:
As @tajo pointed out the `body` element is not interactive and therefore should not be focusable. We use `onFocus` and `onBlur` events to handle auto-dismiss when any of the inner elements get focus. Please note that in React even without `tabIndex` the `onFocus` handler is still called on the element when any of its inner elements is focused.
If your use-case requires the `Toast`'s container to be focusable I'd recommend adding `tabIndex` using `overrides` API and adding a target check in the handler.
Thanks for bringing up this question and I'll be closing it as "won't fix".

PR #3425, opened 2020-06-17
fix(button): prevent button height change when isLoading

jcready:
#### Description
The height of a button was changing when setting `isLoading` in some sites. The cause was tracked down to sites applying global styles like `* { box-sizing: border-box; }`. To ensure that all sites have a consistent button height regardless of these global styles we are setting the LoadingSpinner's `<span>` to have its `box-sizing` set to `content-box`.

#### Scope
Patch: Bug Fix

### DIFF
--- a/src/button/styled-components.js
+++ b/src/button/styled-components.js
@@ -141,6 +141,7 @@ export const LoadingSpinner = styled<SharedStylePropsT>(
       borderLeftColor: background,
       borderBottomColor: background,
       borderRightColor: background,
+      boxSizing: 'content-box',
       display: 'inline-block',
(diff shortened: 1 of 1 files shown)


chasestarr:
Thank you @jcready

src/button/styled-components.js
```diff
@@ -141,6 +141,7 @@ export const LoadingSpinner = styled<SharedStylePropsT>(
       borderLeftColor: background,
       borderBottomColor: background,
       borderRightColor: background,
+      boxSizing: 'content-box',
```
nadiia, line 144: Normally we'd use `border-box`. Is there a reason for `content-box` instead?

PR #3427, opened 2020-06-18
fix(input): add step property to ts defintions

hsnyildiz:
I added step support for Input

Fixes #1

#### Description

I try to use the Input component with type number and step property but I got the typescript error for props. I added optional props.

#### Scope
Patch: Bug Fix

### DIFF
--- a/src/input/index.d.ts
+++ b/src/input/index.d.ts
@@ -62,6 +62,7 @@ export interface BaseInputProps<T> {
   placeholder?: string;
   required?: boolean;
   size?: SIZE[keyof SIZE];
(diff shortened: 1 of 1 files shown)


gergelyke:
thank you @hsnyildiz! I wonder, is it something we should add to the Flow types too?

nadiia:
How do you pass the `step` property? I think unless you pass it through the `Input` overrides it won't be applied to the underlying element. Currently, we don't support the `step` prop in the `Input`s top level API.

nadiia:
Closing this one based on the last comment above ^

PR #3446, opened 2020-06-30
feat(select): add alias for tag override

gergelyke:
Historically, the name `MultiValue` seemed to confuse folks - `Tag` would be probably a better name, as it directly shows what underlying component is being overriden

### DIFF
--- a/src/select/select-component.js
+++ b/src/select/select-component.js
@@ -656,7 +656,7 @@ class Select extends React.Component<PropsT, SelectStateT> {
             key={`value-${i}-${value[this.props.valueKey]}`}
             removeValue={() => this.removeValue(value)}
             disabled={disabled}
-            overrides={{MultiValue: overrides.MultiValue}}
+            overrides={{Tag: overrides.Tag, MultiValue: overrides.MultiValue}}
             {...sharedProps}
             $disabled={disabled}
           >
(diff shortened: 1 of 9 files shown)


src/select/select-component.js
```diff
@@ -656,7 +656,7 @@ class Select extends React.Component<PropsT, SelectStateT> {
             key={`value-${i}-${value[this.props.valueKey]}`}
             removeValue={() => this.removeValue(value)}
             disabled={disabled}
-            overrides={{MultiValue: overrides.MultiValue}}
+            overrides={{Tag: overrides.Tag}}
```
chasestarr, line 659: Is this a regression? Will it need to union the two values like `overrides.MultiValue || overrides.Tag`?

gergelyke: oh yeah, good catch, thanks Chase!

nadiia:
I think naming it `Tag` couples it with the Tag component and makes it less flexible for future changes if say multi-value presented in a different way.

gergelyke:
Yeah, that's a good point - but at the same time, if we'd change the underlying component, that would be a breaking change anyway, right? (thinking about overrides)

nadiia:
Yeah, good point on the complexity of replacing it and overrides API contract. I personally like the semantic name better but I guess it's ok to have an alias.

PR #3460, opened 2020-07-06
feat(button): add Root alias for BaseButton

gergelyke:
All other components have a `Root` overrides - this PR ensures consistency across components

### DIFF
--- a/src/button/button.js
+++ b/src/button/button.js
@@ -64,8 +64,8 @@ class Button extends React.Component<
       ...restProps
     } = this.props;
     // Get overrides
-    const [BaseButton, baseButtonProps] = getOverrides(
-      overrides.BaseButton,
+    const [Root, rootProps] = getOverrides(
+      overrides.BaseButton || overrides.Root,
       StyledBaseButton,
     );
     const [LoadingSpinner, loadingSpinnerProps] = getOverrides(
@@ -84,7 +84,7 @@ class Button extends React.Component<
       $isFocusVisible: this.state.isFocusVisible,
     };
     return (
-      <BaseButton
+      <Root
         ref={forwardedRef}
         data-baseweb="button"
         {...(isLoading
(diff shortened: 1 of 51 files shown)


src/button/button.js
```diff
@@ -64,9 +64,9 @@ class Button extends React.Component<
       ...restProps
     } = this.props;
     // Get overrides
-    const [BaseButton, baseButtonProps] = getOverrides(
-      overrides.BaseButton,
-      StyledBaseButton,
+    const [Root, RootProps] = getOverrides(
+      overrides.Root || overrides.Root,
```
tajo, line 68:
```suggestion
      overrides.Root || overrides.BaseButton,
```

gergelyke: yeah, I was too eager replacing stuff :D  fixing all these now

gergelyke: I am having seconds thoughts on if we should change the name in JSX from BaseButton to Root - probably folks would have to update jest snapshots tests

maybe not a big deal, but by not changing it, it would be less of a headache - what do you all think?

nadiia:
I'd keep it as `BaseButton`. It also styles an actual button element (can be replaced tho).

gergelyke:
would it make sense to add an alias without updating the rest of the codebase?

houmark:
Just adding my opinion here. Changing this would be breaking I believe, right?

Maybe adding an alias now, with a deprecation warning to switch to `Root` before the next major version would be the best path?

gergelyke:
> Just adding my opinion here. Changing this would be breaking I believe, right?
> 
> Maybe adding an alias now, with a deprecation warning to switch to `Root` before the next major version would be the best path?

This PR only adds an alias :)

gergelyke:
going to close this in favor of just adding an alias

gergelyke:
https://github.com/uber/baseweb/pull/3462

PR #3465, opened 2020-07-07
fix(build): update react-extract-types packages

chasestarr:


chasestarr:
bumping build deps to try addressing vercel failure in https://github.com/uber/baseweb/pull/3239

PR #3472, opened 2020-07-10
feat: add skeleton module

jiaoy14:
#### Description
Add a simple Skeleton component, not complete yet. Welcome your suggestions!


src/skeleton/__test__/skeleton.scenario.js
```diff
@@ -0,0 +1,43 @@
+
```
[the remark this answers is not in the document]
chasestarr: Also, I recommend installing the [prettier vs-code extension](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode) and enabling `formatOnSave`

src/skeleton/styled-components.js
```diff
@@ -0,0 +1,20 @@
+
+import * as React from 'react';
+import {styled } from '../styles/index.js';
+
+
+
+export const StyledRoot = styled('div', {
+    display: 'flex',
+    flexDirection: 'row',
```
chasestarr, line 9: why does this need to be a flex layout?

src/skeleton/type.js
```diff
@@ -0,0 +1,17 @@
...
+
+export type SkeletonPropsT = {
+
+  overrides?: OverrideT,
+
+  children?: React.Node,
+
+  rows: number,
+
+  name?: string
```
chasestarr, line 15: For overrides, see this example as a reference: https://github.com/uber/baseweb/blob/master/src/button/types.js#L13-L20

It should be something like:
```js
overrides?: {|
  Root?: OverrideT,
  Row?: OverrideT,
|}
```

What purpose do the `children` and `name` props serve?

src/skeleton/skeleton.js
```diff
@@ -0,0 +1,33 @@
...
+
+class Skeleton extends React.Component<SkeletonPropsT> {
+  render() {
+    const { overrides = {} } = this.props;
+    const [Root, rootProps] = getOverrides(overrides.Root, StyledRoot);
+    const [Row, rowProps] = getOverrides(overrides.Row, StyledRow);
+
+    if (typeof this.props.rows === 'number') {
+      return (
+        <Root {...rootProps}>
```
chasestarr, line 19: It may get complex to render root differently when rows are present or not. Maybe this should be a new override called RowsContainer?

src/skeleton/__test__/skeleton.scenario.js
```diff
@@ -0,0 +1,43 @@
...
+import * as React from 'react';
+import {Skeleton} from "../index.js";
+
+
+export default function Scenario() {
+  return (
+    <div>
+      <Skeleton
```
chasestarr, line 9: Could you post a screenshot of what this looks like?

src/skeleton/styled-components.js
```diff
@@ -0,0 +1,20 @@
...
+});
+
+export const StyledRow = styled('div', {
+  backgroundColor: '#eee',
+  height:'100px',
+  width: '20px',
+  marginRight: '0px',
+  marginLeft: '0px',
+  marginTop: '0px',
+  marginBottom: '0px',
```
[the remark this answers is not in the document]
jiaoy14: I will modify it.

src/skeleton/__test__/skeleton.scenario.js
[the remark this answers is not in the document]
jiaoy14: Yes, thanks. I install prettier and it helps a lot.

src/skeleton/__test__/skeleton.scenario.js
jiaoy14: I have replaced the old example with a new one(youtube style). And now it will look like this:
Screen Shot 2020-07-11 at 10 36 06 AM

src/skeleton/type.js
jiaoy14: Modified.

jiaoy14: I also delete these two prop. It seems no need for now.

jiaoy14: Modify the override.

jiaoy14:
Is anybody familiar with  how to add animation in the CSS? I add codes like this:

```ruby
export const StyledRow = styled('div', {
animation: 'mymove 1.5s ease-in-out 0.5s infinite',
  '@keyframes mymove': {
    '0%': {
      opacity: 1,
    },
    '50%': {
      opacity: 0.4,
    },
    '100%': {
      opacity: 1,
    },
  },
});
```
But it doesn't work.

src/skeleton/styled-components.js
jiaoy14: I changed the default flexDirection to 'column'.  I thought  'flex' can help better arrange rows in a container. What's your opinion? Do I need it?

src/skeleton/skeleton.js
```diff
@@ -8,31 +8,29 @@ LICENSE file in the root directory of this source tree.
...
     const [Root, rootProps] = getOverrides(overrides.Root, StyledRoot);
-    const [Row, rowProps] = this.props.animation
-      ? getOverrides(overrides.Row, StyledAnimationRow)
-      : getOverrides(overrides.Row, StyledRow);
+    const [Row, rowProps] = getOverrides(overrides.Row, StyledRow);
 
     if (typeof this.props.rows === 'number') {
       return (
-        <Root testid={'loader'} {...rootProps}>
+        <Root $rows={this.props.rows} testid={'loader'} {...rootProps}>
```
chasestarr, line 25: Nice job applying arguments to the styletron components

src/skeleton/skeleton.js
```diff
@@ -8,31 +8,29 @@ LICENSE file in the root directory of this source tree.
...
-  StyledRoot,
-  StyledRow,
-  StyledAnimationRow,
-} from './styled-components.js';
+import {StyledRoot, StyledRow} from './styled-components.js';
 
 class Skeleton extends React.Component<SkeletonPropsT> {
   static defaultProps: SkeletonPropsT = {
-    rows: 1,
+    rows: 0,
```
chasestarr, line 15: Can we please remove the default value here?

src/skeleton/__tests__/skeleton-animation.scenario.js
```diff
@@ -19,6 +19,7 @@ export default function Scenario() {
           <div key={index}>
             <Skeleton
               animation={true}
+              rows={1}
```
chasestarr, line 22: We spoke about how these scenarios should not have row props provided, and the override should be renamed to `Root`. Is that possible?

chasestarr:
LGTM! Let's make a follow up PR adding this component to the documentation site

src/skeleton/types.js
```diff
@@ -0,0 +1,26 @@
...
+};
+
+export type SkeletonPropsT = {
+  overrides?: OverridesT,
+  /** Defines the number of row element in a skeleton */
+  rows: number,
+  /** Deinfes if the skeleton has an animation default is false*/
+  animation: boolean,
+  /** Deinfes the height of the skeleton container*/
+  height?: string,
```
conradreuter, line 23: Are `width` and `height` necessary? To me it seems like this would be a better fit for `className`.

```
const skeletonStyles = {width: '100px'}

const FooSkeleton = styled(Skeleton, skeletonStyles)
 // or alternatively
<Skeleton className={css(skeletonStyles)} />
```

src/skeleton/types.js
```diff
@@ -0,0 +1,26 @@
...
+
+export type OverridesT = {
+  Row?: OverrideT,
+  Root?: OverrideT,
+};
+
+export type SkeletonPropsT = {
+  overrides?: OverridesT,
+  /** Defines the number of row element in a skeleton */
+  rows: number,
```
conradreuter, line 19: nit: `rowCount`

src/skeleton/types.js
```diff
@@ -0,0 +1,26 @@
...
+  /** Deinfes if the skeleton has an animation default is false*/
+  animation: boolean,
```
conradreuter, line 21: nit: `isAnimated`, `hasAnimation`

src/skeleton/skeleton.js
```diff
@@ -0,0 +1,58 @@
...
+    const [Row, rowProps] = getOverrides(overrides.Row, StyledRow);
+
+    if (typeof this.props.rows === 'number') {
+      return (
+        <Root
+          $height={this.props.height}
+          $width={this.props.width}
+          $animation={this.props.animation}
+          $rows={this.props.rows}
+          testid={'loader'}
```
conradreuter, line 30: Does this set a `data-testid` attribute? Otherwise I would imagine React will just complain about `div` not accepting this prop.

conradreuter:
Judging from the visual snapshots, looks pretty cool!

One idea: I would let the `Skeleton` just be the most basic thing, a simple box. More complex shapes (multiple rows, circles) can then be built on top of it. Basically, I suggest dropping the `rows` part from `Skeleton` and move that to a different component. I don't have a strong opinion on this though, it's very nice as it is. 🙂

src/skeleton/types.js
jiaoy14: agree with you.

src/skeleton/types.js
jiaoy14: agree with you.

src/skeleton/types.js
jiaoy14: Because the height and width are the most important attribute for skeleton (the one who uses it will need to set it). I prefer to just leave them there and developers could use it conveniently.

PR #3481, opened 2020-07-14
Add tabs-motion module

sandgraham:
This PR introduces a new `Tabs` component via the `tabs-motion` module.

---

I'd suggest starting a review at `src/tabs-motion/tabs.js`, where most of the relevant logic resides. The rest of the PR is mainly new scenarios, tests and docs.

---

There are two main differences from the old `Tabs` component:
1. This one has an animated active tab highlight.
2. This one follows the latest [WAI-ARIA Tabs spec](https://www.w3.org/TR/wai-aria-practices-1.1/#tabpanel).

It would have been a breaking change to introduce these features to the old component so now we have `tabs-motion`. I kept the API nearly identical, but a few things to watch for if switching:

- The override structure and underlying markup is quite a bit different. Manual migration is likely needed for styling overrides.
- The `stateReducer` prop for `StatefulTabs` has a slightly different signature: `(state, action) => state`. This is so that it matches the `React.useReducer` hook.

---

I tried a version of this where all of the logic was in context and hooks. It was nifty since you could compose the various pieces of the UI however you wanted and the aria-roles/focus/keyboard stuff was taken care of. I ended up ditching this approach because it seemed like too great a departure from our current API. It seems like people are fairly happy with our current `Tabs` so there wasn't a lot of reason to mess with it.


src/tabs-motion/tabs.js
```diff
@@ -0,0 +1,353 @@
...
+  const sharedStylingProps = {
+    $orientation: orientation,
+    $fill: fill,
+  };
+
+  // Helper for parsing directional keys
+  const [, theme] = useStyletron();
+  const parseKeyDown = React.useCallback(
+    event => {
+      if (orientation === ORIENTATION.horizontal) {
```
sandgraham, line 133: use utils

src/tabs-motion/tabs.js
```diff
@@ -0,0 +1,353 @@
...
+  const activeTabRef = React.useRef();
+  const [highlightLayout, setHighlightLayout] = React.useState({
+    length: 0,
+    distance: 0,
+  });
+  React.useEffect(() => {
+    if (activeTabRef.current) {
+      // Note, we are using clientHeight/Width here, which excludes borders.
+      // getBoundingClientRect includes borders but returns a fractional value.
+      // This leads to the highlight being slightly misaligned every so often.
```
sandgraham, line 84: clarify or remove

src/tabs-motion/tabs.js
```diff
@@ -0,0 +1,353 @@
...
+        setHighlightLayout({
+          length: activeTabRef.current.clientWidth,
+          distance: activeTabRef.current.offsetLeft,
+        });
+      }
+    }
+  }, [activeKey]);
+
+  // Scroll active Tab into view.
+  // We have separate scroll styles for mount and key change.
```
sandgraham, line 100: clarify

src/tabs-motion/tabs.js
```diff
@@ -0,0 +1,353 @@
...
+  };
+
+  // Helper for parsing directional keys
+  const [, theme] = useStyletron();
+  const parseKeyDown = React.useCallback(
+    event => {
+      if (orientation === ORIENTATION.horizontal) {
+        if (theme.direction === 'rtl') {
+          switch (event.keyCode) {
+            case 39:
```
sandgraham, line 136: alias key codes for legibility?

src/tabs-motion/tabs.js
```diff
@@ -0,0 +1,353 @@
...
+      <TabList
+        data-baseweb="tab-list"
+        role="tablist"
+        aria-orientation={orientation}
+        {...sharedStylingProps}
+        {...TabListProps}
+      >
+        {React.Children.map(children, (child, index) => {
+          if (!child) return;
+
```
sandgraham, line 178: nit remove newline

src/tabs-motion/tabs.js
```diff
@@ -0,0 +1,353 @@
...
+          // Keyboard focus management
+          const handleKeyDown = React.useCallback(event => {
+            // WAI-ARIA 1.1
+            // https://www.w3.org/TR/wai-aria-practices-1.1/#tabpanel
+            // We use directional keys to iterate focus through Tabs.
+
+            // Find all tabs eligible for focus
+            const availableTabs = [
+              ...event.target.parentNode.childNodes,
+            ].filter(node => !node.disabled && node.dataset.baseweb === 'tab');
```
sandgraham, line 231: use `role` not `data-baseweb`

src/tabs-motion/tabs.js
```diff
@@ -0,0 +1,353 @@
...
+            key={key}
+            role="tabpanel"
+            id={getTabPanelId(uid, key)}
+            aria-labelledby={getTabId(uid, key)}
+            aria-expanded={isActive}
+            hidden={!isActive}
+            {...sharedStylingProps}
+            {...TabPanelProps}
+          >
+            {!isActive && !renderAll ? null : children}
```
sandgraham, line 343: maybe clearer: `isActive || renderAll ? children : null`

gergelyke:
looks great! a few things I've noticed:

- should we use `cursor: pointer` on click targets?
- should we also update the original tabs doc page, possibly marking that as deprecated, and tell folks to use the new version?

gergelyke:
One more thing 😬 

should we also add some animation when a tab is hovered? similarly what github has for the tabs above (isses, pull requests, etc..)

src/tabs-motion/types.js
```diff
@@ -0,0 +1,84 @@
...
+
+export type onChangeT = (params: {activeKey: React.Key}) => void;
+
+export type TabsPropsT = {|
+  children: React.Node,
+  activeKey?: React.Key,
+  disabled?: boolean,
+  fill?: FillT,
+  orientation?: OrientationT,
+  keyboardActivation?: KeyboardActivationT,
```
chasestarr, line 61: Why an enum vs a boolean? Are there upcoming specs suggesting another activation option?

src/tabs-motion/types.js
```diff
@@ -0,0 +1,84 @@
...
+export type StatefulTabsPropsT = {|
+  ...TabsPropsT,
+  initialState?: StatefulTabsStateT,
+  stateReducer?: StatefulTabsReducerT,
+|};
+
+export type TabPropsT = {
+  title?: React.Node,
+  key?: React.Key,
+  tabRef?: React.Ref<'button'>,
```
chasestarr, line 76: interesting, didn't know that this pattern was available 😮

src/tabs-motion/types.js
```diff
@@ -0,0 +1,84 @@
...
+|};
+
+export type TabOverridesT = {|
+  Tab?: OverrideT,
+  ArtworkContainer?: OverrideT,
+  TabPanel?: OverrideT,
+|};
+
+export type StatefulTabsStateT = {|
+  activeKey: React.Key,
```
chasestarr, line 40: Interesting use of `React.Key`. Is it just convenient to use this versus `string | number`? Not sure if it's worth worrying about them changing the type offsetting the cost of a bit of code-reuse

tajo:
Functional nitpicks:
- Should the cursor change to `pointer` when hovering over the tab? There is now no indication that the tab is clickable. We use pointer in the older version and different color.
- Keyboard: There is no reason for the tab content to be focusable?
- When using vertical orientation, the middle divider changes position slightly (left <-> right) based on the active tab width.
- Should we take this as a chance to use something better for default keys other than "0", "1" ... ? It's not super intuitive, I think we got a lot of questions about that in past. Maybe just using numbers 0, 1 would be better?
- Is there any strategy for mobile / scrolling?

The new design looks great!

tajo:
`Tab` override doesn't seem to work (for changing background / color).

src/tabs-motion/types.js
tajo: I have to google "flow react ref" every single time and always struggle to type them lol.

chasestarr: I have this page bookmarked :) https://www.saltycrane.com/cheat-sheets/flow-type/latest/

sandgraham:
Okay- thanks for all the feedback folks. I think there are some good points so I'll just go through 'em all here.

- I just added a hover BG to the Tabs. I would resist using `cursor: pointer` as that is [apparently reserved for links](https://www.w3.org/TR/CSS21/ui.html#propdef-cursor). I think the hover gives enough of an indication that the Tab is interactive now.
- Right now I'm using `React.Key` for the `activeKey` value. I thought it made sense to just set it to whatever React uses for keys since users are either setting `key` on each Tab or we are assigning it. This means you can use a `number | string` for the `activeKey`. Internally the Tabs component doesn't care which you use- it uses `==` for the comparison. I suppose React might change the key type, but if its a breaking change we are likely to need a major as well- otherwise we should match whatever people can pass as a `key`.
- I made the active TabPanel focusable so that keyboard folks could get to the content without "overshooting" it. The [reach-ui](https://reach.tech/tabs/) library does this as does the [Mozilla page on tab accessibility](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Tab_Role). The WAI-ARIA spec itself doesn't offer much guidance on this. I agree its a bit noisy though. I just changed it so that it also uses `focusVisible` to avoid highlighting on click. It may be a bit better now since it will only be apparent on keyboard navigation.
- I'm still waiting on design for mobile/overflow helpers. I'm not super hopeful that we'll get these though so I'm inclined to land without that feature built in. On the flip side, we could propose a web specific approach and get it approved. Some ideas I've heard are a floating arrow button or a box-shadow overlay.
- I forgot, but I'll update the old Tabs page with a deprecation warning.
- I'll update all the nits Chase and I left.

@tajo I looked into those two bugs:
- I think the vertical wobble happens in the yard compiler because it is flex centered vertically/horizontally. I _think_ in normal cases this shouldn't be an issue- or if it is you could modify the container. Is there a way to modify the css on the container in `react-view`?
- The `Tab` override not working is cause the `Tab` override needs to be applied to each `Tab`, not the top-level `Tabs`. I put that in there on accident. Separately I need to see if I can add a way to override each  `Tab` through the yard override UI. Not sure if we do something similar elsewhere.

tajo:
> I think the vertical wobble happens in the yard compiler because it is flex centered vertically/horizontally. I think in normal cases this shouldn't be an issue- or if it is you could modify the container. Is there a way to modify the css on the container in react-view?

You can try to update the css [here](https://github.com/uber/baseweb/blob/master/documentation-site/components/yard/index.tsx#L95), `Compiler` accepts `className` prop

src/tabs-motion/types.js
sandgraham: yeah i also google it every time. somehow stumbled on this syntax somewhere in a stack overflow comment and it worked!

src/tabs-motion/tabs.js
sandgraham: sounds good- ill add a todo and make a ticket

src/tabs-motion/types.js
sandgraham: If React ever updated `key` I'd actually *prefer* to update the Tabs component to still use that. I really like that the currently selected Tab is based on the React key. It feels like a nice way to leverage a pattern folks should already be familiar with, if that makes any sense. I imagine if they ever made `key` more complicated they'd also have a way of validating equality.

src/tabs-motion/types.js
sandgraham: good point - changed this to a bool `activateOnFocus`.

PR #3484, opened 2020-07-14
fix(accordian): remove :never from type file

### LINKED ISSUE #3477: Type wrong for ToggleIcon's component in Accordian component's override when used with Typescript
## Current Behavior
This problem only occurs when used with Typescript. Within the Accordian component's override prop, I overrode the ToggleIcon I want to be displayed to be an "up" or "down" chevron (depending on whether the Accordian is expanded or collapsed) instead of the default "plus" sign. 
The problem I believe is due to the type of the "component" prop of ToggleIcon, which does not expects the argument `$expanded`. (The type of `$expanded` is `never` as can been seen below when I hover over the error) When I use `$disabled` as the argument instead, the error goes away. Hence, I believe there is just a typing error with the component prop.

## Expected Behavior
There should not be any typing error if I pass in `$expanded` as the argument.

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v9.85.2  |
| React   |    16.13.1     |
| browser |    NA     |

- [ ] I have searched the [issues](https://github.com/uber/baseweb/issues) of this repository and believe that this is not a duplicate.
TS error
TS error verbose
TS no error if argument changed

peironggg:
Fixes #3477 

#### Description
Removed the `$expanded: never` from the index.d.ts type file of the Accordian component.

#### Scope
Patch: Bug Fix

### DIFF
--- a/src/accordion/index.d.ts
+++ b/src/accordion/index.d.ts
@@ -33,7 +33,7 @@ export interface AccordionProps {
   disabled?: boolean;
   initialState?: AccordionState;
   onChange?: (args: {expanded: React.Key[]}) => any;
-  overrides?: AccordionOverrides<SharedProps & {$expanded?: never}>;
+  overrides?: AccordionOverrides<SharedProps>;
   stateReducer?: StateReducer<AccordionState>;
   renderPanelContent?: boolean;
   renderAll?: boolean;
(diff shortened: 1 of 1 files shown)


gergelyke:
thank you!

PR #3486, opened 2020-07-14
feat(rating): make sizes of emoticons configurable

TanyaChauhan9:
`Fixes #1, Fixes #2`


src/rating/emoticon-rating.js
```diff
@@ -50,7 +50,7 @@ class EmoticonRating extends React.Component<
   renderRatingContents = () => {
     const {overrides = {}, value = -1} = this.props;
```
gergelyke, line 51:
```suggestion
    const {overrides = {}, value = -1, size} = this.props;
```

src/rating/emoticon-rating.js
```diff
@@ -50,7 +50,7 @@ class EmoticonRating extends React.Component<
...
     const {previewIndex} = this.state;
-
+    const size = this.props.size;
```
gergelyke, line 53:
```suggestion
```

src/rating/emoticon-rating.js
```diff
@@ -71,6 +71,7 @@ class EmoticonRating extends React.Component<
           tabIndex={isFocusable ? '0' : '-1'}
           aria-setsize={5}
           aria-checked={x === value}
+          size={size}
```
gergelyke, line 74: to follow the shared prop naming convention, this prop should start with the $ prefix, so it should be `$size={size}`

gergelyke:
Thank you for taking this one! A few additional things that I forgot to tell you, but we should include in this PR:

* you have to add the `$size` prop here: https://github.com/uber/baseweb/blob/master/src/rating/types.js#L44
* you have to add the top-level `size` prop in these two defintions: https://github.com/uber/baseweb/blob/master/src/rating/types.js#L17-L33
* as well as here https://github.com/uber/baseweb/blob/master/src/rating/index.d.ts#L14-L25
* to make sure the same feature is avilable for the star rating too, can you make the same changes here? https://github.com/uber/baseweb/blob/master/src/rating/star-rating.js
* lastly (I promise!), please add a `size` prop here too, so it shows up on the doc site https://github.com/uber/baseweb/blob/master/documentation-site/components/yard/config/rating.ts#L17-L21

Thanks a lot!

src/rating/types.js
```diff
@@ -30,6 +32,8 @@ export type EmoticonRatingPropsT = {
   value?: number,
   /** Callback that's called with the newly selected value. */
   onChange?: ({value: number}) => mixed,
+  /**Size of emoticons. */
+  size: number,
```
chasestarr, line 36:
```suggestion
  size?: number,
```

src/rating/types.js
```diff
@@ -22,6 +22,8 @@ export type StarRatingPropsT = {
   numItems: number,
   /** Callback that's called with the newly selected value. */
   onChange?: ({value: number}) => mixed,
+  /** Size of star icons. */
+  size: number,
```
chasestarr, line 26:
```suggestion
  size?: number,
```

documentation-site/components/yard/config/rating.ts
```diff
@@ -28,7 +28,11 @@ const ratingConfig: TConfig = {
         into: 'value',
       },
     },
-
+    size: {
+      value: 44,
```
gergelyke, line 32: I believe it should be 22

gergelyke:
Screen Shot 2020-07-15 at 12 56 08 PM

it seems like we also have to modify the padding between the items?

TanyaChauhan9:
I'm attaching the screenshots for different star sizes. The padding between the stars( and emoticons) adjusts according to the size inputted.

109007458_275594020442931_8665548441154984560_n
109225172_594462294588668_3939600718253537883_n
110887676_328946528262650_706961464928085257_n

src/rating/styled-components.js
```diff
@@ -57,15 +56,17 @@ export const StyledStar = styled<StyledRatingItemPropsT>(
       marginTop: 0,
       marginBottom: 0,
       marginRight: $theme.sizing.scale300,
-      width: '22px',
-      height: '20px',
+      width: $size,
```
nadiia, line 59: The above it says that the type for the `size` is specified as `number` therefore we'd need to add `px` here if type is `number`.

src/rating/types.js
```diff
@@ -22,6 +22,8 @@ export type StarRatingPropsT = {
   numItems: number,
   /** Callback that's called with the newly selected value. */
   onChange?: ({value: number}) => mixed,
+  /** Size of star icons. */
+  size?: string,
```
nadiia, line 26: TS types say it's a number, we should make it consistent in Flow and TS.

chasestarr:
I went ahead and copied this PR here: https://github.com/uber/baseweb/pull/3505 and landed the change

Issue #3492, opened 2020-07-15
[Combobox] Support an `onSearch` method for input execution


jonsadka:
## Use case

A user can type in the input bar and hit `Enter` or click the Start / End Enhancer to execute a search without choosing an option. This is a common design paradigm for search bars.

## How to achieve this with Combobox today

The current workaround for this is to pass an Input override of `onKeyUp` that listens to `Enter` events. However, even with this approach the Combobox cannot be closed because[ `setIsOpen`](https://github.com/uber/baseweb/blob/master/src/combobox/combobox.js#L50) is not exposed.

## Proposal

Expose a prop that allows a developer to pass an optional function prop into `Combobox` (something like `onSearch`). When an execution event is performed in the Input (via an `Enter` or click on a start / end enhancer), invoke the passed function and closes the component.

#### Pseudocode

```js
if (props.onInputSearch){
  props.onInputSearch(value);
  setIsOpen(false);
}
```

PR #3500, opened 2020-07-16
feat(combobox): adds onSubmit callback and close handler

chasestarr:
closes https://github.com/uber/baseweb/issues/3492

#### Description

Adds a new `onSubmit` callback that will fire in the same step as a parent form element's `onSubmit` attribute. This callback includes a `closeListbox` handler that can be used in response to pressing enter when not focused on a listbox option

@jonsadka Take a look at the included `.scenario.js` file. I think this should do what you're looking for. Clicking on a separate button should already close the listbox. If that's not working for you, can you please share a repro?

#### Scope
Minor: New Feature


jonsadka:
@chasestarr this is great! 

Any way we can pass the input value through to the `onSubmit` callback also?

PR #3507, opened 2020-07-17
feat: add mask for pin-code

jiaoy14:
`Fixes #1, Fixes #2`


src/pin-code/pin-code.js
```diff
@@ -33,7 +33,12 @@ export default class PinCode extends React.Component<PropsT, StateT> {
       if (inputRef && inputRef.focus) inputRef.focus();
     }
   }
-
+  getMaskStyle(i: number) {
+    if (this.props.values[i] == '') return this.props.values[i];
+    if (typeof this.props.mask === 'boolean')
+      return this.props.mask ? '*' : this.props.values[i];
```
chasestarr, line 39: Can we render a solid circle if boolean `true` value is provided, rather than the star symbol?

src/pin-code/__tests__/pin-code-mask.scenario.js
```diff
@@ -0,0 +1,22 @@
+/*
```
chasestarr, line 1: I recommend adding an 'interaction' to [this file](https://github.com/uber/baseweb/blob/master/vrt/config.js#L11) to prepare the component before taking a screenshot. The interaction should type into the pin-code and we should see the mask displayed in the screenshot

src/pin-code/pin-code.js
chasestarr: This may be done best by exploring the `type="password"` attribute

vrt/config.js
```diff
@@ -13,7 +13,7 @@ const config = {
     interactions: [
       {
         name: 'openedMenu',
-        behavior: async page => {
+        behavior: async (page) => {
```
sandgraham: Are all of these `page` -> `(page)` changes intentional?

src/pin-code/pin-code.js
```diff
@@ -34,6 +35,14 @@ export default class PinCode extends React.Component<PropsT, StateT> {
     }
   }
 
+  getMaskStyle(i: number) {
+    if (this.props.values[i] == '') return this.props.values[i];
```
sandgraham, line 39:
```suggestion
    if (this.props.values[i] === '') return this.props.values[i];
```

src/pin-code/pin-code.js
```diff
@@ -34,6 +35,14 @@ export default class PinCode extends React.Component<PropsT, StateT> {
...
+    if (typeof this.props.mask !== 'boolean') {
```
sandgraham, line 40:
```suggestion
    if (typeof this.props.mask === 'string') {
```

src/pin-code/pin-code.js
```diff
@@ -45,7 +54,10 @@ export default class PinCode extends React.Component<PropsT, StateT> {
     );
     const baseOverrides = {
       Root: {component: StyledInputOverrideRoot},
-      Input: {component: StyledInputOverrideInput},
+      Input: {
+        component: StyledInputOverrideInput,
+        props: {type: this.hasMask ? 'password' : 'text'},
```
sandgraham, line 59: Does this need to be an override? We can pass it directly to the `Input` below.

src/pin-code/pin-code.js
```diff
@@ -128,8 +140,7 @@ export default class PinCode extends React.Component<PropsT, StateT> {
               positive={this.props.positive}
               required={this.props.required}
               size={this.props.size}
-              type="text"
```
sandgraham, line 131: We can set the type here like so:
```
type={typeof this.props.mask === 'boolean' && this.props.mask ? 'password' : 'text'}
```

src/pin-code/pin-code.js
```diff
@@ -23,6 +23,7 @@ export default class PinCode extends React.Component<PropsT, StateT> {
 
   _inputRefs = new MultiRef<number, HTMLInputElement>();
 
+  hasMask = typeof this.props.mask === 'boolean' ? this.props.mask : false;
```
sandgraham, line 26: Not sure we need to store this if only used once - the name is a bit misleading. if `this.props.mask` is a string i would expect `hasMask` to be true

src/pin-code/types.js
```diff
@@ -49,6 +49,8 @@ export type PropsT = {
   manageFocus: boolean,
   /** An array of values respective to each pin code input. */
   values: string[],
+  /** Mask for pin. Default is no mask. Set it true then mask is ".". Also accept string input as custom mask style.  */
+  mask: boolean | string,
```
sandgraham, line 53: we also need to update the `index.d.ts` for typescript defs

src/pin-code/pin-code.js
sandgraham: Could also combine these conditionals a bit:
```js
if (this.props.values[i] !== '' && typeof this.props.mask === 'string') {
  return this.props.mask;
} else {
  return this.props.values[i];
}
```

vrt/config.js
jiaoy14: Sorry, the editor changes it when auto save and I don't notice it .

src/pin-code/pin-code.js
jiaoy14: If we pass it this way, the <MaskToggleShowIcon> will appear.

src/pin-code/pin-code.js
jiaoy14: Yeah, I will delete it.

src/pin-code/pin-code.js
sandgraham: Oh okay that makes sense.

src/pin-code/types.js
jiaoy14: OK

src/pin-code/pin-code.js
jiaoy14: Yeah, thanks, this style is much better.

documentation-site/pages/components/pin-code.mdx
```diff
@@ -89,6 +90,12 @@ You can control the size of the inputs with the `size` prop.
 If you donʼt like the default placeholder (`○`), you can change that too.
 Notice how each placeholder disappears when any of the inputs have focus.
 
+<Example title="Adding mask" path="pin-code/mask.js">
+  <PinCodeMask />
+</Example>
+
+You can add a default mask(`.`) by setting `mask=true`. If you don't like the default mask style, you can pass a customized string to `mask`.
```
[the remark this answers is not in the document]
jiaoy14: Thanks! Modify it!

jiaoy14:
Also please don't merge this PR , I am still working on the docs.

PR #3510, opened 2020-07-17
fix(tabs-motion): add cursor:pointer

gergelyke:


chasestarr:
Graham made a comment in the previous PR documenting the decision https://github.com/uber/baseweb/pull/3481#issuecomment-658990506 but it should be noted that default cursor is a deviation from the previous tabs component

gergelyke:
> Graham made a comment in the previous PR documenting the decision [#3481 (comment)](https://github.com/uber/baseweb/pull/3481#issuecomment-658990506) but it should be noted that default cursor is a deviation from the previous tabs component

oh, thanks Chase! I missed that comment

I still believe we should add it, both the [React.js reference implementation](https://github.com/reactjs/react-tabs), as well as other libraries follow this approach ( [React bootstrap](https://react-bootstrap.github.io/components/tabs/), [Material UI](https://material-ui.com/components/tabs/), [Semantic UI](https://react.semantic-ui.com/modules/tab/) or [Garden](https://garden.zendesk.com/react-components/tabs/)). I am a bit worried, that users expect to see the pointer, as that's what they are used to. @sandgraham what do you think?

gergelyke:
Another reason I would add it is to be consistent - we've been using the pointer cursor in other components already, that are not links (like the Rating or the Select component for example).

If we feel strongly about it, we may want to do a sweeping change to change it across the board?

houmark:
I personally like the pointer more than not, even if it's a tab, rating, or other similar use-cases. In my experience, inexperienced users are often confused by the normal cursor in these situations.

PR #3514, opened 2020-07-19
feat(styles): add expandBorderRadiusStyles

leyanlo:
Add expandBorderRadiusStyles and use both expandBorderRadiusStyles and expandRadiusStyles when possible.


chasestarr:
Thanks @leyanlo - was this causing bugs for you? We're working on moving the codebase to typescript with some progress already started. If it's just cleaning up the source code, we may choose to leave this PR open until the conversion has finished

leyanlo:
> Thanks @leyanlo - was this causing bugs for you? We're working on moving the codebase to typescript with some progress already started. If it's just cleaning up the source code, we may choose to leave this PR open until the conversion has finished

No bugs, just would like to add a similar utility for border-radius, and clean up usages while I was at it. No rush.

sandgraham:
When we do merge this in, maybe it can be a more generic `expand` function so it accepts common short hand properties?
```js
expand({
  padding: '12px',
  border: `solid 1px ${theme.colors.borderOpaque}`,
  borderRadius: '2px'
});
```
Although at a certain point maybe Styletron should just expand shorthand properties itself in the atomic engine.

leyanlo:
> When we do merge this in, maybe it can be a more generic `expand` function so it accepts common short hand properties?
> 
> ```js
> ...(expand({
>   padding: '12px',
>   border: `solid 1px ${theme.colors.borderOpaque}`,
>   borderRadius: '2px'
> }))
> ```
> 
> Although at a certain point maybe Styletron should just expand shorthand properties itself in the atomic engine.

Ideally we remove all this code and have Styletron do its job handling styles. I would really like to give it whatever rules I want, and have Styletron figure out what classes to add based on JS Object key order. E.g.
```js
css({padding: '12px', paddingBottom: 0})
// should return classname for
{padding: '12px 12px 0'}

// and
css({paddingBottom: 0, padding: '12px'})
// should return classname for
{padding: '12px'}
```
That’s my wishlist, though. Not sure it will ever get there :-/

chasestarr:
Sorry for keeping this around for so long. We're going to close. The team prefers to be explicit about what rules are applied to the classes

PR #3516, opened 2020-07-20
chore(react): consistently import React from 'react'

leyanlo:
Consistently import React from 'react'. The * as React was only useful if we wanted to get the flow type from React, but we can get types directly from Flow without importing React, e.g. using React$Node or React$ElementRef. This also lets us remove imports from a few types.js files.


chasestarr:
Thanks for helping to standardize the source code here, but it looks like the import * pattern will the primary recommendation https://github.com/facebook/react/pull/18102 I'm going to close for now, but let me know if there is a stronger reason for converting

leyanlo:
> Thanks for helping to standardize the source code here, but it looks like the import * pattern will the primary recommendation [facebook/react#18102](https://github.com/facebook/react/pull/18102) I'm going to close for now, but let me know if there is a stronger reason for converting

Oh wow didn’t realize they were moving away from default exports. Good to know!

PR #3518, opened 2020-07-20
docs(popover): proper type for mountNode

gergelyke:


chasestarr:
It's not quite a ref, it's just the HTMLElement :(

Issue #3519, opened 2020-07-20
[TabsMotion] If Tabs are wrapped in form, the form submits on tab change


houmark:
## Current Behavior

Please see this Stackblitz for example:
https://stackblitz.com/edit/react-szam3j?file=src/example.js

This is different behavior than the old tabs component.

## Expected Behavior

Only the tab changing, no submission of a wrapping form.

I think the solution is simple, and that is to add `type="button"` to the Tab button, if no type is set, it will default to submit.

I have not yet tested that solution, as I don't have BaseWeb set up in development mode (I really should get to that).

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v9.87.0  |
| React   |    16.8     |
| browser |    Chrome     |

- [x] I have searched the [issues](https://github.com/uber/baseweb/issues) of this repository and believe that this is not a duplicate.

houmark:
That's what I call fast fixing :) Early adopter here, loving this new component in terms of design and animation, and besides this one, it was very simple to swap the components out.

gergelyke:
sure, you are welcome 👍 

and yeah, you should set up a local dev env for Base Web :) should be pretty simple, but let us know if you run into any issues!

PR #3520, opened 2020-07-20
fix(tabs-motion): buttons should not trigger form submit

gergelyke:
closes https://github.com/uber/baseweb/issues/3519


houmark:
That's what I call fast fixing :) Early adopter here, loving this new component in terms of design and animation, and besides this one, it was very simple to swap the components out.

PR #3538, opened 2020-07-22
chore: uses new ts converted styletron + some initial fixes

tajo:
`Fixes #1, Fixes #2`


gergelyke:
@tajo should we close this one?

PR #3549, opened 2020-07-27
fix(form-control): don't pass non-DOM props to DOM children

airhorns:
#### Description

FormControls can be passed a single child of any type to contain the actual controls. The `error` and `positive` props are passed down to that single child, which is useful when it's an `<Input/>` or what have you. If it's a `<div/>` or something like that however, React logs a warning (I think only in Strict mode) about passing the non-DOM props `error` and `positive` to a div. You can ideally avoid this warning by not using FormControl's `error` or `positive` functionality at all when using non-baseweb children, but prior to this change the warning would still be logged even when not using those props.

This changes `<FormControl/>` to not pass an `error` or `positive` prop to it's single child unless those props are actually set to something. No more warnings!

#### Scope

Patch: Bug Fix

### DIFF
--- a/src/form-control/form-control.js
+++ b/src/form-control/form-control.js
@@ -40,8 +40,6 @@ export default class FormControl extends React.Component<
     overrides: {},
     label: null,
     caption: null,
-    error: false,
-    positive: false,
     disabled: false,
   };
   state = {captionId: getBuiId()};
--- a/src/form-control/types.js
+++ b/src/form-control/types.js
@@ -29,9 +29,9 @@ export type FormControlPropsT = {
   /** Displays label in light gray color if true */
   disabled?: boolean,
   /** Error state of the input. If an error prop passed it will be rendered in place of caption as an error message. */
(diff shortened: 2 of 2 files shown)


airhorns:
@sandgraham oh, I thought they were optional already! Are they intended to be optional? They are in the TypeScript types, but not in the Flow types as you mentioned.

And just to clarify, this change should not affect the types of the component really, it should still work the same as it did before. This change just affects what it passes to children -- before it would pass false down to the child by default, now it will be undefined.

sandgraham:
@airhorns [Flow has a quirk](https://flow.org/en/docs/react/components/#toc-using-default-props) where if you define default props it makes them optional even though the type reads as required - so when you remove the default props flow just thinks they are required. It's kind of confusing but Flow users will basically see no change if we update the types in this case.

airhorns:
Ah, learn somethin' every day. Updated to mark those props as optional in the flow types as well.

PR #3552, opened 2020-07-28
fix(icons): update titles to not contain the word chevron

gergelyke:


chrisrauh:
This works.

Just wanted to note that this is a much broader change than changing it on the carousel so I am not 100% sure since I don’t know if there are uses where folks might actually need "Chevron" in there. It can happen, but not likely.

Issue #3557, opened 2020-07-29
[Menu] Keyboard events must bubble for menu key navigation to work


airhorns:
Right now, `<Menu/>` attaches keyboard events to the document in order to implement moving the focused menu item around. This means that if any parent component wants to stop the propagation of the event, it can't, without breaking the Menu functionality. A parent component might want to stop the propagation of the event because there are other grandparents also listening for events.

Imagine a `<Menu/>`, or more realistically a `<Select/>`, inside a popover that opens up from another thing that is keyboard navigable, like a tree, or a canvas of draggable components, or whatever. The outer component system might listen for keyboard events too, and when you press a key that both systems are listening for, the user would expect that only one should respond. The normal way to accomplish this would be to have some container for the inner component listen for the event and then call `event.stopPropagation()` on it, so that it didn't bubble out and trigger the outer navigation component's behaviour as well.

But, `<Menu/>` needs its events to make it all the way to the [document](https://github.com/uber/baseweb/blob/master/src/menu/stateful-container.js#L83) for them to be acted on. This means `<Menu/>`s and `<Select/>`s nested inside anything that are also listening to keyboard events either don't work or trigger both behaviours.

## Current Behavior

Menu keyboard navigation doesn't work if keyboard event are `stopPropagation`'d anywhere along the chain to the document, which might be necessary to prevent double keyboard navigation behaviours in nested trees

## Expected Behavior

Menu keyboard navigation should work if an ancestor `stopPropagation`s. It should be able to do this by attaching keyboard event listeners to container nodes inside the `<Menu/>`, instead of on the document.

## Context

AFAICT there is a good reason `<Menu/>` works the way it does surrounding nested menus responding properly to keyboard events, but I am not deeply familiar with the issue at hand. Maybe @chasestarr could shed some light on why it was necessary to attach listeners to the document, and perhaps we could puzzle out a solution to that issue that keeps the event listeners local to the `<Menu/>` so that they don't require full propagation! 

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | 9.90.0 |
| React   |  16.13.1   |
| browser | Chrome latest  |

- [X] I have searched the [issues](https://github.com/uber/baseweb/issues) of this repository and believe that this is not a duplicate.

chasestarr:
[...]

PR #3558, opened 2020-07-29
fix(menu): listen to keydown on list element rather than doc

chasestarr:
#### Description

Begins to address https://github.com/uber/baseweb/issues/3557

#### Scope
Patch: Bug Fix


jiaoy14:
This is right the problem I met last week. When I add type-ahead for `Menu`, both `Menu` and `Select` focus on keyboard input and may cause some trouble. I also add something like `props.typeAhead`. Besides this problem, other risks may exist. `Select` dynamically changes the length of list which may cause `Menu` 's attr `highlightedIndex` out of range. As a result, I definitely support that we mute keyboard focus when in need.

PR #3559, opened 2020-07-29
feat(breadcrumbs): add prop to display trailing separator

shannonmoeller:
#### Description

Adds the option to show a trailing separator after the last breadcrumb.

#### Scope

Minor


src/breadcrumbs/__tests__/breadcrumbs.test.js
```diff
@@ -38,6 +38,19 @@ describe('Breadcrumbs', () => {
     ).toMatchSnapshot();
   });
 
+  it('displays trailing separator', () => {
```
gergelyke, line 41: can we add a visual regression test instead? something like this: https://github.com/uber/baseweb/blob/master/src/breadcrumbs/__tests__/breadcrumbs.scenario.js

gergelyke:
thank you Shannon! one more thing: would you mind also adding it to the react-view config? https://github.com/uber/baseweb/blob/master/documentation-site/components/yard/config/breadcrumbs.ts

src/breadcrumbs/__tests__/breadcrumbs.test.js
shannonmoeller: You bet

shannonmoeller:
On it.

PR #3560, opened 2020-07-30
feat(menu): add type-ahead function for menu

jiaoy14:
`Fixes #1, Fixes #2`


tajo:
How do you test it? What is the sequence of steps?

I tried it here https://baseweb-git-add-type-ahead-menu.uber-ui-platform.vercel.app/components/menu/.

jiaoy14:
> How do you test it? What is the sequence of steps?
> 
> I tried it here https://baseweb-git-add-type-ahead-menu.uber-ui-platform.vercel.app/components/menu/.

Yeah, I think it works on your link. I am also adding a e2e test. The process I expect is check `aria-select` , then keyboard input then check `aria-select` again.

src/menu/stateful-container.js
```diff
@@ -149,11 +154,70 @@ export default class MenuStatefulContainer extends React.Component<
         }
         this.handleEnterKey(event);
         break;
+      default:
+        if (this.props.typeAhead) this.findItemToFocus(event);
+        break;
     }
   };
+  findItemToFocus = (event: KeyboardEvent) => {
```
chasestarr, line 162: An idiomatic way to handle the timeout would be the example below

```js
class MyComponent {

  typeaheadCharacters = '';
  typeaheadTimeout = null;

  handleKeyDown = (event) => {
    // ignoring arrow keys
    clearTimeout(this.typeaheadTimeout);
    this.handleAlphaDown(event);
  }

  handleAlphaDown = (event) => {
    this.typeaheadCharacters += event.key;
    this.typeaheadTimeout = setTimeout(() => {
      this.typeaheadCharacters = '';
    }, 500);

    // search for the matching item here
  }
}
```

jiaoy14: Yeah. Thanks!

src/menu/stateful-container.js
```diff
@@ -149,11 +154,70 @@ export default class MenuStatefulContainer extends React.Component<
...
+    this.charClear = setTimeout(
+      function() {
+        this.charsSoFar = '';
+        this.charClear = null;
+      }.bind(this),
+      500,
+    );
+  };
+
+  findMatchInRange = (list: *, startIndex: number, endIndex: number) => {
```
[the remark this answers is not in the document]
jiaoy14: I originally did this in your way. But I switch to the two range search method after seeing https://w3c.github.io/aria-practices/examples/listbox/js/listbox.js.
I compare these two methods and found  searching in two separate range is better. My reasons are as follow:
1. people may not interested in content before current focus point. They may be more interested in the contents below current focus point because they haven't seen them. Searching in in two separate range can help search the contents after current point first.
2. If a person is looking for all content with the substring `a` . They can just use keyboard to walk over all the options and press `enter` to select one. If we choose the one you mention, we will all stay in the same place. We cannot allow users to press very long string to perfectly match because we must clean out the string  after some time.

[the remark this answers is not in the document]
chasestarr: I would prefer to mirror default select behavior as shown in this example https://codesandbox.io/s/confident-lumiere-2m64e?file=/index.html

Chrome, safari, and edge all have deterministic selection and firefox works closer to how this is currently implemented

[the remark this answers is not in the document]
jiaoy14: Yeah, it's reasonable. I will change it.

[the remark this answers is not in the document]
tajo: Yea, I would use the default behavior of Chrome at https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select. It's strange to type item and get the focus moving. Also, they have some sort of fulltext search. (If my first character doesn't match first letter in any item, it's doing fulltext.)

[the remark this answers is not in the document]
jiaoy14: yeah, that's a great idea. Will also add fulltext match. Thank you!

src/menu/stateful-container.js
```diff
@@ -149,11 +154,65 @@ export default class MenuStatefulContainer extends React.Component<
...
+      highlightedIndex: nextIndex,
+    });
+
+    if (this.refList[nextIndex]) {
+      scrollItemIntoView(
+        this.refList[nextIndex].current,
+        // $FlowFixMe
+        rootRef.current,
+        nextIndex === 0,
+        nextIndex === this.getItems().length - 1,
```
chasestarr, line 208:
```suggestion
        nextIndex === items().length - 1,
```

src/menu/stateful-container.js
```diff
@@ -149,11 +154,65 @@ export default class MenuStatefulContainer extends React.Component<
...
+        if (this.props.typeAhead) this.handleAlphaDown(event);
+        break;
+    }
+  };
+  handleAlphaDown = (event: KeyboardEvent) => {
+    const rootRef = this.props.rootRef ? this.props.rootRef : this.rootRef;
+    const prevIndex = this.state.highlightedIndex;
+
+    this.typeAheadChars += event.key;
+    this.typeAheadTimeOut = setTimeout(() => {
```
chasestarr, line 167: It looks like there is no `clearTimeout` included in this diff. Check the earlier example I wrote on how to handle this feature. It should `clearTimeout` in the keydown event listener. Currently if my text input takes longer than `500ms` I would not be able to search for longer strings

src/menu/stateful-container.js
```diff
@@ -149,11 +154,65 @@ export default class MenuStatefulContainer extends React.Component<
...
+      this.typeAheadChars = '';
+    }, 500);
+
+    var nextIndex = prevIndex;
+    var list = this.getItems();
+    var notMatch = true;
+    for (let n = 0; n < list.length; n++) {
+      let label = 'label' in list[n] ? list[n].label : list[n].id;
```
chasestarr, line 175: We should remove `id` here and in the other loop below. All of the menu elements have a randomly generated id

src/menu/stateful-container.js
jiaoy14: Thanks！

src/menu/stateful-container.js
jiaoy14: Yeah, thanks! I will modify it.

src/menu/stateful-container.js
```diff
@@ -149,11 +154,71 @@ export default class MenuStatefulContainer extends React.Component<
...
+        rootRef.current,
+        nextIndex === 0,
+        nextIndex === list.length - 1,
+      );
     }
   };
 
   // Handler for arrow keys
   handleArrowKey = (event: KeyboardEvent) => {
+    this.typeAheadChars = '';
```
chasestarr, line 221: I believe this reset and the similar one in `handleEnterKey` can be removed if the `clearTimeout` call on line 153 was moved from inside the default case up to the top of the `keydown` event handler

jiaoy14: yes! You are right.

PR #3566, opened 2020-07-30
feat(tag): update designs

### LINKED ISSUE #3452: [Tag] isHovered and isFocused don't do anything
isFocused and isHovered are in the prop types but don't change the hover or focused state

sandgraham:
Fixes #3452

- Simplifies Tag design by removing `light` variant and hover states on the `X` toggle.
- Adds `purple`, `brown`, and `orange` tag colors. These are hard-coded to our colors in the `tokens` module. I also added "primitive" (blue/green/etc) aliases for each "semantic" (accent/positive/etc) `KIND` value.
- Removes `light` variants since we no longer have designs for them. If you have `light` selected, you get `outline` now.
- Simplifies hover states.
- `neutral` looks the same as `primary` by default now (customizable in theme still) 
- Improve logic for determining `custom` shades.
- Deprecates `isHovered` and `isFocused`.
- Adds new Tag sizes.


sandgraham:
@chasestarr design requested that. I actually double-checked because I also wondered if it was intentional. That said, I was happy to oblige due to how much is simplifies the theming/styling.

gergelyke:
Screen Shot 2020-08-12 at 2 55 50 PM

there seems to be an error on the doc size when you change the `size` prop

PR #3569, opened 2020-07-31
feat(tree): add type-ahead for tree

jiaoy14:
`Fixes #1, Fixes #2`


src/tree-view/utils.js
```diff
@@ -189,7 +189,71 @@ export const toggleIsExpanded = (
...
+      nodeId,
+      chars,
+      closestOmmer,
+      getId,
+      false,
+      false,
+    )[0];
+  return foundid;
+};
+export const matchString = (
```
tajo, line 220: It would be great to write some unit tests for function

tajo:
Vercel couldn't build this PR. Some [syntax error](https://vercel.com/uber-ui-platform/baseweb/mcun4vczg)?

src/tree-view/utils.js
jiaoy14: I will change that function. I combined range search into DFS . As suggested by Chase (https://github.com/uber/baseweb/pull/3560#discussion_r463752404), I will change that to a direct search.

jiaoy14:
> Vercel couldn't build this PR. Some [syntax error](https://vercel.com/uber-ui-platform/baseweb/mcun4vczg)?

yes, I may need some type check.

src/tree-view/utils.js
chasestarr: The e2e test looks good. Can you also add a unit test for this function like here? https://github.com/uber/baseweb/blob/b16b9845f984ed62413e0069aef719ea4c1ab1b3/src/tree-view/__tests__/utils.test.js

tajo:
It seems not to be working for some examples like:

https://baseweb-435iz1oqp.vercel.app/components/tree-view/#adding-custom-labels

jiaoy14:
> It seems not to be working for some examples like:
> 
> https://baseweb-435iz1oqp.vercel.app/components/tree-view/#adding-custom-labels

Yes, for now it can't support customized label.

chasestarr:
How could we best support customized labels? Combobox for example has a `mapOptionToString` prop so that we can run text filters on the list

jiaoy14:
> How could we best support customized labels? Combobox for example has a `mapOptionToString` prop so that we can run text filters on the list

Emm, the existing structure gives developer much freedom for customizing.  

> How could we best support customized labels? Combobox for example has a `mapOptionToString` prop so that we can run text filters on the list

Now `tree` has a prop `getID` which is used for getting id from customized labels. It behaves similarly like `mapOptionToString`.  I can imagine we have another prop `getLabelName`  used for customized label. But I  am not sure if this is necessary because it needs developers to define it  and this only used for type-ahead when label is customized.

PR #3570, opened 2020-07-31
feat(select): add config to clear input value on backspace

chasestarr:
#### Description

When a value is selected and a user presses backspace, by default it will pop the last character from that value string and add the apply that text to the input element. When `backspaceClearsInputValue` is true it will set the input value to an empty string instead.

cc @ankitjaincst 

#### Scope
Minor: New Feature


gergelyke:
what's the difference between this new prop and `backspaceRemoves `?

chasestarr:
> what's the difference between this new prop and `backspaceRemoves `?

When `backspaceRemoves` is false, the backspace key will not remove the selected value

gergelyke:
> > what's the difference between this new prop and `backspaceRemoves `?
> 
> When `backspaceRemoves` is false, the backspace key will not remove the selected value

oh okay, understood. thanks for the clarification!

PR #3577, opened 2020-08-04
feat(button-group): use primary button for selected

gergelyke:


src/button-group/button-group.js
```diff
@@ -97,7 +96,7 @@ export class ButtonGroupRoot extends React.Component<{|
                   this.childRefs[nextIndex].current.focus();
               }
             },
-            kind,
+            kind: isSelected(selected, index) ? KIND.primary : kind,
```
chasestarr, line 99: What do you want to happen if the dev specifies a `kind` that's not `primary`? `tertiary` for example. Should that also have primary buttons when it's selected?

src/button-group/button-group.js
```diff
@@ -97,7 +106,7 @@ export class ButtonGroupRoot extends React.Component<{|
                   this.childRefs[nextIndex].current.focus();
               }
             },
-            kind,
+            kind: computedKind,
```
chasestarr, line 109:
```suggestion
            kind: !kind && isSelected(selected, index) ? KIND.primary : kind,
```

chasestarr: If kind is not specified and the button is selected change it to primary, otherwise fallback to the specified kind. This code may be incorrect because there may be a default value for kind, but it described the intention

PR #3595, opened 2020-08-05
fix(build): reads alpha version correctly

chasestarr:


chasestarr:
vercel is failing due to figma api fails - we should consider removing from this docs site once Graham's back

PR #3607, opened 2020-08-07
fix(select):avoid no foucs when index out of range

jiaoy14:
`Fixes #1, Fixes #2`


chasestarr:
Can you please add a description about why this is important to add? Is this an improvement that you are suggesting or was this reported by a baseui user?

jiaoy14:
A description of this issue.
Screen Shot 2020-08-07 at 10 44 59 AM

When we use `Select`, which leverages `Menu` as its dropdown list, dynamically changes the length of the menu. For multi-select, we want it to  have an auto-foucs on the first element on after one input. However, if we use keyboard navigation, the menu will set the `highlightItem` value the same as the previous value. And when we change the length of the menu, it will sometimes be out of range as picture 3 shows. 

This is not a serious problem because when we lose focus , we can use mouse to get focus again and it will reset the auto-focus to the where we want. But I think it  is  better to solve the problem and users can just use keyboard to choose selections.

chasestarr:
Can you share the code that produces the screenshots above? When I try to reproduce myself, the list of items hides when enter is pressed

jiaoy14:
> Can you share the code that produces the screenshots above? When I try to reproduce myself, the list of items hides when enter is pressed

I do it on the storybook 's `select-search-multi`. Also  I add an e2e test here:
```
it('subsequent multi select dropdown opens highlights first value after keyboard navigation', async () => {
    await mount(page, 'select-search-multi');
    await page.waitFor(selectors.selectInput);
    await page.click(selectors.selectInput);
    await page.waitFor(selectors.selectDropDown);
    await page.keyboard.press('ArrowDown');
    await page.keyboard.press('ArrowDown');
    await page.keyboard.press('ArrowDown');
    await page.keyboard.press('ArrowDown');
    await page.keyboard.press('Enter');
    await page.type(selectors.selectInput, 'be');
    await page.keyboard.press('Enter');
    const text = await page.$eval(
      selectors.selectedList,
      select => select.textContent,
    );
    expect(text).toBe('AzureBeige');
  });
```
The original code cannot pass this test.

chasestarr:
Oh I see now. This bug only shows up when `closeOnSelect` is `false`. Could we take a more simple approach? How about remove the prop to configure this feature, and if the index is out of bounds, reset it to the first or last item?

jiaoy14:
> Oh I see now. This bug only shows up when `closeOnSelect` is `false`. Could we take a more simple approach? How about remove the prop to configure this feature, and if the index is out of bounds, reset it to the first or last item?

Yes,  I will try. Thanks!

src/select/dropdown.js
```diff
@@ -165,6 +165,7 @@ export default class SelectDropdown extends React.Component<DropdownPropsT> {
             highlightedIndex: highlightedIndex,
           }}
           typeAhead={false}
+          focusFirstAfterKeyEnter={true}
```
chasestarr, line 168:
```suggestion
```

src/menu/stateful-container.js
```diff
@@ -286,6 +286,11 @@ export default class MenuStatefulContainer extends React.Component<
       event.preventDefault();
       onItemSelect({item: items[highlightedIndex], event});
     }
+    if (this.getItems().length >= this.state.highlightedIndex) {
```
chasestarr, line 289:
```suggestion
    if (items.length >= this.state.highlightedIndex) {
```

chasestarr: Does this bug also come up when clicking with the mouse? Does it need to be handled somewhere else other than the `handleEnterKey` method?

jiaoy14: You are definitely right here, the safest way should be to put it in right after we update the menu value.

PR #3618, opened 2020-08-08
feat(progress-bar): add stepped progress-bar option

nadiia:
#### Description

Add `stepped` property to the progress bar that renders a sectional bar.
It includes adding an additional div container around the progress bar.

#### Scope
Minor: New Feature


documentation-site/components/yard/config/progress-bar.ts
```diff
@@ -53,6 +53,12 @@ const ProgressBarConfig: TConfig = {
         },
       },
     },
+    stepped: {
+      value: undefined,
+      type: PropTypes.Number,
+      description:
+        'Renderes a dashed progress bar. Value should be set to a positive number larger than one.',
```
gergelyke, line 60:
```suggestion
        'Defines how many steps the progress bar has. Defaults to 1.',
```

documentation-site/components/yard/config/progress-bar.ts
```diff
@@ -53,6 +53,12 @@ const ProgressBarConfig: TConfig = {
         },
       },
     },
+    stepped: {
```
gergelyke, line 56: to me, `stepped` implies that it's a Boolean value. Should we name it `steps`?

nadiia: I don't have a strong opinion on the name. 'stepped' is the term design uses if we want to stay aligned. Also, I'd like to avoid introducing two props for it like one that defines it to be stepped and the second defines the number of steps. Otherwise, I'm ok with renaming it.

gergelyke: Yeah, I'd just rename it, and wouldn't introduce another prop

PR #3622, opened 2020-08-10
feat(progress-bar): update animation for infinite progress bar

nadiia:
#### Description

Update animation of the infinite progress bar. It now uses 2 elements and animates the background-position of the linear gradient to get the intended effect. This also adjusts the stepped progress bar to use flex instead of inline-block due to baseline alignment. It also enables the size value for infinite progress bars. 

#### Scope
Minor: New Feature


chasestarr:
When viewing the infinite loader, I noticed that the filled blue jumps in somewhat abruptly. Is that intentional?

chasestarr (approved):
Looking good 👍

PR #3627, opened 2020-08-11
fix: inputtextdisabled theme color

### LINKED ISSUE #3584: inputTextDisabled theme property does not work
I have noticed that the theme property `inputTextDisabled` does not seem to do anything. This can be replicated at https://baseweb.design/components/input/ by changing the colour of `inputTextDisabled` in the Theme tab and then ticking disabled in the Props tab. You will see that the colour set in the Theme tab does not get applied to the input's text.

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v9.91.0  |

rohitb4:
Fixes #3584 

#### Description

When `Input` is disabled, set `BaseInput` color to `inputTextDisabled` of the theme.

#### Scope
Patch: Bug Fix


src/input/styled-components.js
```diff
@@ -365,7 +365,7 @@ export const InputContainer = styled<SharedPropsT>(
 function getInputColors($disabled, $isFocused, $error, colors) {
   if ($disabled) {
     return {
-      color: colors.contentSecondary,
+      color: colors.inputTextDisabled,
```
nadiia, line 368: We should stick to using semantic color tokens and slowly deprecate component-specific tokens.

src/themes/light-theme/color-component-tokens.js
```diff
@@ -179,7 +179,7 @@ export default (
   inputFillDisabled: themePrimitives.mono200,
   inputFillActive: themePrimitives.mono200,
   inputFillPositive: themePrimitives.positive50,
-  inputTextDisabled: themePrimitives.mono600,
+  inputTextDisabled: themePrimitives.mono800,
```
nadiia, line 182: `mono600` is the correct token here that equals `#AFAFAF` per design spec.

src/input/styled-components.js
rohitb4: Does that mean we don't need to support `inputTextDisabled` token?

nadiia: I believe this token is still used in some form components.

sandgraham:
@rohitb4 We recently replaced a lot of the one-off variables (ex: `inputTextDisabled`) with a robust system of semantic variables (ex: `contentDisabled`). This means a number of theme variables are actually quietly deprecated. We don't remove them so as to not break anyone's code. In a future major we can do a bunch of clean up to streamline the size of the theme as well as confusion over which variables actually work.

All this is to say, the intent behind this PR isn't inline with our planning. I'm going to close this for now. Instead I think we should figure out a better way to signify which theme variables work and which are deprecated/to-be-removed.

PR #3629, opened 2020-08-12
fix(theme): add missing core spacing sizes

chrisrauh:
#### Description

Add missing spacing sizes.

image

#### Scope

Patch: Bug Fix


gergelyke:
thank you! would you mind adding them to the typescript definitions too?

chrisrauh:
> thank you! would you mind adding them to the typescript definitions too?

Done.

PR #3630, opened 2020-08-12
fix(input): add step definition to number input

ProjectBarks:
I added step support for Input

Fixes #1

#### Description
I try to use the Input component with type number and step property but I got the flow error for props. I added optional props. Follow up to, https://github.com/uber/baseweb/pull/3427
 
#### Scope
Patch: Bug Fix


gergelyke:
@ProjectBarks would you mind adding this to the TS type definitions too?

ProjectBarks:
- [x] Added TS Definitions.
- [x] Flattened commits. 
- [x] Tested E-E
- [x] Linted

ProjectBarks:
- [x] Fix unit tests

PR #3631, opened 2020-08-13
fix(menu):avoid listening to doc

jiaoy14:
`Fixes #1, Fixes #2`


airhorns:
Anything I can do to help this along? Tests are passing except for the one commented one, just the flow types that are wonky!

jiaoy14:
> Anything I can do to help this along? Tests are passing except for the one commented one, just the flow types that are wonky!

What's your opinion for this solution? Does this meet your requirement? @chasestarr makes the `menu` listens to itself and I add a new `props.listenRef` and we can dynamically assign a node ref from father to child(`menu`). So `menu` could listen to other node in father component or itself(default) rather than doc.

airhorns:
Why's the listenRef necessary in the first place? I am not super familiar with the baseweb internals enough to know when Menu gets used, but when is a parent focused but the child should respond to keyboard events? And, how would the parent know a child was listening and to not respond to events itself? If it's possible, I think the child should just get the focus, but I am not sure if that's realistic.

jiaoy14:
> Why's the listenRef necessary in the first place? I am not super familiar with the baseweb internals enough to know when Menu gets used, but when is a parent focused but the child should respond to keyboard events? And, how would the parent know a child was listening and to not respond to events itself? If it's possible, I think the child should just get the focus, but I am not sure if that's realistic.

`listenRef` can help  the child(`menu`) to be focused. Sometimes It's hard to let the child be focused if it is in a father component and only listens to itself. (This is the reason why some tests fails here https://github.com/uber/baseweb/pull/3558). For example, `menu` is a child of `select`. If menu only listens to itself, the menu can only be focused when it's clicked  and that's not we want. If `menu` could listen to the `input` of` select` and correspond when we want to text  , things can be much better. So `listenRef` can help child easily be focused or correspond with other nodes in the father component. At the same time , `listenRef` just adds flexibility and doesn't affect anything if you just want the `menu` to listen to itself(if you don't pass anything then it will listen to itself).

airhorns:
Ah that makes sense, I see why it's needed. What do you think about doing it the other way around, where a Select gets a ref to a Menu and the Select calls functions on the menu invoked by event handlers the Select added to the input? I'd see that as the other option. Both would solve my issue of not requiring bubbling. 

I think if we're being super duper purists, the Menu's event handlers should be self contained and not need to reach outside the component and the imperative style would be "better", but in practice, I don't think that is going to be much fun to set up or maintain. Every menu consumer that wanted to nest it would have to wire up the same thing and would have to know about Menu's available movements anyways, so I think passing in a `listenRef` is a simple and powerful way to fix the issue. The downside would be that if you aren't careful you could have multiple components listening to the same DOM element's events and reacting which would probably violate user expectations. I feel like that's pretty unlikely and already possible so not really diverging from the status quo though.

👍 from me but I don't really count cause I'm just a user :)

src/menu/types.js
```diff
@@ -122,6 +123,7 @@ export type StatefulContainerPropsT = {
   onItemSelect: OnItemSelectFnT,
   /** Ref for the menu container element. Used to capture key events for navigation */
   rootRef?: RootRefT,
+  listenRef: listenRefT,
```
chasestarr, line 126: Can you please change `listenRef` throughout the PR to `controlNode` or `keyboardControlNode`? I'd like the label to convey that it doesn't listen to every event, just key events and this is not a ref box, but rather the unboxed dom node

src/select/__tests__/select.e2e.js
```diff
@@ -307,7 +307,7 @@ describe('select', () => {
 
     expect(matchArrayElements(actual, expected)).toBe(true);
   });
-
+  /*
```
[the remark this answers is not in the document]
jiaoy14: Oh, this test can be passed. I forget to uncomment it.

src/select/select-component.js
```diff
@@ -350,8 +350,8 @@ class Select extends React.Component<PropsT, SelectStateT> {
       case 13: // enter
         event.preventDefault();
         event.stopPropagation();
-        if (!this.state.isOpen) {
-          this.setState({isOpen: true});
+        if (this.state.isOpen && !this.props.multi) {
```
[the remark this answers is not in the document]
jiaoy14: I think this is not the right logic. When we press "enter", if it's a single select we need to close the menu and if it's a multi select we need to keep the menu open. Originally we make the menu open in any cases after "enter" and I don't know why. 

Besides, If we don't do this modification, the test (https://github.com/uber/baseweb/pull/3631#discussion_r470065126) (https://github.com/uber/baseweb/blob/7196b43a94bd22f2a1d8a519845cad5db861cc0e/src/select/__tests__/select.e2e.js#L311) above cannot pass.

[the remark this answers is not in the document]
chasestarr: I don't really understand why the e2e would fail. Can you identify how the keydown listener change broke this test? Ideally we would not make any behavior changes in this diff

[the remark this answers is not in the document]
jiaoy14: Yes, it's quite strange here. I just found when we listen to `document` the menu will close automatically after "enter" and when we listen to a specific dom node the menu will stay open. I cannot figure out why.

src/menu/menu.js
```diff
@@ -28,6 +28,7 @@ export default function Menu(props: StatelessMenuPropsT) {
     focusMenu = () => {},
     unfocusMenu = () => {},
     handleMouseLeave = () => {},
+    handleKeyDown = (event: FocusEvent | MouseEvent | KeyboardEvent) => {},
```
chasestarr, line 31: Does this need all of the other event types?
```suggestion
    handleKeyDown = (event: KeyboardEvent) => {},
```

src/menu/types.js
```diff
@@ -34,6 +34,7 @@ export type GetProfileItemImgTextFnT = (item: ItemT) => string;
 export type SetRootRefFnT = (ref: React$ElementRef<*>) => void;
 
 export type RootRefT = React$ElementRef<*>;
+export type listenRefT = React$ElementRef<*>;
```
chasestarr, line 37:
```suggestion
```

src/menu/types.js
```diff
@@ -122,6 +123,8 @@ export type StatefulContainerPropsT = {
   onItemSelect: OnItemSelectFnT,
   /** Ref for the menu container element. Used to capture key events for navigation */
   rootRef?: RootRefT,
+  /** Ref for menu's keyboard listener. Default is null and keyboard listens to itself */
+  keyboardControlNode: listenRefT,
```
chasestarr, line 127:
```suggestion
  keyboardControlNode: React$ElementRef<*>,
```

src/menu/types.js
```diff
@@ -122,6 +123,8 @@ export type StatefulContainerPropsT = {
   onItemSelect: OnItemSelectFnT,
   /** Ref for the menu container element. Used to capture key events for navigation */
   rootRef?: RootRefT,
+  /** Ref for menu's keyboard listener. Default is null and keyboard listens to itself */
```
chasestarr, line 126:
```suggestion
  /** Node for menu's keyboard listener. Default is null and keyboard handlers will listen on menu root. */
```

src/menu/types.js
```diff
@@ -178,6 +182,7 @@ export type SharedStatelessPropsT = {
   rootRef?: RootRefT,
   focusMenu?: (event: FocusEvent | MouseEvent | KeyboardEvent) => mixed,
   unfocusMenu?: () => mixed,
+  handleKeyDown?: (event: FocusEvent | MouseEvent | KeyboardEvent) => mixed,
```
chasestarr, line 185:
```suggestion
  handleKeyDown?: (event: KeyboardEvent) => mixed,
```

src/select/select-component.js
```diff
@@ -350,9 +350,6 @@ class Select extends React.Component<PropsT, SelectStateT> {
       case 13: // enter
         event.preventDefault();
         event.stopPropagation();
-        if (!this.state.isOpen) {
```
chasestarr, line 353: Please remove the whole switch case

jiaoy14:
Will add a test.

PR #3632, opened 2020-08-13
fix(tabs-motion): handle conditional tabs

sandgraham:
Fixes an issue where conditionally rendering a `Tab` can cause hook mismatches between renders. The fix is to move the inlined `React.Children.map` "components" into actual components and move the early return condition out.


jiaoy14:
Great to learn some new knowledge from this PR! I didn't  know about hook mismatches before. So is this a bug of react?

sandgraham:
@chasestarr i thought we had those rules enabled already! that explains some things i suppose.

@jiaoy14 check out [this page](https://reactjs.org/docs/hooks-rules.html) on the rules of hooks. the tldr is that hooks have to be called in the exact same order between each render of a component. in this case i had an early return condition which caused the hooks to be called differently between the rendering of tabs.

PR #3634, opened 2020-08-13
fix(timepicker): accessing id of undefined when delete is pressed

grrowl:
#### Description

In TimePicker fields, if you press the Delete key it will incorrectly try to access `params.value[0].id` when `param.value` is `[]`

You can recreate the bug on the baseweb documentation site: https://baseweb.design/components/time-picker#timepicker

With this fix it should correctly deal with a no-value scenario, and if props `nullable` it'll update the value. Hopefully the implementation suits you!

#### Scope
Patch: Bug Fix


grrowl:
@chasestarr updated — onChange is now called with `null` to match the prop `nullable`, types updated

chasestarr:
Thanks for making that change @grrowl It looks like there were other flow problems as a result of the type change. You can run `yarn flow` locally to quickly test

grrowl:
@chasestarr thanks for the heads up, pushed a fix 👍

PR #3635, opened 2020-08-13
feat(form-control): adds htmlFor support

shannonmoeller:
#### Description

Adds support for directly setting the `for` attribute of the rendered `<label>` element. Useful when the id cannot be inferred from the child component.

#### Scope

Minor


shannonmoeller:
TS types updated.

chasestarr:
Thank you @shannonmoeller

PR #3636, opened 2020-08-13
feat(combobox): adds support for id attribute

shannonmoeller:
#### Description

Add support for setting the `id` attribute of the rendered `<input>`.

#### Scope

Minor


gergelyke:
should we add it to the TS type defs too?

shannonmoeller:
TS types updated.

shannonmoeller:
Thank you!

PR #3644, opened 2020-08-15
feat(tabs-motion): add $active state to 'Tab'

lextas:
#### Description

In the baseweb Slack channel there was a question how to change the color of the active tab in the `tabs-motion` component. I wrote a simple workaround by adding the `Tabs` to an array and then check the `activeKey` against the `tab index`. I noticed that the `Tab` from `tabs-motion` already contains an `isActive` boolean but it doesn't add it to the properties. This change adds that state so you can use this as follows.

```
import * as React from "react";
import { Tabs, Tab } from "baseui/tabs";

export default () => {
  const [activeKey, setActiveKey] = React.useState("1");
  return (
    <Tabs
      onChange={({ activeKey }) => {
        setActiveKey(activeKey);
      }}
      activeKey={activeKey}
    >
      <Tab
        title="First"
        overrides={{
          Tab: {
            style: ({ $isActive }) => {
              /* change to your own needs */
              return {
                color: $isActive ? "red" : "black"
              };
            }
          }
        }}
      >
        Content 1
      </Tab>
      <Tab title="Second">Content 2</Tab>
      <Tab title="Third">Content 3</Tab>
    </Tabs>
  );
}
```

#### Scope

Minor: New Feature

#### Bonus

To prevent code-repetition I would still recommend to put the tabs in an array so the `$isActive` style should only be defined once.

```
import * as React from "react";
import { Tabs, Tab } from "baseui/tabs-motion";

export default () => {
  const [activeKey, setActiveKey] = React.useState("0");
 
  const TabsData = [
    {
      title: "First",
      component: () => {
        return (<div>Content 1</div>)
      }
    },
    {
      title: "Second",
      component: () => {
        return (<div>Content 2</div>)
      }
    },
    {
      title: "Third",
      component: () => {
        return (<div>Content 3</div>)
      }
    }
  ]

  return (
    <Tabs
      activeKey={activeKey}
      onChange={({ activeKey }) => {
        setActiveKey(activeKey);
      }}
      activateOnFocus
    >
      {
      TabsData.map((tab) => {
        return (
          <Tab key={tab.title} title={tab.title} overrides={{
            Tab: {
              style: ({$isActive }) => ({
                color: $isActive ? 'red' : 'black'
              })
            }
          }}>
            { tab.component() }
          </Tab>
        )
      })
    }
    </Tabs>
  );
}
```


sandgraham:
@lextas LGTM. Couple small things:
- Can we change `$active` to `$isActive` so it is consistent with the `isActive` variable passed around?
- Can we update the types for the `StyledTab` component in the `styled-components.js` file?

lextas:
> 
> 
> @lextas LGTM. Couple small things:
> 
>     * Can we change `$active` to `$isActive` so it is consistent with the `isActive` variable passed around?
> 
>     * Can we update the types for the `StyledTab` component in the `styled-components.js` file?

I updated the files as requested. I didn't add a 'default' style for the `$isActive` state but this could be an improvement (I'll leave this up to the baseweb (design) team).

src/tabs-motion/styled-components.js
```diff
@@ -79,13 +79,15 @@ export const StyledTab = styled<{
   $orientation?: OrientationT,
   $fill?: FillT,
   $focusVisible?: boolean,
+  $isActive?: Boolean,
```
sandgraham, line 82:
```suggestion
  $isActive?: boolean,
```

lextas: Sorry about that, must have been autocomplete. Will fix.

sandgraham:
Thanks @lextas - I'll merge in once CI is green (running now).

Issue #3651, opened 2020-08-18
Chinese calendar dates wrong


akre54:
Viewing a Datepicker from a Chinese locale displays only the character `星` for the day of week in the CalendarHeader. 

## Current Behavior
<img width="456" alt="Screen Shot 2020-08-03 at 5 16 11 PM (1)" src="https://user-images.githubusercontent.com/931368/90534067-87bcc380-e136-11ea-9c31-eac1e9709818.png">

Displayed: 星, 星, 星, 星, 星, 星, 星

The offending function appears to be [`getWeekdayMinInLocale`](https://github.com/uber/baseweb/blob/e8f95c91498fbbc023a41f29ca6159fd6ca5d5c1/src/datepicker/utils/date-helpers.js#L125) which uses `charAt(0)`. 

I'm not a Chinese speaker, but according to [this website](http://www.learnchineseeveryday.com/2009/08/01/days_of_the_week/) the days of the week always start with 星. I wonder if this can / should be configurable through the locale, maybe falling back to `charAt(0)` for locales where this property holds.

According to the reporter, it looks like in Chinese they refer to the dates of the week by the rightmost character (I'm not sure if this is a top-to-bottom issue when written right-to-left or this is the usual way these characters are written).

## Expected Behavior
Displayed: 日, 一, 二, 三, 四, 五, 六

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v9.67.0  |
| React   |  16.12.0 |
| browser | Chrome 84 |
| locale | zn-ch |

- [X] I have searched the [issues](https://github.com/uber/baseweb/issues) of this repository and believe that this is not a duplicate.

PR #3652, opened 2020-08-18
fix(datepicker): use short weekday format for calendar header

chasestarr:
Fixes: https://github.com/uber/baseweb/issues/3651

#### Description

Calendar header was formatting according to the full weekday term (Monday, Tuesday) and clipping the first character. As described in the linked issue, this behavior is not appropriate for all languages. date-fns locale includes the intended abbreviated weekdays. This PR updates the format string from the full weekday term to the abbreviated.

#### Scope
Patch: Bug Fix


akre54:
Thanks for the quick fix, @chasestarr 👍

src/datepicker/__tests__/datepicker-i18n-chinese.scenario.js
```diff
@@ -0,0 +1,21 @@
...
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+// @flow
+
+import * as React from 'react';
+
+import {StatefulCalendar} from '../index.js';
+import {zhCN} from 'date-fns/locale/index.js';
```
akre54, line 12: FYI it doesn't look like the snapshot tester supports unicode (characters are displaying as black boxes).

PR #3662, opened 2020-08-19
feat(input):design change

jiaoy14:
`Fixes #1, Fixes #2` 

#### Description

Change the design of input to :
image


jiaoy14:
It seems we don't have 27px in our theme. So I adjust the spacing to 28px.

PR #3692, opened 2020-08-25
feat(combobox): add `onFocus` prop to combobox

jonsadka:
This PR adds an `onFocus` prop to the combobox which returns the current input value, similar to the `onBlur` behavior.

This is useful when trying to manipulate app state outside of the component based on focus/blur behavior

### DIFF
--- a/documentation-site/components/yard/config/combobox.ts
+++ b/documentation-site/components/yard/config/combobox.ts
@@ -43,10 +43,6 @@ const ComboboxConfig: TConfig = {
       value: 'nextValue => setValue(nextValue)',
       type: PropTypes.Function,
       description: 'Callback for when input focus changes.',
-      propHook: {
-        what: 'nextValue',
-        into: 'value',
-      },
     },
     size: {
       value: 'SIZE.default',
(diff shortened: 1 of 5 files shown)


src/combobox/combobox.js
```diff
@@ -178,6 +179,7 @@ function Combobox<OptionT>(props: PropsT<OptionT>) {
     setSelectionIndex(-1);
     setTempValue(value);
     if (onBlur) onBlur(event);
+    if (onFocus) onFocus(event);
```
sandgraham, line 182: I think this should be in the `handleFocus` function above.

jonsadka: Yeah good call out. Updated

documentation-site/components/yard/config/combobox.ts
```diff
@@ -39,6 +39,15 @@ const ComboboxConfig: TConfig = {
...
       },
     },
+    onFocus: {
+      value: 'nextValue => setValue(nextValue)',
+      type: PropTypes.Function,
+      description: 'Callback for when input focus changes.',
+      propHook: {
+        what: 'nextValue',
+        into: 'value',
+      },
```
chasestarr, line 49: Can you please remove this? To reproduce the bug, go to the interactive example [here](https://baseweb-git-jonsadka-focus-combobox.uber-ui-platform.vercel.app/components/combobox) select an option, blur, then focus and you will see `[object Object]` in the input field.

```suggestion
```

chasestarr: Looks like this was missed in an earlier PR. Can you please this same `propHook` value from L29 above while looking at the file? Sorry about the confusion

jonsadka: Done

PR #3723, opened 2020-09-02
feat(popover):add popover margin prop

jiaoy14:
`Fixes #1, Fixes #2`


chasestarr (commented):
Can you please add a VRT validating the feature?

sandgraham (commented):
We should add this prop to the sandbox on the doc site. You can do that in:
`baseui/documentation-site/components/yard/config/popover.ts`

chasestarr (approved):
Awesome! this lgtm

src/popover/types.js
```diff
@@ -109,6 +109,8 @@ export type BasePopoverPropsT = {
   showArrow?: boolean,
   /** Whether to toggle the popover when trigger is clicked or hovered. */
   triggerType: TriggerTypeT,
+  /** Margin of the popover */
+  popoverMargin: number,
```
chasestarr, line 113: Sorry I missed this, we need to make the prop optional because it will fail flow check for existing applications

chasestarr (changes requested):

chasestarr (approved):
thanks for the type fix

PR #3724, opened 2020-09-02
feat(fonts): add mono font face

chasestarr:
#### Description

Adds mono font to the theme and updates couple of places where platform mono fonts were previously used.

Screen Shot 2020-09-02 at 1 44 00 PM

#### Scope
Minor: New Feature


src/themes/move-theme/typography.js
```diff
@@ -15,29 +15,34 @@ export const fontTokens: FontTokensT = {
 const secondaryFontFamily =
   'UberMove, UberMoveText, system-ui, "Helvetica Neue", Helvetica, Arial, sans-serif';
 
+const monoFontFamily = 'UberMoveMono, "Lucida Console", Monaco, monospace';
```
sandgraham, line 18: Does design have preferences for backup fonts?

sandgraham (commented):
I think this would be a breaking change for folks who construct their own theme object. @chasestarr do you think we should add a fallback font object for the places we use the font? We could remove the fallback in a future major.

src/themes/move-theme/typography.js
chasestarr: They approved this while working on the data-table, but I can bring it up specifically

chasestarr:
> I think this would be a breaking change for folks who construct their own theme object. @chasestarr do you think we should add a fallback font object for the places we use the font? We could remove the fallback in a future major.

We support themes not constructed with `createTheme`? Maybe I'm misunderstanding. Can you provide an example?

sandgraham:
I haven't seen anyone make their theme entirely from scratch. Most will extend one of our default themes if not using `createTheme` directly. That said, there isn't anything stopping someone from passing an object they've constructed entirely manually.

I think it's fine though. It is probably quite rare and it shouldn't actually crash anything if you don't have the new properties. `const foo = {...undefined}` doesn't throw, it just doesn't add anything to `foo`.

PR #3726, opened 2020-09-02
fix(button): focus respects rounded corners

### LINKED ISSUE #3647: Make button's focus outline consistent with its shape
## Current Behavior

Using a theme with custom `buttonBorderRadius`: **(baseweb is using the `outline` css property here)**

Peek 2020-08-17 17-27

When I set the button's shape to `pill` it looks right, but it ends up overriding the `buttonBorderRadius` theme property: **(baseweb is now using `box-shadow` instead)**

## Expected Behavior

The focus outline should not ignore `buttonBorderRadius` thus having a consistent shape. Since the `outline` css property doesn't follow `border-radius`, it should be replaced in favor of `box-shadow` even if the button shape is not set to `pill`.

Relevant:

https://github.com/uber/baseweb/blob/e8f95c91498fbbc023a41f29ca6159fd6ca5d5c1/src/button/styled-components.js#L43

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v9.95.1 |
| React | v16.13.1 |
| browser | Chromium v84.0.4147.125 |

sandgraham:
Uses `box-shadow` instead of `outline` to respect when folks use `buttonBorderRadius` in their theme.

Fixes #3647 .


chasestarr (approved):
looks good 👍

tajo (approved):

PR #3729, opened 2020-09-03
feat(data-table): add locale support

konradkalemba:
## Description

Hey! I've been using `DataTable` component for a while now and love it. I'd be even more satisfied if it'd have locale support. So I gave it a try to implement it myself, but I bumped into few obstacles which are the following:

- In [`column-datetime.js`](https://github.com/uber/baseweb/blob/master/src/data-table/column-datetime.js) there are constants with localized labels, week days etc. - it makes them hard to localize. I can propose a solution, but wanted to ask beforehand for desired approach if there's any.
- The component's tests are using hardcoded labels. For example:
https://github.com/uber/baseweb/blob/9b088394203b4610d029a73a4200343af057e5cb/src/data-table/__tests__/column-categorical.test.js#L96 I'm not sure if it should stay the same or it should get those translations from locale context.

Thanks!

#### Scope
Minor: New Feature


chasestarr:
Thanks for the PR @konradkalemba glad you are enjoying the component.

- For date/time related localization, you may want to look at how it's handled in datepicker. It get's kind of complex since there are adapters for the various js date libs. There are likely some things that can be reused and I imagine some new helpers will need to be added.

- For tests, I'm not worried too much but if you want to feel free to import the string from the locale file

src/data-table/column-categorical.js
```diff
@@ -56,16 +58,17 @@ function InputBefore() {
 }
 
 function FilterQuickControls(props: {
+  locale: LocaleT,
```
chasestarr, line 61: Can this component `useContext` rather than shuttle the locale through props? No need to include the type if that's possible

chasestarr (commented):
I have some naming nit-picks to call out closer to when this is ready to land, but my general feedback would be to essentially namespace the labels by what column-type/component they address. For example there is a `selectAll` and `clearSelection` in the `column-categorical.js` file. I would prefer those being named `categoricalFilterSelectAll` and `categoricalFilterSelectClear`.

src/data-table/column-categorical.js
konradkalemba: Sure. 👍

konradkalemba:
> * For date/time related localization, you may want to look at how it's handled in datepicker. It get's kind of complex since there are adapters for the various js date libs. There are likely some things that can be reused and I imagine some new helpers will need to be added.

@chasestarr Thanks for the tip. I've made use of the existing datepicker's helpers and added one. To make it work properly I had to add a `locale` option to `DatetimeColumn` function. It's also passed to the filter's datepicker. 

Unfortunately, date-fns library (and other libraries) doesn't support year halves. Therefore I added two translations in locale for each of halves.

Regarding other constants (`RANGE_OPERATIONS` and `CATEGORICAL_OPERATIONS`) I introduced a `localeLabelKey` property which is pretty self-explanatory. When passing those constants as select's options this property's value is used to get the correct translation from the locale object.

src/data-table/column-categorical.js
```diff
@@ -167,6 +171,7 @@ export function CategoricalFilter(props: CategoricalFilterProps) {
           }}
         >
           <FilterQuickControls
+            locale={locale}
```
chasestarr, line 174:
```suggestion
```

konradkalemba:
I added a feature to respect locale's week starting day and list weekdays checkboxes accordingly.

chasestarr:
Thanks @konradkalemba this is looking good. Can you please review these VRT changes and merge the images into this PR after you've taken a look? https://github.com/konradkalemba/baseweb/pull/1

konradkalemba:
@chasestarr Oh, I was quite confused why are there visual changes as the PR shouldn't have introduced ones. Apparently it was because of a typo I made, so konradkalemba#1 is irrevelant now as I fixed the typo.

Regarding locale's support for data table component, AFAIK there is one more place to add locale support - it's filters' descriptions. Before starting working on this I wanted to ask if the following changes are acceptable:

- Instead of `WEEKDAY - 2` there would be `Weekday - Tuesday`
image
- Instead of `GT 322` there would be `> 322`
image

chasestarr:
> Instead of WEEKDAY - 2 there would be Weekday - Tuesday
Thank you for catching this bug.

> Instead of GT 322 there would be > 322
This sounds good to me.

konradkalemba:
@chasestart OK, I'm done with filters descriptions. Can you take a look?

I added locale support for boolean column type. However there is one potential problem there with my implementation. I didn't want to change the original approach too much, so I just pretty much replaced `true` and `false` with `locale.datatable.booleanFilterTrue` and `locale.datatable.booleanFilterFalse` respectively. Therefore if in custom locale there are no values provided for these keys, the component will crash as the boolean filter heavily depends on these. We can obviously ignore this and assume that developers supply all the values of a locale object.

chasestarr (approved):
Thanks for the feature @konradkalemba We will get a release out this week

PR #3737, opened 2020-09-08
Amend typo in documentation

thiagodebastos:
#### Description
Added the word 'you' to the following sentence in the `useStyletron` docs:
> Doing so saves **you** from needing to import the custom theme type around. See below for an example.

#### Scope
Patch: Bug Fix


chasestarr (approved):
Thank you

PR #3742, opened 2020-09-10
chore(button): update ModalButton margin

leyanlo:
Update ModalButton margin to use `:not(:last-child)`, which is slightly more semantic than `:nth-last-child(n+2)`. Should have the same specificity.


leyanlo:
> sgtm, thank you. Was this causing a bug in your application or is it just a style change?

Just a style change. We had updated our usages of `:nth-last-child(n+2)` to `:not(:last-child)` for readability in our codebase, so thought I would do the same here :-)

PR #3757, opened 2020-09-15
feat(snackbar): initial layout

chasestarr:
### Description

Begins to implement the snackbar component. No positioning just yet, but this diff includes some layout rules related to long button content

Screen Shot 2020-09-14 at 4 57 22 PM

#### Scope
Minor: New Feature


sandgraham:
@chasestarr I tried out a layout solution using only CSS. Check out [this codepen](https://codepen.io/sandgraham/pen/rNeZxdP?editors=0100).

The main issue is that it doesn't fully work in IE11 - you would need to use a fallback for `min-content` so that the widths are either always split evenly or always on separate lines.

Not sure if that would be a deal breaker, but if you're happy with the JS solution already then no worries- thought I'd give it a shot.

src/snackbar/index.js
```diff
@@ -0,0 +1,182 @@
...
+      observer.observe(rootRef.current);
+    }
+    return function cleanup() {
+      observer.disconnect();
+    };
+  }, []);
+
+  const actionMeasureRef = React.useRef(null);
+  const [actionMeasureWidth, setActionMeasureWidth] = React.useState(0);
+  React.useEffect(() => {
```
sandgraham, line 80: I wonder if [`useLayoutEffect`](https://reactjs.org/docs/hooks-reference.html#uselayouteffect) would allow us to avoid the offscreen render?

chasestarr:
Awesome I can give that a try

chasestarr:
I spent some time today trying to get the alternate css working but I'm finding some issues where it's unexpectedly wrapping. Here's a branch with those changes. https://github.com/uber/baseweb/compare/snackbar-alternate-css Running storybook and viewing the provider stories for repro

sandgraham:
I'm think it's because the start enhancer is not included in the 50% min width of the content. If they were nested in a "left side" div with the flex-grow, min-content, etc I think it works.

```
<Root>
  <Main>
    <StartEnhancer />
    <Content />
  </Main>
  <Action />
</Root>
```

All that said, it seems like the JS is going well so no sweat about dropping the CSS stuff.

chasestarr:
oh nice that totally worked after some massaging. Closing this and opening a PR with the other branch

tajo (approved):

PR #3759, opened 2020-09-15
fix(styletron): relax styletron-component to accept props

### LINKED ISSUE #3659: [styled] relax prop type when no style object provided
Instantiating a styletron-component like `styled('div')` is valid (although not very useful) but will type props as exactly `{||}`. This type makes it so the created element will not accept any props. This should be relaxed to `{}`.

https://github.com/uber/baseweb/blob/master/src/styles/styled.js#L38

RisingGeek:
Fixes #3659

#### Scope
Patch: Bug Fix

### DIFF
--- a/src/styles/styled.js
+++ b/src/styles/styled.js
@@ -35,7 +35,7 @@ export type StyletronComponent<
 };
 
 type StyleFn<Theme> = {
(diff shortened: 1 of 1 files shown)


chasestarr (approved):
Thank you

PR #3767, opened 2020-09-16
feat(app-nav-bar): add MenuItem event handlers

refrigeratorstory:
adds the possibility to pass onNavItemClick, onNavItemAuxClick, onNavItemKeyDown props to MenuItem which previously couldn't be. for example, onNavItemAuxClick is particularly useful when a middle mouse click occurs, for example, a new tab can be opened, etc.

#### Description
added appropriate static types.
every new prop is optional, so it doesn't break anything.
changed old onClick and onKeyDown type definitions to also include the current item (and also event). [this doesn't break anything, since they weren't even passed to MenuItem before]

one thing to consider: what should be done with UserMenu? the new props aren't passed to UserMenu since StatefulMenu (to which props are further passed) doesn't support them. shouldn't be an issue for now. imo, it would be a very opinionated change. same goes for MobileNav.

one thing more to note is that, onNavItemAuxClick *doesnt* call onNavItemSelect, because it obviously shouldn't.

#### Scope
Minor: New Feature

### DIFF
--- a/src/app-nav-bar/app-nav-bar.js
+++ b/src/app-nav-bar/app-nav-bar.js
@@ -56,6 +56,9 @@ export default function AppNavBar(props: AppNavBarPropsT) {
         active={active}
         item={item}
         key={index}
+        onClick={props.onNavItemClick}
+        onAuxClick={props.onNavItemAuxClick}
+        onKeyDown={props.onNavItemKeyDown}
         onSelect={onNavItemSelect}
       />
     );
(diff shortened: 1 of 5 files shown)


tajo:
TS types should be updated too.

refrigeratorstory:
> TS types should be updated too.

updated TS defs.

also added missing navPosition prop in MainNavItemT def and missing nav prop in UserNavItemT def.

one thing more to consider - should these function just return void, that is, return nothing?

tajo:
> one thing more to consider - should these function just return void, that is, return nothing?

I think `mixed` is good option.

williamernest:
It looks like there's some merge conflicts on this PR due to it's age. If this is still something you'd like to see implemented can you handle the conflicts and ping me for a review. My apologies on how long this has been left open.

williamernest:
The best that I can tell this feature has never been requested outside of this PR, so we're going to close it until we have a concrete request to implement this.

PR #3776, opened 2020-09-18
fix(data-table): reflect applied filters correctly in datetime filter

konradkalemba:
#### Description

Fixes datetime column's filter not showing applied categorical filters. The problem was with `if (input.range)` condition which always evaluated to `true` - In case when categorical filter was applied, `input.range` was an empty array which still evaluates to `true`.

#### Scope
Patch: Bug Fix


chasestarr (approved):
Thank you

PR #3791, opened 2020-09-29
fix(tabs-motion): use string key

Deliaz:
#### Description

I've noticed a bug in docs for [Tabs (Motion)](https://baseweb.design/components/tabs-motion/) example. 
The property `activeKey` isn't updating in yard when switching between the tabs:
tab motion issue demo

As I dug, it happens because the component returns the`activeKey` prop as a number, which causes this issue. 

This fix based on the code implementation from the similar Tab component, where it explicitly converts index to string: https://github.com/uber/baseweb/blob/b144ca682059421fe2fd9cfe0e45af62ad9c3beb/src/tabs/tabs.js#L44

#### Scope
Patch: Bug Fix

### DIFF
--- a/src/tabs-motion/tabs.js
+++ b/src/tabs-motion/tabs.js
@@ -293,7 +293,7 @@ function InternalTab({
   onChange,
   ...props
 }) {
-  const key = childKey || childIndex;
+  const key = childKey || String(childIndex);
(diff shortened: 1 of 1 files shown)


chasestarr (approved):
thank you

PR #3804, opened 2020-10-02
fix(link): add prop types for styled link component

### LINKED ISSUE #3362: Feature: StyledLink Prop Types
## Feature description

`StyledLink` doesn't have proper prop types. I can put any props that i want. Would be better if it has a prop type instead of `<any>`. 

For example, If i use `target` i can put anything that i want (string, bool, etc). I expect that i can only put a string there or even more specific `'_blank' | '_self' | '_parent' | '_top' `.  Even `href` itself, in the documentation the href prop type said that its a string, but i can also put anything that i want and it wont complain.

Having proper types would give a better developer experience.

git-ced:
Fixes #3362 

#### Description

Adds proper prop types for `StyledLink` component 

#### Scope
Patch: Bug Fix


tajo (approved):

chasestarr:
Thanks @git-ced happy to merge once CI passes

documentation-site/examples/table-grid/cell-navigation.js
```diff
@@ -191,7 +191,9 @@ export default function() {
                 {...getCellProps(0, primaryRowIndex)}
                 $gridRow={`span ${row[2].length}`}
               >
-                <StyledLink href={row[1]}>{row[0]}</StyledLink>
+                <StyledLink href={String(row[1])}>
```
chasestarr, line 194: My guess is this is not necessary in the flow-typed files due to it's type inference feature. Do you mind confirming this and related changes are required?

git-ced: I did this because `yarn tsc` fails when it is not wrapped with `String`. Although after checking it again, `tsx` files are the only ones that throw an error. So yeah, its not needed for the flow-typed files.

What would you prefer in this case, `String(row[1])` or `row[1] as string`

chasestarr: To be clear, this is a `.js` file and should not have any impact on `yarn tsc`. Can you confirm that `yarn flow` fails if this is not stringified?

git-ced: Yeah, it doesn't have any impacts on `yarn tsc`. Just asking what should i do for the `tsx` files, should i stringify it or just assert the type string.

chasestarr: It's fine to leave for the ts files, I would just like to simplify examples if it's not needed. the flow-typed example files are used to create the simple js example files on the docs site and that's the most common action on the site

git-ced: Aight, thanks for letting me know and for the review!

chasestarr (approved):
Thank you

PR #3805, opened 2020-10-02
fix(typography): add mono-font components typescript definitions

konradkalemba:
#### Description

Adds missing typescript definitions for typography components with mono font.

#### Scope
Patch: Bug Fix


tajo (approved):

chasestarr (approved):
Thank you

PR #3813, opened 2020-10-07
feat(table-semantic) add aria label prop for head cell

Deliaz:
#### Description
The TableBuilderColumn supports passing React.Node in header prop. 
But the area-label generated for `th` tag doesn't render text properly in this case:
```
<th ... aria-label="[object Object], not sorted">...
```
In this PR I am proposing to add a new optional property `tableHeadAriaLabel` to be shown in aria-label

#### Scope
Patch: Bug Fix


chasestarr (approved):
Thank you

PR #3821, opened 2020-10-08
chore(test): adds react@next CI job

chasestarr:
React 17 will be releasing soon and with that we will likely get an influx of related github issues. This PR adds additional CI jobs for unit-tests and e2e-tests with the next react version


chasestarr:
Unit tests currently fail because there is no enzyme adapter published yet

PR #3822, opened 2020-10-12
feat(table-semantic): add loading and empty messages

andriuss:
#### Description

Added loading and empty message support for semantic-table.
`Table` and `TableBuilder` accepts three new properties `isLoading`, `loadingMessage` and `emptyMessage`.

The loading message will be displayed when a property `isLoading` is set to true. Defaults value is `Loading...`
It will expand to the full table width using the `colSpan` attribute of total columns in the table.

The empty message will be displayed when the data array is zero length and `loadingMessage` is provided as property. By default, message value is not set and it won't be shown.

The messages are rendered inside `tr > td` tag to complain to table semantics.

semantic-table-messages

#### Scope
Minor: New Feature


chasestarr (approved):
Thank You

PR #3829, opened 2020-10-16
chore(test): removes enzyme from tests

chasestarr:
#### Description

We've stopped writing enzyme tests and enzyme doesn't have an adapter available for the next react version. So felt like an appropriate time to take care of moving everything over


tajo:
What a great effort!

tajo (approved):

PR #3834, opened 2020-10-17
fix(tabs-motion): reduce initial re-rendering

sandgraham:
Removes the 100ms debounce on updates coming from the resize observer. Instead, track tab dimensions after each render and compare them when the resize observer fires.

This fixes an issue where the Tabs can render, a Tab changes width, and then 100ms later the highlight updates. This causes a jump slightly after rendering and can cause VRTs to be flakey.

### DIFF
--- a/src/tabs-motion/tabs.js
+++ b/src/tabs-motion/tabs.js
@@ -39,17 +39,7 @@ const KEYBOARD_ACTION = {
   previous: 'previous',
 };
 
-const debounce = (fn, wait) => {
-  let timeout;
-  return (...args) => {
-    clearTimeout(timeout);
-    timeout = setTimeout(() => {
-      fn(...args);
-    }, wait);
-  };
-};
-
-const getHighlightLayoutParams = (el, orientation) => {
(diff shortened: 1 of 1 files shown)


src/tabs-motion/tabs.js
```diff
@@ -116,26 +106,15 @@ export function Tabs({
...
+  // Create a shared, memoized callback for tabs to call on resize.
+  const updateHighlight = React.useCallback(() => {
     if (activeTabRef.current) {
-      setHighlightLayout(
-        getHighlightLayoutParams(activeTabRef.current, orientation),
-      );
+      setHighlightLayout(getLayoutParams(activeTabRef.current, orientation));
     }
-  }, [activeKey, orientation, children]);
+  }, [activeKey, orientation]);
```
chasestarr, line 114:
```suggestion
  }, [activeTabRef.current, orientation]);
```

I think this can be changed to the referenced element so that we don't need to reason about how activeKey affects this code

chasestarr (approved):

tajo (approved):

Issue #3741, opened 2020-09-09
How to disable mouse hover event in emojin and start rating?


dheroying:
## Feature description
I need to disable the emoji and start rating in some cases.
So I removed the onclick event and only set the value as static.
But when I hover the mouse on emoji and starts, still I can see that the color of emojis are changing.
So how to make emojis and start as read only?

chasestarr:
We haven't had a need to do this in our projects but would be happy to accept a pull request. I would start by looking here https://github.com/uber/baseweb/tree/master/src/rating

PR #3836, opened 2020-10-18
feat(rating): allow read-only rating #3741

Kwintenvdb:
Implements enhancement #3741

#### Description
This new minor feature implements a `readOnly` option for the `StarRating` and `EmoticonRating` components.
This allows one to show a rating without allowing user interactivity with this rating.

I've also added this option to the `EmoticonRating` for consistency, although it does make a little less sense here. But I see no issue with providing this option to the user either way.

#### Scope
Minor: New Feature


chasestarr (approved):
Looks great Thank you @Kwintenvdb

PR #3841, opened 2020-10-20
chore(build): use fix and cache eslint options in lint:code command

andriuss:
#### Description

While working on the project I saw that lint command takes a very long time. Around `30 - 40 seconds` on my machine.
The suggestion is to add two options to eslint command:
- **cache**. Only check changed files. The default cache location is `.eslintcache` file which is already ignored in `.gitignore` [docs](https://eslint.org/docs/user-guide/command-line-interface#caching)
- **fix**. Automatically try to fix as many issues as possible without breaking the code, example adding semicolon ar removing space [docs](https://eslint.org/docs/user-guide/command-line-interface#-fix)

Time differences:
Screenshot 2020-10-20 at 11 00 04

#### Scope
Patch: Bug Fix

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
   "module": "./esm/index.js",
   "repository": "uber/baseweb",
   "scripts": {
-    "lint:code": "eslint ./",
+    "lint:code": "eslint ./ --fix --cache",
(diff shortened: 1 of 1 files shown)


chasestarr (approved):
Thank you

PR #3842, opened 2020-10-20
fix: change duration of snackbar from short to medium

### LINKED ISSUE #3838: [SnackBar] Allow to change default duration
## Feature description

Currently the default duration of SnackBar is `DURATION.short`. I'd like to change it to `DURATION.medium` globally.

encetroc:
Fixes #3838 

#### Description

- I changed the default duration of snackbar from short to meduim
- I updated the documentation to reflect the changes made

#### Scope

Patch: Bug Fix

### DIFF
--- a/src/snackbar/snackbar-context.js
+++ b/src/snackbar/snackbar-context.js
@@ -47,6 +47,7 @@ export default function SnackbarProvider({
   children,
   overrides = {},
   placement,
+  defaultDuration,
 }: SnackbarProviderPropsT) {
   const [css] = useStyletron();
 
@@ -58,7 +59,7 @@ export default function SnackbarProvider({
   const [containerHeight, setContainerHeight] = React.useState(0);
   const containerRef = React.useRef(null);
 
-  function enqueue(elementProps, duration = DURATION.medium) {
+  function enqueue(elementProps, duration = defaultDuration || DURATION.short) {
     setSnackbars(prev => {
       if (prev.length === 0) {
         enter(duration);
--- a/src/snackbar/snackbar-context.js
+++ b/src/snackbar/snackbar-context.js
@@ -58,7 +58,7 @@ export default function SnackbarProvider({
   const [containerHeight, setContainerHeight] = React.useState(0);
   const containerRef = React.useRef(null);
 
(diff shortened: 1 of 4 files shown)


chasestarr:
Thanks for looking into this. Based on our design guidelines snackbars with one line of text should use the short duration. I assume that will be the most common case so the default should remain as short. When working on a particular application I may know that my snackbars more often have two lines. As a developer experience improvement I should be able to supply the `SnackbarProvider` with a `defaultDuration` prop so that I can save some typing and only configure the outliers.

```
<SnackbarProvider />

// this queued snackbar should render for the short duration
enqueue({message: 'hello world'});
```

```
<SnackbarProvider defaultDuration={DURATION.medium} />

// this queued snackbar should render for the medium duration
enqueue({message: 'hello world'});
```

encetroc:
Thanks for your clarification as this is my first time trying to contribute to open source. I updated my PR and added the possibility to change the default duration using the provider.

src/snackbar/snackbar-context.js
```diff
@@ -47,6 +47,7 @@ export default function SnackbarProvider({
   children,
   overrides = {},
   placement,
+  defaultDuration,
```
chasestarr, line 50:
```suggestion
  defaultDuration = DURATION.short,
```

src/snackbar/snackbar-context.js
```diff
@@ -58,7 +59,7 @@ export default function SnackbarProvider({
   const [containerHeight, setContainerHeight] = React.useState(0);
   const containerRef = React.useRef(null);
 
-  function enqueue(elementProps, duration = DURATION.short) {
+  function enqueue(elementProps, duration = defaultDuration || DURATION.short) {
```
chasestarr, line 62:
```suggestion
  function enqueue(elementProps, duration = defaultDuration) {
```

src/snackbar/snackbar-context.js
chasestarr: This way it only applies short if `defaultDuration` is undefined

chasestarr (commented):
Thanks for making this change!

chasestarr (approved):
Thank you

PR #3848, opened 2020-10-22
fix(data-table): consistent toLowerCase comparators in textQueryFilter

klyap:
#### Description
There are cases where case-agnostic string equality is checked by comparing the results of `toLocaleLowerCase` to the results of `toLowerCase`. This works in most cases, but [for some locales, the output can be different](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toLocaleLowerCase#:~:text=toLocaleLowerCase()%20does%20not%20affect,may%20be%20a%20different%20result.), even if the input strings are identical.

#### Scope
Patch: Bug Fix


chasestarr (approved):
Thank You

PR #3857, opened 2020-10-23
fix(progress-steps): remove background color from progress steps container

### LINKED ISSUE #3820: [Progress Steps] Design update issues
See comment here: https://github.com/uber/baseweb/pull/3703#issuecomment-698003566

> Here are a before and after picker of the progress steps in my app
> https://imgur.com/a/USOW6ui
> 
> this update made the margins much larger than they were before for steps we are not currently on.
> 
> Also if you view https://baseweb.design/components/progress-steps/ in dark mode, there is a weird, non-transparent black box surrounding the circles.

- Need to fix dark mode node backgrounds
- Margin between nodes seems a bit large

akshatnitd:
Fixes #3820 

#### Description
remove background color from progress steps container

#### Scope
Patch: Bug Fix


sandgraham (approved):
👍

PR #3876, opened 2020-10-29
fix(docs): replace stackblitz with codesandbox

chasestarr:
codesandbox fixed the issues we had with it a few months ago, so switching back


tajo (approved):

EricSimons:
Hey @chasestarr- one of our users just notified me that these no longer open in StackBlitz :( Any chance you could have "open in SB" listed alongside CSB, similar to what Material UI does?

PR #3895, opened 2020-11-05
fix(icon): improve prop pass through logic

### LINKED ISSUE #3410: Icons add extra attributes to dom
Icons add props to the dom that are unnecessary.

E.g.
```js
export default () => {
  return <ArrowUp size={64} color="red" />;
}
```

Renders as:
```html
<svg data-baseweb="icon" viewBox="0 0 24 24" title="Arrow Up" size="64" color="red" class="cu s4 q3 pm pn" />
```

The size attribute is not necessary, as it is added as width/height rules by Styletron. Color is also not necessary, as it is added as color/fill rules by Styletron.

I believe this is due to the nature of React.forwardRef(), where setting the prop will get passed all the way down to the svg.

sandgraham:
Fixes #3410 

These changes address a few concerns around attributes which end up on icons.

- Fix extraneous props (`size`, etc) from ending up on the default icons.
- Fix styling props (`$foo`) from being passed to non styletron components (provided through the theme)
  - Note, you still need to handle `title`, `size`, and `color` - more on that below...

I think these changes bring the theme customization interface closer to what we have documented. That said, I think we can be more explicit about what is passed to your custom icon implementation. With the changes in this PR, the props passed through look like this:

Should definitely be handled (passed by our components)
- `title`
- `size`
- `color`

Can be handled or used within custom implementation
- `data-baseweb`: Always `"icon"`
- `viewBox`: Always `"0 0 24 24"`
- `children`: The children of the default icon
- Anything else you pass or override on the icon instance

If you pass a Styletron component as the custom icon you will also get
- `$style`: Either `undefined` or a style object which you passed through overrides


sandgraham:
Hm looks like some VRT failed - will debug that.

sandgraham:
@tajo @chasestarr should be ready for review now 👍

src/icon/__tests__/icon-attributes.scenario.js
```diff
@@ -0,0 +1,52 @@
...
+// Verify that we are not applying extraneous attributes (color/width/height)
+// to our default icons. Also verify that custom icons do not recieve dollar prefixed
+// props, either from Icon or from directly passing them.
+export default function Scenario() {
+  return (
+    <ThemeProvider theme={theme}>
+      {/* A default icon */}
+      <Check size={100} color="red" />
+      {/* A custom implementation */}
+      <Upload size={100} color="red" $testing="123" data-testing="123" />
```
chasestarr, line 49: Have you considered making a unit test that asserts no exceptions are raised? We won't know if there are future regressions in this behavior.

chasestarr (approved):
One testing suggestion, but otherwise LGTM

PR #3901, opened 2020-11-07
chore(docs): add table guide

chasestarr:
#### Description

Adds a 'table guide' to compare the different options base web publishes


sandgraham (approved):
Very nice

PR #3902, opened 2020-11-07
Check that ref for ID exists in treeItemRefs

jayantbh:
#### Description

This PR will resolve this error that I keep running into:
image

#### Scope
Patch: Bug Fix


sandgraham (approved):

sandgraham (changes requested):
Ah actually we don't have a loader for `?` yet. Can we do this the ol' fashioned way?

jayantbh:
Aha. Noted. Will do.

jayantbh:
@sandgraham, done.

chasestarr (approved):
Thank You

PR #3906, opened 2020-11-09
added aria label prop and attribute to checkbox

### LINKED ISSUE #3905: Adding aria-label attribute for checkbox component (for screen readers)
The checkbox component does not have an aria-label attribute. Adding one would make it more accessible for screen readers as the title attribute is often insufficient.
## Feature description
Adding a single attribute/prop (aria-label) to the checkbox component to increase accessbility.

mattcfilbert:
Fixes #3905

#### Description

Added a single attribute and prop to the checkbox component to allow an aria-label to be used. This will allow greater accessibility for the component as the title attribute is not consistently reliable for all screen readers. (Discovered while getting my 508 accessibility training while working on an app using BaseUI as primary component library.)

#### Scope

Minor: New Feature


tajo:
Should be added to `d.ts` file too

mattcfilbert:
Thank you for your feedback @tajo - as was apparent, I am a new contributor and am happy to learn the proper pattern. Is there anything else that I need to do to make buildkite run? I believe I made the necessary code changes, but see that the build is blocked

chasestarr (approved):
Thank You

PR #3912, opened 2020-11-11
fix(type): Bug with theme sizing type

### LINKED ISSUE #3911: Bug with theme sizing type
## Current Behavior

in `theme.ts` interface Sizing doesn't include scale650, Typescript complains about it. I checked git blame and this bug exists for 2 years already if I remember clearly. It might affect some people using Typescript.
## Expected Behavior
I believe it's supposed to, because if I print out Sizing, it has scale650 field
image

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v9.104.0  |
| React   | 17.0.1    |
| browser |      Chrome Version 86.0.4240.111 (Official Build) (64-bit)   |

- [X] I have searched the [issues](https://github.com/uber/baseweb/issues) of this repository and believe that this is not a duplicate.

Shawnxkwang:
Fixes #3911

#### Description
See issue

#### Scope

Bug Fix

### DIFF
--- a/src/theme.ts
+++ b/src/theme.ts
@@ -652,6 +652,7 @@ export interface Sizing {
   scale500: string;
   scale550: string;
   scale600: string;
+  scale650: string;
   scale700: string;
   scale750: string;
   scale800: string;


tajo (approved):

chasestarr (approved):
Thank you

PR #3914, opened 2020-11-11
fix(datepicker): add with mask example. Fix timezone bug

andriuss:
#### Description

This PR includes 4 changes to Datepicker documentation page:

1. adds placeholder for `Datepicker with localization` example
2. adds `Datepicker with custom mask` example. It clarifies that three props (formatString, placeholder, mask) for datepicker must be set to work correctly. If `formatString` is not set, manual date input (typing) will not work, example - https://stackblitz.com/edit/react-xvthsy-ebrvmn?file=src/example.js
3. moves `Datepicker with no mask` description text after input
4. fixes a bug where a page crash if a date is removed in `Datepicker with timezone selection` example

datepicker

#### Scope
Patch: Bug Fix

### DIFF
--- a/documentation-site/examples/datepicker/datepicker-with-timezone.js
+++ b/documentation-site/examples/datepicker/datepicker-with-timezone.js
@@ -16,10 +16,12 @@ export default () => {
     <React.Fragment>
       <FormControl label="Date">
         <DatePicker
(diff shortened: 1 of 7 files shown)


chasestarr (approved):
Thank you

PR #3940, opened 2020-11-19
Nested overrides playground

tajo:
This PR adds support for nested overrides into our interactive playground and makes the whole UI more condensed:

Screen Shot 2020-11-19 at 1 29 16 PM

## Features
- It works recursively, so it should work even for most complex cases. 
- Accordion now uses the stateless version that's always in sync with the code snippet. For example if you delete an override, it will close the related tree branch in the playground. This was buggy before because we had only stateful accordion.
- There is a `nested` tooltip you can hover over and get more info.
- It is very easy to configure this for each component (this PR updates only Select -> Tag to demo it).

```
overrides: {
  custom: {
    names: [
      'Root',
      TagConfig,
```

Before, the `names` could be only strings. Now, it can be also a component configuration to represent the nested override. 

- Default style object for override now omits `return {}` to limit code nesting (as you can see in the screenshot). There was actually a bug in `@babel/generator` that did not correctly parenthesized returning object of the arrow function `() => ({})` and resulted in syntax error. Fortunately, it was recently [fixed](https://github.com/babel/babel/pull/12082) so that's the only reason why `react-view` needed to be bumped.


documentation-site/components/yard/overrides.tsx
```diff
@@ -66,58 +78,118 @@ const Overrides: React.FC<TOverridesProps> = ({
...
-              set={set}
-            />
-          </Panel>
-        ))}
+              title={
+                <span>
+                  {overrideKey}
+                  {nested ? (
+                    <NestedTooltip
+                      name={componentName}
```
chasestarr, line 145: There are some nested overrides that don't have the same key as the component name. For example there are multiple 'Select' components in the datepicker. May need to add some key value pair in the config file

https://github.com/uber/baseweb/blob/master/src/datepicker/types.js#L29

chasestarr (approved):
Works well for this particular nested override. Feel free to ignore the comment I made until a later pr

documentation-site/components/yard/overrides.tsx
tajo: for that case it should be enough to do

```
overrides: {
  custom: {
    names: [
      'Root',
      {...TagConfig, componentName: 'TagRenamed' }
```

if not, will come up with some different fix later

gergelyke:
this is really cool 👏

PR #3949, opened 2020-11-23
refactor(slider): make it work with a single numeric value

jayantbh:
#### Description

This PR allows the slider component to work even if you just provide a single number as the value prop.
If you do so, the value returned by the onChange handler will also be a single number.

All existing functionality remains the same.

#### Scope

Patch: Refactor

### DIFF
--- a/documentation-site/examples/slider/overrides.tsx
+++ b/documentation-site/examples/slider/overrides.tsx
@@ -1,15 +1,19 @@
 import * as React from 'react';
-import {Slider} from 'baseui/slider';
+import {Slider, SliderState} from 'baseui/slider';
 
 export default () => {
-  const [value, setValue] = React.useState([70]);
+  const [value, setValue] = React.useState<SliderState['value']>([
+    70,
+  ]);
   return (
     <Slider
       value={value}
       onChange={({value}) => setValue(value)}
       overrides={{
         InnerThumb: ({$value, $thumbIndex}) => (
-          <React.Fragment>{$value[$thumbIndex]}</React.Fragment>
+          <React.Fragment>
+            {Array.isArray($value) ? $value[$thumbIndex] : $value}
+          </React.Fragment>
         ),
         ThumbValue: () => null,
         Thumb: {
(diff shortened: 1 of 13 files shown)


documentation-site/examples/slider/overrides.tsx
```diff
@@ -1,15 +1,19 @@
...
+  ]);
   return (
     <Slider
       value={value}
       onChange={({value}) => setValue(value)}
       overrides={{
         InnerThumb: ({$value, $thumbIndex}) => (
-          <React.Fragment>{$value[$thumbIndex]}</React.Fragment>
+          <React.Fragment>
+            {Array.isArray($value) ? $value[$thumbIndex] : $value}
```
tajo, line 15: Frankly, I am not a big fan of having ambiguous prop types. This LOC is a good example why. Also it makes static types harder to read and use.

I get that there it might be slightly more convenient if you never use double-thumb slider but if it's that important to you, you can wrap/re-export it as your own component.

We also use this API approach with other components like select.

jayantbh: >you can wrap/re-export it as your own component

I like this idea.

> Frankly, I am not a big fan of having ambiguous prop types

Initially I didn't feel like it was ambiguous much, but, probably only if you're using Flow/TS.
So, yeah, unless I type it out better, I should probably export a separate component. But that means adding a separate component just to handle one minor change in the input. Doesn't seem ideal to me.

Do you have any suggestions about how to incorporate this change in a better way?

jayantbh: Maybe I can just wrap things in functions and specify their types more directly? If that makes sense?

tajo: > But that means adding a separate component just to handle one minor change in the input. Doesn't seem ideal to me.

You can spread props through. We also export all types. It should be pretty straightforward to wrap/re-export?

chasestarr:
Thanks for the PR, we are going to keep the api as-is

Issue #3955, opened 2020-11-25
Add Optional Close Button to Filter


mattcfilbert:
Providing the option to have a close button for table filters allows devs to:
- better address potential accessibility concerns in the absence of an explicit close button
- better customize the filter component to match the rest of an application (which may, as a norm, have close buttons on pop-ups)

## Feature description
Adding an optional close button, determined by a boolean prop on the filter, gives developers greater customization for design/accessibility.

mattcfilbert:
Thanks @chasestarr for your help getting this through!

PR #3956, opened 2020-11-25
added optional filter close button plus tests

mattcfilbert:
Addresses issue #3955 

**Description**
Added an optional close button, determined by a boolean prop on the filter. This gives developers greater customization for design/accessibility when using the filter.

**Scope**
Minor: New Feature


mattcfilbert:
Happy to remove the close button from the example if you would prefer - put it there for extra visibility of the optional feature

mattcfilbert:
Checks all passed - updated latest master so they need to re-run. Thank you for the prior update @chasestarr

chasestarr (approved):
👍 thanks!

PR #3967, opened 2020-12-05
chore(docs): update next to version 10

chasestarr:


tajo:
Did we need to make all exports not anonymous?

tajo (approved):

chasestarr:
> Did we need to make all exports not anonymous?

Didn't _need_ to, but it logs an annoying warning for each example file when running locally

PR #3970, opened 2020-12-08
fix(select): onblur handles unmount

chasestarr:
click outside handler was not removed on component unmount

#### Scope
Patch: Bug Fix


tajo:
Nice! LGTM. Here's the [CodeSandbox](https://codesandbox.io/s/basic-usage-forked-g761h?file=/example.js) testing Ardo's usecase.

PR #3983, opened 2020-12-14
Use uniform margins for buttons in groups

tajo:
This simplifies margins and fixes the RTL scenario.


src/button-group/button-group.js
```diff
@@ -127,36 +127,15 @@ export default class ButtonGroup extends React.Component<PropsT> {
...
-                      if (index === children.length - 1) {
-                        return {
-                          borderTopLeftRadius: 0,
-                          borderBottomLeftRadius: 0,
-                          marginLeft: '1px',
-                        };
-                      }
-                      // inner button(s)
                       return {
-                        borderTopRightRadius: 0,
```
chasestarr, line 155: Why remove the additional styles? I believe this is here to support older rounded corner buttons

tajo: I think the current also doesn't work for RTL. We probably could add even more conditions but it feels strange adding all this logic trying to support possible override changes? Also, it's been a while since we had round buttons.

tajo: If people wish to have round buttons in a group with inner squared button, they should imo implement their own button group, not trying to rely on our styles that are based around differently shaped buttons.

chasestarr (approved):

PR #3985, opened 2020-12-14
fix(tag): reorder text for rtl direction

enjoylife:
#### Description
The tag component currently does not respect the `rtl` direction.
Reproduce here: https://codesandbox.io/s/base-web-tag-forked-58oq0?file=/src/example.js

image

#### Scope
Patch: Bug Fix


chasestarr (approved):
Thank You

PR #3990, opened 2020-12-16
docs(table-semantic): adds example for alternating table rows

shannonmoeller:
#### Description

Adds an example of a semantic table builder with alternating row background colors.

#### Scope
Docs


tajo:
Nice example. Thanks!

shannonmoeller:
Thanks for merging!

Issue #4007, opened 2020-12-31
Design fix for the date when peek next month is set to true


hardikmodi1:
Reproducible in the documentation site too.

## Current Behavior
Currently it shows the border around the out side month date. Moreover border does not appear on first and last date of the outside month. Also this does not appear if range is set to true. 
<img width="1776" alt="Screenshot 2020-12-31 at 2 40 10 PM" src="https://user-images.githubusercontent.com/22965398/103403460-2ee64880-4b76-11eb-9abb-bee0c5a0edf3.png">

## Expected Behavior
<img width="1775" alt="Screenshot 2020-12-31 at 2 43 45 PM" src="https://user-images.githubusercontent.com/22965398/103403588-ad42ea80-4b76-11eb-9c26-f41644ef71fb.png">
Instead of showing border, just changing the color to `calendarForegroundDisabled` makes it to look much nicer

- [x] I have searched the [issues](https://github.com/uber/baseweb/issues) of this repository and believe that this is not a duplicate.

hardikmodi1:
If expected behaviour mentioned above is fine or some changes required then will raise PR for the same.

PR #4013, opened 2021-01-06
fix(datepicker): design for outside month dates

hardikmodi1:
Closes: https://github.com/uber/baseweb/issues/4007

#### Description
Change the styles for outside month dates when peekNextMonth prop is passed as true. Attaching here before and after screenshot

Before - 
before

After -
after

#### Scope
Patch: Bug Fix


chasestarr (approved):
Thanks for fixing!

PR #4031, opened 2021-01-20
fix(docs): use overflow: auto instead of scroll for code blocks

leyanlo:
Use overflow: auto instead of scroll for code blocks so scrollbars only appear when needed.

Before: 
scrollbars-before

After:
scrollbars-after


tajo (approved):

chasestarr:
Thanks!

PR #4036, opened 2021-01-22
fix(docs): include guidance to migrate way from overrides props functions

### LINKED ISSUE #4035: Overrides Props as a function" warning does not provide guidance on what to do.
Ran into the warning which informs that a function passed into the override props is about to be removed in the next version.

For the specific context, I was needing access to the onChange event occurring within the datepickers inner quick select components, e.g. knowing when a quick selection actually used.
https://codesandbox.io/s/base-web-datepicker-override-quickselect-onchange-w9tnv?file=/src/example.js
## Current Behavior
Warning is emitted `baseui:Overrides Props as a function will be removed in the next major version. ` But doesn't say what one can do to fix it.

## Expected Behavior
The warning provides guidance on alternative approaches.

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v9.107  |
| React   |         |
| browser |         |

enjoylife:
* Clarify and provide guidance for users seeing the warning: `baseui:Overrides Props as a function will be removed in the next major version`
* Add complex component prop merging example for overrides and datepicker docs.

Fixes #4035

### DIFF
--- a/src/helpers/overrides.js
+++ b/src/helpers/overrides.js
@@ -110,7 +110,8 @@ export function getOverrides(
     // TODO(v11)
     if (__DEV__) {
       console.warn(
-        'baseui:Overrides Props as a function will be removed in the next major version.',
+        'baseui:Overrides Props as a function will be removed in the next major version. Override the whole component instead. ' +
+          'See https://baseweb.design/guides/understanding-overrides/#override-the-entire-subcomponent',
       );
     }
     const DynamicOverride = React.forwardRef((props, ref) => {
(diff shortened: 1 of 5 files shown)


chasestarr:
Addressed that this feature is buggy and will not be supported in the future. Because of that, we don't want to point anyone towards using it before the feature is removed.

src/helpers/overrides.js
```diff
@@ -110,7 +110,8 @@ export function getOverrides(
     // TODO(v11)
     if (__DEV__) {
       console.warn(
-        'baseui:Overrides Props as a function will be removed in the next major version.',
+        'baseui:Overrides Props as a function will be removed in the next major version. Override the whole component instead. ' +
```
enjoylife, line 113: @chasestarr

chasestarr (approved):
Thank You

chasestarr:
Can you please run `yarn flow` and `yarn tsc` to typecheck the new example?

enjoylife:
fixed lint

Issue #4030, opened 2021-01-19
[Drawer] showBackdrop prop has no effect


chasestarr:
If `showBackdrop = false`, I expect the backdrop to not 'darken' the rest of the application. It looks to me that this prop never worked in the first place. Something like https://codesandbox.io/s/base-web-drawer-forked-vy3or?file=/src/example.js:229-230 works for now. A fix for this should essentially perform the same action that this override is doing but based on the `showBackdrop` prop

mooreInteractive:
I'm going to see if I can apply a fix for this.

chasestarr:
Thanks for fixing!

PR #4040, opened 2021-01-26
fix(drawer): showBackdrop prop fix from Issues

mooreInteractive:
Fixing Issue: https://github.com/uber/baseweb/issues/4030

#### Description
The showBackdrop prop was being used to determine if the backdrop was rendered. The desired behavior is for the backdrop to still render, so it can catch clicks and close the drawer, but be fully transparent instead of partially transparent. 

After inspecting the code, I found where we could pass along the prop into the styled component and use it as an additional check for setting the opacity to 0. 

Since the backdrop should now always be rendered, I've added a new test to ensure that is the case.

I've also turned on this prop in the documentation page as it seems to be have been turned off because it wasn't working properly yet.

#### Scope
Patch: Bug Fix

### DIFF
--- a/documentation-site/components/yard/config/drawer.ts
+++ b/documentation-site/components/yard/config/drawer.ts
@@ -87,10 +87,10 @@ const DrawerConfig: TConfig = {
       hidden: true,
     },
     showBackdrop: {
-      value: false,
+      value: true,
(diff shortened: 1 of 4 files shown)


tajo (approved):
LGTM. Thanks!

Issue #4039, opened 2021-01-25
Full-height DataTable double-scroll bar


supersonicclay:
## Current Behavior

https://codesandbox.io/s/datatable-double-scrollbar-rtmms
Notice the double vertical scroll bars

![image](https://user-images.githubusercontent.com/469582/105734922-5e2b7380-5ef0-11eb-81de-83b6cbef6e08.png)

## Expected Behavior

This worked properly in v9.44.2
https://codesandbox.io/s/datatable-no-double-scrollbar-fhptb

![image](https://user-images.githubusercontent.com/469582/105734945-65528180-5ef0-11eb-92b3-f3e037c09cf5.png)

## Your Environment

https://codesandbox.io/s/datatable-double-scrollbar-rtmms
baseui v9.108.0
Chrome

- [x] I have searched the [issues](https://github.com/uber/baseweb/issues) of this repository and believe that this is not a duplicate.

FYI, this (or a similar issue) has been fixed in the past: https://github.com/uber/baseweb/issues/2592

chasestarr:
Thank you for reporting. A fix was included in 1.109.0

PR #4041, opened 2021-01-26
fix(data-table): full height table

chasestarr:
fixes: https://github.com/uber/baseweb/issues/4039

Fixed and added snapshot test

#### Scope
Patch: Bug Fix


src/data-table/data-table.js
```diff
@@ -633,6 +633,8 @@ function MeasureScrollbarWidth(props) {
   return (
     <div
       className={css({
+        height: 0,
+        overflow: 'scroll',
```
chasestarr, line 637:
```suggestion
```

tajo (approved):

PR #4050, opened 2021-02-03
Add a 'uid' option to tabs

stephencookdev:
#### Description

Currently the `tabs-motion` component leverages `useUid` hook from `react-uid`

Although `useUid` is documented to be SSR-safe, this isn't entirely true — [as also documented regarding code splitting](https://www.npmjs.com/package/react-uid#code-splitting).

Unfortunately, in some frameworks (e.g. Next.js) details of SSR and parts and ordering of code-splitting is black-box, and not entirely within your control. As such, the use of `useUid` with no hooks to override makes the `Tabs` component difficult to work with, as it generates SSR errors (as the UID ends up being different between SSR and client rendering).

This PR adds the ability to specify a `uid` custom override, so it's possible to break out of the logic handled by `useUid`.

#### Scope

Minor: New Feature


tajo (approved):

stephencookdev:
I'm not 100% sure why the VRT testing is reporting image diffs — nothing has changed visually in any way I can repro 🤔

chasestarr:
> I'm not 100% sure why the VRT testing is reporting image diffs — nothing has changed visually in any way I can repro 🤔 would appreciate help debugging that, if it's not a known issue

The results don't make sense to me and have not seen it this incorrect before. I've restarted the job to confirm it's not flakey behavior. Will see soon

chasestarr:
Thank you

PR #4053, opened 2021-02-04
Add the ability for star-ratings to have partial stars

stephencookdev:
#### Description

Currently the following

```jsx
import { StarRating } from "baseui/rating";

export default () => {
  return (
    <StarRating value={3.9} />
  );
}
```

will return a component showing 3/5 stars filled.

This is fine, but it's fairly common behaviour for star ratings to show half-stars (i.e. in the example, 3 and a half stars)

This PR adds in support for this, making the example above show 3 and a half stars, rather than a flat 3

#### Scope

Minor: New Feature


src/rating/star-rating.js
```diff
@@ -79,6 +79,9 @@ class StarRating extends React.Component<StarRatingPropsT, RatingStateT> {
           $isActive={
             previewIndex !== undefined ? x <= previewIndex : x <= value
           }
+          $isPartialActive={
```
stephencookdev, line 82: I can see the argument for making this an opt-in prop. Happy to add if you like. IMO omitting the prop is simpler, and it's fairly unsurprising default behaviour.

But very happy to make this opt-in if people prefer 🙂

chasestarr:
This change LGTM besides the flow error. Can you please run `yarn flow` locally and update?

stephencookdev:
@chasestarr whoops — fixed those now 🙂

chasestarr (approved):
Looks good, thank you

PR #4069, opened 2021-02-10
Ladle: Fix minor story errors

tajo:
Went through all the stories and found/fixed some minor issues. Also, adding netlify deploy.


chasestarr (approved):

src/snackbar/snackbar-element.js
```diff
@@ -81,7 +81,9 @@ export default function SnackbarElement({
   React.useEffect(() => {
     if (__BROWSER__) {
       const observer = new window.ResizeObserver(([entry]) =>
-        setRootWidth(entry.contentRect.width),
+        window.requestAnimationFrame(() =>
```
chasestarr, line 84: I'm seeing jittery performance in the snackbar. Do you see as well? https://deploy-preview-4069--baseweb.netlify.app/?story=snackbar--provider-overrides

tajo: Yea, I see it in this example now too. Without RAF, it was throwing a runtime error: https://stackoverflow.com/questions/49384120/resizeobserver-loop-limit-exceeded

Probably because setRootWidth is too expensive operation?

tajo: hm, actually seeing this error even with RAF in this specific example

PR #4071, opened 2021-02-11
Fix StarRating line-height issues

stephencookdev:
#### Description

Currently if you change the `line-height` of your `body`, you can break the `StarRating` component. The reason for this is is that by default `li` and `ul`'s `line-height` inherit. This then causes the following issues:

1. the line-height of the `li` containing the star pseudoelements is set to a dynamic value (i.e. outside the control of this library)
2. the pseudo-element rendering the star actually renders outside of the `li` containing it, because although we set the `height` of the `li`, we don't `overflow: hidden`, so don't notice that the star actually exceeds its `li`
3. we _do_ set `overflow: hidden` on the partial-star pseudoelement, so this can become cut-off, depending on the `line-height` involved

You can repro the root issue here by creating a `<Rating value={3.5} />` with a global styling `body { line-height: 2 }`, although I imagine there could be other issues caused by the pseudoelements rendering outside of their `li`

#### Scope

Patch: Bug Fix


chasestarr (approved):
thank you

PR #4072, opened 2021-02-12
Don't scroll the page down on tabs-motion mount

stephencookdev:
#### Description

Currently, in tabs-motion we run `scrollIntoView` when a tab changes, or on mount.

Unfortunately, `scrollIntoView` actually changes the _page_ scroll position, as well as just the parent scroll position. (IMO this is a fairly poorly documented function, but you can easily test this by calling `scrollIntoView` any element not currently in the viewport).

I'm guessing this is a bug, rather than this being intended behaviour — as I can't think of any reasonable UX where we would _always_ want to scroll the page down to the tabs (especially as having multiple tabs on a page isn't at all something advised against).

To fix this, I've copied the logic aiming to centre the element in the parent view, but manually applying the `scroll` so that the scroll logic is scoped _just_ to the parent element, rather than the whole viewport.

#### Scope
Patch: Bug Fix


chasestarr (commented):
Nice find. Is there a test possible to validate this behavior? Maybe a vrt interaction like these here? https://github.com/uber/baseweb/blob/master/vrt/config.js

stephencookdev:
@chasestarr good shout — I've updated the VRTs in 6db6132

It looks like there was actually already a test looking at this, but the snapshot must have been updated at some point with an erroneous snapshot (i.e. it looks like the scrolling to the element on mount was a regression that was accidentally snapshotted).

I've tried to add some in-DOM explanation to help prevent this in future.

chasestarr (approved):
👍

PR #4073, opened 2021-02-12
Ladle for E2E tests

tajo:


.buildkite/pipeline.yml
```diff
@@ -138,7 +130,7 @@ steps:
   - name: e2e
     # we have to install puppeteer here to trigger the install.js script
     # and also to make sure we do not add this to the released build artifact
-    command: yarn add puppeteer && yarn e2e:test:ci
+    command: yarn e2e:test:ci
```
chasestarr, line 133: IIRC this was added because users where finding puppeteer pulled in when getting baseui from npm https://github.com/uber/baseweb/pull/1909/files

chasestarr (approved):
LGTM after verifying the puppeteer thing

tajo:
It is a dev dependency? Not sure how it would endup in the baseui bundle.

Issue #2872, opened 2020-02-18
[Block] Setting `className` on a `Block` element results in a TypeScript error


jonsadka:
Setting `className` on a `Block` element results in a TypeScript error:

https://codesandbox.io/s/block-basic-usage-dny94

<img width="649" alt="Screen Shot 2020-02-18 at 10 47 12 AM" src="https://user-images.githubusercontent.com/6504944/74767621-0d20f300-523c-11ea-9780-c3a027adfe7b.png">

## Current Behavior

TypeScript error

## Expected Behavior

No TypeScript error

- [x] I have searched the [issues](https://github.com/uber/baseweb/issues) of this repository and believe that this is not a duplicate.

gergelyke:
I think this is by design? Currently, I don't think any component supports this

jonsadka:
Seems odd that this `className` wouldnt be a supported prop / attribute since it's available on nearly all JSX elements. 

I know it can be set as an override but this behavior is inconsistent with FlowJS which does not throw an error

jonsadka:
@gergelyke @chasestarr any further thoughts about this? 

Seems odd I'd get an error doing something like:
```javascript
import {useStyletron} from 'baseui';

const [css] = useStyletron();
<Block className={css({color: 'red'})}/>
```

chasestarr:
I believe it's just because it's left out of the TS props definition here https://github.com/uber/baseweb/blob/master/src/block/index.d.ts#L215 I can put up a diff later today

nick-khoury:
> I believe it's just because it's left out of the TS props definition here https://github.com/uber/baseweb/blob/master/src/block/index.d.ts#L215 I can put up a diff later today

can you expose the `id` prop as well?

jonsadka:
> > I believe it's just because it's left out of the TS props definition here https://github.com/uber/baseweb/blob/master/src/block/index.d.ts#L215 I can put up a diff later today
> 
> can you expose the `id` prop as well?

@nick-khoury if you update to the latest version of BaseWeb you should see the `id` prop supported (was part of https://github.com/uber/baseweb/pull/4081)

x-EricH-x:
Hi, sorry for reviving this thread, I notice `id` and `className` is not exposed on other components as well,
it is by design or just missing TS definition like this one?

PR #4081, opened 2021-02-19
feat(block): add id prop to block TS def

jonsadka:
Setting `id` on a `Block` element results in a TypeScript error:

Screen Shot 2021-02-19 at 9 37 41 AM

### Current Behavior

TypeScript error

### Expected Behavior

No TypeScript error

### Related issue for `className`
https://github.com/uber/baseweb/pull/3620
https://github.com/uber/baseweb/issues/2872

### DIFF
--- a/src/block/index.d.ts
+++ b/src/block/index.d.ts
@@ -271,6 +271,7 @@ export interface BlockProps {
   gridTemplateColumns?: Responsive<string>;
   /** available values: https://developer.mozilla.org/en-US/docs/Web/CSS/grid-template-rows */
   gridTemplateRows?: Responsive<string>;
+  id?: String;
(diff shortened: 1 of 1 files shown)


chasestarr (approved):
👍

nick-khoury:
I am getting `Property 'id' does not exist on type 'IntrinsicAttributes & ButtonProps & RefAttributes<HTMLButtonElement>'` on the `Button` element when I try setting the `id` prop.

jonsadka:
@nick-khoury this diff is was for the `Block` attribute. props that are missing on other elements would need to their respective definitions updated. In your case, this is the one for the `Button` component https://github.com/uber/baseweb/blob/master/src/button/index.d.ts

PR #4094, opened 2021-02-25
Publish 'next' tag with unique semver

stephencookdev:
#### Description

Attempt 3 to get https://github.com/uber/baseweb/pull/4089 working properly...! (Sorry for the multiple PRs, this is inherently hard to test end-to-end)

So, apparently my understanding of `npm publish --tag x` was incorrect. Apparently, it just performs the normal publish flow, but then _also_ tags the end result. This means if that package version already exists, `npm publish` will result in a 403

This PR follows the same logic that the alpha release follows, ensuring that we create a unique package version for the release before hitting `npm publish`

#### Scope
Patch: Bug Fix

### DIFF
--- a/scripts/publish-bleeding-edge.js
+++ b/scripts/publish-bleeding-edge.js
@@ -11,4 +11,7 @@ LICENSE file in the root directory of this source tree.
 // @flow
 
 const publishToNpm = require('./publish-to-npm.js');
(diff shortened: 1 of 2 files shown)


tajo (approved):

chasestarr (approved):
No worries, thanks!

PR #4099, opened 2021-03-01
feat(select): adds ignoreCase

### LINKED ISSUE #4097: Case-sensitivity prop for creatable multi Select
## Feature description

Select currently uses a [case-insensitive comparison against the current options and value](https://github.com/uber/baseweb/blob/v9.110.0/src/select/select-component.js#L867) to allow creating a new value. This makes it impossible to create both "abc" and "ABC" values.
Propose adding a `ignoreCase` (default true) prop to Select to toggle the .toLowerCase() behavior.

Even though default filterOptions also does a case-insensitive search, I don't think that needs to change based on this prop because there's already a prop to override the whole function.

ksnyder9801:
fixes #4097 

#### Description

Adds a case-sensitive comparison prop to Select. This allows adding multiple creatable values that only differ in case.

#### Scope
Minor: New Feature

### DIFF
--- a/documentation-site/components/yard/config/select.ts
+++ b/documentation-site/components/yard/config/select.ts
@@ -218,6 +218,13 @@ const SelectConfig: TConfig = {
       description:
         'Defines if currently selected options are filtered out in the dropdown options. Only for `multi` select.',
     },
+    ignoreCase: {
+      value: true,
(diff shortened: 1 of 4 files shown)


chasestarr (approved):
Thank you

PR #4100, opened 2021-03-02
feat(select): keep dropdown options highlighted

### LINKED ISSUE #4098: Automatically highlight first item in creatable multi Select
## Feature description
We are using Select with `creatable`, `multi` and empty `options` as a multiple-value entry. This works great except that pressing Enter to add the typed value "appears" to randomly do nothing (or worse, submit the parent `<form>` without the typed value).

For our use case, ideally, typing into Select's input box would always highlight either the last selected option (if it's still there) or the most relevant option, so hitting Enter would select it.

Two things prevent this and cause the unexpected behavior
1. onMouseLeave removes the highlight

- This is [Menu's](https://github.com/uber/baseweb/blob/v9.110.0/src/menu/stateful-container.js#L331) fault, and I think the behavior probably makes more sense for a "context menu" triggered on click, but not Select's dropdown.

2. "No results" removes the highlight
- This is also due to code in [Menu](https://github.com/uber/baseweb/blob/v9.110.0/src/menu/stateful-container.js#L117). [Repro](https://codesandbox.io/s/gmv9c?module=src/example.js&file=/src/example.js): Type abc, enter (works), type abcd, enter (doesn't work). The second time you type "abc", there are "No results", removing the highlight before you type "d". This is pretty common if you are entering several values with a shared prefix (numbers, identifiers...)

Propose adding a new prop `autoHighlightFirstItem` (to both Select and Menu) for this, unless you think that we can change the default behavior - then maybe only Menu needs the prop.

ksnyder9801:
Fixes #4098

#### Description

Adds a "internal" forceHighlight prop to StatefulMenu that Dropdown can use.
The prop disables the mouseLeave highlight removal and sets the highlight to the first item if there are 1 or more items but no highlight.

#### Scope
Patch: Bug Fix

### DIFF
--- a/src/menu/stateful-container.js
+++ b/src/menu/stateful-container.js
@@ -41,6 +41,7 @@ export default class MenuStatefulContainer extends React.Component<
     removeMenuFromNesting: () => {},
     getParentMenu: () => {},
     getChildMenu: () => {},
+    forceHighlight: false,
   };
 
   state: StatefulContainerStateT = {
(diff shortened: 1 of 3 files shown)


chasestarr (approved):
Tested and works well. Thank you

PR #4109, opened 2021-03-05
chore: test faster snowpack?

tajo:
`Fixes #1, Fixes #2`


adenta:
@tajo, are you trying to get baseweb to work in Snowpack? Biggest problem on my end is `create-snowpack-app` pins react to v17, which causes issues for baseweb.

tajo:
> @tajo, are you trying to get baseweb to work in Snowpack? Biggest problem on my end is create-snowpack-app pins react to v17, which causes issues for baseweb.

It already [works](https://baseweb.netlify.app). I was just testing some pre-release stuff in this PR. We are testing Base Web in React 17 in our CI, is there some more specific problem?

adenta:
Randomly while developing, I was getting weird, errors from styletron. Looking through my google search history, I kept coming back to [this](https://github.com/mui-org/material-ui/issues/9178) page, I remember the errors had looked something like: ` Uncaught TypeError: (0 , _styles2.default) is not a function ` , and by removing styletron, they went away.

So I guess not _directly_ any issues with baseweb, but styletron does not play nice.

adenta:
Ight, found the exact message: `Uncaught TypeError: (0 , _hyphenateStyleName2.default) is not a function`

This is coming from styletron, happens somewhat randomly, and it seems its just a matter of luck to get it to go away.

tajo:
@adenta Can you try to create a repro and maybe create an issue in styletron? We definitely want to look into that. Thanks!

PR #4123, opened 2021-03-15
fix(datepicker): pass order sharedprop to prev and next button (#4122)

### LINKED ISSUE #4122: Clicking on previous icon button is not working for second month when monthsShown > 1
Issue can be reproducible in the documentation by setting the monthsShown > 1.

## Current Behavior
Screenshot 2021-03-15 at 4 07 20 PM
Highlighted button is not working

## Expected Behavior
Behaviour is fine, as we are subtracting one month from that month and considering that month as a base month. But `order` prop is not passed to `PrevButton` and `NextButton`, consumer can use the order prop if want to hide that buttons based on order.

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v9.111.1  |
| React   |         |
| browser |         |

hardikmodi1:
Pass order as shared prop from calendar header to previous and next button

#### Description
Pass order as shared prop so that consumer can style the respective buttons based on prop. For example one may not want to show the previous button for 2nd month when monthsShown is equal to 2. For the purpose she/he can style that button using this $order prop.

Fixes #4122


chasestarr (approved):
thank you

Issue #4137, opened 2021-03-19
Input clear button is missing for number type input that uses a number value.


ajbogh:
When an Input is a number type and includes the clearable property to show the X button, and the value is numeric, then the X button is missing.

If the value is changed to a string then the X button returns.

https://codesandbox.io/s/base-web-input-forked-qu2u5?file=/src/example.js

## Current Behavior

The X button for clearable numeric inputs is missing when the value is a number.

## Expected Behavior

The X button should render no matter what the value type is.

## Your Environment

https://codesandbox.io/s/base-web-input-forked-qu2u5?file=/src/example.js

| Tech    | Version |
| ------- | ------- |
| browser |      Version 89.0.4389.90 (Official Build) (64-bit)   |
| OS |  Ubuntu 18.04 |

Using string "5"
![Screenshot_2021-03-19_12-48-39](https://user-images.githubusercontent.com/161761/111834900-7b0d7300-88b1-11eb-9e89-3d11fc8c975e.png)

Using number 5
![Screenshot_2021-03-19_12-48-24](https://user-images.githubusercontent.com/161761/111834904-7ba60980-88b1-11eb-8aaf-742bae7fc352.png)

- [ X ] I have searched the [issues](https://github.com/uber/baseweb/issues) of this repository and believe that this is not a duplicate.

PR #4138, opened 2021-03-19
fix(input): render clear when initial value is number

chasestarr:
fixes: https://github.com/uber/baseweb/issues/4137

#### Scope
Patch: Bug Fix


src/input/base-input.js
```diff
@@ -236,9 +236,16 @@ class BaseInput<T: EventTarget> extends React.Component<
 
   renderClear() {
     const {clearable, value, disabled, overrides = {}} = this.props;
-    if (!clearable || !value || !value.length || disabled) {
+    if (
+      disabled ||
+      !clearable ||
+      typeof value === null ||
+      typeof value === undefined ||
```
ajbogh, line 243:
``` 
typeof value === null ||
typeof value === undefined ||
```

This can be reduced to 

```
typeof value == null
```

PR #4143, opened 2021-03-23
fix(list): avoid strict height value, min-height

chasestarr:


chasestarr:
There's some issues with the VRTs. Will need to debug and ensure this doesn't cause issues with downstream applications

PR #4149, opened 2021-03-26
Adding Global Override Support

### LINKED ISSUE #106: [Theming] Allow component specific global theme overrides
There seems to be a need to override global styles per component due to design language differences and variations. Per component global overrides should be on a theme or the ThemeProvider level.

### Current Behavior
Currently, in order to override styles on a specific component across the whole application, you'd need to create and use a customized component that passes needed overrides to the BaseUI component.

### Potential Solutions
1. Let the `styled`function accept a third argument, that is a per component style override hook. A POC can be found [here](https://github.com/uber-web/baseui/commit/2bf7c99c3a2eba23682af1be44e9d0b9c590f28c).
   Some cons here:
   - Component overrides would be css declaration blocks or functions that return a css declaration block in order to support dynamic styles. As opposed to the theme that contains style values only and no css declarations.
2. Have optional component specific style values in the theme obj. [POC](https://github.com/uber-web/baseui/commit/411e18f7baf806b0c581063433d139292051f587).
   Cons:
   - The styled components code get's messy

tolgaouz:
Fixes #106

#### Description

As discussed in the issue, I've created a HOC to wrap components while exporting them. This allows components to get global overrides passed from `ThemeProvider` or `BaseProvider` object. 

There are couple things to discuss and which I need help with regarding this solution though.

- What should we use as name convention? There are some differences on docs and component names. e.g. `DragAndDropList` is used on component class but it's exported as `dnd-list` my preference would be to use `DragAndDropList` in this example since it's the class name.
- There are many problems when it comes to compatibility of the new HOC with Flow. Probably this is the part where I spent most time while working on this PR. The problem occurs when I'm using forwardRef inside the HOC. Flow complains about ref type being `mixed`, hence not compatible with `HTMLElement`. Even though I followed official docs for using Flow + React while using HOCs, I wasn't able to solve the issue with it. So just to pass the unit-tests and type-checks I put `//$FlowFixMe` to necessary lines.
- I didn't apply HOC to certain components like `data-table` and `flex-grid` since they are not supposed to accept overrides (I guess?) anyways. There wasn't a section for overrides in their docs, also `flex-grid` tests failed when I added override support to it because the nature of the test.
- I don't know how/if I should add more tests to this PR. Since most of the functionality this HOC uses is already tested (overrides and theme tests) I thought checking if this global override HOC passes overrides to children components would be enough. Please advise on that matter.

#### Scope
Minor: New Feature


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "baseui",
-  "version": "9.112.0",
+  "version": "9.111.2",
```
tajo, line 3: You should not update this.

tolgaouz: Just converted it back

airhorns:
It'd be really great to get this in so us downstream users can adjust things like the input border thickness and whatnot. Anything @tolgaouz can do to help get this where you folks want it to be?

tajo:
Hey, yea we will review/test it in our applications first, should get in next 2 weeks.

src/helpers/overrides.js
```diff
@@ -206,13 +211,53 @@ export function useOverrides(
...
+    constructor(props) {
+      super(props);
+    }
+
+    render() {
+      // Grab forwardedRef from props and pass it down
+      const {forwardedRef} = this.props;
+      // mergeOverrides from themeContext and props
+      const overrides = mergeOverrides(
+        this.context.overrides?.[componentName],
```
tajo, line 247: this causes a flow error https://buildkite.com/uberopensource/baseweb/builds/18428#9050979e-44d2-49d6-b921-3d25a5063830/128-132

src/styles/theme-provider.js
```diff
@@ -8,16 +8,23 @@ LICENSE file in the root directory of this source tree.
...
+export const ThemeContext: React.Context<ThemeProviderT> = React.createContext({
+  theme: LightTheme,
+});
 
-const ThemeProvider = (props: {theme: ThemeT, children: ?React.Node}) => {
-  const {theme, children} = props;
+const ThemeProvider = (props: {
+  theme: ThemeT,
+  children: ?React.Node,
+  overrides?: ThemeLevelOverridesT,
```
tajo, line 21: I am not a big fan of this type name. Maybe `componentOverrides?: ComponentGlobalOverridesT` is more descriptive?

tajo:
Looks great, thanks! Left some comments. Also

- Typescript types should be updated [here](https://github.com/uber/baseweb/blob/master/src/index.d.ts#L75).
- There should be also a test that makes sure a component level override takes precedence over a global override.
- `DragAndDropList` is fine. We'll need to create some documentation for this feature and maybe even a blog post.
- We don't use overrides in all components so it's fine not applying the HOC there.
- Ad flow types: I'll test it more in some applications before publishing but as long as it works and preserves current level of prop types it should be fine.

tajo:
I fixed the types for button this way:

before

```
//$FlowFixMe
export default withOverrides(ForwardedButton, 'Button');
```

after

```
export default withOverrides<ButtonPropsT, HTMLButtonElement>(
  ForwardedButton,
  'Button',
);
```

Can we just do the same for all other components to fix types everywhere?

src/helpers/base-provider.js
```diff
@@ -13,8 +13,10 @@ import type {BaseProviderPropsT} from './types.js';
 const BaseProvider = (props: BaseProviderPropsT) => {
   const {children, overrides, theme, zIndex} = props;
   return (
-    <LayersManager zIndex={zIndex} overrides={overrides}>
-      <ThemeProvider theme={theme}>{children}</ThemeProvider>
+    <LayersManager zIndex={zIndex}>
+      <ThemeProvider theme={theme} overrides={overrides}>
```
airhorns, line 17: This is an API change we'll have to document right? Previously the overrides were for the LayersManager?

src/helpers/__tests__/global-overrides.test.js
```diff
@@ -0,0 +1,138 @@
...
+      const ref = React.createRef();
+      const handleClick = () => {
+        if (!ref.current) return;
+        mockFn();
+        ref.current.textContent = 'called';
+      };
+      return (
+        // $FlowFixMe
+        <ThemeProvider
+          overrides={{
```
airhorns, line 113: What do you think about making the overrides a property of the theme instead of a prop passed to the ThemeProvider? That seems a bit more in line with the idea of the theme as a value that you can swap around -- you might have different overrides for the light theme vs the dark theme for example.

src/helpers/__tests__/global-overrides.test.js
```diff
@@ -0,0 +1,138 @@
...
+      >
+        {/* $FlowFixMe */}
+        <TestComponent />
+      </ThemeProvider>,
+    );
+    const testComponent = getByTestId(container, 'global-overrides');
+    expect(testComponent).toBeTruthy();
+    expect(testComponent).toHaveStyle('color: red');
+  });
+  test('passes global overrides through BaseProvider', () => {
```
airhorns, line 59: I think this needs a test that uses an actual baseui component instead of just components defined inline here, like setting the `borderWidth` for an `Input` component.

src/helpers/__tests__/global-overrides.test.js
tajo: >  you might have different overrides for the light theme vs the dark theme for example

I think the preferred way is using theme color values in your global overrides to do light vs dark.

tajo:
Closing this. Currently, it's really hard to have a proper typing for this feature without significant rewrites.

airhorns:
Rewrites of the code underneath or rewrites of this PR?

tajo:
> Rewrites of the code underneath or rewrites of this PR?

The code underneath. To get a good typing we would probably have to use hooks to access the context but most of our component are uses classes. Rewriting classes into functional components is a lot of work.

PR #4167, opened 2021-04-07
feat(combobox): adds inputRef to <Combobox/>

nick-mcdonald:
#### Description

Adds a ref to the Combobox's inner input. This is useful if you want to give focus to the Combobox or manipulate the inner Input control in other ways.

#### Scope

Minor: New Feature


chasestarr:
Thanks @nick-mcdonald looks like this will be all set once flow typing is fixed

nick-mcdonald:
Thanks @chasestarr, the flow typing should be fixed but please let me know if there's a better way to handle it

PR #4168, opened 2021-04-07
fix(radio): fix navigation in NVDA browse mode

chasestarr:
#### Description

Radio component cannot be operated on with NVDA using browse mode without tabbing to previous element, then tabbing back to the radio.

#### Scope
Patch: Bug Fix


src/radio/radio.js
```diff
@@ -117,51 +117,54 @@ class Radio extends React.Component<RadioPropsT, RadioStateT> {
...
+
           <Input
-            aria-invalid={this.props.error || this.props.isError || null}
+            // aria-invalid={this.props.error || this.props.isError || null}
             checked={this.props.checked}
-            disabled={this.props.disabled}
-            name={this.props.name}
-            onBlur={this.props.onBlur}
-            onFocus={this.props.onFocus}
+            // disabled={this.props.disabled}
```
tajo, line 146: Should we rather just delete instead of commenting?

chasestarr: Sorry for not being clear on this diff, I'm pushing here so that I can test on my windows computer. Attempting to remove all of the extra stuff from the elements to isolate the issue

PR #4183, opened 2021-04-16
feat(timepicker): add support for min/max times

### LINKED ISSUE #4130: TimePicker should include a minTime and maxTime feature, similar to DatePicker's options
TimePicker should include minTime and maxTime properties that take a Date object as input.

## Feature description

The minTime and maxTime properties should take a Date object as input and disable any time not equal to or between the values. 

Both minTime and maxTime should be optional.

The code doesn't need to validate minTime and maxTime besides the Date input type (typically handled by Flowtype and Typescript). By the names, the 3rd party developer should know that minTime must be before maxTime, but if maxTime (by some fluke) ends up before minTime (or vice versa) then the code shouldn't prevent that situation. In that case the UI should disable all options.

williamernest:
Fixes #4130

#### Description

Adds support for a minTime and maxTime to be provided for the time picker. When provided, any times (steps) that fall outside of the min/max range will not be rendered. 

#### Scope
Minor: New Feature


chasestarr (approved):
Code looks good and the feature works well. May want to consider adding an end time assertion.

src/timepicker/types.js
```diff
@@ -35,6 +35,8 @@ export type TimePickerPropsT<T = Date> = {
   nullable?: boolean,
   placeholder?: string,
   size?: SizeT,
+  minTime?: T,
+  maxTime?: T,
```
chasestarr, line 39: Can you please add this to the typescript types as well? https://github.com/uber/baseweb/blob/master/src/timepicker/index.d.ts

PR #4191, opened 2021-04-20
fix(tooltip): set focus related props to false by default

### LINKED ISSUE #3840: Interactive elements inside of a StatefulTooltip do not receive focus
## Current Behavior

Inside Modal, interactive elements wrapped inside StatefulTooltip is not getting focused while navigating via tab.
[Issue Demo](https://stackblitz.com/edit/react-hykhyi?file=package.json)

## Expected Behavior

Inside modal, every interactive element should get focused when navigating with tab.

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v9.102.0  |
| React   |    v16.14.0     |
| browser |     Chrome    |

williamernest:
Fixes #3840

#### Description
Change defaults to false, since they're boolean attributes. 

#### Scope
Patch: Bug Fix


chasestarr (approved):
Looks like this fixes the reported issue. May want to verify behavior when using a screenreader. Does it read the content?

williamernest:
> Looks like this fixes the reported issue. May want to verify behavior when using a screenreader. Does it read the content?

The screenreader still works for tooltips to the same extent it did before (`aria-describedby` is added to the focused element and the `role` is set on the tooltip). There is some inconsistent behavior with it correctly reading the tooltip in some instances (shift+tab back from the aside and it fails to read tooltips, applies to all popovers afaict), but those issues exist on master as well.

williamernest:
Looks like shift+tab may have been a chrome vox issues. Voice over correctly reads tooltips and popover content if shown.

chasestarr:
👍 thanks for checking

PR #4043, opened 2021-01-26
feat(table): reverse table sorting icons

andriuss:
`Fixes #3962`

#### Description

Reverse table sort icon to make it more consistent with the industry.
More details can be found in  #3962
The change includes all BaseWeb tables - `semantic table`, `data table`, `grid table` and `flex table`.

table

#### Scope
Minor: New Feature

PR #4200, opened 2021-04-21
fix(data-table): initialSortOrder value

andriuss:
#### Description

Fix `initialSortOrder` property by inverting its meaning.
A case is covered by E2E tests - `src/data-table/__tests__/data-table-initial-sort.e2e.js`

Match sort icons meaning to actual sort order which was mismatched after changes in - #4043

#### Scope
Bug Fix


chasestarr (approved):
Thank you

PR #4203, opened 2021-04-25
docs(combobox): add popover overrides to combobox

matteo-hertel:
#### Description

This PR updates the `Combobox` documentation to the reflect the fact that overrides are available for the `Popover` component.

There is no functionality change, it's just documenting an undocumented feature

The change of `componentName` is not ideal as it's mutating the object but the default value exported is `StatefulPopover` and won't work if left unchanged, I'm more than happy to update the PR with some guidance

Screenshot 2021-04-25 at 22 20 18
Screenshot 2021-04-25 at 22 20 25

#### Scope
Patch: Docs Fix


chasestarr (approved):
Thank You

PR #4214, opened 2021-05-04
Datepicker style changes

williamernest:
Adjusts the styles to match new version of Datepicker component in figma. 
Adds support for a label below individual dates.


tajo (approved):
dateLabel looking good

PR #4248, opened 2021-05-13
fix(data-table): ensure row is defined before showing its actions

konradkalemba:
#### Description

Fixes a bug in Data Table component where row's actions remain visible after the row is deleted. 

Base Web - Examples

#### Scope
Patch: Bug Fix


chasestarr (approved):
👍

PR #4251, opened 2021-05-15
fix(rating): use semantic theming for rating

akshatnitd:
#### Description
- Use semantic theming for rating
- Update related docs


chasestarr:
Thank you very much for the contribution. This is unfortunately going to break some applications where the rating400 is being customized. We're going to keep the code as-is

akshatnitd:
@chasestarr As part of this PR, can we atleast add theming support for the inactive state of rating/emoticon and the star/emoticon's stroke color? Since they are not supported as of now.

And also, should we instead mark the `rating400`/`rating200` move to the deprecated fields as part of the eslint plugin?

chasestarr:
Those proposed changes sound good to me

akshatnitd:
@chasestarr Made changes as discussed above.

akshatnitd:
The snapshots have changed as, before the PR, the inactive state color for the star and emoticon was different, but now it has been made the same for consistency.

akshatnitd:
@chasestarr If this looks good, can we merge it?

chasestarr:
> @chasestarr If this looks good, can we merge it?

Sorry for the delay. I just fixed some testing in our internal JS code so that this change can be tested in visual regression tests before merging

chasestarr (approved):
Thank you

PR #4252, opened 2021-05-16
fix(popover): fix onClickOutside event handler for web component

### LINKED ISSUE #4181: [popover] The "onClickOutside" event does not work properly in a web component
The `onClickOutside` event being fired when user clicks on a popover target or a content if the popover used inside a web component.
[Demo](https://codesandbox.io/s/base-web-popover-stateless-example-forked-8rxdf?file=/src/example.js)

The reason why it happens is [the event target](https://github.com/uber/baseweb/blob/v9.112.0/src/popover/popover.js#L254) of the `onDocumentClick` event. Such approach works only when the root element is `document`, for a popover which is rendered in a web component `event.target` returns whole web component host element. The solution which will work for web components as well is `event.composedPath`:
```js
const target = event.composedPath()[0];
```

## Expected Behavior
The `onClickOutside` should not be fired when user clicks on a popover target or a content 

## Your Environment
| Tech    | Version |
| ------- | ------- |
| Base UI | latest |
| React   |  latest       |
| browser |   any      |

tusharmoraye:
Fixes #4181 

#### Description

I have fixed the firing of `onClickOutside` when the user clicks on a popover target or a content in a web component with the approach mentioned in the issue. I have tested this in a web component.

#### Scope
Patch: Bug Fix

### DIFF
--- a/src/popover/popover.js
+++ b/src/popover/popover.js
@@ -251,7 +251,8 @@ class Popover extends React.Component<PopoverPropsT, PopoverPrivateStateT> {
   };
 
   onDocumentClick = (evt: MouseEvent) => {
-    const target = evt.target;
+    //$FlowFixMe
+    const target = evt.composedPath ? evt.composedPath()[0] : evt.target;
     const popper = this.popperRef.current;
     const anchor = this.anchorRef.current;
     // Ignore document click if it came from popover or anchor


chasestarr (approved):
👍

PR #4260, opened 2021-05-18
Global overrides

tajo:
based on https://github.com/uber/baseweb/pull/4149

this might be impossible to flow type in a reasonable way :/


vasdhara:
@tajo I would like to work from this branch and try and get the PR approved by updating all the flow typing for the global overrides issue fix. Just adding a comment here to reach out and check if you have any sugesstions for this. Be great to hear your thoughts on the same.

tajo:
The feature works but we didn't find a reasonable way to keep flow types working, since wrapping everything with na HOC makes things messy. Not breaking the current types is the priority.

Issue #4277, opened 2021-05-25
[popover], [select], [modal], [combobox], [datepicker] don't work properly inside of shadow DOM


gevgeny:
A few components which rendered in a separate Layer now don't work properly inside of shadow DOM. The reason is that `event.target` is not always event target when handling event outside of shadow DOM, which may cause incorrect behaviour of the "click outside" functionality. `event.composedPath()` may be used to fix the problem. [Here is a good guide](https://pm.dartus.fr/blog/a-complete-guide-on-shadow-dom-and-event-propagation) on event propagation with shadow DOM for more details.

Here is the components I have faced the problem with:

- Popovover (fixed in https://github.com/uber/baseweb/pull/4252 and described in https://github.com/uber/baseweb/issues/4181)
- Select
- Modal
- Combobox
- Datepicker

PR #4278, opened 2021-05-25
Shadow DOM support

gevgeny:
Fixes https://github.com/uber/baseweb/issues/4277

Additionally added:
- e2e tests for shadow DOM cases of the updated components
- e2e test utils for shadow DOM testing 
- Warning message when the library used inside of shadow DOM with `mode="closed"` since in this cause even `event.composedPath()` will not fetch a proper event target

#### Scope
Patch: Bug Fix


gevgeny:
@chasestarr @tajo @williamernest Could you please help me to run build, so I can verify all the checks and tests?

gevgeny:
@williamernest Could you please help me with tests? The all are passed locally, but fails on this CI 😐

gevgeny:
@williamernest Looks like the merge didn't help, do you have any ideas why test may not pass?

chasestarr:
Can you please try running locally again? We've made some changes recently to align the puppeteer versions between local and CI. I'm not familiar with shadow dom, so not able to help out very effectively.

gevgeny:
@chasestarr Yes, now the tests fail locally too. I also see that now e2e tests run on react v16 while before your update they run on react v17. And looks this is the reason of the problem, react works properly with shadow DOM only starting from v17 (see https://reactjs.org/blog/2020/08/10/react-v17-rc.html#changes-to-event-delegation) and in order to baseUI's mouse events work within shadow DOM user needs to run baseui with react v17. 
Does baseUI support react 17 officially?

gevgeny:
@chasestarr What about adding such util into `helpers/index.js`:
```js
...

const describeIfReactGreaterThan16 = React.version.startsWith('17')
  ? describe.skip
  : describe;
```

and then in `select-shadow-dom.e2e.js`:
```js
const { describeIfReactGreaterThan16 } = require('../../../e2e/helpers/index.js');

describeIfReactGreaterThan16('select in shadow DOM', () => {
  it('should pass', () => { ... }
});
```

chasestarr:
That sounds fine to me

gevgeny:
@chasestarr Added a condition. Let's try to build this one.

chasestarr:
Sorry to say, but we're going to decline adding broad support for shadow-dom. We may consider landing this change once our company is updated to react v17 but for the time-being we'd like to focus on the plain dom support. Apologies again for not reaching this decision sooner before you wrote the code.

gevgeny:
@chasestarr What If I remove all the tests and leave only `event.target` -> `event.composedPath()[0]`replacement? It is fully back compatible and all the old tests pass for both react16 and react17.

PR #4300, opened 2021-06-03
chore(vrt): reduce timeout, update erroneous screenshots

williamernest:
Ran 5 consecutive builds with no flakes.
https://buildkite.com/uberopensource/baseweb/builds?branch=tweak_timeouts

Updates incorrect screenshots and disables puppeteer option to capture beyond the viewport which causes some inconsistencies with rendering. The screenshots now should match what the developer sees when viewing locally. 

Reduces the timeout to 100ms.


vrt/tests.vrt.js
```diff
@@ -76,7 +76,7 @@ describe('visual snapshot tests', () => {
 });
 
 async function snapshot(identifier, viewport = VIEWPORT.desktop) {
-  const image = await page.screenshot();
+  const image = await page.screenshot({captureBeyondViewport: false});
```
chasestarr, line 79: Just wondering, what inconsistency does this solve? Would this be an issue with playwright as well? Looks like there's no equivalent flag.

chasestarr (approved):

vrt/tests.vrt.js
williamernest: Related issue: https://github.com/puppeteer/puppeteer/issues/6985

In our project it looks like the page refreshes and the elements are initially rendered incorrectly and the screenshot is taken before the components get updated.  

Playwright doesn't appear to have the same issue, so unless we see any issues come up I would think we'd be fine.

chasestarr: 👍 Nice job getting to the bottom of these flakes

PR #4315, opened 2021-06-09
feat(table-semantic): add dividers and cell spacing

chasestarr:
#### Description

Adds two new props for configuring table styles. First is `divider` for specifying how cell dividers should be rendered. Second is `size` which controls the space afforded to cell padding.

dividers:
Screen Shot 2021-06-09 at 1 36 31 PM

sizes:
Screen Shot 2021-06-09 at 1 36 41 PM

#### Scope
Minor: New Feature


williamernest (approved):

cmelone:
Hi there, 

I'm using the `TableBuilder` component in one of my apps, which works great, but after updating to a new version of the package, the table dividers are different, I assume as a result of this change. I saw that both the dividers and size are modifiable for the semantic table through props, but that doesn't seem to be the case for `TableBuilder` (which is an abstraction on the semantic table, correct?). Any suggestions on how to resolve? 

Cheers

Deliaz:
@cmelone I have the same issue. Seems to be a breaking change

PR #4321, opened 2021-06-11
fix(avatar): render avatar on client side to avoid broken images (#4306)

### LINKED ISSUE #4306: onError is not called in avatar
onError is not called in avatar component in some case. Reproducible in docs refresh the page 1-2 times.
Screenshot 2021-06-07 at 9 11 41 PM

## Current Behavior
When image does not exist, avatar component show the initials by setting the state in onError handler of image. But in some cases this handler is not called and avatar gets stuck with broken image and alt text

## Expected Behavior
It should show the initials in every case.

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v9.116.0  |
| React   |         |
| browser |         |

hardikmodi1:
Fixes #4306 

#### Description
onError does not work with server side rendering, so it renders the broken images some times instead of initials.

Before 
Screenshot 2021-06-11 at 4 29 35 PM

After
Screenshot 2021-06-11 at 4 30 01 PM

### DIFF
--- a/src/avatar/avatar.js
+++ b/src/avatar/avatar.js
@@ -32,9 +32,39 @@ export default class Avatar extends React.Component<PropsT, StateT> {
     size: 'scale1000',
   };
 
+  active: boolean;
+
   constructor(props: PropsT) {
     super(props);
     this.state = {noImageAvailable: !this.props.src};
+  }
+
+  componentDidMount() {
+    // using this approach instead of attaching onError to image to make sure that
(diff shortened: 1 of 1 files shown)


hardikmodi1:
Checked on Safari too. It is working as expected.

chasestarr:
Thank you for starting this PR. I've taken a different direction here: https://github.com/uber/baseweb/pull/4326 so will close this one

PR #4340, opened 2021-06-21
fix(docs): explicitly scroll to cheat sheet entry

chasestarr:
Clicking an 'exports' list item would navigate to the cheat sheet page, but in many situations scrolling to an element based on the url hash was not working. This change now explicitly scrolls to the element and does not rely on browser behavior


williamernest:
I'm seeing an error when navigating to the cheat-sheet with this change
image

chasestarr:
Thank you, only tested my local version

chasestarr:
Looking into fixing the flakey snapshot in this PR

tajo (approved):

PR #4352, opened 2021-06-23
docs(theme): update theming.mdx overrides example

zhongeric:
Update example for Button overrides to use BaseButton instead of Base per https://baseweb.design/components/button/

#### Description

Was looking through theming docs and noticed the example given for overrides for <Button> was `Base` when `BaseButton` should be used instead.

#### Scope
Patch: Bug Fix


williamernest (approved):

williamernest:
Thanks for the contribution.

PR #4353, opened 2021-06-25
Radius updates

tajo:


houmark:
I personally think these types of major design changes should also be major version updates? The card design will change a lot going from no radius to a fairly large radius. I know this can be overridden, but some people may miss that if this is released as a minor update.

tajo:
> I personally think these types of major design changes should also be major version updates? The card design will change a lot going from no radius to a fairly large radius. I know this can be overridden, but some people may miss that if this is released as a minor update.

Yes, we'll probably make a few more components rounder as well and release this change as a major release.

houmark:
> Yes, we'll probably make a few more components rounder as well and release this change as a major release.

Major would be v10?

williamernest (approved):

houmark:
I don't think we will ever go as aggressive on the border-radius as Uber seems to be pushing towards. One of the things we really liked about BaseWeb was that things were more edged/square than other open-source libraries.

Overriding on all the components that will be rounded, would introduce a lot of code, so maybe the Baseweb team would consider adding an overall template configuration for "rounding"? This way, changing one theme configuration, all components would round accordingly?

tajo:
> I don't think we will ever go as aggressive on the border-radius as Uber seems to be pushing towards.

For now we are adjusting just a limited set of components related to surfaces. For example, there are no changes planned to buttons or inputs.

> maybe the Baseweb team would consider adding an overall template configuration for "rounding"

The changes are pretty minor right now. There is some vision of a system that has various levels of roundness based on nesting. When that pans out, we might add some more centralized override solution.

houmark:
I don't feel a rounded `Modal` with 16px border-radius is a minor visual change, coming from 0px border-radius. `Card` is similar.

houmark:
Hey guys,

Sorry for continuing on this one, but maybe a middle way could be found here?

For each of these components a SHAPE could be introduced just like in the list. It could default to ROUND but if set to SQUARE or similar, it will render without the border-radius.

Or, alternatively, instead of adding the border-radius directly to the theme as `$theme.borders.radius500` an intermediate variable is added for each component (since all radius are the same on one component, only one variable is needed). This would enable us to override the radius directly in `createTheme` by overriding that variable.

As is, we need to add an override on every single component that loads the override for each of the radiuses, and that's quite a lot of work, and also not very pretty.

What do you think?

PR #4356, opened 2021-06-28
chore(checkbox): add react-hook-form scenarios

lhbrennan:
Related to WPT-7537

#### Description
Adding scenarios for using checkbox with react-hook-form

#### Scope
chore


chasestarr (approved):
Not sure what's wrong with codesandbox but the rest looks good

PR #4357, opened 2021-06-29
Cancel event in Link disabled

apt-exploit:
This allows choosing the Link component and disabling the button to cancel the link event.

`Fixes #1, Fixes #2`


chasestarr:
Hi @apt-exploit you may be able to fix the behavior you're looking for by using overrides https://baseweb.design/guides/understanding-overrides/

We cannot add this css rule to button for all users. Something like below may work for you

```js
<Button
  overrides={{
    BaseButton: {
      style: {
        pointerEvents: 'none',
      }
    }
  }}
/>
```

PR #4360, opened 2021-06-30
fix(color-tokens): update colors for better contrast

chasestarr:
#### Description
Fixes color contrast due to an error in the tool we were using to measure it

In light mode:
contentPositive (currently green400) needs to change to green500
backgroundPositive (currently green400) needs to change to green500
contentWarning (currently yellow500) needs to change to yellow700

In dark mode:
contentOnColor (currently flips to gray100) needs to change to white
contentOnColorInverse (currently flips to gray900) needs to change black

#### Scope
Patch: Bug Fix


williamernest:
The color change from green400 -> green500 causes the date picker to fail WCAG AA standards
image

chasestarr:
Yeah I noticed that as well when reviewing the snapshots, but that story is meant to validate some overrides are working correctly

williamernest (approved):

PR #4365, opened 2021-06-30
fix(button-group): prevent overriding isSelected on children within button group 

jonmcclanahan:
`Fixes #1, Fixes #2` 

#### Description

- Currently ButtonGroup clones the child and overrides isSelected. This should likely not be done, as there are cases in which respecting isSelected at the child level makes a lot of sense. For instance, let's assume we have an array of objects that define a group of buttons. We only want to mark a button as selected if one of those attributes matches a string. With the current setup we would need to loop through the array of objects to find the index of the matching value and then put it into the array that is past down to ButtonGroup (on top of looping through to render the respective elements). In my opinion, the far simpler and flexible approach would be to allow users to override the default selected behavior by respecting the isSelected prop if past.


williamernest:
Can you sign the CLA ?

jonmcclanahan:
> Can you sign the CLA ?

Sure, I just signed it

chasestarr (approved):
Thank you for contributing

PR #4402, opened 2021-07-16
fix(select): open non-searchable select when label clicked

chasestarr:
#### Description

Non-searcable select never renders an input element. Elements associated with a label 'for' attribute will only focus on label-click when the element is '[labelable](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#attr-for)'. Labelable elements are essentially a set of form elements like input, textarea, etc. Div elements are not labelable. The fix here is to add a hidden input element used to capture the focus event.

Earlier broken behavior: https://stackblitz.com/edit/react-rrtinv?file=src/example.js

#### Scope
Patch: Bug Fix


williamernest (approved):

shannonmoeller:
This change broke the ability to click on `<Select>` components when using a native `<label>` as a container. If you click the `<label>`, the dropdown opens. If you click the `<Select>`, nothing happens.

https://stackblitz.com/edit/react-rrtinv-qdtswm?file=src%2Fexample.js

williamernest:
In this instance both the wrapping label and the select are catching the click and triggering the dropdown, which causes it to open and then close in rapid succession.

shannonmoeller:
Found a temporary workaround by putting a click handler on the `<label>` which seems to confirm that theory, @williamernest:

```js
function handleClick(event) {
  if (event.target.closest('[data-baseweb="select"]')) {
    event.preventDefault();
  }
}
```

williamernest:
Will a FormControl component and an id value not work for your use case? Since our select click handler is on a div, it will always interpret 2 clicks when wrapped by a label. To get around that we've been using the for attribute, which this bug fix was related to .

PR #4407, opened 2021-07-20
feat(card): card title to respect HeadingLevel

### LINKED ISSUE #3470: [Card] title ignores HeadingLevel and renders as <h1> all the time
Hello. First of all, thank you for the base web!

## Current Behavior
Example code:

```
<HeadingLevel>
  <Heading>Header H1</Heading>
  <HeadingLevel>
    <Heading>Header H2</Heading>
    <Card title="Card title still H1">
      <StyledBody>
            Proin ut dui sed metus pharetra hend rerit vel non
            mi. Nulla ornare faucibus ex, non facilisis nisl.
            Proin ut dui sed metus pharetra hend rerit vel non
            mi. Nulla ornare faucibus ex, non facilisis nisl.
      </StyledBody>
    </Card>
  </HeadingLevel>
</HeadingLevel>
```
`<Card title="This is card title">` renders as `<h1>` anyway.

Stackblitz to reproduce [Link]( https://stackblitz.com/edit/react-yd3ejd?file=src%2Fexample.js)

## Expected Behavior

`<Card title="This is card title">` renders as `<h2>`.

Moreover, I suggest adding `<StyledHeader>` like `<StyledBody>` and `<StyledAction>`. It would be more intuitive instead of the title prop. And maybe card title shouldn't be rendered as a header element at all, render a `<div>` like `StyledBody` and let the developer decide what to put there `<Label1>` or `<Heading>`.

All the best, Sergei

lhbrennan:
Fixes #3470

#### Description
Previously the `Card` title was always an `h1`. Now it uses the heading tag that matches the `HeadingLevel`, using `h1` as a default.

#### Scope
Patch: Bug fix


src/card/__tests__/card-header-level.scenario.js
```diff
@@ -0,0 +1,42 @@
+/*
```
williamernest, line 1: This test won't catch a regression since we're trying to validate that the semantically correct element is used, but the styling doesn't change. A unit test would more appropriately catch a regression with this change.

williamernest (commented):
You'll need to run the component size update script to update the component sizes after a clean build. `scripts/component-sizes.js`

williamernest:
I don't think this is a breaking change, it should probably be just a bug fix since it's just ensuring the card header is used the correct html element.

lhbrennan:
> I don't think this is a breaking change, it should probably be just a bug fix since it's just ensuring the card header is using the correct html element.

Ok that's easier. I just reread [Hyrum's Law](https://www.hyrumslaw.com/) and worried something might still somehow break for someone.

williamernest (approved):

PR #4419, opened 2021-07-23
feat(layout-grid): add compact type to layout grid

phou87:
Adds the "compact" version of the grid

#### Description
This adds a new prop, gridStyle, which currently has two possible values.
"default", which is also the default value if the prop is not specified, will take the grid values from the theme as before.
"compact" will ignore the grid values in the theme and instead use a hardcoded set of values representative of the compact version, although the values can still be overridden by props passed directly to Grid and Cell.

#### Scope
Minor: New Feature


chasestarr (approved):
Thank you @phou87 LGTM

PR #4423, opened 2021-07-27
fix(combobox): change options type to an array

ChoJongHoon:
#### Description

I found that the `options` type was `OptionT`, but `options` should be an array. So I changed this type and opened a pull request.

#### Scope

Patch: Bug Fix


chasestarr (approved):
Thank you

PR #4437, opened 2021-08-04
fix(datepicker): allow empty placeholder

Deliaz:
#### Description

This PR implements the ability to use an empty string as a placeholder in the datepicker. 

#### Scope
Patch: Bug Fix


williamernest:
Thanks for the contribution.

PR #4442, opened 2021-08-08
Global overrides flow fixes

### LINKED ISSUE #106: [Theming] Allow component specific global theme overrides

vasdhara:
Fixes #106 

#### Description

This branch is based on the PRs #4149 and #4260 . Have tried to fix all flow errors resulting from the global overrides feature addition.

- Components have been fixed using [Config](https://flow.org/en/docs/react/types/#toc-react-config)  to type the default props
- any instead of mixed has been used in combobox because flow throws error with the type declaration for mapOptionToString. Setting as ReadOnly does not work, so any was used.
- Components with generic types in props have any used as well in declaring the config. mixed doesnt work with generic types either.
- some intersection types have been converted to exact spreads ba on documentation provided here. [exact object spread](https://flow.org/en/docs/types/objects/#toc-exact-object-types)
#### Scope
Minor: New Feature


airhorns:
@tajo does this approach seem viable to you?

airhorns:
@tajo (maybe also @chasestarr) we discussed this in #106 as a viable approach for setting properties like input border width without having to add new theme properties -- it'd be really great to get this in and as you can see it touches a lot of stuff so the merge conflict list keeps growing and growing. Can you help us understand if this makes sense to you and what else we might need to do to get it in? Thanks!!

chasestarr:
Hi, this feature is not something we plan on investing time into. I'll go ahead and close the PR to make that clear. Thank you all for exploring the potential.

PR #4464, opened 2021-08-17
Fix startEnhancer & endEnhancer zero render

Deliaz:
#### Description
Since React renders everything with the type of number, there is an unexpected behavior when passing `0` into `startEnhancer` or `endEnhancer`. 
The issue leads to zero number being showed as an unstyled element in the following components: `Input`, `Tag` and `ListItem`. 

Consider the screenshots: 

---
The fix for Tag and ListItem would be to do an explicit check. 
However, for the Input enhancers, since it supports React.Nodes, the logic can be extended to show zero number as correct enhancer: 

**After fix:** 

#### Scope
Patch: Bug Fix


Deliaz:
Hey Team, 
Something is stuck in the build phase. Can you help me to investigate please?  
Thanks

chasestarr (approved):
Thank you for the contribution

PR #4466, opened 2021-08-17
fix(select): handle open click when inside native label

chasestarr:
#### Description

Fixes bug where `Select` would not open listbox on click when inside a native `label` element. For reference, see reporting comment here: https://github.com/uber/baseweb/pull/4402#issuecomment-899857492

#### Scope

Patch: Bug Fix


src/select/__tests__/select-searchable-form-control.scenario.js
```diff
@@ -48,6 +50,32 @@ export default function Scenario() {
...
+          value={nativeNotSearchable}
+          options={options}
+          onChange={params => setNativeNotSearchable(params.value)}
+        />
+      </label>
+
+      <label htmlFor="searchable-native-label">
+        Native searchable
+        <Select
+          id="searchable-native-label"
```
shannonmoeller, line 70: Could we include a test that does not specify `id` and `htmlFor` attributes? One of the benefits of wrapping form elements in `<label>` is that the browser automatically wires up the connection. I realize that's not going to work for every complex form component, but it's a valid option for `<Select>` after the change in #4402.

williamernest (approved):
Thanks for the quick turnaround!

shannonmoeller:
Thank you!

PR #4469, opened 2021-08-18
fix(block): Add background related props for Block Components

Losses:
#### Description

The document said there are several props about background for block element, but they are not available on the type definition. This PR fixed this problem.

#### Scope
Patch: Bug Fix


chasestarr (approved):
Thank you

PR #4490, opened 2021-08-30
fix(datepicker): add prop to always render 6 rows in datepicker (#4465)

### LINKED ISSUE #4465: Prevent height change in date picker
Need a way to avoid jumping behaviour in datepicker

## Current Behavior
Currently, when we navigate between months in datepicker, it causes heights to change. This is because some months only needs 5 rows and some 6. 

## Expected Behavior
We can add a support to always render 6 rows. I checked the implementation, where we only render a weeks till `weekStart` is within month. We may consider to add a prop based on which we will always render 6 rows to stop the height changing while navigating.

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v?.?.?  |
| React   |         |
| browser |         |

hardikmodi1:
Fixes #4465 

#### Description
Added prop named `fixedHeight` in datepicker to always render 6 rows in calendar

#### Scope
Minor: New Feature

### DIFF
--- a/src/datepicker/types.js
+++ b/src/datepicker/types.js
@@ -116,7 +116,9 @@ export type WeekPropsT<T = Date> = {
   value: ?T | Array<T>,
 };
 
-export type MonthPropsT<T = Date> = WeekPropsT<T>;
+export type MonthPropsT<T = Date> = WeekPropsT<T> & {
+  fixedHeight?: boolean,
+};
 
 export type CalendarInternalState<T = Date> = {
   highlightedDate: T,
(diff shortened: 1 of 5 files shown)


src/datepicker/types.js
```diff
@@ -116,7 +116,9 @@ export type WeekPropsT<T = Date> = {
   value: ?T | Array<T>,
 };
 
-export type MonthPropsT<T = Date> = WeekPropsT<T>;
+export type MonthPropsT<T = Date> = WeekPropsT<T> & {
+  fixedHeight?: boolean,
```
williamernest, line 120: Can you add this prop to the `index.d.ts` file as well?

williamernest (commented):
Can you add a story test for this, just to catch any regressions in the future? Other than that I think this looks good to go.

williamernest (approved):

PR #4496, opened 2021-09-02
feat(list): add new ListHeading component

lhbrennan:
#### Description

Build a new `ListHeading` component to complement the `ListItem` component.

**Specifications:**

- Four sections: Heading (upper left), Subheading(lower left), End Enhancer (upper right), End Enhancer Description (lower right)
- When Heading is one line and there is no Subheading:
    - The heading top padding is 24 from the top of the heading text
    - bottom padding is 8 from the bottom of the heading
- When Heading is two lines or there is a Subheading
    - The heading top padding is 16 from the top of the heading text
    - bottom padding is 8 from the bottom of the subheading
- When endEnhancer is text content, it should be bottom aligned with heading, and the (optional) endEnhancer description text should be top-aligned with the subheading.
- When endEnhancer is not text content (e.g. button, icon, etc), it should be center-aligned with the heading (regardless of whether there is a subheading).
- No endEnhancerDescription will be rendered when endEnhancer is not text
- Heading can be either 1 or 2 lines (determined via maxLines prop)
- End Enhancer and End Enhancer description always truncated after just 1 line
- Minimum padding/margin between left and right sides is 16px

image
image

#### Scope
Minor: New Feature


williamernest (approved):
LGTM. Not really a nit on your changes but the file seems inconsistent with regards to using static pixel values or theme sizing. Might be a good time to make it more consistent (or at least documented why it's inconsistent).

src/list/list-heading.js
```diff
@@ -0,0 +1,134 @@
...
+  StyledHeadingContent,
+  StyledHeadingContentRow,
+  StyledHeadingEndEnhancerContainer,
+  StyledHeadingEndEnhancerDescriptionContainer,
+  StyledHeadingMainHeading,
+  StyledHeadingSubHeading,
+} from './styled-components.js';
+import type {HeadingPropsT} from './types.js';
+
+function RenderNode(props) {
```
chasestarr, line 25: 👍 This node | abstractcomponent approach sgtm. Maybe pull this out into a shared utility in a follow up PR

chasestarr (approved):

lhbrennan:
> LGTM. Not really a nit on your changes but the file seems inconsistent with regards to using static pixel values or theme sizing. Might be a good time to make it more consistent (or at least documented why it's inconsistent).

Thanks, resolved by using theme sizing throughout.

PR #4502, opened 2021-09-03
fix(layout-grid): better checks for custom grid props

phou87:
Use a better check for when custom grid props are passed in.

#### Description

Responsive props allow for single values, even outside of an array (e.g. gridMargins={0}), so if a falsey value was passed it would be ignored. This PR fixes the check to allow falsey values such as 0 for grid props.

#### Scope
Patch: Bug Fix


chasestarr:
Thank you. Alpha tests look good

PR #4504, opened 2021-09-06
Remove duplicate import from Snackbar types

simPod:
#### Description

The removed import seems to be duplicate of the above one


williamernest (commented):
Removed line is not an import

simPod:
olol

PR #4506, opened 2021-09-08
fix(slider): update handle color and size

lhbrennan:
#### Description

Update the slider so that it is up to spec:
- increase size of thumb/handle from 24px to 28px
- thumb should be white when enabled but not active
- thumb should be dark (theme.primaryA) when active/pressed

#### Scope
Patch: Bug Fix


lhbrennan:
Design is reconsidering these changes. Closing this for now.

PR #4509, opened 2021-09-10
fix(button-group): add example for responsive styling

### LINKED ISSUE #3448: ButtonGroup responsive styling
ButtonGroup components do not line wrap - they should line wrap responsively if the container is smaller than the group.

## Current Behavior
buttons do not wrap to next line

## Expected Behavior
buttons should be pushed onto next line

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v?.?.?  |
| React   |         |
| browser |         |

- [ ] I have searched the [issues](https://github.com/uber/baseweb/issues) of this repository and believe that this is not a duplicate.

im-adithya:
Fixes #3448

#### Description
Added an example using flex-wrap property to button-group so that the buttons can wrap on small screens

#### Scope
Minor: Docs

### DIFF
--- a/documentation-site/examples/button-group/wrappable.js
+++ b/documentation-site/examples/button-group/wrappable.js
@@ -1,0 +1,20 @@
+// @flow
+import * as React from 'react';
+import {Button} from 'baseui/button';
+import {ButtonGroup} from 'baseui/button-group';
+
+export default function Example() {
+  return (
+    <ButtonGroup
+      overrides={{
+        Root: {
+          style: {flexWrap: 'wrap'},
+        },
+      }}
+    >
+      <Button>Label</Button>
+      <Button>Label</Button>
+      <Button>Label</Button>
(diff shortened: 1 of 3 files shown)


williamernest:
This change seems more appropriate as an override and not something we want as part of our core use case.

im-adithya:
@williamernest That's what I initially thought of, but then I proceeded to add it as the issue (#3448) wasn't closed.
Should I close this then?

williamernest:
We've decided to add an example on the doc site of this behavior, but to leave the component alone. Since this is such a small /straightforward change, it doesn't make a lot of sense to hide it behind a prop so overrides seems like the best approach. Thanks for the contribution. We're happy to help you land a doc-site example if you're still interested in resolving the issue.

im-adithya:
Done!

williamernest (approved):

PR #4510, opened 2021-09-10
fix(textarea): add vertical resizing

### LINKED ISSUE #4054: Text area with auto-sizing
It would be really nice to have a text area with auto-sizing out of the box (ideally with minRows and maxRows props).

Examples from other UI libraries:
https://material-ui.com/components/textarea-autosize/
https://ant.design/components/input/#components-input-demo-autosize-textarea

im-adithya:
Fixes #4054  

#### Description
Changed `resize: none` to 'resize: vertical` so as to enable auto-sizing in textarea component

#### Scope
Patch: Bug Fix

### DIFF
--- a/documentation-site/components/yard/config/textarea.ts
+++ b/documentation-site/components/yard/config/textarea.ts
@@ -27,6 +27,11 @@ const TextareaConfig: TConfig = {
       'inputMode',
       'autoComplete',
     ]),
+    resize: {
+      value: false,
+      type: PropTypes.Boolean,
+      description: 'If true, adds vertical resizability to textarea component.',
+    },
     overrides: {
       value: undefined,
       type: PropTypes.Custom,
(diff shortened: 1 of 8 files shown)


williamernest:
Ideally, we don't want this behavior on our default component so this change should be behind a `resize` prop and it's probably best to implement min/max rows with this change as well.

im-adithya:
Sure! I'll change this to a prop then!
But how do we set max rows for a `textarea` component? (or is it max-height that you are referring to indirectly?)
Because the existing rows prop would kinda set the minimum rows (which is subject to resize unless min-height and max-height are set...)

im-adithya:
Screenshot from 2021-09-16 01-29-57


Done @williamernest! Please review :)

chasestarr:
Thank you for putting up this diff. We've added this feature here https://github.com/uber/baseweb/pull/5103

PR #4516, opened 2021-09-14
feat(ts): export the Theme type for TS users

### LINKED ISSUE #4511: Export Theme type?
I'd like to put my styles into separate files:
```
SignupForm.tsx
SignupFormStyles.ts
```

That would require passing the theme as a variable:

```tsx
// SignupFormStyles.ts
import { StyleObject } from 'styletron-react';
import type { Theme } from 'baseui';

export const signupBox = (theme: Theme): StyleObject => ({
  backgroundColor: theme.colors.background,
  borderRadius: '4px',
  boxShadow: 'rgba(0, 0, 0, 0.1) 0px 4px 12px',
  padding: '10px 20px',
  width: '400px',
});
```

But the Theme type is not exported. Would it be possible to export it?

I didn't find any guidelines in the docs or open issues about this, but this pattern seems reasonable to me?

Zevgon:
Fixes #4511

#### Export the Theme type for TS users. An example use case is described in the bug.

Scope
Minor: New Feature

I'm not sure if tests are required for this change? I verified the fix in my own project, here are some screenshots:

**Before:**
image

**After:**
image

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -51,6 +51,7 @@ export function styled<
     P
 >;
 
+export { Theme } from './theme';
 export const LightTheme: Theme;
 export const LightThemeMove: Theme;
 export const lightThemePrimitives: ThemePrimitives;


Zevgon:
Sorry, closing this PR, see bug for discussion.

PR #4528, opened 2021-09-16
Add membership token, update primitive and semantic tokens for improved accessibility

lhbrennan:
#### Description

We're changing yellow600 from #997328 to #996F00 to better pass accessibility so that we can use it for our Uber One pass brand and lighten up our contentWarning token

Update primitive yellow600 from #997328 to #996F00
Add token for membership to map to yellow600, this token does not need to flip in night mode
Update contentWarning to map to yellow600 in light mode

**Gray colors**
Update gray500 from #757575 to #6B6B6B

**In dark mode**
contentPrimary update to white
contentSecondary update to gray300
contentTeriary update gray400
contentInversePrimary update to black
contentInverseSecondary to gray700
contentInverseTertiary to gray600


chasestarr (approved):
👍

PR #4535, opened 2021-09-20
docs(data-table): add example for customized column sort

### LINKED ISSUE #3488: [DataTable] Create example documenting custom column filter/sort
There should be an example of this on the docs site, but right now it's only hidden in integration tests:  https://github.com/uber/baseweb/blob/master/src/data-table/__tests__/data-table.scenario.js#L137-L220

im-adithya:
Fixes #3488  

#### Description
Added an example for customized column sort as requested in the above issue.

#### Scope
Minor: New example


chasestarr (commented):
This looks good! Do you mind reducing the example scope to only highlight the custom column?

im-adithya:
@chasestarr Done! Does this seem good to you?

Screenshot from 2021-09-25 01-10-16

chasestarr (approved):
Thank you

PR #4544, opened 2021-09-22
Add docs for list heading

lhbrennan:
#### Description
Extend the docs page for `list` to include `ListHeading`.


src/list/list-heading.js
```diff
@@ -78,6 +82,9 @@ const ListHeading = React.forwardRef<HeadingPropsT, HTMLLIElement>(
         'endEnhancerDescription will not be rendered if endEnhancer is not a string',
       );
     }
+    if (maxLines && !isMaxLinesValid(maxLines)) {
```
chasestarr, line 85: Nit: Can we wrap this is a `__DEV__` if statement, maybe do the same for `isEndEnhancerString` check?

chasestarr (approved):

src/list/list-heading.js
lhbrennan: Good catch, thanks.

PR #4546, opened 2021-09-24
feat(select): add setDropdownOpen to Select API

lhbrennan:
#### Description
Allow the user to imperatively open and close the `Select` dropdown.
This is achieved with a new `methods` prop that acts as a handle to expose internal component methods to the user. (For now we are just exposing `setDropdownOpen`.)

#### Scope
Minor: New Feature

### DIFF
--- a/src/select/__tests__/select-methods.scenario.js
+++ b/src/select/__tests__/select-methods.scenario.js
@@ -1,0 +1,51 @@
+/*
+Copyright (c) Uber Technologies, Inc.
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+// @flow
+
+import React from 'react';
+
+import {StatefulSelect} from '../index.js';
+import {Button} from '../../button/index.js';
+
+export default function Scenario() {
+  const methods = React.useRef();
+
+  return (
+    <div style={{width: '360px'}}>
+      <div
+        style={{
+          display: 'flex',
+          justifyContent: 'space-between',
+          marginBottom: '20px',
+        }}
+      >
+        <Button
+          onClick={() => {
+            methods.current && methods.current.setDropdownOpen(true);
+          }}
+        >
+          Open Dropdown
+        </Button>
+        <Button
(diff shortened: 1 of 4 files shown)


src/select/__tests__/select-methods.scenario.js
```diff
@@ -0,0 +1,51 @@
...
+        <Button
+          onClick={() => {
+            methods.current && methods.current.setDropdownOpen(false);
+          }}
+        >
+          Close Dropdown
+        </Button>
+      </div>
+      <StatefulSelect
+        methods={methods}
```
chasestarr, line 42: bike shedding on the prop name.. maybe `methodsRef` to suggest that a ref should be provided?

src/select/types.js
```diff
@@ -134,6 +134,8 @@ export type PropsT = {
   maxDropdownHeight: string,
   /** Defines if multiple options can be selected. */
   multi: boolean,
+  /** Handle for accessing internal methods. */
+  methods?: React.ElementRef<*>,
```
chasestarr, line 138: I think something like below may work without the deprecated `*` (I can't even find docs on the syntax anymore)

```
type ImperativeFnsT = {|
  setDropdownOpen: (boolean) => mixed,
|};

type PropsT = {|
  methods?: {| current: ImperativeFnsT | null |},
|};
```

chasestarr (commented):
I think we may want to get decent unit/integration test coverage on this feature. I'm not really sure how this may effect all of the different state changes in the component

src/select/__tests__/select-methods.scenario.js
lhbrennan: Please always be unrestrained with the bike shedding when reviewing my diffs :)

Thanks for the suggestion. I was thinking the same thing actually.

lhbrennan:
> I think we may want to get decent unit/integration test coverage on this feature. I'm not really sure how this may effect all of the different state changes in the component

Yep -- just wanted to be sure first that this is the route we want to go.

PR #4553, opened 2021-09-29
feat(date-picker): adds month shared prop (#4513)

### LINKED ISSUE #4513: outsideMonth styles are not working when `peekNextMonth` is true
outsideMonth date used to have different styles, but after upgrading to v9.115.0 there is no separate styles for outsideMonth date when peekNextMonth is true

## Current Behavior
It shows the dates in month and outside month with the same styles

Screenshot 2021-09-13 at 1 58 19 PM

## Expected Behavior
It should show the dates outside month as following.

Screenshot 2021-09-13 at 1 59 12 PM

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v10.1.1  |
| React   |         |
| browser |         |

DhruvilSutariya:
`Fixes #1, Fixes #2` 
Fixes #4513 

#### Description
Adds `$month` shared prop to enable styling for the Day which are outside of month

#### Scope
Patch: Bug Fix

### DIFF
--- a/src/datepicker/types.js
+++ b/src/datepicker/types.js
@@ -231,6 +231,7 @@ export type SharedStylePropsT = {
   $isHighlighted: ?boolean,
   $isHovered: ?boolean,
   $isFocusVisible: ?boolean,
+  $month: ?boolean,
   $outsideMonth: ?boolean,
   $outsideMonthWithinRange: ?boolean,
   $peekNextMonth: ?boolean,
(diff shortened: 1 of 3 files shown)


src/datepicker/types.js
```diff
@@ -231,6 +231,7 @@ export type SharedStylePropsT = {
   $isHighlighted: ?boolean,
   $isHovered: ?boolean,
   $isFocusVisible: ?boolean,
+  $month: ?boolean,
```
chasestarr, line 234: should this be a `?number` type?

DhruvilSutariya: Yes, it was a mistake @chasestarr. Thanks for pointing it out ☺️

PR #4573, opened 2021-10-13
Split datepicker dropdown

### LINKED ISSUE #4430: Separate month and year selection on DatePicker dropdown
Right now, there is a single selection dropdown for Month and Year combined on DatePicker UI. 
This often leads to a bad user experience because of increased scroll times and long list of values. 

## Feature description
Could we split the selection to Month and Year separately? Thanks! 

Screen Shot 2021-07-29 at 2 24 37 PM

lhbrennan:
Fixes #4430 

#### Description

Split the datepicker's month-year dropdown into two separate dropdowns.

Also, update the icons to be consistent with design.

###### Before
image

###### After
image

#### Scope
Minor: New Feature


williamernest:
Some UI observations:
- When opening either month or year, the currently selected option is not centered in the dropdown (the current implementation centers the current month/year). 
- I believe the selected value should be indicated in some way when the dropdown is opened for a11y sake. Double check with design if this is intentional.
- Overall, the calendar feels sluggish and not as responsive as prior versions. This is likely due to the several new filtering/array populating functions that are run on every render. It should be possible to prevent this logic from running on every hover event within the existing calendar, since the dropdown shouldn't change. Also, I feel like we should be able to parse this logic once instead of on every render but that may be a wider scope task we can file to resolve later.

williamernest (approved):

PR #4584, opened 2021-10-19
fix(card): card title font-weight should be configurable from theme values

Nathan-Fenner:
#### Description

We're using a font that has a slightly different range of "intended" fontWeight values than baseweb's default theme. As a result, "500" is not very bold, an appropriate level for something like a Card title would be 700 or 800.

Luckily, we can just set this value in the `$theme.typography.HeadingSmall` value, which will configure all of the components throughout the project to "know" what proper bolded text looks like!

Except, unfortunately, the `Card`'s `<Title>` component _overrides the value configured in the theme_ with 500, which means that the `fontWeight` configured in the theme is ignored.

---

To fix this, we want to reorder the fields so that `fontWeight` is taken from the `HeadingSmall` theme, and then only if it's absent there, it _defaults_ to the 500 value. This leaves control of the value in the theme instead of being an unconfigurable hard-coded constant.

#### Scope

Minor: all of the default themes in this repository set their `HeadingSmall` font-weight to be `500`, so this has no effect on them.

However, it's possible that elsewhere there is some user-provided them that relies on this transformation, where the `Card` `<Title>` is forced to be more- or less-bold than other `HeadingSmall` values elsewhere in their application. In this case, there would be a minor styling regression caused by changing from `500` to the theme-configured value.

Since 500 is the current `fontWeight` for `HeadingSmall` value in the baseweb-provided default themes, it's relatively unlikely that this would affect many applications, if any, since they would have had to have changed that default weight _and_ preferred that for whatever reason this change had no effect on their Card titles.


Nathan-Fenner:
@williamernest I've signed the CLA now, thanks. It took longer than expected to get the approval to sign.

src/card/styled-components.js
```diff
@@ -70,9 +70,9 @@ export const Thumbnail = styled<{}>('img', ({$theme}) => ({
 }));
 
 export const Title = styled<{}>('h1', ({$theme}) => ({
+  fontWeight: 500,
```
williamernest, line 73: Since fontWeight is already included in HeadingSmall, I think we can completely eliminate this line.

Nathan-Fenner: That sounds reasonable to me too. I've just updated.

It does look like this new change broke the build, though? Although the break doesn't seem to have anything to do with this change.

williamernest (approved):

PR #4585, opened 2021-10-19
fix(table-semantic): right-align column headers for numbers

leyanlo:
Right-align column headers for numbers. A common best practice is to align headers with their data. See https://www.darkhorseanalytics.com/blog/clear-off-the-table/ or https://medium.com/mission-log/design-better-data-tables-430a30a00d8c.

Before:
column-headers-before

After:
column-headers-after


shannonmoeller:
> Because there’s no room on the right?

I'm not sure this is the appropriate way to handle this. We may need design input. I think always having the caret full-right was a mistake on my part in the original implementation and we should correct that so that the caret is always inline to the right of the text instead of on the right of the cell.

shannonmoeller:
That's why I'm thinking we should as the Base design team. They have strong preferences.

PR #4592, opened 2021-10-27
Update Ladle - Snowpack replaced by Vite

tajo:
This upgrades @ladle/react, the tool that powers https://baseweb.netlify.app/. It builds the prod app twice as fast than snowpack version and outputs only optimized bundles (instead everything mounted). This is a [change set](https://github.com/tajo/ladle/compare/068cc7d..6094707) in ladle (snowpack -> vite).

**Two necessary changes in BaseWeb:**

1. Renamed `#Select.isMounted` to `isItMounted` since I was seeing this error in Ladle:

```
cannot set property`isMounted of #<Component> which has only a getter
```

I believe `isMounted` is a reserved name in React Classes. I think the reason I started seeing this in new ladle is because it replaces `__DEV__` even in node_module dependencies (React in this case), so it triggers this error/warning.

Anyway, since it's an internal change, no concerns for breaking anything. 

2.

```
export {arrayMove, arrayRemove} from 'react-movable';
```

is not currently correctly handled by vite if the dependency is commonjs ([an issue filed by me](https://github.com/vitejs/vite/issues/5521)). I decided not to use ESM version of react-movable yet because we would need to update some jest setup here and we could face some issues down the stream with web-code, since we use older version of node and yarn berry is still struggling with ESM.

Also, at least Node 14 is needed across all environments since ladle/vite now uses latest eslint deps and they require that as an engine (or `y install` fails)

**Note:** Initial run of `yarn ladle serve` does some pre-bundling/caching and can spit out some errors. If that ever happens, restart should fix it and once deps are cached there are no issues at all and startup is instant.


chasestarr (approved):
🟢

PR #4598, opened 2021-11-01
fix(popover): improve keyboard a11y for hover popovers

### LINKED ISSUE #4557: Accessibility Issue: Keyboard focus for interactive elements inside Tooltip
Navigating to a tooltip via keyboard will display the tooltip but not autoFocus on the first interactive element. 

Repro: go to this section for [Tooltip docs](https://baseweb.design/components/tooltip/#tooltip-stateful-example-with-complex-content) and hit tab once. This will open the 'complex content' tooltip which contains a link. The link is not focused, and cannot be clicked via keyboard navigation in any way.

## Current Behavior

Keyboard navigation skips the interactive element inside the tooltip. In Base docs, this can be seen in the example that contains a link. In my development environment, I have observed this behavior with buttons and inputs as well.

## Expected Behavior

Keyboard navigation includes navigating to the interactive element.

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v10.1.1  |
| React   | 16.14 |
| browser |  Chrome 94 |

Similar to https://github.com/uber/baseweb/issues/3840, but that issue and fix did not cover accessibility via keyboard navigation.

williamernest:
Fixes #4557

#### Scope
Patch: Bug Fix


chasestarr:
Does it address the reported issue? Tab focus the tooltip anchor on the page, how can I navigate to the link contained within the tooltip?

williamernest:
Not directly, but I'll update the example and add a blurb that tooltips with interactive content should be used with autoFocus or focusLock. Since the next element in the DOM order won't be the tooltip (since our tooltip appends to the end of the DOM), if the desired state is to move the tooltip content the keyboard flow developers should do this explicitly.

chasestarr (approved):

PR #4605, opened 2021-11-04
feat(datepicker): support separated startDate/endDate inputs

lhbrennan:
#### Description
When implementing a ranged datepicker (where `range === true`), the user can now specific that the input field be divided into two separate fields, one for Start Date, one for End Date.

This is to bring the component in line with design. The combined input field may be deprecated in the future.

#### Scope
Minor: New Feature
Major: Breaking change


src/datepicker/datepicker.js
```diff
@@ -13,10 +13,33 @@ import Calendar from './calendar.js';
...
 import type {DatepickerPropsT} from './types.js';
 import DateHelpers from './utils/date-helpers.js';
 import dateFnsAdapter from './utils/date-fns-adapter.js';
+import type {LocaleT} from '../locale/types.js';
+
+export const DEFAULT_DATE_FORMAT = 'yyyy/MM/dd';
+
+const INPUT_DELIMITER = ' – ';
+
+const combineSeparatedInputs = (
```
lhbrennan, line 26: I'm concerned the purpose of this function might not be clear even after seeing it used in context. Let me know if that's the case and I'll leave a comment above it (I'm usually reluctant to do that because comments can get cluttery, and fall out of sync with the code itself).

src/datepicker/datepicker.js
```diff
@@ -436,33 +520,27 @@ export default class Datepicker<T = Date> extends React.Component<
...
+              {}
+              <InputWrapper
+                {...inputWrapperProps}
+                $separateRangeInputs={
+                  this.props.range && this.props.separateRangeInputs
+                }
+              >
+                {this.props.range && this.props.separateRangeInputs ? (
+                  <>
+                    <div style={{marginRight: '8px'}}>
```
lhbrennan, line 532: Let me know if this way of styling is not ideal.

src/datepicker/datepicker.js
```diff
@@ -436,33 +521,27 @@ export default class Datepicker<T = Date> extends React.Component<
...
+              {}
+              <InputWrapper
+                {...inputWrapperProps}
+                $separateRangeInputs={
+                  this.props.range && this.props.separateRangeInputs
+                }
+              >
+                {this.props.range && this.props.separateRangeInputs ? (
+                  <>
+                    <div style={{width: '100%', marginRight: '8px'}}>
```
chasestarr, line 533: We may want to extract this div and below out into styled components. Do you think developers will need to customize them?

chasestarr (commented):
Just one comment on the newly added elements, but otherwise the component works well

src/datepicker/types.js
```diff
@@ -220,6 +220,10 @@ export type DatepickerPropsT<T = Date> = CalendarPropsT<T> & {
   /** Called when calendar is closed */
   onClose?: () => mixed,
   mask?: string | null,
+  /** Determines if startDate and endDate should be separated into two input fields. Ignored if `range` is not true. */
+  separateRangeInputs?: boolean,
+  startDateLabel?: string,
+  endDateLabel?: string,
```
chasestarr, line 226: Please also update the override types here and in the typescript file

chasestarr (approved):

PR #4621, opened 2021-11-16
fix(datepicker): render correct formatString when mask is undefined

### LINKED ISSUE #4008: [Datepicker] improve format string documentation
Using appropriate letters from the date-fns documentation, it doesn't seem like the formatString functionality works.

## Current Behavior

With formatString set to 'EEEE, MMMM dd yyyy', documented in [the date-fns website](https://date-fns.org/docs/format), the text in the input box shows **"  31, 2020        "**

https://codesandbox.io/s/kgwib?module=src/example.js&file=/src/example.js

## Expected Behavior

The text input should show something like "Thursday, December 31 2020".

## Your Environment

https://codesandbox.io/s/kgwib?module=src/example.js&file=/src/example.js:355-373

lhbrennan:
Fixes #4008

#### Scope
Patch: Bug Fix


src/data-table/__tests__/data-table-columns.e2e.js
```diff
@@ -487,8 +487,7 @@ describe('data table columns', () => {
 
     await datepicker.click({clickCount: 3});
     await page.keyboard.press('Backspace');
-    await datepicker.type('07132014');
-    await datepicker.type('07132014');
+    await datepicker.type('07–13–2014 – 07–13–2014');
```
chasestarr, line 490: Is it possible to maintain the hyphen auto fills? Please make double check that the default keyboard input behavior continues to work as before

src/datepicker/__tests__/datepicker-keyboard-nav.e2e.js
```diff
@@ -153,7 +153,7 @@ describe('Datepicker - keyboard navigation', () => {
     // check that selected date shows up in the input
     let inputValue = await page.$eval(selectors.input, input => input.value);
     // get rid of any whitespaces in the value to compare
-    expect(inputValue.replace(/\s/g, '')).toBe('2019/03/12–//');
+    expect(inputValue.replace(/\s/g, '')).toBe('2019/03/12');
```
chasestarr, line 156: I wonder how many integration/unit tests assert on the hyphen and slash symbols and will need to be updated after this change? Is it possible to keep it the same?

src/datepicker/__tests__/datepicker-unreliable.e2e.js
```diff
@@ -161,7 +161,7 @@ describe('Datepicker', () => {
 
     // input mask
     let selectedValue = await page.$eval(selectors.input, input => input.value);
-    expect(selectedValue).toBe('    /  /  ');
+    expect(selectedValue).toBe(' –     /  /  ');
```
chasestarr, line 164: I don't think there are any range datepickers in https://github.com/uber/baseweb/blob/9197fe57e2abb1276ca80dbd59344f6601088042/src/datepicker/__tests__/datepicker.scenario.js Why do you think we're getting a hyphen here?

chasestarr (approved):
Adding a mask to the data-table datetime column would be appreciated but not required for this PR if you'd prefer including it in a follow-up

PR #4630, opened 2021-11-18
fix(popover): autoFocus without focusLock

lhbrennan:
#### Description

Previously, autoFocus did not work unless focusLock was true. Now, autoFocus and focusLock function correctly, independently of one another.

#### Scope
Patch: Bug Fix


src/app-nav-bar/user-menu.js
```diff
@@ -120,6 +120,7 @@ export default function UserMenuComponent(props: {|
           {...userMenuProps}
         />
       )}
+      autoFocus={false}
```
lhbrennan, line 123: There are multiple instances throughout the repo where we've used `Popover` in ways that assume autoFocus is false, when really it is true by default but just wasn't working correctly. In these cases we need to explicitly pass in `autoFocus={false}` to preserve existing functionality.

src/popover/__tests__/popover-focus-loop.scenario.js
```diff
@@ -32,7 +32,7 @@ export function Scenario() {
         </div>
       )}
     >
-      <button>click</button>
```
lhbrennan, line 35: The trigger for this popover is set to "hover", so this button should be named to reflect that.

williamernest:
The failed tests indicate a time issue with the tests or the component. Can you investigate?

src/popover/__tests__/popover-autofocus.e2e.js
```diff
@@ -0,0 +1,42 @@
...
+    selectedEl,
+  );
+  activeEl.dispose();
+  return equal;
+};
+
+const inputSelector = 'input';
+
+describe('popover', () => {
+  it('contents are auto-focused but not focus-locked', async () => {
```
chasestarr, line 28: Do you mind adding a test for the new scenario file asserting that when the popover opens, the input selector is focused and after tab key press, the input selector is still focused?

chasestarr (approved):
Noting that this does change the default behavior of popover to auto-focus the first focusable element. This was always intended to be the default behavior, and I believe it was the default behavior until a regression was introduced. Given the historical context, this seems fine to merge now as a bug fix rather than waiting for a major release sometime early next year

PR #4643, opened 2021-12-03
chore(datepicker): update calendar styles

lhbrennan:
#### Description
Bring Datepicker's Calendar styles up to date with "Web Medium" design in Figma


src/datepicker/utils/date-helpers.js
```diff
@@ -40,7 +40,7 @@ class DateHelpers<T> {
           monthNumber: 'M',
           dayOfMonthNumber: 'd',
           weekday: 'EEEE',
-          weekdaymin: 'EEEEE',
+          weekdaymin: 'EEEEEE',
```
lhbrennan, line 43: Here's I'm changing the weekday headers from "S M T W..." to "Su Mo Tu We...".  I'm not 100% sure if this is the only place I should be making this change.

chasestarr (approved):
VRTs look good 👍

PR #4652, opened 2021-12-03
fix(documentation:select) Removes `<ValueT>` from JavaScript example

scarabaeus:
#### Description

Removes errant type `<ValueT>` from the JavaScript "[Select with Custom Labels](https://baseweb.design/components/select/#select-with-custom-labels)" documentation example that was preventing it from working on code sandbox.

#### Scope
Patch: Bug Fix


chasestarr (approved):
Thank you. It looks like this generic parameter was not removed by the babel transform.. Not quite sure why.

PR #4655, opened 2021-12-06
feat(datepicker): implement dense calendar

lhbrennan:
#### Description
Adds a "dense" version of the calendar

#### Scope
Minor: New Feature


src/datepicker/constants.js
```diff
@@ -27,3 +27,8 @@ export const STATE_CHANGE_TYPE = Object.freeze({
 export const WEEKDAYS = [0, 1, 2, 3, 4, 5, 6];
 
 export const DEFAULT_MONTHS = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11];
+
```
lhbrennan, line 30: Please suggest alternatives to the variable names I've used. In Figma the term "Medium" corresponds with what I've called "default" here. According to Josh there may be other levels of density for the calendar, so I chose not to use a boolean for this prop.

chasestarr (approved):
LGTM

Unrelated to this specific PR but relevant to a recent datepicker change I noticed that the calendar header does not quite line up in dark mode shown at the bottom in this VRT https://github.com/uber/baseweb/blob/master/vrt/__image_snapshots__/datepicker--calendar-multi-month__dark-snap.png

I believe this was introduced here https://github.com/uber/baseweb/commit/cdd7381bd4a21e95873f2f066a2ab6c1337d6b6b#diff-c5d1412322c4a03f85cfac919b4fc07b12e73a0c446e386da96994feb99d8470 Do you mind looking into that in a follow up PR?

lhbrennan:
> I noticed that the calendar header does not quite line up in dark mode shown at the bottom in this VRT https://github.com/uber/baseweb/blob/master/vrt/__image_snapshots__/datepicker--calendar-multi-month__dark-snap.png
> 
> I believe this was introduced here [cdd7381#diff-c5d1412322c4a03f85cfac919b4fc07b12e73a0c446e386da96994feb99d8470](https://github.com/uber/baseweb/commit/cdd7381bd4a21e95873f2f066a2ab6c1337d6b6b#diff-c5d1412322c4a03f85cfac919b4fc07b12e73a0c446e386da96994feb99d8470) Do you mind looking into that in a follow up PR?

Thanks for catching that!

PR #4657, opened 2021-12-06
fix(timepicker): handle min/max datetimes outside of value date

chasestarr:
#### Description

Addressing issue raising in this PR https://github.com/uber/baseweb/pull/4645

#### Scope
Patch: Bug Fix


williamernest (approved):
Looks good. I think one thing we could add is a note in the documentation that the date portion of the min/max dates will be updated to match the value in the event they're different.

src/timepicker/timepicker.js
```diff
@@ -200,24 +201,45 @@ class TimePicker<T = Date> extends React.Component<
...
-    if (!min) {
-      min = midnight;
+    let {minTime: min, maxTime: max, ignoreMinMaxDateComponent} = this.props;
+    const dayStart = this.setTime(this.props.value, 0, 0, 0);
+    const dayEnd = this.setTime(this.props.value, 24, 0, 0);
+
+    if (
+      !min ||
+      (this.props.adapter.toJsDate(min) <
+        this.props.adapter.toJsDate(dayStart) &&
```
abhimanyu-singh-uber, line 211: small nit: can we use `this.props.adapter.isBefore` & `this.props.adapter.isAfter` instead of less than or greater than, I think the code will be more readable then.
```suggestion
      (this.props.adapter.isBefore(min, dayStart) &&
```

src/timepicker/timepicker.js
```diff
@@ -200,24 +201,45 @@ class TimePicker<T = Date> extends React.Component<
...
+        this.props.adapter.getMinutes(min),
+        this.props.adapter.getSeconds(min),
+      );
     }
-    if (!max) {
-      max = this.setTime(this.props.value, 24, 0, 0);
+
+    if (
+      !max ||
+      (this.props.adapter.toJsDate(max) > this.props.adapter.toJsDate(dayEnd) &&
```
abhimanyu-singh-uber, line 226:
```suggestion
      (this.props.adapter.isAfter(max, dayEnd) &&
```

src/timepicker/timepicker.js
chasestarr: Thank you

abhimanyu-singh-uber (approved):
Thanks Looks good to me

williamernest (approved):

PR #4674, opened 2021-12-20
fix(modal): update returnFocus type to match with returnFocus type of FocusLock

abhimanyu-singh-uber:
#### Description

update returnFocus type to match with returnFocus type of FocusLock.
So that we don't have to do flow-ignore or ts-ignore.

#### Scope
Patch: Bug Fix


chasestarr (approved):
Thank you

PR #4694, opened 2022-01-11
chore(datepicker): accommodate null date values

lhbrennan:
#### Description
This change allows the datepicker to handle null date values, allowing for more flexibility in how ranged datepickers work. This changed is required for use-cases where the end-user may want to select the end date before selecting a start date, or when the user nullifies a previously selected start date without changing the end date.

This change is required as a prerequisite for [WPT-7624](https://t3.uberinternal.com/secure/RapidBoard.jspa?rapidView=582&projectKey=WPT&view=detail&selectedIssue=WPT-7624)

#### Scope
Minor: New Feature


src/data-table/column-datetime.js
```diff
@@ -410,7 +410,7 @@ function DatetimeFilter(props) {
                     if (Array.isArray(date)) {
                       if (!date.length) return;
                       const nextDates = date.map((d, i) =>
-                        applyDateToTime(rangeDates[i], d),
+                        d ? applyDateToTime(rangeDates[i], d) : null,
```
lhbrennan, line 413: `d` can now be `undefined` or `null`, sp we need to make sure we don't send a non-date value to `applyDateToTime`.

src/datepicker/day.js
```diff
@@ -234,10 +245,12 @@ export default class Day<T = Date> extends React.Component<
       return false;
     }
     // fix flow by passing a specific arg type and remove 'Array.isArray(value)'
-    if (Array.isArray(value) && value.length > 1) {
+    if (Array.isArray(value) && value[0] && value[1]) {
       return this.dateHelpers.isDayInRange(
         this.clampToDayStart(date),
+        // $FlowFixMe -- value[0] can't be null or undefined
```
lhbrennan, line 251: I played around with this a bit but I can't figure out a good way to re-write the logic to obviate the need for the `FlowFixMe`s. Open to suggestions!

src/datepicker/stateful-container.js
```diff
@@ -32,11 +33,11 @@ class StatefulContainer<T = Date> extends React.Component<
 
   constructor(props: PropsT<T>) {
     super(props);
-    const value = props.range ? [] : (null: ?T);
+    const value = props.range ? [null, null] : (null: ?T);
```
chasestarr, line 36: Is it possible for this line to remain unchanged? `[]` and `[null, null]` should both resolve to the type `Array<?T>`

src/datepicker/calendar.js
```diff
@@ -346,27 +350,34 @@ export default class Calendar<T = Date> extends React.Component<
...
-        return newTimeState[index];
-      });
+      const start = data.date[0]
+        ? this.dateHelpers.applyDateToTime(newTimeState[0], data.date[0])
+        : null;
+      const end = data.date[1]
+        ? this.dateHelpers.applyDateToTime(newTimeState[1], data.date[1])
+        : null;
+      newTimeState[0] = start;
+      newTimeState[1] = end ? end : newTimeState[1];
```
chasestarr, line 367: maybe this line should be moved to within the if statement below to avoid the inline-ternary with the same predicate

src/datepicker/datepicker.js
```diff
@@ -351,7 +351,7 @@ export default class Datepicker<T = Date> extends React.Component<
           if (displayValueAtRangeIndex === 0) {
             left = date;
             if (!right) {
-              onChange({date: [left]});
+              onChange({date: [left, null]});
```
chasestarr, line 354: is setting this to null necessary, or is leaving [left, undefined] sufficient?

I've made similar comments regarding this idea in various places so far, and the point I want to make is that there seems to be an implicit difference between `undefined` and `null` values internally, and for consumers. It would be less complicated if we can operate as if undefined is essentially the same as null regarding the date value

src/datepicker/day.js
```diff
@@ -87,9 +87,20 @@ export default class Day<T = Date> extends React.Component<
     const {range, value} = this.props;
     let date;
     if (Array.isArray(value) && range) {
-      if (!value.length || value.length > 1) {
-        date = [selectedDate];
-      } else if (this.dateHelpers.isAfter(selectedDate, value[0])) {
+      if ((!value[0] && !value[1]) || (value[0] && value[1])) {
```
chasestarr, line 90: Can you please add some integration tests that validate the behavior here, and below?

src/datepicker/day.js
chasestarr: I believe if you assign value[0] and value[1] to local variables, flow can make more assertions on whether the value changes or not. So maybe on L247 doing something like `const [start, end] = value;` would help out here and for the instances below

src/datepicker/__tests__/datepickers-composed-range.scenario.js
```diff
@@ -29,7 +29,7 @@ function printTime(dt) {
 
 export function Scenario() {
   const [css, theme] = useStyletron();
-  const [dates, setDates] = useState<Array<Date>>([START_DATE, END_DATE]);
+  const [dates, setDates] = useState<Array<?Date>>([START_DATE, END_DATE]);
```
chasestarr, line 32: This and the changes below may indicate a substantial change required to application code. Do you have an idea of how many instances there are in internal code that would need updating? Try running the alpha test in CI

src/datepicker/day.js
lhbrennan: Fantastic, thank you.

src/datepicker/calendar.js
lhbrennan: yep, thanks for the suggestion.

src/datepicker/datepicker.js
lhbrennan: Yes, `undefined` and `null` are treated the same internally. My thinking in using `null` (even though not technically necessary) was that it was more explicit, but I get what you're saying and I totally agree. Using `null` in cases like the above makes it seem like there may actually be a difference between the two, which could be confusing for future devs reading the code. I'll review all the changes and make sure I'm not using `null` needlessly.

chasestarr (approved):
👍

PR #4699, opened 2022-01-14
fix(eslint): add jsx check if import was renamed

williamernest:
Fix instance where import was renamed and an old typography component name was used within the component (H1 as StyledH1 and later H1 is a valid local variable that should not be renamed). See test for minimally reproduced test case

### DIFF
--- a/packages/eslint-plugin-baseui/__tests__/deprecated-component-api.test.js
+++ b/packages/eslint-plugin-baseui/__tests__/deprecated-component-api.test.js
@@ -442,6 +442,38 @@ const tests = {
         }));
       `,
     },
+    // H1, renamed with a different component using same name
+    {
+      code: `
+        import { H1 as StyledH1 } from "baseui/typography"
+        import ReactMarkdown from 'react-markdown';
+        import { getOverrides } from 'baseui/helpers/overrides';
+
+        const MarkdownRender = ({ overrides = {}, text }: Props) => {
+           const [H1, h1Props] = getOverrides(overrides.H1, StyledH1);
+        
+          return <ReactMarkdown
+            components={{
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin-baseui/__tests__/deprecated-component-api.test.js
```diff
@@ -442,6 +442,38 @@ const tests = {
...
+        import { HeadingXXLarge as StyledH1 } from "baseui/typography"
+        import ReactMarkdown from 'react-markdown';
+        import { getOverrides } from 'baseui/helpers/overrides';
+
+        const MarkdownRender = ({ overrides = {}, text }: Props) => {
+           const [H1, h1Props] = getOverrides(overrides.H1, StyledH1);
+        
+          return <ReactMarkdown
+            components={{
+              h1: ({ node, ...rest }) => <H1 {...h1Props} {...rest} />
```
chasestarr, line 472: before, this H1 was changing to StyledH1?

chasestarr (approved):

packages/eslint-plugin-baseui/__tests__/deprecated-component-api.test.js
williamernest: No, it was changing it to HeadingXXLarge. The JSXIdentifier selector was checking if a typography component was imported, but not checking if it was still using the original name. So when the JXSIdentifier got to this `H1`, it saw that `H1` was imported and it went ahead and renamed it. My fix adds a check to ensure that it was imported and is still named `H1` before replacing it. If it's renamed, we do nothing since we already replaced the name on import in the ImportIdentifier selector.

PR #4702, opened 2022-01-15
fix(eslint): add more tests and rewrite typography logic

williamernest:
Rewrites the typography eslint logic by pulling it one level higher (executes once per import statement instead of on each import in a statement). It tracks previous imports, and renames a component to reuse the previous import instead of duplicating imports. 

Updates the typography array with a map, and tracks what components need actions in the current document (based on import presence) so performance should be much better.


tajo (approved):

packages/eslint-plugin-baseui/__tests__/deprecated-component-api.test.js
```diff
@@ -475,6 +475,49 @@ const tests = {
...
+
+        const MarkdownRender = () => {
+          return <div><HeadingXXLarge>Large</HeadingXXLarge><HeadingXXLarge>Large</HeadingXXLarge></div>
+          }
+      `,
+    },
+    // oldName imported and renamed, and newName imported and used
+    {
+      code: `
+        import { HeadingXXLarge as Hello, H1 } from "baseui/typography"
```
chasestarr, line 502: I believe something like this may avoid the need to manually remove the import 
https://astexplorer.net/#/gist/06f1acbd8bc8feb3b6c28f5bc656d579/f99268da03756b87c48c556eaf38c8a429912a06

williamernest: Rewriting the entire import statement using the current node state causes other import changes to be discarded. I've updated the rule to prune from the end of the previous specifier to the end of the current specifier which fixes the issue.

PR #4713, opened 2022-01-19
fix(eslint): check object owns property for Identifier, add test

williamernest:
Adds a test case and fix for instances of `toString`, which exist on Object prototype causing erroneous code changes.


chasestarr (approved):
Approving, since I assume you know what this is fixing but I'm not quite sure how this test exercises the code change - could you elaborate for my own benefit?

tajo (approved):

williamernest:
Sure. The `Identifier` method was triggering on `toString`, since `toString` technically qualifies as an identifier. Since the map is just an object, toString is a prototype function, so looking up `identifiersToRename['toString']` results in a function and not undefined. Line 497 was evaluating true, which caused the `toString` text to be replaced with the string value of the function (`function toString() { [native code] }`). 

The fix leverages `hasOwnProperty`, since it doesn't check the prototype chain for the property and only checks properties assigned to the object itself.

chasestarr:
Ah ok thanks. I overlooked that it's also traversing over identifiers. Thought it was just JsxIdentifier

PR #4725, opened 2022-01-26
fix(datepicker): remove boxShadow from outsideMonth date(#4724)

### LINKED ISSUE #4724: Mysterious blue box appears on some dates.
Mysterious blue box appears on some dates.

## Current Behavior
when we have range enabled and we navigate via keyboard from one month to another and then again to previous month. From primary investigation I have found out that bug is introduced in version 9.115.0. and it is because `$isFocusVisible` is true for that day and as `onBlur` will be never called for that day as it really does not exist. So that variable never set to false again

https://user-images.githubusercontent.com/22965398/151122535-fc9c3f84-943b-40fe-9d21-bf463766a893.mov

## Expected Behavior
Blue box should not appear

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v10.8.0  |
| React   |         |
| browser |   Chrome      |

hardikmodi1:
Remove boxShadow from date out side month when peekNextMonth is false

Fixes #4724


chasestarr:
How about inverting the logic by applying the `$peekNextMonth` and `$outsideMonth` when [setting the box-shadow](https://github.com/uber/baseweb/blob/master/src/datepicker/styled-components.js#L499), rather than unsetting it later?

chasestarr (approved):
Thank you

PR #4729, opened 2022-01-28
feat(rating): add rating kind secondary and primary

ProjectBarks:
This adds a secondary style to ratings, splitting the two kinds by `KIND` constants.

#### Description

Have seen use cases where base could warrant a secondary style especially for a read-only mode.

image
image
image

#### Scope
Minor: New Feature


chasestarr:
Thank you for contributing. We've updated rating star colors to black since this PR was opened. Please re-assess and open again if this is something we still want

PR #4735, opened 2022-01-31
fix(popover): replace mouse events with pointer events (#4692)

### LINKED ISSUE #4692: Tooltip: does not work when child element is in disabled state
StatefulTooltip does not work with child element in `disabled` state. For example, tooltip is not rendered on disabled buttons. 

Codesandbox for this: https://codesandbox.io/s/base-web-statefultooltip-forked-7dlpl?file=/src/example.js

## Expected Behavior

Tooltip should be supported on disabled elements. Currently have to manage tooltip's state with `onPointerEnter` and `onPointerLeave` events on the child to achieve the desired behaviour.

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v10.6.0  |
| React   | 17.0.2 |
| browser | Chrome |

- [X] I have searched the [issues](https://github.com/uber/baseweb/issues) of this repository and believe that this is not a duplicate.

ujjawalpabreja99:
Fixes #4692

#### Description

Replaces `onMouseEnter` and `onMouseLeave` events with `onPointerEnter` and `onPointerLeave` events on the anchor element so as to support Popover/Tooltip even when the anchor is in disabled state

#### Scope
Patch: Bug Fix


ujjawalpabreja99:
@chasestarr should we be re-naming the props to reflect the change (eg: `onMouseEnterDelay` -> `onPointerEnterDelay`) ? Since it'll be a breaking change across baseui, wanted to confirm first.

src/popover/types.js
```diff
@@ -219,8 +219,8 @@ export type AnchorPropsT = {
   onBlur?: (e: Event) => mixed,
   onClick?: (e: Event) => mixed,
   onFocus?: (e: Event) => mixed,
-  onMouseEnter?: (e: Event) => mixed,
-  onMouseLeave?: (e: Event) => mixed,
```
chasestarr, line 223: Please restore these prop types. I would like to maintain the current API. I think even after it has changed to pointer events the top level api describes essentially the same behavior

chasestarr (commented):
Thank you, we can merge after CI and api change

yashatgit:
@chasestarr can you please help with the next steps on this PR

src/popover/types.js
chasestarr: Consider if an application creates a custom Anchor component rather than providing an element https://codesandbox.io/s/base-web-statefulpopover-forked-hs1t44?file=/src/example.js

ujjawalpabreja99: @chasestarr restored the prop types. Please review the changes. Thanks! 😄

chasestarr:
Thank you. I'll need to spend some time next week reviewing how this will affect our internal projects. As shown in this PR, some tests may need to change

Laegas:
What's the status on this? I'm using a workaround wrapping the StatefulTooltip children, including a disabled Button, in `<div>`s but that has another problem I reported [here](https://github.com/uber/baseweb/issues/4979).

Laegas:
@chasestarr @ujjawalpabreja99 Can you guys please share status on this one? 😇

chasestarr:
We're likely not going to land this PR due to the disruption it would cause in our internal tests. Thank you for submitting

Laegas:
@chasestarr So there's no fix coming for [this issue](https://github.com/uber/baseweb/issues/4979)? Or is there an alternative way how to have a tooltip on a disabled button?

PR #4737, opened 2022-02-03
feat(form-control): add counter end enhancer

ProjectBarks:
#### Description

There is many example in Uber Applications where users count the number of characters for a text input. This standardizes the appearance and function of such a tool.

A new `counter` prop is introduced to the `form-control` component that will override an end-enhancer with a count of the characters for the sub-input. It will also leverage a nested `maxLength` child prop if it is available to display the length as a ratio of the maximum length. 

If the form-control cannot infer length or maxLength the user can pass in a custom max length and length property.

image

#### Scope
Minor: New Feature


src/form-control/form-control.js
```diff
@@ -83,17 +93,57 @@ export default class FormControl extends React.Component<
...
+    if (counter) {
+      let maxLength: ?number = null;
+      let length: ?number = null;
+
+      if (typeof counter === 'object') {
+        length = counter.length;
+        maxLength = counter.maxLength;
+      }
+
+      maxLength = maxLength ?? onlyChildProps.maxLength;
```
chasestarr, line 106: Can you please use the older ternary syntax?

src/form-control/form-control.js
```diff
@@ -83,17 +93,57 @@ export default class FormControl extends React.Component<
       }
     }
 
+    let labelEndEnhancer = this.props.labelEndEnhancer;
+    if (counter) {
```
chasestarr, line 97: Should this override developer-provided `props.labelEndEnhancer`? Maybe include a check to apply this if-block only when `props.labelEndEnhancer` is falsey

src/form-control/form-control.js
```diff
@@ -83,17 +93,57 @@ export default class FormControl extends React.Component<
...
+            >
+              {typeof label === 'function' ? label(sharedProps) : label}
+            </Label>
+            {labelEndEnhancer && (
+              <LabelEndEnhancer
+                {...sharedProps}
+                {...getOverrideProps(LabelEndEnhancerOverride)}
+              >
+                {typeof labelEndEnhancer === 'function'
+                  ? labelEndEnhancer(this.props)
```
chasestarr, line 142:
```suggestion
                  ? labelEndEnhancer(sharedProps)
```

Can you please change args to `sharedProps`? This will follow the existing pattern in label, positive, negative props

chasestarr (commented):
Couple small comments, but looks good overall

src/form-control/form-control.js
ProjectBarks: This is a good point

src/form-control/form-control.js
```diff
@@ -83,17 +97,82 @@ export default class FormControl extends React.Component<
...
+        sharedProps.$counterError = true;
+      }
+      console.log(
+        sharedProps,
+        children,
+        typeof onlyChildProps.error !== 'undefined'
+          ? onlyChildProps.error
+          : sharedProps.$error,
+      );
+    }
```
chasestarr, line 148:
```suggestion
```

chasestarr (approved):

PR #4738, opened 2022-02-03
feat: added gitpod setup

Ayush7614:
* This PR contains

- Setup of Gitpod
- In Gitpod how you run and build the project


chasestarr:
Hi @Ayush7614 we will not add gitpod config to this repo, but you can look into using github codespaces that's already available here

PR #4740, opened 2022-02-04
Map marker label enhancer

NameFILIP:
#### Description

* Move default color values for label enhancer and badge enhancer into the styled-components.js file
* Wrap `console.warn` into `__DEV__` condition to remove it from the production bundle
* Downgrade `react-map-gl` to v5 in order to use `mapbox-gl-js` v1 (v2 charges for each map initialization)
* Replace dynamic map tiles with the static background color. This will ensure snapshot tests are not flaky

#### Scope
Patch: Bug Fix


src/map-marker/__tests__/map-style.js
```diff
@@ -0,0 +1,50 @@
...
+
+type GeoJSONPointFeature = {
+  type: 'Feature',
+  geometry: {
+    type: 'Point',
+    coordinates: [number, number],
+  },
+};
+
+export function getMapStyle(
```
NameFILIP, line 17: Moved the source and the layer configurations for debug points into this separate file

src/map-marker/fixed-marker.js
```diff
@@ -107,24 +107,35 @@ const FixedMarker = ({
...
-    );
+    ![
+      PINHEAD_SIZES_SHAPES.xxSmallCircle,
+      PINHEAD_SIZES_SHAPES.xxSmallSquare,
+    ].includes(size);
+
+  if (__DEV__) {
+    if (
+      needle !== NEEDLE_SIZES.none &&
+      [
```
NameFILIP, line 126: Note, this condition is slightly different from the one above for the `renderNeedle` so that it doesn't show the warning when the `needle` is `NEEDLE_SIZES.none`

PR #4746, opened 2022-02-08
fix(modal): modal will not lock focus within an iframe only

### LINKED ISSUE #4743: Weird Behavior in iframe when popover or modal is open
## Feature description

This is that cross between a bug and an enhancement. We have an iframe that loads an app running BaseWeb components. We operate more or less like a codesandbox or repl.it where users run apps in the iframe and can manipulate it externally. If a popover or modal is open on the app in iframe, any behavior on the parent frame conflicts with the app due to the focus lock. This causes a flicker as the conflicts are fighting for focus.

I think the easiest solution would be to use `react-focus-lock's` undocumented `crossFrame` property https://github.com/theKashey/react-focus-lock/issues/104 and setting it to false. It essentially ensures the component maintains focus within the iframe, but allow the parent window to take focus.

kmcgrady:
Fixes #4743

#### Description

This change allows a focus-locked component (modal, drawer, popover, table filter) to lose focus only when the focus exits the frame. This will only occur if the component is within an `iframe`

See prop details here https://github.com/theKashey/react-focus-lock/commit/486a7e08428487286d09ef7bfeaa701e8376318d

#### Scope
Patch: Bug Fix

### DIFF
--- a/src/drawer/drawer.js
+++ b/src/drawer/drawer.js
@@ -253,7 +253,13 @@ class Drawer extends React.Component<DrawerPropsT, DrawerStateT> {
       <LocaleContext.Consumer>
         {locale => {
           return (
-            <FocusLock returnFocus autoFocus={autoFocus} noFocusGuards={true}>
+            <FocusLock
+              // Allow focus to escape when UI is within an iframe
+              crossFrame={false}
+              returnFocus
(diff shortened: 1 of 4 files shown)


chasestarr (approved):
Thank you

PR #4749, opened 2022-02-09
refactor(data-table): performance improvements for wide tables

### LINKED ISSUE #4564: [DataTable] initial load freezes the page with large data
When loading a page that contains a data-table component and when that table attempts to render lots of data the page can hang. This is due to the component rendering many cells off-screen to measure column width. Filing this issue to track any improvements we can make here. Can we ensure that the page does not freeze or can we speed up the measurement process?

enjoylife:
Fixes #4564

#### Description

When rendering data with a large number of columns, the initial page load and scrolling performance was poor.
* First page load took 5+ seconds to see any data.
* Scrolling horizontally was often blank.

The root cause was due to the `MeasureColumnWidths` component.
image

This component was not showing any data until it had measured all of the `(num_sampled_rows * num_columns)` bounding rects. 

To fix this,  we now only have to measure a single parent div, which holds the header and sampled row cells. This reduces the expensive dom measurements to just `(num_columns)`.

In addition we also chunk the number of times we refresh the widths, allowing visible column data to be shown quickly,  before the rest of the columns measurements come in.

For testing the following was done:
* Adds a scenario with a large amount of data in both the rows, as well as the number of columns.
* The scenario, has numeric and string data with varying sizes, and records the median render times to the console.

#### Scope
Patch: Refactor


src/data-table/measure-column-widths.js
```diff
@@ -12,46 +12,60 @@ import {useStyletron} from '../styles/index.js';
...
-  const [ref, dimensions] = useDimensions();
-
-  React.useEffect(() => {
-    onDimensionsChange(dimensions);
-  }, [dimensions, onDimensionsChange]);
+  }, []);
 
-  return React.cloneElement(props.item, {ref});
+  return (
+    <div
```
chasestarr, line 30: Awesome! just measuring the parent is such a painfully obvious improvement. Thank you

src/data-table/measure-column-widths.js
```diff
@@ -152,61 +154,26 @@ export default function MeasureColumnWidths({
...
-            key={`measure-${columnIndex}-${rowIndex}`}
-            onDimensionsChange={dimensions =>
-              handleDimensionsChange(columnIndex, rowIndex, dimensions)
-            }
-            item={
-              <Cell
-                value={columns[columnIndex].mapDataToValue(rows[rowIndex].data)}
-                isMeasured
-                onSelect={
-                  isSelectable && columnIndex === 0 ? () => {} : undefined
```
chasestarr, line 175: I believe this should be forwarded to where the cell is being rendered

chasestarr (commented):
Thank you for making this improvement. There seems to be a problem with columns expanding their width to fill the available table width. Can you please investigate that? See the example stories below, and change the url to baseweb.netlify.app to see how it renders on master.

https://deploy-preview-4749--baseweb.netlify.app/?story=data-table--batch-action
https://deploy-preview-4749--baseweb.netlify.app/?story=data-table--initial-selected-rows
https://deploy-preview-4749--baseweb.netlify.app/?story=data-table--initial-sort
https://deploy-preview-4749--baseweb.netlify.app/?story=data-table--loading

src/data-table/measure-column-widths.js
enjoylife: Good call. forwarded it to the header, which has the impact on the size.

enjoylife:
> There seems to be a problem with columns expanding their width 

Fixed that up.

chasestarr:
> Ahh, hold up the empty message is kinda janky. Or perhaps these screenshots are stale? [enjoylife#2](https://github.com/enjoylife/baseui/pull/2)

I just restarted CI so we will see VRT diffs soon

enjoylife:
Ahh I see whats up, I needed to create a deterministic random generator otherwise the snapshots will be different for every run. Fixed.

enjoylife:
Anything blocking the PR?

chasestarr:
> Anything blocking the PR?

Green CI and we'll be good!

chasestarr (approved):
Thank you!

PR #4754, opened 2022-02-10
fix: passing style props to element to not be present on DOM

danestves-reworth:
Fixes #4753 

#### Description

This PR fix the error of passing props directly to the DOM element of `FlexGridItem` and use the prop directly to avoid errors

#### Scope
Patch: Bug Fix


chasestarr:
Hi @danestves-reworth we will not be adding more props to the block component. We recommend using `useStyletron` as shown on the block documentation page here https://baseweb.design/components/block

PR #4755, opened 2022-02-10
feat(button-group): add selected disabled visual state

ProjectBarks:
Adds a new visual state for button groups for when they are both "selected & disabled"

#### Description

Left: Selected
Right: Selected & Disabled

image
image#### Scope

Minor: New Feature


chasestarr (commented):
Looks good after the snapshots are updated

PR #4773, opened 2022-02-15
feat(progress-bar): add minValue and maxValue

tajo:
Adds `minValue` and `maxValue` while `successValue` still works as a fallback to `maxValue`. I think it makes sense to preserve the range functionality since the [progress element](https://html.spec.whatwg.org/multipage/form-elements.html#the-progress-element) has `max` attribute and having `min` is nice to have so the user doesn't have to normalize values manually.

Some other component libraries like [react-spectrum](https://react-spectrum.adobe.com/react-spectrum/ProgressBar.html) also gives user `minValue` and `maxValue`.

### DIFF
--- a/documentation-site/components/yard/config/progress-bar.ts
+++ b/documentation-site/components/yard/config/progress-bar.ts
@@ -65,11 +65,17 @@ const ProgressBarConfig: TConfig = {
       description:
         'Defines how many steps the progress bar has. Defaults to 1.',
     },
-    successValue: {
-      value: 100,
+    minValue: {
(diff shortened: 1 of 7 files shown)


williamernest:
Probably makes sense to remove the deprecated prop from the eslint plugin as well
https://github.com/uber/baseweb/blob/7f25bf19647dc7c2c272b3f8fb5999f0d57694e8/packages/eslint-plugin-baseui/src/deprecated-component-api.js#L292

williamernest (approved):

PR #4775, opened 2022-02-15
feat(data-table): histogram for numerical filter

enjoylife:
#### Description
* Adds a histogram for filtering on a numerical column
* Removes the inequality option.
* Moves the exclude up a line for all filters

#### Scope
Minor: New Feature


src/data-table/filter-shell.js
```diff
@@ -13,16 +13,33 @@ import {Checkbox, STYLE_TYPE} from '../checkbox/index.js';
 import {useStyletron} from '../styles/index.js';
 import {LocaleContext} from '../locale/index.js';
 
+type ExcludeKind = 'value' | 'range';
+
+// If modifying this, take a look at the histogram and adjust. see HISTOGRAM_SIZE
+const FILTER_SHELL_WIDTH = '320px';
```
chasestarr, line 19: maybe extract this value out into the `constants.js` file, and subtract some margin when reading in the numerical filter?

src/data-table/__tests__/column-numerical.test.js
```diff
@@ -132,18 +132,44 @@ describe('numerical column', () => {
...
+    fireEvent.click(getByText('Single Value'));
+    fireEvent.click(getByText('Exclude Value'));
+    fireEvent.click(getByText('Apply'));
+
+    const [filterParamsSecond] = mockSetFilter.mock.calls[1];
+    expect(filterParamsSecond.exclude).toBe(false);
+    expect(filterParamsSecond.description).toBe('= 3');
+  });
+
+  it.skip('hides 2nd input if filtering single value', () => {
```
chasestarr, line 172: Please remove if this is not needed anymore

src/data-table/filter-shell.js
```diff
@@ -13,16 +13,33 @@ import {Checkbox, STYLE_TYPE} from '../checkbox/index.js';
 import {useStyletron} from '../styles/index.js';
 import {LocaleContext} from '../locale/index.js';
 
+type ExcludeKind = 'value' | 'range';
```
chasestarr, line 16: Was this requested by design? I don't particularly like the idea of special-casing this shared wrapper to support one filter's use-case, and I think there is some benefit to maintain consistency. Not too big of a deal, if you feel strongly about it

chasestarr (commented):
Thanks for making this improvement @enjoylife. One bug, but otherwise this looks and operates well. Repro steps below

[scenario](https://deploy-preview-4775--baseweb.netlify.app/?story=data-table--data-table)
1. open filter menu and select numerical filter
2. switch to single-value and click apply
3. click on tag component documenting the applied filter to edit it
4. view that the filter opens to range selection, it should remember that this is a single select

src/data-table/filter-shell.js
enjoylife: Requested by design yeah, 

> It’s more clear to label the toggle “Exclude range” / “Exclude value”. It doesn’t look as nice but it is easier to understand. 

As for how we inform the shared wrapper, it was the minimal choice, although it turns the wrapper into a leaky abstraction. I don't feel too strongly, but I would say, if we are to update the other filters, the wording on the excludes could be refined there as well.

src/data-table/filter-shell.js
enjoylife: extracted out to the constants

enjoylife:
Fixed the state bugs and added a visual snapshot.

src/data-table/column-numerical.js
```diff
@@ -338,6 +352,13 @@ function NumericalFilter(props) {
                 margin: '0 -7px',
               }),
             },
+            InnerTrack: {
+              style: ({$theme}) => ({
+                // For range selection we use the color as is, but when selecting the single value,
+                // we don't  want the track to standout so we mute its color
+                background: isRange ? undefined : theme.colors.mono400,
```
chasestarr, line 359: Looks like this is backwards based on https://deploy-preview-4775--baseweb.netlify.app/?story=data-table--numerical-column

enjoylife: ahh.. thanks for catching, glad I added the visual snapshot. Updating the visuals.

enjoylife:
@chasestarr not sure what is failing the vercel deployment?

chasestarr:
> @chasestarr Hopefully that fixes it, as I started going down the route of upgrading our usage of next to v11, but that was painful.

Thank you. I also have made the mistake of thinking a next upgrade would be pretty quick. We should take care of it soon

chasestarr (approved):
Thank you @enjoylife

PR #4790, opened 2022-02-18
fix(map-marker): mistake during merging conflicts

NameFILIP:
#### Description
The following piece had to be removed:
```
<Block
                  display="flex"
                  flexDirection="column"
                  height="150px"
                  key={index}
                >
                  <Label3 color={contentSecondary}>{label}</Label3>
```

#### Scope
Patch: Bug Fix


src/map-marker/__tests__/tile-grid.js
```diff
@@ -53,21 +53,14 @@ const TileGrid = ({children, customizerOptions, cols}: TileGridPropsT) => {
               >
                 <LabelSmall color={contentSecondary}>{label}</LabelSmall>
```
NameFILIP, line 54: Block and Label are duplicated above and below. Instead, `Label3` had to be changed to `LabelSmall`.

PR #4795, opened 2022-02-21
fix(data-table): fix slider scale

### LINKED ISSUE #4796: Numerical column slider does not allow picking float's
## Current Behavior
When the data sent into the numerical column is high precision, the current step size for the slider does not allow picking values smaller than 1. 

## Expected Behavior
The slider allows choosing values smaller than 1.

enjoylife:
Fixes #4796
#### Description

Allows the numerical column filtering to work for data with high precision and allows selecting values with step sizes smaller than 1.
 
#### Scope
Patch: Bug Fix


chasestarr (approved):
LGTM

PR #4801, opened 2022-02-23
revert(data-table): reverts recent data-table changes 

williamernest:
Reverts data-table changes. 

#4749 reverted due to table rows not rendering as expected in fusionjs tests
#4775 reverted due to ESM issues with D3 (nextjs doesn't support 3rd party ESM, also throwing similar errors in fusionjs) 
#4795 reverted due to being built on top of the other 2


chasestarr (approved):

enjoylife:
@williamernest I understand the d3 related ones with d3 and esm. But I'm not seeing how #4801 needs to be reverted.

PR #4804, opened 2022-02-23
Add locked calendar behavior (deprecated)

lhbrennan:
#### Description
The adds "locked behavior" for ranged Datepickers. When a user selects the Start Date input field, they will only be modifying the Start Date. When the user picks the End Date input field, they will only be modifying the End Date.

#### Scope
Minor: New Feature


lhbrennan:
Oops, this should go to the V11 branch, not master.

PR #4808, opened 2022-02-24
Add locked calendar behavior 2

lhbrennan:
#### Description
The adds "locked behavior" for ranged Datepickers. When a user selects the Start Date input field, they will only be modifying the Start Date. When the user picks the End Date input field, they will only be modifying the End Date.

#### Scope
Minor: New Feature


src/datepicker/day.js
```diff
@@ -83,39 +84,64 @@ export default class Day<T = Date> extends React.Component<
       : this.props.month;
   };
```
lhbrennan, line 86: The logic for determining what happens when a user clicks on a day in the calendar is spread out across three components:`Day`, `Calendar`, and `Datepicker`. 

`Day`'s `onSelect` method determines whether a selected day is considered a Start Date or End Date.
`Calendar`'s `handleDateChange` merges time values in with the day values to produce the full date.
`Datepicker`'s `onChange` method handles edge cases like Start Dates that fall after End Dates, and also manages whether the calendar should stay open or not after a selection.

It might make sense to organize this logic in a different way in the future, but for now I've tried to conform to the existing rough separation of concerns.

src/datepicker/datepicker.js
```diff
@@ -253,7 +262,6 @@ export default class Datepicker<T = Date> extends React.Component<
     }
 
     if (range && !separateRangeInputs) {
-      // TODO(LUKE): use INPUT_DELIMITER below
```
chasestarr, line 256: use `INPUT_DELIMITER` like the code above?

chasestarr (approved):
couple nits, but overall looks good!

src/datepicker/__tests__/datepicker-range-locked-behavior.e2e.js
```diff
@@ -0,0 +1,206 @@
...
+    // select a day for the End Date
+    endDateInput.click();
+    await page.waitForSelector(selectors.calendar);
+    await page.click(selectors.day);
+    await page.keyboard.press('Escape');
+    const endDateValue = await page.$$eval(
+      selectors.input,
+      inputs => inputs[1].value,
+    );
+    expect(endDateValue).toBe('2019/03/28');
```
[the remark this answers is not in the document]
lhbrennan: Yes, good suggestion.

src/datepicker/datepicker.js
lhbrennan: So it's unclear to me whether `INPUT_DELIMITER` should be ` – ` (as it is now) or `–`, which also shows up. For now I used the former, and only in cases that pertained to the dates themselves (rather than to the mask). To completely remove hardcoded en dashes from the file, I should use the latter instead, which would mean using spaces around INPUT_DELIMITER in many cases, as in ` ${INPUT_DELIMITER} `. I'll save this change for a separate PR.

lhbrennan: Changed my mind, I'll switch to using `INPUT_DELIMITER = '–'` for this PR.

PR #4810, opened 2022-02-24
fix(timezone-picker): correct offset value to align with prev versions

williamernest:
Fixes the onChange param type to align with previous versions, and Date.getTimezoneOffset(). Adds tests to verify any future changes.

### DIFF
--- a/src/timezonepicker/__tests__/timezone-picker.test.js
+++ b/src/timezonepicker/__tests__/timezone-picker.test.js
@@ -1,0 +1,65 @@
+/*
+Copyright (c) Uber Technologies, Inc.
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+// @flow
+
+import React from 'react';
+import {render, screen, fireEvent} from '@testing-library/react';
+
+import {TimezonePicker} from '../index.js';
+import {TestBaseProvider} from '../../test/test-utils.js';
+
+const mockedDate = new Date('2021-01-01T00:00:00Z');
+const Example = ({setDate}) => {
+  return <TimezonePicker value={mockedDate} onChange={setDate} />;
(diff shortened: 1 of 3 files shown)


src/timezonepicker/timezone-picker.js
```diff
@@ -73,10 +73,13 @@ class TimezonePicker extends React.Component<
...
         }
 
+        const offsetMinutes = offset * 60;
+
         timezones.push({
           id: zoneName,
           label,
-          offset,
+          // offset output is in minutes, difference of UTC and this zone (negative for hours ahead of UTC, positive for hours behind)
+          offset: offsetMinutes === 0 ? 0 : offsetMinutes * -1,
```
lhbrennan, line 82: would `offset: offsetMinutes * -1` work instead?

lhbrennan (approved):

src/timezonepicker/timezone-picker.js
williamernest: offsetMinutes* -1 returns -0 for 0 which isn’t accurate or what we would want to return in this case.

src/timezonepicker/__tests__/timezone-picker.test.js
```diff
@@ -0,0 +1,65 @@
...
+*/
+// @flow
+
+import React from 'react';
+import {render, screen, fireEvent} from '@testing-library/react';
+
+import {TimezonePicker} from '../index.js';
+import {TestBaseProvider} from '../../test/test-utils.js';
+
+const mockedDate = new Date('2021-01-01T00:00:00Z');
```
williamernest, line 15:
```suggestion
const mockedDate = '2021-01-01T00:00:00Z';
```

chasestarr (approved):

PR #4812, opened 2022-02-24
Range selection disallowed when includes excluded dates

lhbrennan:
#### Description
Previously, ranged Datepicker allowed a ranged selection to span across dates that have been explicitly excluded by the user via the `excludeDates` prop. This change disallows such range selections.

#### Scope
Patch: Bug Fix


src/datepicker/datepicker.js
```diff
@@ -104,6 +104,16 @@ export default class Datepicker<T = Date> extends React.Component<
           } else {
             nextDate = [start, start];
           }
+        } else if (
+          this.dateHelpers.dateRangeIncludesDates(
+            // $FlowFixMe Cannot call `this.dateHelpers.dateRangeIncludesDates` with `nextDate` bound to the first parameter because  read-only array type [1] is incompatible with  array type [2]
```
lhbrennan, line 109: It thinks `DateValueT` is a read-only array, but the actual type is `DateValueT<T> = ?T | Array<?T>`

chasestarr: I did a bunch of testing to see why this is erroring but no luck :(

src/datepicker/datepicker.js
```diff
@@ -104,6 +104,16 @@ export default class Datepicker<T = Date> extends React.Component<
...
+            nextDate,
+            this.props.excludeDates,
+          )
+        ) {
+          nextDate = this.props.value;
+          console.warn('Selected date range includes excluded dates');
```
chasestarr, line 115: Let's either remove or wrap in a `__DEV__` fence. End users will very likely not open up the console to see this message

chasestarr (approved):

src/datepicker/datepicker.js
lhbrennan: 👍

PR #4819, opened 2022-03-01
Remove radio group overrides pass through

lhbrennan:
#### Description
Previously, `RadioGroup` accepted overrides for `Radio` which it passed through to `Radio`. This change deprecates that behavior. Users should instead apply `Radio` overrides directly to instances of the `Radio` component itself.

#### Scope
Major: Breaking change


williamernest:
Should this be closed in favor of #4828

PR #4822, opened 2022-03-02
docs(roadmap): add button minimal -> tertiary to roadmap

williamernest:


chasestarr (approved):

bsplosion:
Is there any documentation why this was done?  Minimal had a distinct style from tertiary and removing it has caused some issues with design alignments and component definitions.  There doesn't appear to be any "transition" possible which retains distinction between the two since you're actually merging them and reducing the number of kinds defined.

williamernest:
@bsplosion Hey Brandon. This was removed from the design system (Base), and we held off on removing it until a major release. We didn't plan to implement a transition state or type since the minimal type was already deprecated in the design system. You can recreate the type using overrides, here's the PR we removed it in so you can reference the style changes from https://github.com/uber/baseweb/pull/4834

williamernest:
The design system is documented at https://base.uber.com/ . I wouldn't say implementing your own version of a removed component is hacking around a problem. Leaving in components/types/styles that are no longer supported becomes a maintenance burden for the team and also incorrectly implies that they're supported. The documentation could have been updated earlier, but upgrading a major version implies there's going to be some friction.

PR #4828, opened 2022-03-03
chore(radio): remove radio group overrides pass through

lhbrennan:
#### Description
Previously, `RadioGroup` accepted overrides for `Radio` which it passed through to `Radio`. This change deprecates that behavior. Users should instead apply `Radio` overrides directly to instances of the `Radio` component itself.

#### Scope
Major: Breaking change

### DIFF
--- a/documentation-site/components/yard/config/radio.ts
+++ b/documentation-site/components/yard/config/radio.ts
@@ -149,15 +149,7 @@ const RadioGroupConfig: TConfig = {
       type: PropTypes.Custom,
       description: 'Lets you customize all aspects of the component.',
       custom: {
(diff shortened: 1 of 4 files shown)


williamernest (approved):
Looks good. Only thing I would consider is adding a note on the radio page indicating this pass through behavior was removed in v11.

lhbrennan:
> Looks good. Only thing I would consider is adding a note on the radio page indicating this pass through behavior was removed in v11.

Good suggestion. Where would you put it? Perhaps above the "When to use" section?

image

williamernest:
Yup, above when to use will work.

PR #4831, opened 2022-03-04
Remove select searchicon override

lhbrennan:
#### Description
The `SearchIcon` override for `Select` is deprecated in favor of `SearchIconContainer`.

### DIFF
--- a/src/select/select-component.js
+++ b/src/select/select-component.js
@@ -866,13 +866,13 @@ class Select extends React.Component<PropsT, SelectStateT> {
     const sharedProps = this.getSharedProps();
 
     return (
-      // TODO(v11): remove searchIconProps from SearchIconContainer
-      <SearchIconContainer
-        {...sharedProps}
-        {...searchIconProps}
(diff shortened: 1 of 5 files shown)


src/select/select-component.js
```diff
@@ -859,20 +859,11 @@ class Select extends React.Component<PropsT, SelectStateT> {
...
-      overrides.SearchIcon,
-      SearchIconComponent,
-    );
     const sharedProps = this.getSharedProps();
 
     return (
-      // TODO(v11): remove searchIconProps from SearchIconContainer
-      <SearchIconContainer
-        {...sharedProps}
-        {...searchIconProps}
```
chasestarr, line 872: The `SearchIcon` override should be maintained but this part should be changed. Notice how this element is getting two different sets of overrides applied to it. View this PR that introduced it for more context https://github.com/uber/baseweb/pull/2397

Sort of tldr on it is one override was previously split into two and to maintain backwards compat the props from the original was applied to both

tajo (approved):

PR #4860, opened 2022-03-21
fix(input): allow for 'any' value for input step

leyanlo:
Allow for 'any' value for input step. See spec:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-step


chasestarr (approved):
Thank you

PR #4864, opened 2022-03-23
feat(select): imperatively change input value

lhbrennan:
#### Description
Previously the `controlRef` prop gave developers direct access to `Select`'s input element. It has been repurposed to offer developers imperative control over multiple internal methods for controlling `Select` imperatively:
- setInputValue
- setDropdownOpen
- setInputFocus
- setInputBlur

#### Scope
Minor: New Feature


src/select/__tests__/select-controlref.test.js
```diff
@@ -73,21 +73,61 @@ describe('setDropdownOpen', function () {
...
 });
+
+describe('setInputValue', function () {
+  it('correctly sets the input value', () => {
+    const options = [
+      { id: 'a', label: 'dragons' },
+      { id: 'b', label: 'unicorns' },
+      { id: 'c', label: 'elves' },
+    ];
+    const methodsRef = React.createRef();
```
williamernest, line 102: Is this test still valid or useful? methodsRef looks like it's removed.

src/select/types.js
```diff
@@ -132,8 +138,8 @@ export type PropsT = {
   id?: string,
   /** Defines if the comparison for a new creatable value should be case-insensitive. */
   ignoreCase?: boolean,
-  /** A ref to access the input element powering the select if it's a search select, or the container div if it isn't. */
-  controlRef?: ReactRefT<HTMLElement>,
+  /** An imperative handle exposing internal methods. To preserve backward compatibility. */
```
williamernest, line 141: This comment seems unfinished. To preserve backwards compatibility with what? The element? This type doesn't support an element.

src/select/__tests__/select-controlref.test.js
lhbrennan: Sorry that was sloppy. `methodsRef` should be `controlRef` here. I'm confused about why the test did not fail.

src/select/types.js
lhbrennan: Good catch, thanks.

PR #4866, opened 2022-03-25
fix(input): fix types to align with element expected

williamernest:


src/tree-view/types.js
```diff
@@ -8,7 +8,7 @@ LICENSE file in the root directory of this source tree.
 import * as React from 'react';
 import type { OverrideT } from '../helpers/overrides.js';
 
-export type ReactRefT<T> = {| current: null | T |};
+export type ReactRefT<T> = { current: null | T } | {| current: null | T |};
```
chasestarr, line 11: Is relaxing the exact constraint sufficient?

chasestarr (approved):
lgtm besides one question

shannonmoeller:
This introduced a ton of `// $FlowFixMe` comments in places where we use `<Input/>`. It feels like this broke more than it fixed. What was the original issue this was introduced to address?

src/tree-view/types.js
Prestaul: I'm trying to understand why `{ current: null | T } | {| current: null | T |}` rather than just `{ current: null | T }`. Can someone explain?

williamernest: createRef returns an exact type. useRef returns an inexact type.

williamernest:
> This introduced a ton of `// $FlowFixMe` comments in places where we use `<Input/>`. It feels like this broke more than it fixed. What was the original issue this was introduced to address? Also, why not just make it always inexact instead of trying to do both?

These were added to align our external code with our internal libraries/rules. It definitely introduced a lot of FlowFixMe comments in areas where typing didn't exist. In almost all instances that were updated with a comment, it was already implicitly using an "any" type (asterisk is unsafe when used across modules). 

As for inexact and exact, React itself returns an exact type for createRef and an inexact type for useRef. Happy to chat offline

shannonmoeller:
None of our cases used `any`. I'll DM you links.

PR #4884, opened 2022-03-31
feat(accordion): support arrow keys, home key, end key

### LINKED ISSUE #4255: Accordion component should support W3C WAI-ARIA Authoring Practices requirements
The accordion currently only supports navigation via TAB/Shift+TAB. It should support arrow keys and home/end.
Besides it has an error with the focus style (v9.115.1): each tabbed panel header gets a visible focus indicator.

## Feature description
See https://w3c.github.io/aria-practices/#accordion & https://w3c.github.io/aria-practices/examples/accordion/accordion.html

lhbrennan:
Fixes #4255 

#### Description

**Down Arrow**: If focus is on an accordion header, moves focus to the next accordion header. If focus is on the last accordion header, does nothing.

**Up Arrow**: If focus is on an accordion header, moves focus to the previous accordion header. If focus is on the first accordion header, does nothing.

**Home**: When focus is on an accordion header, moves focus to the first accordion header.

**End**: When focus is on an accordion header, moves focus to the last accordion header.

#### Scope
Minor: New Feature


src/accordion/accordion.js
```diff
@@ -61,12 +64,52 @@ export default class Accordion extends React.Component<AccordionPropsT, Accordio
...
+    }
+    if (e.keyCode === END) {
+      const lastItem = itemRefs[itemRefs.length - 1];
+      lastItem.current && lastItem.current.focus();
+    }
+    if (e.keyCode === ARROW_UP) {
+      const activeItemIdx = itemRefs.findIndex((item) => item.current === document.activeElement);
+      if (activeItemIdx > 0) {
+        const prevItem = itemRefs[activeItemIdx - 1];
+        prevItem.current && prevItem.current.focus();
```
williamernest, line 91: The page scrolls when using the arrow keys, so we should be preventing the default event behavior for each of the arrow keys.

lhbrennan: Thanks -- the same must be true for `home` and `end`, so they should also get prevent default yea?

lhbrennan: 🤔 strange. Working in ladle. I'll play around with it some more to see if I can reproduce unexpected behavior.

src/accordion/stateless-accordion.js
```diff
@@ -22,13 +23,54 @@ function StatelessAccordion({
 }: StatelessAccordionPropsT) {
   const { Root: RootOverrides, ...PanelOverrides } = overrides;
   const [Root, rootProps] = getOverrides(RootOverrides, StyledRoot);
+  const itemRefs = React.useRef([]);
+
+  const handleKeyDown = (e: KeyboardEvent) => {
```
williamernest, line 28: This functionality doesn't seem like it belongs in the stateless variant.

williamernest (approved):

PR #4906, opened 2022-04-08
fix(datepicker): prevent early parsing of enclosing date in range setup

### LINKED ISSUE #4891: Date picker input with range, unable to type days higher than 9
https://baseweb.design/components/datepicker/ 
with the range props selected on Chrome 
## Current Behavior
Type in the second part (end) of the range 2022/11/11, and you get 2022/11/01 instead, and you can't change it unless you select the date via the calendar.

Typing for the /dd (days) parts return the following.

- 30 return 00
- 1* return 01
- 2* return 02 
- N return 0N ... etc

Typing anything beyond 09 is not doable.

## Expected Behavior
I should be able to type the date without having to rely on selecting it in the calendar, especially for long range date across multiple months/years as it involves a lot of click.

## Your Environment

Version 99.0.4844.82 (Official Build) (64-bit)

In firefox I had no such issue.

| Tech    | Version |
| ------- | ------- |
| Base UI | v10.12.1  |
| React   |         |
| browser | Chrome  |

atroncy:
Fixes #4891 

#### Description

There is an hack used on the user input date when typing the days part of the date.
This hack was not replicated in the case of the user choosing the range options.
We applied this hack to the endDate of the range option (the hacks seems unecessary for the startDate somehow)

I'm not familiar on where/how I should implement test for this. 
Any guidance, appreciated.

#### Scope

Patch: Bug Fix


src/datepicker/datepicker.js
```diff
@@ -323,15 +332,7 @@ export default class Datepicker<T = Date> extends React.Component<DatepickerProp
...
-      const formatString = this.props.formatString;
-
-      // Prevent early parsing of value.
-      // Eg 25.12.2 will be transformed to 25.12.0002 formatted from date to string
-      if (dateString.replace(/(\s)*/g, '').length < formatString.replace(/(\s)*/g, '').length) {
-        date = null;
-      } else {
-        date = parseDateString(dateString);
-      }
+      date = parseEnclosingDate(dateString);
```
lhbrennan, line 335: Do we need line 334? Seems strange to define `date` and then immediately redefine it in the next line.

lhbrennan (commented):
As far as I can tell, this issue only surfaces when `Datepicker` pre-populates part of a date range (either the start date but not the end date, or the end date but not the start date). If you type in the full date range (either by deleting what's already in the input or configuring `Datepicker` without an initial date of any kind), it behaves as expected. Just noting this for completeness.

Along those lines, you mentioned that this issue only affects the end date, but it actually also effects the start date if the end date is populated but the start date is null. So I think it makes sense for `parseEnclosingDate` to factored into `parseDateString` so that it's applied uniformly. This also makes the code easier to reason about.

src/datepicker/datepicker.js
atroncy: I followed your feedback, I did not realize the same issue was present for startDate when typing it afterwards endDate

lhbrennan:
This change is causing flow and unit test failures. Can you run the tests locally and to determine the issue?

chasestarr:
Closing this in-progress pr to clean up the list. Please let us know if this is still an issue

Issue #4818, opened 2022-02-28
[FormControl] Boolean attributes `positive` and `error` considered `non-boolean`


simPod:
## Current Behavior

`<FormControl label={label}>...</FormControl>`

Gives me 

```
Warning: Received `false` for a non-boolean attribute `positive`.

If you want to write it to the DOM, pass a string instead: positive="false" or positive={value.toString()}.

If you used to conditionally omit it with positive={condition && value}, pass positive={condition ? value : undefined} instead.
```

```
Warning: Received `false` for a non-boolean attribute `error`.

If you want to write it to the DOM, pass a string instead: error="false" or error={value.toString()}.

If you used to conditionally omit it with error={condition && value}, pass error={condition ? value : undefined} instead.
```

in development (react.dom.development)

## Expected Behavior

No error

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v10.9.2  |
| React   |  v17       |
| browser |  chrome |

- [x] I have searched the [issues](https://github.com/uber/baseweb/issues) of this repository and believe that this is not a duplicate.

williamernest:
Can you reproduce this with a codesandbox? I'm not seeing an error when using the code you've provided.

simPod:
I'll have to investigate how react dom development kicks in. Will look into
it

houmark:
This seems to happen if you wrap a `<FormControl>` with a direct child such as a `div` or `<Block>`.

simPod:
Ah, cool! @williamernest https://codesandbox.io/s/base-web-formcontrol-forked-n0ptwf?file=/src/example.js

andriuss:
My use case is to use FormControl to display a label for the custom ButtonGroup + Button element.
**Tags** section in the form:
<img width="1236" alt="Screenshot 2022-04-05 at 10 01 13" src="https://user-images.githubusercontent.com/6738026/161697425-4fa8285c-a178-42ac-822f-42608f65c11a.png">

<img width="548" alt="Screenshot 2022-04-05 at 10 01 30" src="https://user-images.githubusercontent.com/6738026/161697418-e35ddd82-7ccd-45c9-a95d-d2e5adc3bf2e.png">

chasestarr:
[...]

PR #4912, opened 2022-04-09
feat: add type of props passed to FormControl child

simPod:
Related to #4818

#### Description

The props passed to child of FormControl seems to have the shape mentioned in https://github.com/uber/baseweb/issues/4818#issuecomment-1093925352

This PR adds the type definition.

#### Scope

Minor: New Feature


chasestarr (approved):
thank you

PR #4949, opened 2022-05-16
fix(typescript): fix HeaderNavigationProps type

Qinruiy:
Add HeaderNavigationProps missing children type defined


chasestarr (approved):
thank you

PR #4953, opened 2022-05-18
fix(popover): fix animateout property

williamernest:


chasestarr (approved):
popover fix lgtm, what's up with the select test changes? is that a result of `addTestStyles`?

williamernest:
> popover fix lgtm, what's up with the select test changes? is that a result of `addTestStyles`?

The select uses the popover to display the menu, but since it's no longer immediately removing those elements from the DOM (it waits until the animation completes), the test was querying elements that were transitioning out before they were gone (and expecting them to not be there)

Issue #4936, opened 2022-05-02
Select - controlRef.setInputBlur does not reset $isFocused


danroberts:
I am running into some issues trying to programmatically blur a select input which uses $isFocused to style a highlight state. I'm not 100% sure this is a bug, but it seems unintuitive to me. 

https://codesandbox.io/s/base-web-imperatively-control-dropdown-forked-8vh7go?file=/src/example.js

## Current Behavior

When you're using a control ref for a Select component, controlRef.current.setInputFocus() will result in a $isFocused state, however, controlRef.current.setInputBlur() does not result in $isFocused = false. 

This means that there is no way to programmatically blur the component if you are applying styles based on $isFocused
## Expected Behavior

setInputBlur() should result in $isFocused = false;

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v11.0.2  |
| React   |     v18.1.0    |
| browser |         |

- [x] I have searched the [issues](https://github.com/uber/baseweb/issues) of this repository and believe that this is not a duplicate.

PR #4960, opened 2022-05-20
fix(select): handleBlur should check event type before bound check

danroberts:
Fixes [#4936](https://github.com/uber/baseweb/issues/4936)

#### Description

When we handle a blur event in the select-component, we want to check if
the event is transferring focus to an element which is a child of the
select component before we un-focus.

Because handle blur takes FocusEvents or MouseEvents, there is a hole in
the logic, where a FocusEvent which does not have a relatedTarget (which
is what happens if you imperatively blur with element.blur() or via the
controlRef.setInputBlur() ) will pass to the second conditional check in
handleBlur, which should only be applied to mouse events (i.e. If you
click on a new element, we check if that new element (event.target) is
inside our anchor).

In the case of a FocusEvent, event.target is the element being blurred
(the input) so it will always be inside our anchor.

#### Scope
Patch: Bug Fix

### DIFF
--- a/src/select/__tests__/select-controlref.test.js
+++ b/src/select/__tests__/select-controlref.test.js
@@ -7,7 +7,7 @@ LICENSE file in the root directory of this source tree.
 // @flow
 import * as React from 'react';
 import { act } from 'react-dom/test-utils';
-import { render, fireEvent, screen } from '@testing-library/react';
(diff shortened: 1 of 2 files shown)


chasestarr (approved):
Thank you

PR #4961, opened 2022-05-23
feat(payment-card): added uatp card support

ayush712:
#### Description
A new card type(UATP) support is added in this PR.

Created a custom card config array in custom-cards.config.js file,
So that in future if someone wants to add a new card type support, 
he/she can add the configuration to custom-cards.config.js file itself.
He would not need to add the configuration in component file.

Because we support card icons as well so he would need to add a new icon 
and add it to CardTypeToComponent variable.

A UATP card has a fixed length of 15 digits and its bin range lies between 1001 and 1999. 
There is no CVV for UATP cards.

#### Scope
Minor: New Feature


chasestarr:
LGTM, please review and merge in the new visual snapshot changes

PR #4964, opened 2022-05-23
feat(badge): implement new badge component

lhbrennan:
#### Description

New badge component. Refer to the [design specs](https://zeroheight.com/6d2425e9f/v/latest/p/286e31-badge/b/69e663) for this component.

#### Scope
Minor: New Feature


src/badge/styled-components.js
```diff
@@ -0,0 +1,151 @@
...
+        color: $theme.colors.contentWarning,
+        backgroundColor: $theme.colors.backgroundLightWarning,
+      },
+    },
+  };
+
+  return COLOR_STYLES[$hierarchy][$color];
+}
+
+export const StyledRoot = styled<{}>('div', () => {
```
lhbrennan, line 66: Is `Root` the best name for this component? I ask because it actually wraps around the component that the badge is positioned relative to (rather than the badge itself, which is what some might expect when seeing "root"). One alternative could be `Container`.

documentation-site/examples/badge/offsets.js
```diff
@@ -0,0 +1,35 @@
...
+  return (
+    <React.Fragment>
+      <Badge
+        content="Promotion"
+        shape={SHAPE.pill}
+        color={COLOR.positive}
+        placement={PLACEMENT.bottom}
+        verticalOffset="8px"
+      >
+        <div style={styles}>
```
chasestarr, line 25: Does this work without the wrapper?

src/badge/types.js
```diff
@@ -0,0 +1,46 @@
...
+};
+
+export type BadgePropsT = {
+  horizontalOffset?: string,
+  verticalOffset?: string,
+  placement?: PlacementT,
+  children: ?React.Node,
+  content: ?React.Node,
+  overrides?: BadgeOverridesT,
+} & InlineBadgePropsT;
```
chasestarr, line 46: I recommend switching to exact types and spreading to union `InlineBadgePropsT` with `BadgePropsT`. I believe in the current state it will resolve the `overrides` property to `InlineBadgeOverridesT` which I assume is not what you wanted.

What's the practical difference between `children: ?React.Node` and `children?: React.Node`? I've only written the latter.

```suggestion
export type InlineBadgePropsT = {|
  color?: ColorT,
  shape?: ShapeT,
  hierarchy?: HierarchyT,
  children?: React.Node,
  overrides?: InlineBadgeOverridesT,
  hidden?: boolean,
|};

export type BadgePropsT = {|
  ...InlineBadgePropsT,
  horizontalOffset?: string,
  verticalOffset?: string,
  placement?: PlacementT,
  children?: React.Node,
  content?: React.Node,
  overrides?: BadgeOverridesT,
|};
```

src/badge/styled-components.js
```diff
@@ -0,0 +1,151 @@
...
+}>('div', ({ $theme, $role, $placement, $horizontalOffset, $verticalOffset }) => {
+  let positionStyle = POSITION_STYLES[$placement][$role];
+
+  if ($verticalOffset) {
+    if (
+      $placement === PLACEMENT.topLeft ||
+      $placement === PLACEMENT.top ||
+      $placement === PLACEMENT.topRight
+    ) {
+      positionStyle = { ...positionStyle, top: $verticalOffset };
```
chasestarr, line 119: nit, but no need to spread when only assigning a value to a property is required. similar could be done for the rest of these cases

```suggestion
      positionStyle.top = $verticalOffset;
```

src/badge/styled-components.js
```diff
@@ -0,0 +1,151 @@
...
+    }
+    if ($placement === PLACEMENT.topRight || $placement === PLACEMENT.bottomRight) {
+      positionStyle = { ...positionStyle, right: $horizontalOffset };
+    }
+  }
+
+  return {
+    ...positionStyle,
+    position: 'absolute',
+    // TODO(LUKE): this seems hacky 🤔
```
chasestarr, line 147: agree :)

src/badge/styled-components.js
```diff
@@ -0,0 +1,151 @@
...
+        : $theme.typography.ParagraphSmall),
+    };
+  }
+);
+
+export const StyledPositioner = styled<{
+  $role: RoleT,
+  $placement: PlacementT,
+  $horizontalOffset: ?OffsetT,
+  $verticalOffset: ?OffsetT,
```
chasestarr, line 109: 🤷 I would just use string since `Offset` is not a union

src/badge/styled-components.js
```diff
@@ -0,0 +1,151 @@
...
+  'div',
+  ({
+    $theme,
+    $shape = SHAPE.rectangle,
+    $color = COLOR.accent,
+    $hierarchy = HIERARCHY.primary,
+    $hidden,
+  }) => {
+    return {
+      visibility: $hidden ? 'hidden' : 'inherit',
```
chasestarr, line 88: In what case would someone want hidden?

src/badge/constants.js
```diff
@@ -0,0 +1,96 @@
...
+  bottomLeft: 'bottomLeft',
+});
+
+export const ROLES = Object.freeze({
+  badge: 'badge',
+  hintDot: 'hintDot',
+  notificationCircle: 'notificationCircle',
+});
+
+export const POSITION_STYLES = {
```
chasestarr, line 42: should this also be frozen?

src/badge/badge.js
```diff
@@ -0,0 +1,74 @@
...
+        'Secondary badges should not be positioned. Use the inline version of this badge instead.'
+      );
+    }
+  }
+
+  const anchor = getAnchorFromChildren(children);
+  // If there's no anchor, render the badge inline
+  if (!anchor) {
+    return (
+      <Badge color={color} shape={shape} hidden={hidden} {...badgeProps}>
```
chasestarr, line 51: should `InlineBadge` be removed if the use-case can be handled here? If not, I would remove this branch and fail to render

documentation-site/examples/badge/offsets.js
lhbrennan: I'm not sure I understand the question. We can certainly remove the div with the styles that wraps Avatar -- Badge will still work as expected. I included it because it a situation with an image or icon inlayed inside a tile or box seems like a situation where someone might want to manually position the badge.

src/badge/badge.js
lhbrennan: Thanks. Having removed `InlineBadge`, everything is much cleaner :)

src/badge/constants.js
lhbrennan: Indeed, good eye!

src/badge/styled-components.js
lhbrennan:
```
  hideBadge ? (
    <Card>Card Content</Card>
  ) : (
    <Badge>
      <Card>Card Content</Card>
    </Badge>
  );
```

VS

```
<Badge hidden={hideBadge}>
  <Card>Card Content</Card>
</Badge>
```

src/badge/styled-components.js
lhbrennan: 👍

src/badge/types.js
lhbrennan: According to the Flow docs, it [returns the type of the first property that matches its name](https://flow.org/en/docs/types/intersections/#toc-intersections-of-object-types). (Definitely seems backwards from a javascript perspective). So, I put `InlineBadgePropsT` after the rest of `BadgePropsT` to avoid the issue of `overrides` resolving to that of `InlineBadgePropsT`.

That said I prefer your approach from a readability standpoint.

_That_ said all of this is mute because I was able to remove `InlineBadgePropsT` and related types now that I've removed `InlineBadge` per your recommendation!

documentation-site/examples/notification-circle/offset.js
```diff
@@ -0,0 +1,22 @@
...
+
+export default function Example() {
+  return (
+    <React.Fragment>
+      <NotificationCircle
+        content={11}
+        color={COLOR.positive}
+        placement={PLACEMENT.topRight}
+        horizontalOffset="0px"
+        verticalOffset="0px"
```
chasestarr, line 14: is setting offsets to 0 necessary?

chasestarr (approved):

documentation-site/examples/notification-circle/offset.js
lhbrennan: The offsets override the default positions given the type of badge and its placement. In this case, the badge is being placed on a Tag, and the Tag component has margin, so the default positioning places the badge floating a few pixels away from the edge of the Tag. It just so happens that setting the offsets to `0px` lines up the badge nicely. But I do see how that's confusing -- perhaps it makes sense to choose a different element as an anchor.

PR #4977, opened 2022-06-01
fix(types): update typescript definitions to better match the implementation

zxbodya:
#### Description

- update typescript version
- update styletron types definitions
- move type definitions files to better match where the source implementation is
- update type definitions to better match the implementation
- remove `--skipLibCheck` which is no longer needed after the fixes


documentation-site/examples/breadcrumbs/pseudo.tsx
```diff
@@ -1,13 +1,14 @@
...
 import {StyledLink as Link} from 'baseui/link';
+import {StyleObject} from 'styletron-react';
 
 export default function Example() {
   return (
     <Breadcrumbs
       overrides={{
         ListItem: {
-          style: ({$itemIndex, $theme}) => {
+          style: ({$itemIndex, $theme}): StyleObject => {
```
chasestarr, line 11: Was this required after the type def changes? Wondering if I should categorize this as a breaking api change

zxbodya: double-checked it - this change is not actually required for current type definitions to work

let me revert it

chasestarr (approved):
thank you

PR #4997, opened 2022-06-22
fix(app-nav-bar): use grid maxWidth

chasestarr:
#### Description

I mistakenly used the large breakpoint value as the app-nav-bar content maxWidth. It was previously set to grid maxWidth and this PR restores that

#### Scope
Patch: Bug Fix


jacquesg:
Still some issues here.

image

jacquesg:
The styling of the Primary and Secondary menu containers are not symmetric. Its not possible currently to change for example the background and font colors of the Primary (say to blue and white) and have the the Secondary say white and black.

https://vimeo.com/723714087/5e29aa0b93

jacquesg:
The tweaks made to this component has significantly broken our code to the extent that we had to replace the components completely. It isn't possible to just change the styling.

jacquesg:
Breakages observed since 11.0.0:

image

Previously:

image

PR #4999, opened 2022-06-22
chore(button): make button borders rounded

lhbrennan:
#### Description

Square and rectangular buttons should have border radius of 8px

#### Scope
Patch: Bug Fix


tajo (approved):

madhusudhand:
Button group design seems to be affected in a negative way after this change.

| Before | After |
| --- | --- |
| image | image |

Somehow they no longer appear to be a group. Feels like individual buttons.

PR #5002, opened 2022-06-23
feat(file-uploader): progress bar, cancel & retry buttons overrides

### LINKED ISSUE #4995: [FileUploader] Cancel, Retry & ProgressBar overrides
## Feature description
Currently `BaseButton` overrides in FileUploader are applied only to `Button` component used before file drop and it's not possible to add custom styles to **Cancel** & **Retry** buttons showed `afterFileDrop`. From what I observed, `prefixedStyledProps` & `buttonProps` are not passed down to those components.

Additionally it would be good to have overrides for `ProgressBar` component too, just like it was already done for `Spinner`.

Is there anyone working on this feature? If not, I would be more than happy to help.

kstoklosa:
Fixes #4995 

#### Description

- Allow to override buttons used after file drop as `RetryButtonComponent` & `CancelButtonComponent`. 
- Overrides can be now applied to `ProgressBar` component
- Documentation update for FileUploader (with the information about new overrides)


chasestarr (approved):
Thank you

PR #5011, opened 2022-07-01
chore(flow): Add flow declarations without implementaion

zxbodya:
#### Description

Removing implementation from `.js.flow` files, merging them to single definition file per component for most cases.

This also fixed issue with flow when in many cases types are ignored, because apparently flow does not allow having implementation in `.js.flow` in some cases.


chasestarr:
Closing this in-progress pr to prep for sync out switch

PR #5012, opened 2022-07-03
fix(styletron): correct StyletronComponent signature

simPod:
#### Description

https://github.com/uber/baseweb/pull/5006 broke TS since type declaration has changed

`TS2314: Generic type 'StyletronComponent' requires 1 type argument(s).`

_I hope [this](https://github.com/uber/baseweb/pull/5009) gets merged so such issues won't happen. Flow was great but it's TS era now (no offense)_

#### Scope
Patch: Bug Fix

### DIFF
--- a/src/accordion/index.d.ts
+++ b/src/accordion/index.d.ts
@@ -123,10 +123,10 @@ export class StatefulPanelContainer extends React.Component<
   ): void;
 }
 
-export declare const StyledRoot: StyletronComponent<any, any>;
-export declare const StyledPanelContainer: StyletronComponent<any, any>;
-export declare const StyledHeader: StyletronComponent<any, any>;
-export declare const StyledContent: StyletronComponent<any, any>;
-export declare const StyledContentAnimationContainer: StyletronComponent<any, any>;
-export declare const StyledToggleIcon: StyletronComponent<any, any>;
-export declare const StyledToggleIconGroup: StyletronComponent<any, any>;
+export declare const StyledRoot: StyletronComponent<any>;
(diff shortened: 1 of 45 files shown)


zxbodya:
that is intended change to have two type arguments in `StyletroinComponent`, it is coming with a new version of styletron(https://github.com/uber/baseweb/pull/5006/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L196-R195)

in `@types/styletron-react` there was only one type argument for component props, but in the new types now available as part of updated styletron-react package there is also element type - which allows to correctly check types when using `$as` property

simPod:
@zxbodya upgrading from 11.1.2 to 11.2.0 fails for TS users, since `StyletronComponent` has one argument. what should TS user do in order to upgrade?

`styled` from `import { styled } from 'baseui';` returns any because the generic signature is wrong now.

I'm using latest 5.0.3 v of `@types/styletron-react`. 

How to upgrade to baseui 11.2 then? It's using wrong types it seems.

zxbodya:
baseui 11.2.0, is to relying on types published in:
 - `styletron-react@6.1.0-beta.1`
 - `styletron-standard@3.1.0-beta.1`

(`@types/styletron-react` should not be used, and can be removed after updating styletron)

simPod:
Shouldn't have been the versions of styletron peer dependencies be raised to 6.1 and 3.1? Cuz it's effectively broken for TS users now.

Though I understand I should update those dependencies before upgrading to 11.2. Hope other users find this too.

PR #5013, opened 2022-07-05
fix(badge): avoid warning for inline badges

plougsgaard:
#### Description

Badges may have 0 or 1 children. This prevents the erroneous error when calling `Badge` with 0 children (inline badge).

#### Scope

Patch: Bug Fix


lhbrennan (approved):
Looks good, thanks!

PR #5014, opened 2022-07-05
fix: allow input focus when readonly

### LINKED ISSUE #4992: Input and Textarea with readOnly prop cannot be focused
Input and Textarea fields cannot be focused if the readOnly prop is true. 
This prevents the read-only fields from visually showing that they are focused and also makes them no longer emit `focus` and `blur` events.

This behavior is correct for fields with disabled attribute, but not for fields with readonly attribute. 

> The difference between `disabled` and `readonly` is that read-only controls can still function and are still focusable, whereas disabled controls can not receive focus and are not submitted with the form and generally do not function as controls until they are enabled.

https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly#attribute_interactions

Sandbox: https://codesandbox.io/s/base-web-input-forked-7f9vwc?file=/src/example.js

## Current Behavior

Input and Textarea fields do not indicate focused state and do not emit focus and blur events when readOnly prop is true. 

## Expected Behavior

Input and Textarea fields should indicate focused state and should emit focus and blur events when readOnly prop is true. 

## Your Environment

| Tech    | Version |
| ------- | ------- |
| Base UI | v11.1.0  |
| React   |  17.0.2        |
| browser | Latest Firefox and Chromium browsers |

- [ ✔] I have searched the [issues](https://github.com/uber/baseweb/issues) of this repository and believe that this is not a duplicate.

selfdocumentingcode:
Fixes #4992

#### Description

Input and Textarea fields cannot be focused if the readOnly prop is true. 
This prevents the read-only fields from visually showing that they are focused and also makes them no longer emit `focus` and `blur` events.

This behavior is correct for fields with disabled attribute, but not for fields with readonly attribute. 

> The difference between `disabled` and `readonly` is that read-only controls can still function and are still focusable, whereas disabled controls can not receive focus and are not submitted with the form and generally do not function as controls until they are enabled.

https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly#attribute_interactions

Sandbox: https://codesandbox.io/s/base-web-input-forked-7f9vwc?file=/src/example.js

#### Scope
Patch: Bug Fix


chasestarr (commented):
Sorry to let this sit for a while. This change looks good to me, but we've converted over to typescript. Do you mind updating to that?

selfdocumentingcode:
No worries. I will, but it'll have to be next week.

selfdocumentingcode:
@chasestarr Hi. I updated the .tsx files.

chasestarr (approved):

PR #5027, opened 2022-07-12
Divider component

lhbrennan:
#### Description

Adds new `Divider` component

image

#### Scope
Minor: New Feature


src/divider/styled-components.ts
```diff
@@ -0,0 +1,35 @@
...
+  const borderResetStyles = {
+    borderRight: 'none',
+    borderBottom: 'none',
+    borderLeft: 'none',
+  };
+
+  if ($size === SIZE.cell) {
+    return {
+      ...borderResetStyles,
+      borderTop: `1px solid ${$theme.colors.borderOpaque}`,
```
tajo, line 22: We should still probably keep using long hand props only to be safe?

tajo (approved):

src/divider/styled-components.ts
lhbrennan: Good catch

src/divider/styled-components.ts
```diff
@@ -0,0 +1,41 @@
...
+Copyright (c) Uber Technologies, Inc.
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+import { styled } from '../styles/index';
+import { SIZE } from './constants';
+import type { DividerPropsT } from './types';
+
+export const StyledDivider = styled<'hr', DividerPropsT>('hr', (props) => {
```
chasestarr, line 11:
```suggestion
import type { DividerProps } from './types';

export const StyledDivider = styled<'hr', DividerProps>('hr', (props) => {
```

src/divider/types.js.flow
```diff
@@ -0,0 +1,13 @@
+/*
+Copyright (c) Uber Technologies, Inc.
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+// @flow
+import { SIZE } from './constants.js';
+
+export type DividerPropsT = {
```
chasestarr, line 10:
```suggestion
export type DividerProps = {
```

PR #5028, opened 2022-07-13
Created File for overrides on label grouped select

Martijnme:
#### Description
When working with the select, I noticed on the website their was no override for the label of the select. 

#### Scope
Minor: New Feature


chasestarr:
Thank you for submitting this pr. I'd like to minimize this use-case and not provide a distinct example

PR #5031, opened 2022-07-14
fix(file-uploader): add missing types for overrides

kstoklosa:
This PR fixes TS error that was displayed while trying to override `ProgressBar`, `CancelButtonComponent` or `RetryButtonComponent` (types for that overrides were missing).

#### Description

`CancelButtonComponent`, `ProgressBar` & `RetryButtonComponent` are now part of `FileUploaderOverrides` interface.

Patch: Bug Fix


chasestarr:
Hi, thank you for putting this up. It has been addressed at some point here https://github.com/uber/baseweb/blob/master/src/file-uploader/types.ts#L27

PR #5032, opened 2022-07-14
fix(types): backward compatibility fixes

zxbodya:
#### Description

Improving backward compatibility with "old" typescript type definitions, by adding aliases and in some cases adding more types.

### DIFF
--- a/src/card/index.ts
+++ b/src/card/index.ts
@@ -5,7 +5,7 @@ This source code is licensed under the MIT license found in the
 LICENSE file in the root directory of this source tree.
 */
 
-import type {CardProps} from '../../legacy-dts/card';
+import type { CardProps } from './types';
 
 export { default as Card, hasThumbnail } from './card';
 export {
--- a/src/card/index.ts
+++ b/src/card/index.ts
@@ -5,6 +5,8 @@ This source code is licensed under the MIT license found in the
 LICENSE file in the root directory of this source tree.
 */
 
+import type {CardProps} from '../../legacy-dts/card';
+
 export { default as Card, hasThumbnail } from './card';
 export {
   Action as StyledAction,
@@ -17,3 +19,5 @@ export {
   Root as StyledWrapper,
 } from './styled-components';
 export * from './types';
+/** @deprecated To be removed in future versions */
+type hasThumbnail = CardProps['hasThumbnail'];
--- a/src/checkbox/index.ts
+++ b/src/checkbox/index.ts
@@ -4,6 +4,8 @@ Copyright (c) Uber Technologies, Inc.
 This source code is licensed under the MIT license found in the
 LICENSE file in the root directory of this source tree.
 */
+import type { CheckboxReducerState } from './types';
+
 export { default as StatefulCheckbox } from './stateful-checkbox';
 export { default as StatefulContainer } from './stateful-checkbox-container';
 export { default as Checkbox } from './checkbox';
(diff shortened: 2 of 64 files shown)


src/card/index.ts
```diff
@@ -17,3 +19,5 @@ export {
   Root as StyledWrapper,
 } from './styled-components';
 export * from './types';
+/** @deprecated To be removed in future versions */
+type hasThumbnail = CardProps['hasThumbnail'];
```
chasestarr, line 23: should CardComponents also be aliased?

src/checkbox/index.ts
```diff
@@ -21,3 +23,5 @@ export { STATE_TYPE, STYLE_TYPE, LABEL_PLACEMENT } from './constants';
 
 // Flow
 export * from './types';
+/** @deprecated use CheckboxReducerState instead. To be removed in future versions.*/
+export type initialState = CheckboxReducerState;
```
chasestarr, line 27:
```suggestion
export type StateT = CheckboxReducerState;
```

src/link/index.tsx
```diff
@@ -47,3 +47,6 @@ export const StyledLink = withWrapper<typeof Link, { animateUnderline?: boolean
       );
     }
 );
+
+/** @deprecated To be removed in future versions.*/
+export type LinkProps = React.ComponentProps<typeof StyledLink>;
```
chasestarr, line 52: I couldn't find a reference to this, is it required?

src/select/index.ts
```diff
@@ -4,6 +4,9 @@ Copyright (c) Uber Technologies, Inc.
 This source code is licensed under the MIT license found in the
 LICENSE file in the root directory of this source tree.
 */
+import type { Optgroups, Options } from './types';
+import { SelectProps } from '../../legacy-dts/select';
```
chasestarr: it looks like this is not used in the file, required?

src/timezonepicker/types.ts
```diff
@@ -28,6 +28,9 @@ export type Timezone = {
    */
   offset: number;
 };
+export type TimezoneOverrides = {
```
chasestarr, line 31:
```suggestion
export type TimezonePickerOverrides = {
```

src/timezonepicker/types.ts
```diff
@@ -45,9 +48,7 @@ export type TimezonePickerProps = {
   mapLabels?: (option: Option) => string;
   /** Callback for when the timezone selection changes. */
   onChange?: (value?: Timezone | null) => unknown;
-  overrides?: {
-    Select?: Override;
-  };
+  overrides?: TimezoneOverrides;
```
chasestarr, line 51:
```suggestion
  overrides?: TimezonePickerOverrides;
```

src/checkbox/index.ts
zxbodya: this is alias to replace old type definition from here:  https://github.com/uber/baseweb/blob/master/legacy-dts/checkbox/index.d.ts#L21-L24 so initialState is indeed a correct name, but I just noticed it is not actually exported… -- will remove this

src/link/index.tsx
zxbodya: it is for type that was exported here - https://github.com/uber/baseweb/blob/master/legacy-dts/link/index.d.ts#L4-L11

src/timezonepicker/types.ts
zxbodya: that is what we have in old type definitions - https://github.com/uber/baseweb/blob/master/legacy-dts/timezonepicker/index.d.ts#L12-L14

zxbodya: but yeah, looks like a good idea to rename it here and create deprecated alias with old name

chasestarr (approved):

PR #5033, opened 2022-07-15
fix(types): accordion, relax children type

zxbodya:
#### Description

Current type does not allow to pass children like this:

```jsx
    <Accordion>
      {[
          <Panel key="1">1</Panel>,
          <Panel key="2">2</Panel>
      ]}
      <Panel key="3">3</Panel>
    </Accordion>
```

Making the type less strict (ReactNode instead of only ReactElement), also adding support for primitive values

### DIFF
--- a/src/accordion/stateless-accordion.tsx
+++ b/src/accordion/stateless-accordion.tsx
@@ -23,9 +23,18 @@ function StatelessAccordion({
   return (
     <Root data-baseweb="accordion" {...rootProps}>
       {React.Children.map(children, (child, index) => {
-        const key = child.key || String(index);
-        return React.cloneElement(child, {
-          disabled: child.props.disabled || disabled,
+        let normalizedChild =
+          typeof child === 'object' ? (
+            (child as  // skip {} from type definition
+              | React.ReactElement<any, string | React.JSXElementConstructor<any>>
+              | React.ReactPortal)
+          ) : (
+            // if primitive value - wrap it in a fragment
+            <>{child}</>
+          );
+        const key = normalizedChild.key || String(index);
+        return React.cloneElement(normalizedChild, {
+          disabled: normalizedChild.props.disabled || disabled,
           expanded: expanded.includes(key),
           key,
           onChange:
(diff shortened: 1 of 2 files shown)


src/accordion/stateless-accordion.tsx
```diff
@@ -23,9 +23,18 @@ function StatelessAccordion({
...
+            (child as  // skip {} from type definition
+              | React.ReactElement<any, string | React.JSXElementConstructor<any>>
+              | React.ReactPortal)
+          ) : (
+            // if primitive value - wrap it in a fragment
+            <>{child}</>
+          );
+        const key = normalizedChild.key || String(index);
+        return React.cloneElement(normalizedChild, {
+          disabled: normalizedChild.props.disabled || disabled,
```
zxbodya, line 37: as an option we can just ignore "unexpected elements", like this:

```js
{React.Children.map(children, (child: React.ReactElement | React.ReactPortal, index) => {
        if (!child.props) return;
```

src/accordion/stateless-accordion.tsx
```diff
@@ -23,9 +23,18 @@ function StatelessAccordion({
   return (
     <Root data-baseweb="accordion" {...rootProps}>
       {React.Children.map(children, (child, index) => {
-        const key = child.key || String(index);
-        return React.cloneElement(child, {
-          disabled: child.props.disabled || disabled,
+        let normalizedChild =
+          typeof child === 'object' ? (
```
chasestarr, line 27: Will `ReactIs.isElement(child)` handle the type cast?

chasestarr: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-is/index.d.ts#L27

chasestarr (approved):

src/accordion/stateless-accordion.tsx
zxbodya: yes, that would allow to remove typecast.

updated to use it, also added check for isPortal - that looks like another allowed case here

PR #5038, opened 2022-07-18
fix(ts): add $style prop to block component

chasestarr:
#### Description

Updated the `BlockComponentType` interface to be closer to what styletron-react publishes. Seeing errors like below after the TS conversion

```
  | 2022-07-15 16:57:13 PDT | src/components/rules-container.tsx:15:7 - error TS2322: Type '{ children: Element[]; flexGridColumnCount: number; flexGridColumnGap: string; flexGridRowGap: string; marginBottom: string; display: "flex"; flex: string; flexDirection: "row"; $style: { ...; }; }' is not assignable to type 'IntrinsicAttributes & { flexGridColumnCount?: Responsive<number>; flexGridColumnGap?: Responsive<Scale>; flexGridRowGap?: Responsive<...>; } & BlockProps<...> & { ...; } & Omit<...>'.
  | 2022-07-15 16:57:13 PDT | Property '$style' does not exist on type 'IntrinsicAttributes & { flexGridColumnCount?: Responsive<number>; flexGridColumnGap?: Responsive<Scale>; flexGridRowGap?: Responsive<...>; } & BlockProps<...> & { ...; } & Omit<...>'. Did you mean 'style'?
  | 2022-07-15 16:57:13 PDT |  
  | 2022-07-15 16:57:13 PDT | 15       $style={{
  | 2022-07-15 16:57:13 PDT | ~~~~~~
```

#### Scope
Patch: Bug Fix


src/block/types.ts
```diff
@@ -403,11 +404,17 @@ export type StyledBlockProps = {
   $whiteSpace?: Responsive<WhiteSpace>;
 };
 
+type BaseProps<P extends {}> = P & {
+  $style?: StyleObject | ((props: P) => StyleObject);
+  className?: string;
+};
+
+type AddStyletronRef<P extends { ref: any }> = P extends { ref: infer R } ? P & { ref?: R } : P;
```
zxbodya, line 412: should this be `P & { $ref?: R }` like it is in styletron?

currently this helper type, does not really do anything(I somehow missed that in previous version of `BlockComponentType` as well)

chasestarr: Yeah I can change to $ref. Is that what you mean by it does not do anything? Or are you saying it should be removed?

zxbodya: yes, changing it to $ref will align it with styletron

(in a current state `P & { ref?: R }` is effectively the same as just `P`, because it is used only when `ref` is already in `P`)

PR #5040, opened 2022-07-19
chore: ladle v2

tajo:
- upgrades to Ladle v2 (Vite v3), even faster, prevents some reloads during cold starts
- removing flow plugin from the Ladle setup since we use typescript now
- some changes to the main provider to correctly sync styletron styles between iframed and standard version of stories


.ladle/components.tsx
```diff
@@ -1,17 +1,44 @@
...
   globalState: {
     theme: string;
     rtl: string;
+    story: string;
+    width: number;
   };
-}> = ({children, globalState}) => {
+}> = ({ children, globalState }) => {
+  const [engine, setEngine] = React.useState(
+    new Styletron({ container: getDocument(globalState.story).head })
```
chasestarr, line 24:
```suggestion
    () => new Styletron({ container: getDocument(globalState.story).head })
```

Does constructing have any side effects? https://reactjs.org/docs/hooks-reference.html

chasestarr (approved):

PR #5055, opened 2022-07-21
chore(banner): export banner override types

austingreco:
#### Description

When trying to add typing for banner overrides, I found the types not exported like other components:

```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/components/common/base-overrides/banner.js:4:65

Cannot import OverridesT because there is no OverridesT export in baseui/banner. [missing-export]
```

#### Scope

Patch


src/banner/types.js.flow
```diff
@@ -33,6 +33,18 @@ export type ArtworkContentT = {|
   type?: ArtworkTypeT,
 |};
 
+export type OverridesT = {
```
chasestarr, line 36: Can you please maintain the exact pipes? https://flow.org/en/docs/types/objects/#toc-exact-object-types

austingreco: :thumbsup: updated

chasestarr (approved):

PR #5068, opened 2022-07-29
feat(message-card): build message card

lhbrennan:
#### Description

New `MessageCard` component

#### Scope
Minor: New Feature


src/message-card/utils.ts
```diff
@@ -0,0 +1,141 @@
...
+const POOR_CONTRAST_COLORS = [colors.red300, colors.gray500, colors.yellow600];
+
+export function getBackgroundColorType(backgroundColor: string) {
+  if (__DEV__ && POOR_CONTRAST_COLORS.includes(backgroundColor)) {
+    console.warn(
+      `The provided value for backgroundColor, ${backgroundColor}, is not supported because \
+        it does not pass accessibility contrast tests for either white or black text.`
+    );
+  }
+  if (LIGHT_COLORS.includes(backgroundColor)) {
```
lhbrennan, line 134: I thought about using a map instead of an array for `LIGHT_COLORS` and `DARK_COLORS` so I can get constant time lookup here. Seems like the performance gain is negligible, but let me know if you disagree.

src/message-card/types.js.flow
```diff
@@ -0,0 +1,71 @@
...
+  backgroundColorType?: BackgroundColorTypeT,
+  image?: {
+    src: string,
+    layout?: ImageLayoutT,
+    objectFit?: ObjectFitT,
+  },
+  overrides?: MessageCardOverridesT,
+};
+
+export type StyledRoot = StyletronComponent<
```
lhbrennan, line 55: I'm not entirely confident that these exports at the bottom here are correctly typed -- please take a close look.

src/message-card/constants.ts
```diff
@@ -0,0 +1,22 @@
+/*
+Copyright (c) Uber Technologies, Inc.
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+export const BACKGROUND_COLOR_TYPE = {
```
lhbrennan, line 7: `BACKGROUND_COLOR_TYPE` is bulky. Welcoming suggestions for alternate ways to name this constant.

src/message-card/index.ts
```diff
@@ -0,0 +1,10 @@
+/*
+Copyright (c) Uber Technologies, Inc.
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+export { default as MessageCard } from './message-card';
+export * from './constants';
+export * from './styled-components';
+export * from './types';
```
lhbrennan, line 10: `MessageCard` uses the `secondary` and `tertiary` values from the `Button`'s `KIND` constant. Does it make sense to duplicate that constant in `MessageCard`'s `constants.ts` file and expose it? Or should devs using `MessageCard` import the `KIND` constant from `baseui/button`?

src/message-card/types.ts
```diff
@@ -0,0 +1,35 @@
...
+  ParagraphContainer?: Override;
+  Button?: Override;
+  Image?: Override;
+};
+
+export interface MessageCardProps {
+  onClick: (a: SyntheticEvent<HTMLButtonElement>) => unknown;
+  heading?: string;
+  paragraph?: string;
+  buttonLabel?: string;
```
lhbrennan, line 25: Design uses the word "Call to Action" instead of "Button". Semantically, "Call to Action" is more accurate since the entire MessageCard surface is a tap target, whereas the button itself is inert. But it seems to me that anyone familiar with Base Web would think of this element as "Button" since it does use the Base Web `Button` component. Open to suggestions.

src/message-card/index.ts
lhbrennan: I ended up creating a `BUTTON_KIND` constant to export from `MessageCard`. I couldn't get the docs site working otherwise.

src/message-card/utils.ts
chasestarr: try using a `Set`

src/message-card/types.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export interface MessageCardProps {
+  onClick: (a: SyntheticEvent<HTMLButtonElement>) => unknown;
+  heading?: React.ReactNode;
+  paragraph?: React.ReactNode;
+  buttonLabel?: React.ReactNode;
+  buttonKind?: keyof typeof BUTTON_KIND;
+  backgroundColor?: string;
+  backgroundColorType?: keyof typeof BACKGROUND_COLOR_TYPE;
+  image?: {
```
chasestarr, line 28: should this have required aria-label?

src/message-card/message-card.tsx
```diff
@@ -0,0 +1,144 @@
...
+    <Root
+      onClick={onClick}
+      $backgroundColor={backgroundColor}
+      $backgroundColorType={backgroundColorType}
+      $imageLayout={layout}
+      {...RootProps}
+    >
+      {image && (
+        <Image
+          $src={src}
```
chasestarr, line 98: I believe a `role="img"` and `aria-label` would be good https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/img_role

src/message-card/utils.ts
lhbrennan: Ah, great idea!

src/message-card/message-card.tsx
lhbrennan: Yea I think you're right. I wasn't eager to add another required prop to the component, but the image really is more than just "background" decoration. I'll add this in.

chasestarr (approved):

PR #5086, opened 2022-08-10
[Bug Fix] group button border radius issue

Gopigk:
`Fixes #1, Fixes #2`


src/button-group/button-group.tsx
```diff
@@ -131,6 +131,32 @@ export default class ButtonGroup extends React.Component<ButtonGroupProps> {
...
+                          borderBottomLeftRadius: '0px',
+                          marginLeft: '1px',
+                        };
+                      }
+
+                      if (index === children.length - 1) {
+                        return {
+                          borderTopLeftRadius: '0px',
+                          borderBottomLeftRadius: '0px',
+                          marginLeft: '1px',
```
madhusudhand, line 156: in case of shape !== default a scale100 margin is expected (line 160).

Capture the border radius values in an object in all the three cases, and append object to the below return statements. Leave margins as is.

Gopigk: Below return means the return statement in line 167, right?

madhusudhand: both 161 and 167.

chasestarr:
Thank you for submitting this PR. This is the prefered behavior from our design team

PR #5092, opened 2022-08-11
chore(button): change mini button radius to 4px

lhbrennan:
#### Description

Previously the `Button` radius for all sizes was _8px_. This changes the radius to _4px_ when `Button` size is `mini`.


chasestarr (approved):

src/button/styled-components.ts
```diff
@@ -9,7 +9,6 @@ import { KIND, SIZE, SHAPE } from './constants';
 import type { SharedStyleProps } from './types';
 import type { Font } from '../themes/types';
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
chasestarr, line 11:
```suggestion
```

PR #5146, opened 2022-08-27
Update remix guide

akazwz:
#### Description
That guide is not very accurate and outdate,  there should be a SSR config in remix with base web.
I have a demo about using remix and base web,
the repo link is https://github.com/akazwz/remix-baseweb-ts
and the online demo is https://remix-baseweb.zwz.workers.dev/
#### Scope
Patch: Bug Fix


chasestarr:
Thank you for submitting. We will not update this link and may remove it in the future

PR #5169, opened 2022-09-21
fix: datepicker focus outline shouldnt be cut off

rBortnikas:
Fixes datepicker’s day focus outline cut off issue.

#### Description

Currently, day's focus outline is partly covered by the following day’s cell. The images and the recording explain this better. This fix makes the datepicker look a little more polished.

#### After

https://user-images.githubusercontent.com/17082436/191592924-67021dce-bdb8-4948-b5d1-c592fc22c402.mp4

#### Scope

Patch: Bug Fix


chasestarr:
Thank you for this PR. We will need to figure out a solution that does not use z-index. Please use another method for creating a stacking context. We almost never use z-index in base web

PR #5173, opened 2022-09-26
feat(data-table): add submit actions

mia-le:
#### Description
Adds the ability to add "Submit actions" for use cases where the data table is used in a form and CTAs based on row selections are needed. `submitActions` are passed in similarly to `batchActions`, but is placed where users would expect CTAs. We display checkboxes when `submitActions` are available, but do not hide the search bar at selection if there are no `batchActions`

#### Scope
Minor: New Feature


src/data-table/types.ts
```diff
@@ -127,6 +127,7 @@ export type StatefulDataTableProps = {
   rowHeight?: number;
   rowHighlightIndex?: number;
   searchable?: boolean;
+  submitActions?: BatchAction[];
```
chasestarr, line 130: Can this be reconfigured into something more reusable? Maybe like below?

```ts
{
  // other types
  selectable: boolean; // defaults false, renders checkboxes even if no batch actions are provided
}
```

chasestarr (commented):
Thank you for contributing this change. I want to enable applications to have similar behaviors as this by making the component more customizable and less opinionated. 

I'm hesitant to add more built-in behaviors like submit-actions because it will often necessitate similar special-case workarounds in the future. 

Could we reconfigure this PR into decoupling row selection from batch actions? See types.js inline comment for specific api suggestion. When no batch actions are provided and selectable = true, on row selection the text query input should remain visible.

Would that enable your use-case?

PR #5179, opened 2022-09-30
fix(types): make SkeletonProps.rows optional to match the docs

alexgorbatchev:
#### Description

According to the [docs](https://baseweb.design/components/skeleton/) `SkeletonProps.rows` is optional, however it's not so according to the [typings](https://github.com/uber/baseweb/blob/master/src/skeleton/types.ts#L14). This PR makes `SkeletonProps.rows` optional.

#### Scope

Patch: Bug Fix


chasestarr (approved):
thank you

PR #5187, opened 2022-10-03
Update button background in dark mode to match design spec

merriechristopher:
`Update the buttonPrimaryFill color token dark mode mapping to match design spec` 

#### Description

Currently all Primary buttons have a component specific color token that maps them to white in a dark mode theme. This change replaces the mapping to the correct color token to match the design spec (`primaryA`)

Note: Long term, we should remap all component color tokens to map to semantic ones (`backgroundInversePrimary`) over the primitives.

Before: 
Screenshot 2022-10-03 at 15 17 40

After:
Screenshot 2022-10-03 at 15 21 09

#### Scope

Bugfix


chasestarr (approved):
thank you!

PR #5193, opened 2022-10-10
Fix display names

lhbrennan:


src/badge/styled-components.ts
```diff
@@ -184,8 +184,6 @@ export const StyledRoot = styled<'div', {}>('div', () => {
   };
 });
 
-StyledRoot.displayName = 'StyledRoot';
```
chasestarr, line 187: Please keep the displayName assignments. Enzyme uses this property to search for components in the react tree

lhbrennan: Ah. My understanding was that the display name is automatically deduced from the name of the function (or the variable the component is assigned to), and thus explicitly assigning displayName is only necessary when you want the component to show up differently in the dom. Is that incorrect?

chasestarr: Not when building with styletron :(

lhbrennan: Gotcha. Added displayName to all the recent new components.

chasestarr (approved):

PR #5205, opened 2022-10-24
Mobile header

lhbrennan:
#### Description
New component: https://t3.uberinternal.com/browse/WPT-8673

#### Scope
Minor: New Feature


src/mobile-header/__tests__/mobile-header.e2e.ts
```diff
@@ -0,0 +1,24 @@
...
+Copyright (c) Uber Technologies, Inc.
+
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+*/
+
+import { expect, test } from '@playwright/test';
+import { mount, analyzeAccessibility } from '../../test/integration';
+
+test.describe('mobile-header', () => {
```
lhbrennan, line 11: Which additional tests make sense for this component?

src/mobile-header/mobile-header.tsx
```diff
@@ -0,0 +1,140 @@
...
+  StyledRoot,
+  StyledNavContainer,
+  StyledTitle,
+  StyledActionButtonsContainer,
+} from './styled-components';
+import { TYPE } from './constants';
+import { useStyletron } from '../styles/index';
+import type { MobileHeaderProps } from './types';
+
+const renderButtonContent = (content) => {
```
lhbrennan, line 20: The design docs indicate that the content for the MobileHeader buttons can be either an icon or a string. My sense is that if the consumer wants to use an icon, the component would ideally size the icon appropriately. But what if they want to pass in something that is't an icon (or a string)? In that case this render function is passing in an erroneous `size` prop.

chasestarr (approved):

PR #5281, opened 2023-04-21
fix(table-semantic): remove padding-right from sortable columns

leyanlo:
- Remove padding-right from sortable columns to match documentation: https://base.uber.com/6d2425e9f/p/901301-data-table/t/4633ea
- Avoid right-aligning text for numeric columns to match documentation.
- Add `background: inherit` to sort icon so it appears in front of text to match documentation.

This should help us with data-dense tables, where we have many columns that need to fit on a page.


leyanlo:
Moving to web-code

PR #5317, opened 2023-11-14
[baseui stepper] fix IncrementButton and IncrementButtonIcon overrides

jinyang1994:
#### Description

When I use overrides to override the buttons of stepper, I find that IncrementButtonIcon and IncrementButton use the DecrementButtonIcon and DecrementButton. 

Please merge them, thanks.

#### Scope
Patch: Bug Fix


jinyang1994:
I don’t know that why you closed it?! @tajo

PR #5329, opened 2024-03-20
fix(icons): add title="" to icons

leyanlo:
add title="" to icons to prevent the wrong name being read on screen readers

E.g. 
Screenshot 2024-03-20 at 2 20 24 PM

Screenshot 2024-03-20 at 3 03 55 PM


chasestarr (approved):
TY!

PR #5394, opened 2026-04-07
[chore]Use updated radio/checkbox for yard knobs

hualf1995:
#### Description
Use updated radio-v2 and checkbox-v2 for doc site yard knobs

Screenshot 2026-04-06 at 5 54 56 PM

#### Scope
no need to bump version since it is only for doc site change


DianaSuvorova (approved):
👍

PR #5403, opened 2026-07-02
Add noMargin prop to Tag component

dyesin:
## Summary
Tag's Root applies a hardcoded 5px margin on all sides, which forces
consumers (including TagGroup itself) to reach for an
overrides.Root.style.margin = 0 hack to lay tags out normally. Add a
first-class noMargin prop instead, and use it in TagGroup to remove
the override hack.

Mirrors the internal noMargin change merged in web-code#30018.

## Issues
WBASE-282


DianaSuvorova (approved):
Nice, Thanks!

PR #5410, opened 2026-09-05
fix: stop corrupting the violations-by-node map in A11y

zigzagdev:
`Fixes` 

#### Description

`segmentViolationsByNode` builds a map from DOM node target to its axe violations. 
When a node already had an entry, it work:

```js
map[node.target] = map[node.target].push(violation);
```

`Array.prototype.push` returns the new length (a number), not the array, so this overwrote the array with a number. Once a node had 2 violations recorded, `map[target]` became a number instead of an array. 
On a 3rd violation for that same node, `map[target].push(...)` threw a `TypeError`, crashing the dev-only `A11y` wrapper.

#### Scope
Patch: Bug Fix


BUG_FINDINGS.md
```diff
@@ -0,0 +1,51 @@
+# Bug & Refactor Findings — `src/a11y` … `src/button`
```
dyesin: Please drop this file before merging. It's not a deliverable, will go
stale immediately, and doesn't belong in the repo.

dyesin (changes requested):

BUG_FINDINGS.md
zigzagdev: @dyesin 

Thank you for reviewing.

I'm so sorry for adding this file.
And Removed it . ([commit](https://github.com/uber/baseweb/pull/5410/commits/25b68dcc4a6fe3c035cf5767b3f6db85ad514065))

PR #5411, opened 2026-09-06
fix: mobile horizontal sub-nav never rendering in `AppNavBar`

zigzagdev:
`Fixes #1, Fixes #2` 

#### Description

Fixing an evaluation-order bug in `AppNavBar` where `secondaryMenu`, `desktopSubNavPosition`, and `mobileSubNavPosition` were plain `let` variables mutated as a side effect inside the desktop `mainItems.map()`.
Because the mobile nav block is evaluated earlier in the same JSX return statement, it always read these variables before that map ran, so `mobileSubNavPosition === POSITION.horizontal` could never be true and the mobile horizontal sub-nav never rendered (a `// @ts-expect-error always false` comment in the source acknowledged this).
Now the active item's secondary menu and positioning are resolved up front via `useMemo`/`find`, so both the mobile and desktop blocks read the same, already-resolved values.

#### Scope

- Patch: Bug Fix


dyesin (approved):
LGTM.

PR #5414, opened 2026-09-07
fix: highlight `BottomNavigation` overflow selector for the first overflow item

zigzagdev:
#### Description

The overflow "More" selector's active state was computed as `isActive={displayOverflow || activeKey > 4}`, which is off-by-one.
Since `navItems.length > MAX_SELECTORS (5)`, only indices 0-3 get their own direct `Selector` (`navItems.slice(0, 4)`); index 4 is already only reachable through the overflow panel/selector. 
However, `activeKey > 4` requires index 5 before the "More" tab lights up, so selecting the first overflow item (index 4) showed its panel while leaving the "More" tab unhighlighted.
Changed the condition to `activeKey >= MAX_SELECTORS - 1`. 
Therefore, it matches the actual cutoff used when slicing selectors.

#### Scope

Patch: Bug Fix


dyesin (approved):
LGTM

PR #5415, opened 2026-09-07
Update card.tsx

MaddipatlaChetan24:
Fixes #1, Fixes #2

#### Description

Hoists the `levels` array (`['', 'h1', 'h2', 'h3', 'h4', 'h5', 'h6']`) out of
the `SemanticTitle` component in `Card` into a module-level constant
(`HEADING_LEVELS`). Previously this array literal was re-allocated on every
render of `SemanticTitle`, which runs for every `Card` with a `title` and
re-runs on every re-render. The array's contents never change, so it only
needs to be allocated once at module load. No change to rendered output or
component behavior.

#### Scope
Patch: Bug Fix


dyesin (changes requested):
Strip the unrelated formatting noise, keep only the hoist.

dyesin (changes requested):
strip the unrelated formatting noise, keep only the hoist.

PR #5418, opened 2026-09-15
Replace posts with ToDense layer implementation

MaddipatlaChetan24:
Removed the existing posts array and added a new ToDense layer implementation for handling composite tensors in Keras models.

#### Description

Simplifies `ToDense.get_config()` to merge `base_config` and `config` using a dict-unpacking merge (`{**base_config, **config}`) instead of converting both to lists, concatenating them, and re-wrapping with `dict(...)`. Verified behaviorally identical, including on the key-overlap case (`config`'s values still take precedence over `base_config`'s, matching the original's `+` ordering) — same output dict either way, just without the two throwaway list allocations. No other changes.

#### Scope
Patch: Cleanup


MaddipatlaChetan24:
please check this !!!

dyesin (changes requested):
What are you trying to do?

PR #5419, opened 2026-09-16
Refactor badge props to use BaseBadgeProps

MaddipatlaChetan24:
Fixes #1, Fixes #2

#### Description

Extracts the fields shared by `BadgeProps`, `NotificationCircleProps`,
and `HintDotProps` (`color`, `hidden`, `horizontalOffset`,
`verticalOffset`, `overrides`, `children`) into a new internal
`BaseBadgeProps` type, and builds each of the three exported types on
top of it via intersection instead of repeating the six fields
independently in each. `content` and `placement`, which are shared by
only two of the three types, stay declared individually on
`BadgeProps`/`NotificationCircleProps` rather than being pulled into
the base, since `HintDotProps` doesn't have them.

`BaseBadgeProps` is not exported — it's purely an internal
organizational detail, not a new public API surface.

Verified structural equivalence with the TypeScript compiler: wrote a
mutual-assignability check (`A extends B ? (B extends A ? ... )`)
between the original and refactored versions of all three types and
confirmed it compiles cleanly under `tsc --strict`, meaning each
refactored type is identical, field-for-field, to what it was before.
No consumer of these types is affected.

#### Scope
Patch: Cleanup


MaddipatlaChetan24:
please check this !!!

PR #5420, opened 2026-09-17
Prevent mutation of whitelisted array during sorting

MaddipatlaChetan24:
Ensure whitelisted array is not mutated by sorting directly.

Fixes #1, Fixes #2

#### Description

`getOverrides` in `overrides.jsx` called `whitelisted.sort()` directly
on the `whitelisted` prop array. `Array.prototype.sort()` sorts in
place and returns the same array reference — it does not return a new
sorted copy. Since `whitelisted` is a prop passed in by the caller
(potentially a shared/reused array reference, e.g. a module-level
constant reused across multiple `<Overrides>` usages to avoid
recreating the array on every render), this silently reordered the
caller's own array as a side effect of simply rendering this component.

Reproduced directly: calling `getOverrides` once with a shared constant
array as `whitelisted` permanently reorders that constant in place.

Fix: `[...whitelisted].sort()` — copies the array before sorting, so
the caller's array is left untouched while still returning the sorted
result. Verified this produces the same sorted output as before while
leaving the original array reference unmodified.

No other changes. (Note: I don't have visibility into whether any
current call site in this codebase actually reuses a shared array
reference for `whitelisted` — I can't confirm this has caused a visible
symptom today. Regardless, mutating a prop array is an unsafe pattern
worth fixing on its own merits.)

#### Scope
Patch: Bug Fix


MaddipatlaChetan24:
please check this !!!

PR #5423, opened 2026-09-19
Update knob.tsx

MaddipatlaChetan24:
Fixes #1, Fixes #2

#### Description

Two override style objects in `Knob.tsx` were being re-created on every
render for no reason:

- `RadioGroupRoot`'s override was written as a function of `$theme`
  (`style: ({ $theme }) => ({...})`) but never actually referenced
  `$theme` in its returned object — a constant disguised as a function.
  Replaced with a plain `RADIO_GROUP_ROOT_OVERRIDE_STYLE` object hoisted
  to module scope, which baseui's `overrides.<Component>.style` accepts
  directly. Removes both the per-render function allocation and the
  function call baseui previously had to make to resolve it.
- `Checkbox`'s `Label` override (`style: { fontWeight: 500 }`) was
  already a static shape but was being re-allocated as a new object
  literal on every render. Hoisted to `CHECKBOX_LABEL_OVERRIDE_STYLE`.

Both changes are behavior-identical — same resolved styles, same
visual output.

Left the two per-`Radio` overrides inside the enum options `.map()`
unchanged. They genuinely reference `$theme` and can't become plain
objects, and hoisting them to module scope risks losing baseui's
contextual type inference for the `$theme` parameter without knowing
this project's exact `tsconfig`/baseui override type setup.

#### Scope
Patch: Cleanup


MaddipatlaChetan24:
please check this !!!

dyesin (changes requested):
Include tests. What is the benefit of this change? Have you run tests, validations.

MaddipatlaChetan24:
Thanks for the questions — answering both directly:

**What is the benefit of this change?**
`Knob` re-renders on every keystroke/interaction (it's driven by
`useValueDebounce`), and both overrides were being reconstructed on
every one of those renders:
- `RadioGroupRoot`'s override was written as `style: ({ $theme }) => ({...})`
  but the returned object never actually referenced `$theme` — so baseui
  was calling a function every render just to get back the same constant
  object. Hoisting it to `RADIO_GROUP_ROOT_OVERRIDE_STYLE` removes both
  the per-render allocation and the unnecessary function call.
- `Checkbox`'s `Label` override (`{ fontWeight: 500 }`) was a plain
  object literal, so it was a fresh object on every render even though
  its shape never changes.

Both are pure allocation/call-count reductions — no visual or behavioral
difference, since the resolved style values are identical to before.

**Have you run tests/validations?**
No — I don't have this repo's dev environment set up, so I haven't run
the existing test suite or done a visual/snapshot check myself. I traced
the change by hand (confirmed `$theme` is unused in the `RadioGroupRoot`
function body, confirmed the `Checkbox` Label object's shape is static),
but that's static reading, not an executed test run.

**Tests included?**
No test was added. Honestly, I'm not sure this specific change is
something a typical test would catch either way — it doesn't change
rendered output, so a snapshot/visual test wouldn't show a diff, and a
unit test asserting "this object is only allocated once" would need
something like a render-count/reference-equality assertion across
re-renders, which I don't know this repo's existing test conventions
for. If there's a lightweight way you'd want that verified (e.g. an
existing render-count helper elsewhere in the test suite), pointing me
at an example would help me add one properly rather than guessing at
a pattern that doesn't match how this codebase actually tests things.

PR #5424, opened 2026-09-20
Update checkbox-v2.ts

MaddipatlaChetan24:
Fixes #1, Fixes #2

#### Description

`CheckboxConfig`'s picked `changeHandlers` list included `onMouseDown`
and the complete `onMouseEnter`/`onMouseLeave` pair, but omitted
`onMouseUp`. Confirmed against baseui's own API docs that `Checkbox`
supports `onMouseUp` as a first-class prop alongside `onMouseDown`,
`onMouseEnter`, `onMouseLeave`, `onFocus`, and `onBlur` — so this was
a real supported handler missing from the playground config, not a
deliberate omission of an unsupported prop.

Fix: added `"onMouseUp"` to the picked list. Assumes `changeHandlers`
in `./common/common` already defines an `onMouseUp` entry parallel to
its existing `onMouseDown` entry — please confirm that file has it
before merging, since I don't have its contents.

#### Scope
Patch: Bug Fix


MaddipatlaChetan24:
please check this !!!

dyesin (changes requested):
Needs a companion edit to common.ts before it can merged. Breaking change.

MaddipatlaChetan24:
The missing companion change dyesin flagged is now available in a
separate PR: #5426 — adds the missing `onMouseUp` entry to
`changeHandlers` in `common.ts`, which this PR's `pick()` call depends
on.

@dyesin — that PR also asks for clarification on what specifically
breaks, since adding a new key to `changeHandlers` shouldn't affect
existing consumers on its own. Happy to fold both into one PR instead
if you'd prefer that over two linked ones — let me know which you'd
rather review.

